PR #15, opened 2016-11-02
Lle

PatrickvL:
A bit of a mess commit-wise, but what do you think of the last commit?


x1nixmzeng:
👍 for the switch macros, although porting from qemu may make it harder to follow

Remove `Cxbx.VC.db`
Squash your commits

PatrickvL:
I'm still trying to get used to git, being a perforce user for fifteen years. I'll have a look how to squash commits.

I suspect in that one file the line ending changed from UNIX to windows style.

As for the macro's, I'm striving to end up with code that could easily be ported over to MAME, once they allow rendering directly through OpenGL. Still a long way to go, as this code still needs to be converted into actual devices, but I wanted to show you my idea.

LukeUsher:
Looks great, anything that makes the code easier follow and maintain is a welcome addition.

As for the line endings, the entire code base is a mess right now, in some cases there are even have Unix and Windows line endings within the same file. I guess this is the result of so many contributors over many years, without an agreed coding style. When I get home from work, I'll create an issue with some thoughts in order to open up a discussion about coding standards.

PR #18, opened 2016-11-15
here you go

PatrickvL:


PatrickvL:
I should have merged all commits into one, squashing the unintentional update and subsequent rollback. Too late now. Never mind

LukeUsher:
Process is all over the place currently, we'll get there.
I'm new to collaborating, I'm usually only working on single developer projects, even at work.

PR #20, opened 2016-11-24
For your consideration ;)

PatrickvL:


LukeUsher:
I'll have a proper look when I get home for work, but right now it looks good.

I am not sure about typedef wchar_t to *WCHAR though, wchar_t is compiler specific and could differ in size if a different compiler is used, on GCC for example it would be 4 bytes rather than the 2 bytes it is with MSVC.

PR #36, opened 2016-12-19
More kernel stuff

PatrickvL:


LukeUsher:
Copy/pasting from Google Hangouts for documentation purposes:

The functions you implemented that use fs:[24h] and such will need to be updated to use the in-memory EmuFS structure we create in EmuFS.cpp.

Accessing the FS registers directly would have worked on 32-bit Cxbx, as the FS register would have been pointing to the Xbox FS thanks to the LDT.

That is not possible on 64-bit as we cannot alter the LDT so instead, we store a pointer to the EmuFS structure in the host FS structure, and use that to look up the EmuFS 

See KfRaiseIrql for an example

LukeUsher:
I just did some digging and it looks like my KPCR implementation is wrong. 
The KPCR should be per processor, so in the case of Cxbx we should only have one KPCR structure as the xbox only has one processor/one core, however, the NT_TIB portion of the KPCR structure should be per thread. Thread Local Storage is working okay because we create a KPCR per thread, and therefore also an NT_TIB per thread, however the rest of the KPCR structure (IRQ data, etc) should be global.

This will cause us issues with IRQ emulation and any other kernel calls that use the KPCR. The solution would be to store the KPCR in a global variable, however we still require a way to make sure the NT_TIB element is per-thread.

I'm not quite sure how to go about implementing this

PR #37, opened 2016-12-19
Extra logging

PatrickvL:
Not finished yet, but since StrikerX3 is working on file API's this might help already


LukeUsher:
Thanks, good work as always

PR #42, opened 2016-12-20
Use IRQL value in the emulated KPCR structure rather than overwriting the Windows TIB structure

LukeUsher:
Note: The entire KPCR is currently per thread, this is not correct as the KPCR should be per processor.
The issue here is that the first part of the KPCR, NT_TIB should be per thread.

I haven't figured out a good way to keep that portion per-thread while keeping the rest of the structure global, that said, it will function fine if only one thread calls the IRQL functions.

### DIFF
--- a/src/CxbxKrnl/EmuKrnl.cpp
+++ b/src/CxbxKrnl/EmuKrnl.cpp
@@ -56,6 +56,12 @@ namespace NtDll
 };
 
 // ******************************************************************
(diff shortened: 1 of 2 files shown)


LukeUsher:
Added an implemenation of KeGetPcr();

This is a macro on the Xbox, rather than an exported kernel function, which is why it lacks the function offset and EXPORTNUM

PR #53, opened 2016-12-22
File information struct conversions

StrikerX3:
And a bunch of other things, such as overall code cleanup/improvements made by PatrickvL


StrikerX3:
This should get `NtSetInformationFile` and `NtQueryInformationFile` working. I still need to work on `NtQueryDirectoryFile`.

Issue #58, opened 2016-12-23
The new file/path funcs break many XBEs (if not all)


BiatuAutMiahn:
```
EmuKrnlNt.cpp (0x30F8): NtOpenFile forwarding to "IoCreateFile"...
EmuKrnlIo.cpp (0x30F8): IoCreateFile(
 OUT FileHandle              : 0xF1EFE3C
   DesiredAccess             : 100001
   ObjectAttributes          : 0x0F1EFE20 -> POBJECT_ATTRIBUTES {
   .RootDirectory: 00000000
   .ObjectName: 0x0006B000 -> PSTRING {
   .Length: 1D
   .MaximumLength: 1E
   .Buffer: \Device\Harddisk0\partition1\}
   .Attributes: 40}
 OUT IoStatusBlock           : 0xF1EFE2C
   AllocationSize            : 0x00000000
   FileAttributes            : 0
   ShareAccess               : 3
   Disposition               : (CREATE_DISPOSITION) FILE_OPEN = 0x1
   CreateOptions             : (CREATE_OPTION) FILE_DIRECTORY_FILE|FILE_SYNCHRONOUS_IO_NONALERT|FILE_OPEN_FOR_FREE_SPACE_QUERY = 0x800021
   Options                   : 0
);
EmuKrnl : IoCreateFile Corrected path...
  Org:"\Device\Harddisk0\partition1\"
  New:"$XbePath\Device\Harddisk0\partition1\"
EmuKrnl (0x30F8): IoCreateFile Failed! (0xC000003A)
EmuKrnlIo.cpp (0x30F8): IoCreateFile returns -1073741766
EmuXapi.cpp (0x30F8): EmuXLaunchNewImage(
   lpTitlePath               : <nullptr>
   pLaunchData               : 0F1EF238
);
EmuMain (0x30F8): Recieved Fatal Message:

* The xbe rebooted to Dashboard and xboxdash.xbe could not be found
```

PatrickvL:
Thanks for testing, it seems I've messed something up. Will look into that soon

LukeUsher:
Just had a look, it would seem that FindNtSymbolicLinkObjectByDevice is returning the wrong symbolic link object, this looks to be because the C symbolic link has XboxSymbolicLinkPath set to an empty string, causing the strncmp to compare 0 characters.

PatrickvL:
[...]

PR #59, opened 2016-12-23
Update symbolic code to be closer to how real hardware does things.

LukeUsher:
This means that we only assign a symbolic link for D:, on a real Xbox the kernel only handles this, the application takes care of the rest, usually via XAPI

Added some comments to explain behavior that is currently incorrect, eg: Drive D should point to the Xbe path, which is not always CDROM, however, as we set CDROM to the Xbe Path, it works anyway.

### DIFF
--- a/src/CxbxKrnl/CxbxKrnl.cpp
+++ b/src/CxbxKrnl/CxbxKrnl.cpp
@@ -504,16 +504,22 @@ extern "C" CXBXKRNL_API void CxbxKrnlInit
 
 	// Create default symbolic links :
 	{
+		// TODO: DriveD should always point to the Xbe Path
+		// This is the only symbolic link the Xbox Kernel sets, the rest are set by the application, usually via XAPI.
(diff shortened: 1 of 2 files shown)


LukeUsher:
This also fixes issue #58

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -527,21 +533,6 @@ extern "C" CXBXKRNL_API void CxbxKrnlInit
 
 		DbgPrintf("EmuMain : XeImageFileName = %s\n", xboxkrnl::XeImageFileName.Buffer);
 
-		CxbxCreateSymbolicLink(DriveA, DeviceCdrom0); // CdRom could be read from A:
```
PatrickvL: perhaps it's a good idea to keep this list around somewhere for documentation purposes

PatrickvL:
Thanks for looking into this

PR #62, opened 2016-12-24
Fix NV2A address mapping

LukeUsher:


src/CxbxKrnl/EmuNV2A.cpp
```diff
@@ -649,125 +649,153 @@ WRITE32_START(USER)
...
 
+typedef struct NV2ABlockInfo {
+		uint32_t offset;
+		uint32_t size;
+		uint32_t(*read)(uint32_t addr);
+		void(*write)(uint32_t addr, uint32_t value);
+} NV2ABlockInfo;
+
+static const NV2ABlockInfo regions[] = {{
+		0x000000,
```
PatrickvL: It's preferrable to replace these hardcoded offsets with their corresponding NV2A constants

LukeUsher:
Merging this for now as it is a good step in the right direction.
Improvements can come in the future once we have implemented an efficient way of dispatching memory functions

Issue #72, opened 2016-12-29
Config options for LLE/HLE sound and graphics


PatrickvL:
LLE is about to start for both sound (APU) and graphics (GPU), but we don't want to delete our HLE just yet.

So, we need a switch in the GUI for both sound and graphics, choosing between HLE and LLE.
This switch needs to be transferred to our CxbxKrnl.DLL, and be put in bLLE_APU vs bLLE_GPU (which were introduced recently).

Default should be HLE for both.
Choosing LLE is purely for development purposes, so these switches might as well only be available in DEBUG builds...

PatrickvL:
Done by pull request #73 (except DEBUG-only availability)

PR #73, opened 2016-12-29
Configurable LLE switches

PatrickvL:


PatrickvL:
Fixes issue #72

LukeUsher:
Looks fine to me. GUI work is a ballache but does need to be done

PR #77, opened 2017-01-03
Enable the use of PVS-Studio (Static Code Analysis Tool) 

LukeUsher:
These comments activate the free for Open Source license for PVS Studio
See http://www.viva64.com/en/b/0457/ for more details.

An initial run of PVS-Studio found the following:
460 high severity issues
146 medium severity issues
157 low severity issues


BiatuAutMiahn:
Nice!

PatrickvL:
Perhaps imports shouldn't be instrumented.

PatrickvL:
A link to the current analysis results would be nice

x1nixmzeng:
Compiling with LLVM does a great deal better than VS. Most of the warnings are to do with the `printf` format specifiers.

PR #78, opened 2017-01-03
Added unimplemented stubs for remaining Xc functions

x1nixmzeng:
This is missing the correct return values (if any) but should be useful in the current state


PatrickvL:
Great work! Thanks for the effort, much appreciated. It looks like this can be accepted as-is for now.
(I'll still do a quick review of the signatures once i find the time, correcting a few tidbits if you don't mind)

LukeUsher:
Thanks, 
We are getting close to the completion of kernel signatures, now we just need implementations ;)

PR #79, opened 2017-01-04
Review results on Xc* kernel functions

PatrickvL:


x1nixmzeng:
Just a few changes then ;)

I understand most of them though - thanks!

PatrickvL:
It wasn't such a big deal; I added return types, made the commenting and indenting consistent with other kernel API's, corrected argument types, and made sure the logging won't render binary argument values as strings.

PR #84, opened 2017-01-05
LLE_OpenGL work

PatrickvL:


LukeUsher:
Merging this as it has no effect on HLE emulation, it is totally opt-in.

PR #85, opened 2017-01-06
Implemented HalReadWritePciSpace

LukeUsher:
On the Xbox this function calls the X86 in/out opcodes, so this implementation calls IORead/IOWrite functions in EmuX86.

Once In/Out opcodes are implemented, they will also be implemented by calling these new IO functions.


PatrickvL:
Which software calls this? Might I suggest adding comments to the code of each function, documenting test cases?

LukeUsher:
All software that access the GPU when LLE is enabled.
It's called by DirectX during hardware init, I haven't checked yet but I am sure it is used elsewhere too

PR #87, opened 2017-01-07
Fixed errors when compiling with the LLVM toolset

x1nixmzeng:
Fixes a few errors from clang:

- Initialization order
- Const-correctness for texts
- Invalid pragma syntax with `#endif`
- Concatenating the date macro


LukeUsher:
Thanks!

Issue #55, opened 2016-12-23
Remove all compiler hints and warnings


PatrickvL:
Having no compiler hints and warnings gives a much cleaner build.

Causing new compiler messages should be forbidden - if any appear, the person responsible for that should fix it.

x1nixmzeng:
Agreed, the flag to treat warnings as errors should be set, and perhaps raise the error level?

However, as there are copies of specific SDKs some files will flagged as deprecated, ie `GetVersionExA` in `multimon.h` from `import\DirectX8\include\multimon.h`

x1nixmzeng:
Fixes from #100 have helped with this issue.

The biggest issue left is the alignment of XBE and EXE structures - as the use of `#include "AlignPosfix1.h"` is not really supported with other compilers.

Otherwise there are just lots of printf-style format specifiers which need updating!

x1nixmzeng:
Ah, also the device enumeration callback issue which I don't know how to solve yet:

```
1>..\..\src\Common\Win32\XBController.cpp(933,28): error : no member named 'WrapEnumGameCtrlCallback' in 'XBController'
1>              &XBController::WrapEnumGameCtrlCallback,
1>               ~~~~~~~~~~~~~~^
1>..\..\src\Common\Win32\XBController.cpp(973,60): error : use of undeclared identifier 'WrapEnumObjectsCallback'
1>          m_InputDevice[m_dwCurObject].m_Device->EnumObjects(WrapEnumObjectsCallback, this, DIDFT_ALL);
1>                                                             ^
```

Margen67:
If we're trying to get rid of warnings maybe the warning level should be raised?
![capture](https://user-images.githubusercontent.com/3462541/40161889-f4fb0b32-5966-11e8-871c-468dd7ab7899.PNG)

PatrickvL:
Our own code should generate warnings, imported code shouldn't.

PatrickvL:
In an ideal world, our code shouldn't raise any warning even when the compiler uses its most thorough validation methods. But right now it might not be feasible to reach that level of confirmation to the highest possible standards

RadWolfie:
Moving comment to here from #1262 conversation.
>There's no way to treat warning as error level to be raise. The option is either treat warning as error or warning. See "Treat Warnings as Errors" option.
>
[...]

PR #88, opened 2017-01-07
Replaced libjpeg with stb_image

x1nixmzeng:
Removes the hack needed for MSVC 2015 (and helps with #55)!

Also supports other image types when/if the app is rebranded 😃


LukeUsher:
Once again, thank you very much!

PR #90, opened 2017-01-10
Switched over to Distorm

PatrickvL:


src/CxbxKrnl/EmuX86.cpp
```diff
@@ -36,7 +36,14 @@
 #define _CXBXKRNL_INTERNAL
 #define _XBOXKRNL_DEFEXTRN_
 
-#include <Zydis.hpp>
+// Link the library into our project.
+#pragma comment(lib, "distorm.lib")
```
PatrickvL: I just followed the example here : https://github.com/gdabah/distorm/wiki/CSample
Also, http://stackoverflow.com/questions/12821391/c-visual-studio-linking-using-pragma-comment mentions one "doesn't need to add this library to the linker dependencies list", so this approach avoids a change to the solution itself. In addition to that, this syntax could also be used to easily switch between versions.
Finally, http://stackoverflow.com/a/6287783/12170 says : "The advantage of #pragma comment is that the user of your library cannot forget to add the setting. Or add the wrong one, it is not uncommon to get lost at the difference between the debug and release build and the /MD vs /MT build."

LukeUsher: Although it has those advantages, this method of linking libraries is MSVC specific not supported on most other compilers.

As it's such a minor thing, I can merge this in for now, and we can address it later if we need to.

RadWolfie:
@PatrickvL, is there a reason why distorm in this pull request is a compiled build?

It is missing distorm.pdb file with 6 warnings repeatedly.

PatrickvL:
Back then, I couldn't get it to work using source only

RadWolfie:
Will you re-build distorm without debugger option into compiled .lib file? Doing this way will stop outputting warnings for distorm.pdb not being found. Or include distorm.pdb file in the repo with reference to it in some way.

PR #100, opened 2017-01-14
Cleanup

x1nixmzeng:
Few minor changes. You might not want to include 2515071 as it makes it harder to diff older codebases.

Also, is there any reason the `.gitignore` excludes `.filters`? It means MSVC does not group files by type!


LukeUsher:
.filters being ignored is indeed accidental.

I don't see any problems with including 2515071, I have yet to see a branch of Cxbx that has changes to the error class, let's wait and see what @PatrickvL things about merging that one, but it's all fine by me.

PatrickvL:
I added a few remarks, nothing significant if not addressed

x1nixmzeng:
Thanks for the comments - I'm still getting familar with the code 🚀 

I've updated `.gitignore` to allow `.vcxproj.filters` to be submitted - but not added my local copy because it doesn't reproduce the folders - it just splits headers and source files. Luke may have an updated version he could add.

PatrickvL:
Thanks for the fixes, let's merge this!

PR #106, opened 2017-01-15
Removed all duplicates OOVPA's

jarupxx:


PatrickvL:
Please know your work is greatly appreciated, but we shouldn't merge this without a thorough review, which will cost time because of the size of this delta.

PatrickvL:
Not criticizing but since you indicated this work is the result of manual labour, so we have to assume there could be a few errors in here. If we don't catch them, software that worked before this change, might not work anymore afterwards, which will be very hard to troubleshoot.

PatrickvL:
I propose for reviewers to mark every change, right or wrong, so we can be sure everything is reviewed before we merge this. Agreed?

PatrickvL:
Let's wait a little with this review until I've finished the HLE self test. If all goes well, that should report on the exact same set of OOVPA's as have been deduplicated in this PR. That would prove these changes can be merged. Otherwise, there's some more work to do.

PatrickvL:
I'm afraid this PR cannot be merged as it stands right now. In PR #110 the HLE verification generates many many more errors than were changed here, but that's not the real reason to reject this.

In it's current state, this PR looses information.

A much simpler approach can be seen in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/110/commits/20ed2384b1a926507acbde64c14d507e6b8353ad which simply uses a #define to alias an older version to a newer version OOVPA.
This doesn't require any change to the OOVPATable, so no information is lost.
Also, this approach sucessfully prevents HLE verification warnings on identical OOVPA's.

So, I'm really really sorry, but it looks like this needs another attempt. (Please, if you're still up for it, start out small, to prevent wasting precious time.)

LukeUsher:
I apologise too, thanks for the contribution but it would probably be better to start smaller, maybe submit a single patch at a time

PR #107, opened 2017-01-15
Improved OOVPA scanning

PatrickvL:


LukeUsher:
Merging the scanning code paths is definitely a good thing, even with the small performance impact, it will reduce the chance of errors occurring,

PR #118, opened 2017-01-19
Add XDK5788

jarupxx:
Add XDK 5788
All-Star Baseball 2005 status Menus.


LukeUsher:
Thanks!
At a quick glance it looks okay, I'll be able to check this over more thoroughly on Saturday but I don't see there being any issue with getting this merged.

Just a quick note: PatrickvL has submitted a pull-request that removes the patches for XGUnswizzleRect and XGSwizzleRect, you might want to remove these patches from your new 5788 OOVPA's too,

PatrickvL:
I'm impressed with the eye for detail you've shown here, and I'd like to thank you for this effort. 

Since this targets a new XDK version, there's no risk of this breaking existing (somewhat) supported software.

Even though I do have a few remarks, these can just as well be addressed after a merge of the PR.

PR #132, opened 2017-01-27
Fixed the pattern XNetStartup, WSAStartup

### LINKED ISSUE #129: [Regression] WSAStartup is not detected
Version fa30aa09d61b4e921f3a08016da8888ff34b76d6
WSAStartup is registered in XOnline.1.0.4627.inl, but detection fails.

I tested it with Whacked 1.0.4831.
[xbe.txt](https://github.com/jarupxx/cxbx_log/blob/master/Whacked%20Xbe.txt)

jarupxx:
Fixed #129 Those patterns no longer use XREF.

### DIFF
--- a/src/CxbxKrnl/HLEDataBase/XNet.1.0.4627.inl
+++ b/src/CxbxKrnl/HLEDataBase/XNet.1.0.4627.inl
@@ -54,10 +54,10 @@ OOVPA_XREF_LARGE(XnInit, 4627, 18,
         { 0x001A, 0xB1 }, // (Offset,Value)-Pair #7
         { 0x001B, 0x11 }, // (Offset,Value)-Pair #8
 
-        // XnInit+0x0041 : push 0x0AF8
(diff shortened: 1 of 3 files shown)


PatrickvL:
Thanks for working on this. I do wonder - could these OOVPA's be kept unmodified if the XRef itself (XnInit) was improved/fixed instead? That would be an even better solution...

jarupxx:
Exactly. These OOVPAs are available without modification.
Surely it is safer not to change existing OOVPA's...

PatrickvL:
What I mean is, what if these two OOVPA's are kept totally unchanged, including their xref to XnInit, but instead update the OOVPA for XnInit. To determine that, a search in the XDK patterns and/or a test xbe like Whacked can be examined with a tool like XbeExplorer for example.

jarupxx:
I modified OOVPA of XnInit. This value can be shared with XDK 4627 and 4831.

PatrickvL:
Cool! Thanks for digging further into it, this is certainly a better solution.

PR #137, opened 2017-02-01
Remove all references to Cxbe

DisorderIy:
Hopefully I did this correctly. I'm still as rusty with Git as ever 😛


DisorderIy:
Okay, that DEFINITELY does not look right. Is there a way to have only that last commit 0d094e7 in the PR?

LukeUsher:
You appear to have started with the wrong base fork, this includes commits from my personal branch (LukeUsher/Cxbx-Reloaded) which should not be included in this pull request.

Ordinarily, we can do a squash-merge so the pull-request becomes a single commit, however, this time that is not possible as the first three commits should not be merged.

The best approach is to create a new fork of this branch, Cxbx-Reloaded/master, and re-create the changes there, then submit a pull-request containing only that change.

x1nixmzeng:
Afting forking this repo, **pull** changes and create a new branch to work on. After doing that you can **push** and create the pull request from that.

Squashing commits can be done using `git rebase` but not recommended in this case!

DisorderIy:
I'm gonna go and close this, since it's not something that can really be fixed on the fly all that easily.

VelocityRa:
Are you still working on this? I could do it if not.

PatrickvL:
As far as I can tell, the previous attempt was cancelled, so please go ahead

PatrickvL:
If so desired, we could try to create more entry level and/or smallish issues for interested newcomers to work on?

VelocityRa:
@PatrickvL I'm still in the process of looking through the project and deciding if I'll be helping but this would be helpful for everyone in the same position, yes.
It can be something as simple as an `easy` or `good-for-contributors` Issue label.

PR #141, opened 2017-02-03
Disable HLE_ENTRY XG 5788

jarupxx:
Not registered in the OOVPATable XG 5788, it causes a crash.
This entry has been disabled.


PatrickvL:
With what title/software does this OOVPA cause a crash?

jarupxx:
For example, All-Star Baseball 2005 version 5788.
There are few software using 5788, so we can test it using 5849 instead.
Enablie the HLE_ENTRY then 5849 software loading.

jarupxx:
When all OOVPA disabled, you disabled the entry.
because it causes a crash?
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/14585e9fec6d2a421cdc052616eeb36391bf5eed

PatrickvL:
Now that the XGIsSwizzledFormat OOVPA regsitration is back, this little pull request became obsolete.

If there continue to be problems, please open an issues with all relevant details.

PR #163, opened 2017-02-13
Added DSound OOVPAs

jarupxx:
To check the software used for testing and OOVPA values, I would refer you to this [documents](https://github.com/jarupxx/cxbx_log/blob/master/Document_pr_fnc-4.txt).


PatrickvL:
Thank you very much!

PatrickvL:
To answer your question on the correct use of ALIAS registration:

For a lot of OOVPA's, the name of the OOVPA and the name of the EmuPatch could be made identical (the OOVPA and EmuPatch macros insert prefixes, so renaming won't cause name collisions).

Once the name of the OOVPA and the name of the EmuPatch are identical, the PATCH registration syntax can be used (for which the fourth argument is no longer necessary and can be removed from the registration).

Only when there are multiple OOVPA's for one EmuPatch, an ALIAS registration is required.
But by renaming as much as possible to identical names, we can reduce the number of ALIAS registrations considerably.

This is preferrable because PATCH registrations are shorter and thus easier to maintain.

jarupxx:
I understood more deeply. found a preferred ALIAS usage in Xapi.

PatrickvL:
If you're going to rename OOVPA and/or EmuPatch names for DSound, take a look at how I did that for D3D, here : https://github.com/PatrickvL/Cxbx-Reloaded/commit/29483bc9892b2a218f8954db108ceb5b65960d8e

jarupxx:
I'm going to rename them, I will reference Dxbx Patterns file.
Since I found such a pair, I name IDirectSoundBuffer_ and DirectSound_CDirectSoundStream_.
```
IDirectSoundBuffer8_SetOutputBuffer vs IDirectSoundBuffer_SetOutputBuffer
CDirectSoundStream_SetOutputBuffer vs DirectSound_CDirectSoundStream_SetOutputBuffer
```

PatrickvL:
Good luck with the renaming, mostly it's just a matter of removing the '8' and making the prefixes and suffixes consistent. Be carefully though, not to create duplicates by mistake. When that happens, rename only one OOVPA+EMUPATCH, and keep the others intact. Those can be renamed in a later rename action.

PR #169, opened 2017-02-15
Dsound refactoring Part 1

jarupxx:
Renamed IDirectSoundBuffer8_ prefix to IDirectSoundBuffer_
For those, registration is now done via PATCH instead of ALIAS.


PatrickvL:
Thank you!

PR #171, opened 2017-02-15
Fixed Turok crash early in menu's

PatrickvL:


elfogames:
is turok evolution playable?i cant start a new game,always get this error

error

PatrickvL:
Not in Cxbx. Dxbx could get in-game, was even playable for the first level. But that was because of the better emulation of pushbuffer and pixel shaders. We'll slowly port stuff back though, so we'll get in-game one day, I'm sure ;)

elfogames:
but i gets in game in cxbx 0.0.1 Poc

PatrickvL:
Oh, that's interesting! Please create a regression issue for that

PatrickvL:
Thanks for reporting that - is exactly those kind of things that help improving our overall emulation, as we can't catch everything. And in this case, I didn't even know Turok could go in-game with an earlier version. So thanks!

elfogames:
you should create the issue yourself,my english is very bad, so it stays more organized

PatrickvL:
No problem, I'll create an issue for this. But please tell if you remember more details, like from when to when Turok worked, etc,

PatrickvL:
Here's the issue : https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/174.
Again, if you remember any more details (like a commit number or anything), please add those details to the issue.

PR #189, opened 2017-02-21
Full EEPROM readout

PatrickvL:


LukeUsher:
I had a quick test and this doesn't seem to break any titles I have. I see no problem merging this.

PR #208, opened 2017-03-02
Separate XDK file is generated for 5344

jarupxx:


LukeUsher:
I only have one 5344 title available to me right now, so I am unable to properly test this at the moment. 
How does this effect titles like Futurama? In some old Cxbx branches it could reach in-game, it would be great if we could match that

jarupxx:
The big effect improves false detection of the SetMinDistance function.
This is one reason why Futurama run only on a specific branch.

When you adapt to 32bit final version https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/tree/231e924bd13ae31a77a14b1f8128541a562ebaaa it reach in-game.

LukeUsher:
Interesting, that implies it doesn't work in the current branch? Do you have any idea which commit broke it?

jarupxx:
Exactly! It seems that it was broken at least before this commit 67c37feebd014a9e2725409f162e327bc24b0d8f that was fixed so that it can be compiled with Visual Studio 2015.
I will open a ticket after investigating this issue.

src/CxbxKrnl/HLEDataBase/D3D8.1.0.5233.inl
```diff
@@ -1665,40 +1347,40 @@ OOVPATable D3D8_5233[] = {
...
 	REGISTER_OOVPA(D3DDevice_SetFlickerFilter, 5233, PATCH),
-	REGISTER_OOVPA(D3DDevice_SetFlickerFilter, 5344, PATCH),
+	//REGISTER_OOVPA(D3DDevice_SetFlickerFilter, 5344, DISABLED),
 	REGISTER_OOVPA(D3DDevice_SetSoftDisplayFilter, 5233, PATCH),
-	REGISTER_OOVPA(D3DDevice_SetSoftDisplayFilter, 5344, PATCH),
+	//REGISTER_OOVPA(D3DDevice_SetSoftDisplayFilter, 5344, DISABLED),
 	// REGISTER_OOVPA(D3DDevice_Unknown1, 5233, PATCH)
 	REGISTER_OOVPA(D3DDevice_SetRenderState_TwoSidedLighting, 4134, PATCH), // Beware of the typo...
     // ******
     // Duplicate references are normally bad, but this XDK is aliased to XDK 5344
```
PatrickvL: this comment can be removed now that 5344 has its own database

src/CxbxKrnl/HLEDataBase/D3D8.1.0.5233.inl
```diff
@@ -1595,7 +1277,7 @@ OOVPATable D3D8_5233[] = {
     // Duplicated functions generally dangerous, but until such time as XDK 5344
     // has it's own OOVP's, fine here
     // *******
-	REGISTER_OOVPA(D3DDevice_SetRenderState_Dxt1NoiseEnable, 5344, PATCH),
+	//REGISTER_OOVPA(D3DDevice_SetRenderState_Dxt1NoiseEnable, 5344, DISABLED),
```
PatrickvL: Why are all 5344 patches commented out AND disabled?

PatrickvL:
Thanks for your hard work. About the remark on missing functions: I've stopped my work on consolidating all versions into one database, because I'd rather work on something else than HLE. However, I still have an idea to complete incomplete HLE databases, for which I'll create an issue later.

src/CxbxKrnl/HLEDataBase/D3D8.1.0.5233.inl
jarupxx: I removed this comment done.

src/CxbxKrnl/HLEDataBase/D3D8.1.0.5233.inl
jarupxx: These are disabled as they can be detected without using 5344 patches.
How can I do it?

PR #215, opened 2017-03-09
LaunchDataPage improvements

PatrickvL:


PatrickvL:
@LukeUsher : Please review, test and merge this when time permits ;)

LukeUsher:
Notice the following changes in behavior:

Innocent Tears triggers HalReturnToFirmware Unimplemented warnings & launches in the wrong mode
When running default.xbe, it should reload to play the intro/menus, instead it ends up at some kind of character management menu.

Dead or Alive Ultimate relaunches itsself multiple times (instead of once) and then attempts to load the dashboard

PatrickvL:
Hmm. These unimplemented warnings in DoA deserve further research. Same for the crash in Innocent Tears, although that might have regressed due to something​else than my LaunchDataPage & HalReturnToFirmware changes. Since I'm quite busy IRL, could you look into this? If not, I'll try to look into it next week or so. Thanks for testing anyway!

PatrickvL:
Oh, that crash might well be fixed already in the other PR, by calling CxbxInitFilePaths() before CxbxRestoreEEPROM()

LukeUsher:
These issues persist in the other branch too, I believe it must b e related to LaunchDataPage/HalReturnToFirmware, I'll see if I can figure out what's going on

PatrickvL:
Great, thanks. I'm sorry that I seem to cause breakage from time to time

LukeUsher:
Further investigation has revealed that HalReturnToFirmware and MmPersistContiguousMemory are working as intended, additionally the Launch data page is being correctly loaded (they are, however, stored in Appdata\Local\CxbxReloaded, where all other files are in AppData\Roaming\Cxbx-Reloaded). This shouldn't matter for now, but we do need to come up with a consistent storage location. I was using AppData\Roaming\ as that's where the %APPDATA% environment variable pointed to.

The unimplemented message we are seeing is because there is a break; statement missing within HalReturnToFirmware., causing the switch/case to fall through, but this is not the reason for the breakage.

Turns out that XLaunchNewImage was only writing the title ID to the launchdata if the dashboard was being launched, this caused XGetLaunchInfo to fail.

Rearranging the code a little in XLaunchNewImage fixes this, and Innocent Tears works as normal. I'll merge these pull-requests and then add another commit to fix this

PatrickvL:
Wow, you fixed it! Much appreciated!

PR #234, opened 2017-03-14
Update project to VS 2017

LukeUsher:
It's always good to keep an up-to-date developer environment.

Additionally, VS 2017 now has a smaller installation footprint than VS2015, and the C++ components can be installed standalone, should be easier for devs to get setup with this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,8 @@ init:
 branches:
   only:
   - master
+image: 
+- Visual Studio 2017  
 configuration:
 - Release
 - Debug
--- a/build/win32/Cxbx.vcxproj
+++ b/build/win32/Cxbx.vcxproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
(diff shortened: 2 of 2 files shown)


LukeUsher:
It seems AppVeyor doesn't support the latest compiler just yet, let's not merge this until after this is implemented

hcorion:
I'm not familiar with appveyor but it looks like you can specify the vs build image: 
See the following appveyor that targets both 2015 and 2017: https://github.com/jagerman/pybind11/blob/91227a8bc3600da14cc090d3c60a4cc5a618460a/.appveyor.yml#L4

JayFoxRox:
I'm not a Windows developer, but wouldn't it make more sense to use something like CMake?
Even on Windows you currently depend on Visual Studio right now I'd assume? Supporting mingw etc. could be interesting and it will overall make it easier to port to other platforms (think `winecc`).

It will also (hopefully) take away the hassle of manually upgrading project files (still won't free you from upgrading code from time to time). You could also generate project files for other IDEs and tools.
(on the other hand, getting CMake to work on Windows can be a huge hassle itself. it's still something you should consider imo)

mirh:
> It seems AppVeyor doesn't support the latest compiler just yet, let's not merge this until after this is implemented

AppVeyor is actually [supporting](https://twitter.com/appveyor/status/839999240763990016) it since like 4 days before your PR. 
You have to [specify](https://www.appveyor.com/docs/build-environment/#choosing-image-for-your-builds) the relative image though, since default is still VS2015 (in the meantime latest Windows SDK moved to 10.0.15063.0 btw)
Alternatively, just put `$(DefaultPlatformToolset)` in PlatformToolset property and call it a day.

LukeUsher:
Cool stuff, I'm new to AppVeyor, this was the first time I used it

PR #239, opened 2017-03-16
XAPI: Patch OutputDebugStringA

LukeUsher:
4x4 Evolution 2 now boots to Intros.
Many other titles are likely affected.

This is required to be patched as the Xbox implementation calls int 2Dh
and int 3, which on the Xbox causes the Kernel to trigger debug output.

On Windows, int 2Dh is skipped over, and we just end up with an
unexpected breakpoint.

### DIFF
--- a/src/CxbxKrnl/EmuXapi.cpp
+++ b/src/CxbxKrnl/EmuXapi.cpp
@@ -1253,3 +1253,15 @@ DWORD WINAPI XTL::EMUPATCH(XMountMURootA)
 
 	RETURN(ERROR_SUCCESS);
 }
+
+// ******************************************************************
+// * patch: OutputDebugStringA
+// ******************************************************************
+VOID WINAPI XTL::EMUPATCH(OutputDebugStringA)
+(
+	IN LPCSTR lpOutputString
+)
+{
+	LOG_FUNC_ONE_ARG(lpOutputString);
+	printf("OutputDebugStringA: %s\n", lpOutputString);
+}
--- a/src/CxbxKrnl/EmuXapi.h
+++ b/src/CxbxKrnl/EmuXapi.h
@@ -750,6 +750,14 @@ BOOL WINAPI EMUPATCH(WriteFileEx)
 	LPOVERLAPPED_COMPLETION_ROUTINE lpCompletionRoutine // completion routine
 );
 
+// ******************************************************************
+// * patch: OutputDebugStringA
+// ******************************************************************
+VOID WINAPI EMUPATCH(OutputDebugStringA)
+(
+	IN LPCSTR lpOutputString
+);
+
 // s+
 /* not necessary?
 // ******************************************************************
--- a/src/CxbxKrnl/HLEDataBase/Xapi.1.0.3911.inl
+++ b/src/CxbxKrnl/HLEDataBase/Xapi.1.0.3911.inl
@@ -156,6 +156,44 @@ OOVPA_NO_XREF(SetThreadPriority, 3911, 10)
 OOVPA_END;
 
 // ******************************************************************
+// * OutputDebugStringA
+// ******************************************************************
+OOVPA_NO_XREF(OutputDebugStringA, 3911, 32)
+		{ 0x00, 0x55 },
+		{ 0x01, 0x8B },
+		{ 0x02, 0xEC },
+		{ 0x03, 0x51 },
+		{ 0x04, 0x51 },
+		{ 0x05, 0x8B },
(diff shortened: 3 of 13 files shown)


PatrickvL:
Can't we use this to our advantage, outputting the text to the VS debugger somehow? Or log from within our exception handler, avoiding a patch + handling all aother occurences of the same int combination?

LukeUsher:
I thought about that, however int 2Dh on Windows is often used to prevent debugging, it behaves differently based on if a debugger is connected or not, so we cannot rely on this behavior.

PatrickvL:
Yeah, but if we catch an INT 3 and check if there's an INT 2Dh before it, we can still recognize and handle it accordingly.

PatrickvL:
PS: What value is put in EAX?

LukeUsher:
eax = 1
ecx = the offset of the string to print

PatrickvL:
Do we know of any other (smaller) test-case?

LukeUsher:
This method of debug output only seems to be used in two places in the retail XDK.
In OutputDebugStringA it appears as
```asm
  lea ecx, [string address]    
  mov eax, 1
  int 2dh  
  int 3
```

and assert():
```asm
mov eax, 5
mov ecx, (string offset)  
int 2dh  
int 3h
```

LukeUsher:
See http://reverseengineering.stackexchange.com/questions/1541/int-2d-anti-forensic-method for why int 2dh makes our life difficult

PatrickvL:
```
bool  EmuX86_Opcode_INT3(LPEXCEPTION_POINTERS e, _DInst& info)
{
	// Is INT 3 prefixed with INT 24h?
	if (((PBYTE)e->ContextRecord->Eip)[-2] == OPCODE_INT_CD)
		if (((PBYTE)e->ContextRecord->Eip)[-1] == 0x2D)
		{
			switch (e->ContextRecord->Eax)
			{
			case DEBUG_PRINT:
				printf(((PANSI_STRING)e->ContextRecord->Ecx)->Buffer);
				return true;
			}
		}

	return false;
}


```

LukeUsher:
If a debugger is attached, int 2dh causes windows to trigger int 3 at that address. so we also need to check if the int 3 instruction WAS int 2dh, or if it is prefixed by int 2dh, then we do a debugprint, otherwise we act as if it's a normal breakpoint.

Also when a debugger is attached, int 3 will cause Visual Studio to catch the breakpoint before our code sees it

PatrickvL:
Also:
```
bool  EmuX86_Opcode_INT(LPEXCEPTION_POINTERS e, _DInst& info)
{
	// TODO : Verify if this is the correct way to check for "0xCD 0x03" (long form of INT 3) in distorm :
	switch (info.flags)
	{
	case 0x03:
		return EmuX86_Opcode_INT3(e, info);
	case 0x2D:
		// Is the next opcode an INT3? 
		if (((PBYTE)e->ContextRecord->Eip)[1] == OPCODE_INT3_CC)
			// Let that one execute
			return true;
	}

	return false;
}
```

PatrickvL:
Can you run with that?

LukeUsher:
This code will never be hit because in EmuException we do this (in order to allow user's to Skip breakpoints):

```C++
	if (e->ExceptionRecord->ExceptionCode == STATUS_BREAKPOINT)
	{
		// notify user
		if (EmuExceptionBreakpointAsk(e))
		{
			// We're allowed to continue :
			g_bEmuException = false;
			return EXCEPTION_CONTINUE_EXECUTION;
		}
	}
```
We need a different approach of handling breakpoints, perhaps doing the above check after EmuX86 has finished processing the exception and if it returns false.

Just to test this, I disabled the above code.

Without Visual Studio attached, however, it does work, if we need to change the if statement to the following:
```C++
	if (((PBYTE)e->ContextRecord->Eip)[-2] == 0xCD || ((PBYTE)e->ContextRecord->Eip)[0] == 0xCD)
		if (((PBYTE)e->ContextRecord->Eip)[-1] == 0x2D || ((PBYTE)e->ContextRecord->Eip)[1] == 0x2D)

```

This is required as Windows sometimes silently converts int 2dh to int3.

HOWEVER:
If Visual Studio is debugging the process, it end up triggering breakpoints, hitting to 'Continue' execution just skips the int 3 instruction so our EmuX86 code never runs. As well as stopping code execution until you manually continue (even if you tick the box to disable this exception type), it also prevents the output from being displayed.

So the result is:
Works if running with no debugger attached to the process
Doesn't work during debugging.

PatrickvL:
Let it be so ;)

PR #254, opened 2017-03-21
MemoryRelatedImprovements

PatrickvL:


LukeUsher:
This doesn't work for me (using the binary from AppVeyor), I get this error:

image

Not sure what the cause is, but it would seem the EXE memory map has changed somehow.

LukeUsher:
I get the same error if I build it myself without going through AppVeyor

PatrickvL:
How bizar! I've been running with that code at least a dozen times​. I'll look into it when I have some time, which might take a few days, so if anyone wants to debug this, be my guest.

LukeUsher:
Found the reason, the if statement seems to be being optimised out, see the following screenshots of both if statements in the disassembly view:

image
image

The second one is missing the cmp instruction

LukeUsher:
Found a solution: Disable optimisations for CxbxKrnlMain();
This won't cause a performance impact as that only sets up the emulation, all other functions are still optimised.

PatrickvL:
Well spotted. Strange, how the optimizer could make such a mistake. Perhaps this is something to submit to the visual studio c++ bug tracker? In any case, thanks for this fix

LukeUsher:
As it's normally impossible to guarantee the address of a variable in memory, I expect the optimiser makes this assumption when processing the if statement. From the compilers point of view, the location of memory not allocated via VirtualAlloc (with an input address) cannot be guaranteed, so a check on the offset would usually be false.

Not sure this would qualify as a bug as we are relying on such undefined behavior

PR #260, opened 2017-03-22
Fixed a bug in EmuIDirectSoundBuffer_SetNotificationPositions.

jagotu:
The QueryInterface was called with *ppvObject instead of **ppvObject.

This makes the call to SetNotificationPositions actually succeed, so for example the XBOX Dashboard logs are now completely free of "EmuWarn: Could not create notification interface!"


PatrickvL:
Thanks!

PR #263, opened 2017-03-23
Avoid including version.h everywhere. Use multiprocessor build.

hrydgard:
Version.h was included everywhere through Cxbx.h, and it's updated every build. This leads to every file being recompiled on every build which got to kill your workflow.

So moved that to CxbxVersion.h, which is now included only in four files. This does not make things perfect, but way, way faster at least.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@
 #Documentation output
 doc/doxygen/html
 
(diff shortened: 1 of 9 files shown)


PatrickvL:
Thanks!

PR #286, opened 2017-03-27
Verified OOVPAs

jarupxx:
Improved the warning "Found OOVPAs on other address than in XRefDataBase!"


LukeUsher:
You're an OOVPA wizard!

PR #287, opened 2017-03-27
Derive Xrefs

PatrickvL:
This enables deriving (directly addressed) Xrefs from succesfully located functions, which can help hardening our OOVPA's even further than previously possible. A few examples are included. Once these are tested and deemed correct, the code locating EmuD3DDeferredRenderState and EmuD3DDeferredTextureState can be cleaned up a little.


src/CxbxKrnl/OOVPA.h
```diff
@@ -99,6 +99,7 @@ const uint08 XRefZero = (uint08)0;
 // This XRefOne constant, when set in the OOVPA.XRefCount field,
 // indicates the OOVPA contains one (1) {offset, XREF_* enum} pair.
 const uint08 XRefOne = (uint08)1;
+const uint08 XRefTwo = (uint08)2;
```
LukeUsher: Why XRefTwo rather than a macro such as "XRefs(2)" that could be used with any number of xrefs?

src/CxbxKrnl/HLEDataBase/D3D8.1.0.3925.inl
```diff
@@ -1059,7 +1059,13 @@ OOVPA_END;
 // ******************************************************************
 // * D3DDevice_SetTextureState_TexCoordIndex
 // ******************************************************************
-OOVPA_NO_XREF(D3DDevice_SetTextureState_TexCoordIndex, 3925, 11)
+OOVPA_XREF(D3DDevice_SetTextureState_TexCoordIndex, 3925, 1+11,
```
LukeUsher: Couldn't we macro this to something more readable? 

Eg.
OOVPA_XREF(D3DDevice_SetTextureState_TexCoordIndex, 3925, XRefs(1), 11, 
or
OOVPA_XREF(D3DDevice_SetTextureState_TexCoordIndex, 3925, XRefs(1), Pairs(11),

src/CxbxKrnl/OOVPA.h
PatrickvL: I added and used an XRefTwo symbol, so all use-cases can quickly be found. But XRefs(2) could also work, or XRefCount(2)

src/CxbxKrnl/HLEDataBase/D3D8.1.0.3925.inl
PatrickvL: The last one is a good suggestion, but would need a sizeable rewording of all OOVPA_XREF occurrences

LukeUsher:
I've tested 8 titles so far across various XDKs and not seen a change of behavior, or any warnings about the address differing, so it seems good to me

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -343,40 +348,81 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
...
-                                patchOffset = 162*4 - 92*4;
-                            }
-
-                            XRefDataBase[XREF_D3DDEVICE]                   = *(DWORD*)((DWORD)pFunc + 0x03);
+								if (BuildVersion == 3925)
+								{
+									DerivedAddr_D3DRS_CULLMODE = *(xbaddr*)(pFunc + 0x25);
+									Decrement = 0x1FC;  // TODO: Clean up (?)
+									Increment = 82 * 4;
+									patchOffset = 142 * 4; // TODO: Verify
```
jarupxx: I have tested with various XDK versions. In many versions it worked, but regression was seen in version 3925.
Would we like this?

Decrement = 0x204;
Increment = 84 * 4
patchOffset = 142 * 4;

PatrickvL: comparing this to the old code, there's indeed a difference in the outcome (how quickly such a mistake can be made!) - sorry for that. your second suggestion would match the old code best, I'll fix that ASAP. thanks for testing!!!

PatrickvL:
@jarupxx : I fixed the bug you found; which 3925 title(s) did you use to test with?

jarupxx:
@PatrickvL I tested it with 3925 titles. It will be scanned very accurately!

LukeUsher:
Confirmed workin in 3911/3925 using 4x4 Evolution

PatrickvL:
Yay!

Issue #285, opened 2017-03-26
Blinx - The Time Sweeper crashes due to UpdateOverlay bugs


jarupxx:
This function seems to crash during full color conversion if ImageSize is large.
Parameters as

Unreal Championship Passed
dwImageSize = 614400, g_dwOverlayP = 1280, g_dwOverlayH = 480
Rayman Arena Passed
dwImageSize = 614400, g_dwOverlayP = 1280, g_dwOverlayH = 480
Blinx - The Time Sweeper Crashed! (issue #235)
dwImageSize = 1048576, g_dwOverlayP = 2048, g_dwOverlayH = 512

jarupxx:
Change Video Resolution to 1280 x 960 32 bit, the movie play.
![blinx](https://cloud.githubusercontent.com/assets/20138367/24350899/e75c3eb8-131e-11e7-80f2-3451a51dca89.jpg)

PatrickvL:
Thanks for testing - good to see it helped;

bhodg22:
I still just get a black screen on Blinx but thats most likely due to GPU

PatrickvL:
@bhodg22 If your GPU supports native yuv overlays, then maybe that's the cause of black screens while playing intro movies. We'll sort it out eventually

Issue #89, opened 2017-01-10
Define coding guidelines


PatrickvL:
Instead of keeping around various issues on the preferred way of coding on cxbx, we should instead write up (and keep up to date) a coding guidelines document which stipulates and explains all those details

x1nixmzeng:
Bit off-topic but parts of the source could do with modernising - there is a mixed use of standard library (`std::string`) and C-style allocations.

So firstly - do we want to keep/enforce the Cxbx ASCII art? 😄

LukeUsher:
I definitely want to modernise the source code, the C++ standard library is my preferred option.
Most of the issues stem with the Cxbx codebase being so old.

So let's start with the ASCII art, I think we can drop it because it doesn't serve a purpose, and increases the scroll distance from the top of the file to the main code.

Secondly, what about line-endings? I believe most are Windows style \r\n but this is not enforced, MSVC supports both that and unix style \n. Whatever is decided doesn't matter to me, so long as we have an agreed style. I'm leaning towards \r\n everywhere because this is a Windows application on a Windows development environment, but most projects I have ever worked on have been Unix style line endings.

Thirdly, for control statements and braces, I do have a preferred standard, all control structures should use braces, even if the following statement is a single line. This eliminates the risk of introducing errors by extending the code without adding correct braces, eg:

```c++
if (var == true) 
    var = false;
}
```

rather than 
```c++
if (var == true)
    var = false;
```

Additionally, I prefer for braces to be on a new line for functions, but the same line for control statements, eg:
```c++
int main(int argc, char* argv[])
{
    for(int i = 1; i <= 100; i++) {
        if (i % 2 == 0) {
            printf("Even");
        } else {
            printf("Odd");
        }
    }

    return 0;
}
```

RadWolfie:
I think ASCII art should be kept on main source file, all other source files should not have them. That's my opinion.

I remember there is a way to enforce line-endings with git as a standard for linux, mac, windows clients sending to git server. I had not done it before, might be worthy to do this way though. As in git server always serve line-endings as \r\n and clients can do what they like with their line-endings via git download and sync to server.

I do agree with @LukeUsher for both braces standard on function and single line process.

PatrickvL:
[...]

PR #289, opened 2017-03-28
UpdateOverlay improvements

PatrickvL:
This could be a fix for issue #285

This PR also contains faster (fixed point) decoding of YUY2 to XRGB, and releasing forgotten backbuffer references.

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -366,6 +366,20 @@ VOID CxbxSetPixelContainerHeader
 		;
 }
 
+VOID CxbxReleaseBackBufferLock()
+{
+	XTL::IDirect3DSurface8 *pBackBuffer = nullptr;
+
+	if (D3D_OK == g_pD3DDevice8->GetBackBuffer(0, XTL::D3DBACKBUFFER_TYPE_MONO, &pBackBuffer))
+	{
+		// TODO : assert(pBackBuffer != nullptr);
+
+		pBackBuffer->UnlockRect();
+		pBackBuffer->Release();
+	}
+}
+
+
 // Direct3D initialization (called before emulation begins)
 VOID XTL::EmuD3DInit()
 {
@@ -4273,14 +4287,7 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_Present)
 
 	HRESULT hRet = S_OK;
 	
-	// release back buffer lock
-	{
-		IDirect3DSurface8 *pBackBuffer;
-
-		g_pD3DDevice8->GetBackBuffer(0, D3DBACKBUFFER_TYPE_MONO, &pBackBuffer);
-
-		pBackBuffer->UnlockRect();
-	}
+	CxbxReleaseBackBufferLock();
 
 	// TODO: Make a video option to wait for VBlank before calling Present.
 	// Makes syncing to 30fps easier (which is the native frame rate for Azurik
@@ -4351,14 +4358,7 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_Swap)
     if(Flags != 0)
         EmuWarning("XTL::EmuD3DDevice_Swap: Flags != 0");
 
-    // release back buffer lock
-    {
-        IDirect3DSurface8 *pBackBuffer;
-
-        g_pD3DDevice8->GetBackBuffer(0, D3DBACKBUFFER_TYPE_MONO, &pBackBuffer);
-
-        if(pBackBuffer) pBackBuffer->UnlockRect();
-    }
+	CxbxReleaseBackBufferLock();	
 
 	// TODO: Make a video option to wait for VBlank before calling Present.
 	// Makes syncing to 30fps easier (which is the native frame rate for Azurik
@@ -6204,6 +6204,7 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
 				}
 
 				pBackBuffer->UnlockRect();
+				pBackBuffer->Release();
 			}
 
 			// Update overlay if present was not called since the last call to
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -6010,6 +6010,13 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_EnableOverlay)
     return;
 }
 
+// Based on http://codereview.stackexchange.com/questions/6502/fastest-way-to-clamp-an-integer-to-the-range-0-255
+inline uint08 ClampIntToByte(int x)
+{
+	int r = x > 255 ? 255 : x;
+	return r < 0 ? 0 : r;
+}
+
 // ******************************************************************
 // * patch: D3DDevice_UpdateOverlay
 // ******************************************************************
@@ -6153,42 +6160,44 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
 				// full color conversion (YUY2->XRGB)
 				else
 				{
+					// Based on https://pastebin.com/mDcwqJV3
 					for(uint32 v=0;v<dwImageSize;v+=4)
 					{
-						float Y[2], U, V;
+						const int K1 = int(1.402f * (1 << 16));
+						const int K2 = int(0.714f * (1 << 16));
+						const int K3 = int(0.334f * (1 << 16));
+						const int K4 = int(1.772f * (1 << 16));
 
-						Y[0] = *pCurByte++;
-						U    = *pCurByte++;
-						Y[1] = *pCurByte++;
-						V    = *pCurByte++;
+						uint8_t Y1 = *pCurByte++;
+						uint8_t U  = *pCurByte++;
+						uint8_t Y2 = *pCurByte++;
+						uint8_t V  = *pCurByte++;
 
-						int a=0;
-						for(int x=0;x<2;x++)
-						{
-							float R = Y[a] + 1.402f*(V-128);
-							float G = Y[a] - 0.344f*(U-128) - 0.714f*(V-128);
-							float B = Y[a] + 1.772f*(U-128);
+						int8_t uf = U - 128;
+						int8_t vf = V - 128;
 
-							R = (R < 0) ? 0 : ((R > 255) ? 255 : R);
-							G = (G < 0) ? 0 : ((G > 255) ? 255 : G);
-							B = (B < 0) ? 0 : ((B > 255) ? 255 : B);
+						int R = (K1*vf >> 16);
+						int G = (K2*vf >> 16) + (K3*uf >> 16);
+						int B = (K4*uf >> 16);
 
-							uint32 i = (dy*LockedRectDest.Pitch+(dx+x)*4);
+						uint32 i = (dy * LockedRectDest.Pitch) + (dx * 4);
 
-							pDest[i+0] = (uint08)B;
-							pDest[i+1] = (uint08)G;
-							pDest[i+2] = (uint08)R;
-							pDest[i+3] = 0xFF;
+						pDest[i + 0] = ClampIntToByte((int)Y1 + B);
+						pDest[i + 1] = ClampIntToByte((int)Y1 - G);
+						pDest[i + 2] = ClampIntToByte((int)Y1 + R);
+						pDest[i + 3] = 0xFF;
 
-							a++;
-						}
+						pDest[i + 4] = ClampIntToByte((int)Y2 + B);
+						pDest[i + 5] = ClampIntToByte((int)Y2 - G);
+						pDest[i + 6] = ClampIntToByte((int)Y2 + R);
+						pDest[i + 7] = 0xFF;
 
-						dx+=2;
+						dx += 2;
 
-						if((dx % g_dwOverlayW) == 0)
+						if ((dx % g_dwOverlayW) == 0)
 						{
 							dy++;
-							dx=0;
+							dx = 0;
 						}
 
 					}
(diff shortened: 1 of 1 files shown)


gatesco:
I want to help in this project, from where should I start learning?

JayFoxRox:
@gatesco not commenting offtopic on other peoples PRs is probably a good idea.. The project uses gitter.im for communication, see README.md for details

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -366,6 +366,20 @@ VOID CxbxSetPixelContainerHeader
 		;
 }
 
+VOID CxbxReleaseBackBufferLock()
+{
+	XTL::IDirect3DSurface8 *pBackBuffer = nullptr;
+
+	if (D3D_OK == g_pD3DDevice8->GetBackBuffer(0, XTL::D3DBACKBUFFER_TYPE_MONO, &pBackBuffer))
+	{
+		// TODO : assert(pBackBuffer != nullptr);
```
JayFoxRox: why is this commented out when this is pretty much working code?

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6010,6 +6010,13 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_EnableOverlay)
     return;
 }
 
+// Based on http://codereview.stackexchange.com/questions/6502/fastest-way-to-clamp-an-integer-to-the-range-0-255
+inline uint08 ClampIntToByte(int x)
+{
+	int r = x > 255 ? 255 : x;
+	return r < 0 ? 0 : (uint08)r;
+}
```
JayFoxRox: I think you shouldn't base this on a 6 year old post on the internet for micro-optimizations like this.
I'd assume that most compilers understand the intention by now and optimize properly.
However, if you go through such great length you might want to use an intrinsic or take another approach to optimization in general.

If you absolutely want to make sure that your intent is clear also consider [std::clamp](http://en.cppreference.com/w/cpp/algorithm/clamp) (C++17) or [std::min / std::max](http://en.cppreference.com/w/cpp/algorithm/min) (C++11)

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6042,12 +6049,10 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
 			DDSURFACEDESC2  ddsd2;
 
 			ZeroMemory(&ddsd2, sizeof(ddsd2));
-
 			ddsd2.dwSize = sizeof(ddsd2);
-
 			if(FAILED(g_pDDSOverlay7->Lock(NULL, &ddsd2, DDLOCK_SURFACEMEMORYPTR | DDLOCK_WAIT, NULL)))
 				EmuWarning("Unable to lock overlay surface!");
-
+			else
```
JayFoxRox: imo these things scream `goto fail` (and they were commonplace when cxbx was started - heck, we still do it in Citra :( ). Not sure if you have code guidelines about this / just wanted to express my concerns about this.

I'll always prefer explicit scoping, even for single statements.

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6069,9 +6074,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
 						pSour += g_dwOverlayP;
 					}
 				}
-			}
 
-			g_pDDSOverlay7->Unlock(NULL);
+				g_pDDSOverlay7->Unlock(NULL);
```
JayFoxRox: what version of C++ do you use? `NULL` -> `nullptr` ?

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6114,87 +6118,109 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
...
 			D3DLOCKED_RECT LockedRectDest;
 
-			IDirect3DSurface8 *pBackBuffer=0;
-
+			IDirect3DSurface8 *pBackBuffer = 0;
 			HRESULT hRet = g_pD3DDevice8->GetBackBuffer(0, D3DBACKBUFFER_TYPE_MONO, &pBackBuffer);
-
 			// if we obtained the backbuffer, manually translate the YUY2 into the backbuffer format
-			if(hRet == D3D_OK && pBackBuffer->LockRect(&LockedRectDest, NULL, NULL) == D3D_OK)
+			if (hRet == D3D_OK && pBackBuffer->LockRect(&LockedRectDest, NULL, NULL) == D3D_OK)
```
JayFoxRox: unintentional change? Functions above still don't have a space there.
So I'd suggest doing the style fixes in a bigger clang-format [or w/e] update to not mess with the blame history too much.

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6114,87 +6118,109 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
...
-				uint32 dwImageSize = g_dwOverlayP*g_dwOverlayH;
+				// Limit the width and height of the output to the backbuffer dimensions.
+				// This will (hopefully) prevent exceptions in Blinx - The Time Sweeper 
+				// (see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/285)
+				uint32 W = min(g_dwOverlayW, BackBufferDesc.Width);
+				uint32 H = min(g_dwOverlayH, BackBufferDesc.Height);
 
-				// grayscale
+				// grayscale - TODO : Either remove or make configurable
 				if(false)
```
JayFoxRox: This should probably be pointed out in the PR description?
It looks like this has to be changed to fully test this PR

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6114,87 +6118,109 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
...
 				{
-					for(uint32 y=0;y<g_dwOverlayH;y++)
+					// Clip to backbuffer height :
+					for(uint32 y=0;y<H;y++)
 					{
-						uint32 stop = g_dwOverlayW*4;
+						uint32 stop = g_dwOverlayW *4;
```
JayFoxRox: Another formatting change?
Why not a space behind the `*`

(I still think you should group formatting changes)

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6114,87 +6118,109 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
...
-							R = (R < 0) ? 0 : ((R > 255) ? 255 : R);
-							G = (G < 0) ? 0 : ((G > 255) ? 255 : G);
-							B = (B < 0) ? 0 : ((B > 255) ? 255 : B);
+							int nCb = Cb - 128;
+							int nCr = Cr - 128;
 
-							uint32 i = (dy*LockedRectDest.Pitch+(dx+x)*4);
+							int Rd =  (K1 * nCr) >> 16;
+							int Gd = ((K2 * nCb) + (K3 * nCr)) >> 16;
+							int Bd =  (K4 * nCb) >> 16;
```
JayFoxRox: I feel these code blocks should be lambdas or functions, K1 - K4 local variables / constants.
The nesting in these files gets kind of insane and the code isn't too descriptive either.

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -366,6 +366,20 @@ VOID CxbxSetPixelContainerHeader
 		;
 }
 
+VOID CxbxReleaseBackBufferLock()
```
JayFoxRox: Why is this called "Lock" but it does `..->UnlockRect` ?

src/CxbxKrnl/EmuD3D8.cpp
hcorion: I wouldn't be so sure.
The clamp mentioned at the top of the article.
```asm
        mov     eax, edi
        not     eax
        sar     eax, 31
        and     edi, eax
        mov     eax, 255
        sub     eax, edi
        sar     eax, 31
        or      eax, edi
        ret
```
The clamp used here. (the recommended faster version)
```asm
        cmp     edi, 255
        mov     eax, 255
        mov     edx, 0
        cmovle  eax, edi
        test    eax, eax
        cmovs   eax, edx
        ret
```
Check it out here: https://godbolt.org/g/9jsGj9

JayFoxRox: my version:

```cpp
uint8_t Clamp3(int n)
{
    return std::min(std::max(n, 0), 255);
}
```

Literally gives the same code:

```
cmp     edi, 255
mov     eax, 255
mov     edx, 0
cmovle  eax, edi
test    eax, eax
cmovs   eax, edx
ret
```

And yet mine is a lot more readable. It also shows intent and in the future this might actually compile to MIN / MAX instructions or CLAMP.

So, my argument was that these optimizations are not useful. Especially if there are language features to do it. Even *IF* this optimization did much, I doubt this single microoptimization (cutting a handful of instructions) helps too much. For example: getting better cache coherence would be a way better optimization target.

(Strangely the `std::clamp` does not work yet on that website [despite having been in 3.9 clang c++ stdlib](https://libcxx.llvm.org/cxx1z_status.html); same for gcc and icc even with `-std=c++1z` - not sure why that is)

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: I haven't gotten around to get assert() compiling - it requires a namespace prefix and I don't know which yet

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: thanks for these suggestions, it's obvious that i (and this project as a whole) still have to learn to apply a few of the newer language features

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: when changing code I'm trying to use nullptr for all things happening on the host side and NULL for all things regarding the Xbox side - that's not a common guideline yet, but it helps keeping a distinction between the two. in this case it should indeed have been nullptr

LukeUsher: I think nullptr should be used everywhere

src/CxbxKrnl/EmuD3D8.cpp
LukeUsher: That's odd, I'm sure assert() should be global scope. Make sure you #include <cassert>

If we require C libraries (cstdio, cstdlib, cassert), the C++ versions should be preferred over the C versions (stdio.h, stdlib.h, assert.h)

src/CxbxKrnl/EmuD3D8.cpp
LukeUsher: Surprisingly, I wasn't aware Lambda expressions existed in C++, that's good to know.

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: I'm from a school that teached me "goto is evil", hence this construct. But if we decide that "goto fail" is the best approach on these type of scenarios, then sure, let's do that

hcorion: image

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: I don't. Look at the openxdk xboxkrnl.h - it specifies NULL. So it would be consistent to use that for all data that's going to be read by Xbox code, like resources, data points, etc. The "nullptr" keyword on the other hand, is more modern, so would be quite applicable to all code running on the outside of the emulator.

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: visual studio automatically updates indents and spacing when modify statement(block)s - we could look into a setting that will make VS use our coding style (once we know what that should be)

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: the grayscale path was already present, I only added a comment indicating that we should do with it

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: I'm curious, how would you write something like this?

I completely agree on the crazy indenting

PatrickvL:
@gatesco Join the chat: https://gitter.im/Cxbx-Reloaded/Lobby

src/CxbxKrnl/EmuD3D8.cpp
JayFoxRox: @PatrickvL `goto fail` refers to https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/ where they didn't use explicit braces in a case similar to this. I feel it's very easy to add another line above the `else` and not realize that it actually corrupts the control flow. Hence I usually prefer explicit scopes.

I'm by no means saying you should add a goto [you should not!].

Sorry for being unclear.

src/CxbxKrnl/EmuD3D8.cpp
JayFoxRox: yes, but it appears as if your newly added code is disabled by default. So if somebody was to test this PR, they'd not even run all changes. So this increases the risk of having strange errors later (when somebody turns this `false` into `true` or an actual condition) which are extremly hard to bisect.

Usually reviewers are lazy and won't check the design, so they will nit about code style and form, then run the code and see if it causes regressions or crashes. If it doesn't they'll consider it fine - so not mentioning this before review is dangerous, even people who review code might not be aware of the consequences of this line.

LukeUsher:
I'll merge this now, but this does highlight the need for fleshing out a style guide. See Issue https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/89

PR #291, opened 2017-03-29
Improve D3DDevice_SetLight_5558 oovpa

phire:
Make it more generic, so it can detect SetLight in Star Wars: KotOR, which is a 5455 game.

### DIFF
--- a/src/CxbxKrnl/HLEDataBase/D3D8.1.0.5558.inl
+++ b/src/CxbxKrnl/HLEDataBase/D3D8.1.0.5558.inl
@@ -291,16 +291,34 @@ OOVPA_END;
 // ******************************************************************
 // * D3DDevice_SetLight
 // ******************************************************************
-OOVPA_NO_XREF(D3DDevice_SetLight, 5558, 8)
+OOVPA_NO_XREF(D3DDevice_SetLight, 5558, 15)
 
-        { 0x1E, 0x00 },
-        { 0x3E, 0x89 },
-        { 0x5E, 0x00 },
-        { 0x7E, 0x07 },
-        { 0x9E, 0xC8 },
-        { 0xBE, 0xC9 },
-        { 0xDE, 0x07 },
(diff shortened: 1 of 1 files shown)


LukeUsher:
This same symbol is reused in the 5849 database too, can you verify that it still gets matched in all three of these XDKs (5455, 5558, 5849)?

jackchentwkh:
I think the version based API detection should be implemented in order to
prevent the conflict/regression between different XDK titles.
make the OOVPAs per XDK version and make each version a complete set, then
scan the xbe per the lib version info in the xbe header.
this shall improve the accuracy of API scanning, and also ease the possible
regression when we modified one OOVPA for one XDK version.

thanks and regards!

Jack Chen
Substrate sales and tech. service
YXsemicon Ltd.
Cell: +886-933311773
Office:+886-7-7536832
Fax:+886-7-7431976

2017-03-29 17:08 GMT+08:00 Luke Usher <notifications@github.com>:

LukeUsher:
We do already use version based scanning, but when a function has either not changed, or has changed very little between XDK versions, we share a symbol. The same when two XDK versions are 99% identical.

PatrickvL:
I'd rather combine all OOVPA databases into one per library, registering all versions of a function in one go, using a macro with a variable number of arguments.
When that's in place, we can derive a subset to scan with at runtime, trailored to the currently emulated version

PatrickvL:
As it turned out, the order of the OOVPA searches matter, so this consolidated database mustn't be sorted though

LukeUsher:
JayFoxRox and I have been discussing the possibility of splitting out the HLE Database into a separate library that can be linked into other projects, this could help generate useful stack traces for XQEMU, as well as facilitate development of reverse engineering tools using our symbol database. 

Just something to keep in mind

phire:
I've confirmed that this matches 5849 too.

LukeUsher:
Thanks!

Issue #283, opened 2017-03-26
MemoryManager bugs


LukeUsher:
Patrick has identified some bugs with my new MemoryManager implementation, specifically the allocation of contiguous memory. These shouldn't cause any issues currently, but will need to be fixed before the removal of more D3D HLE patches.

* ~~First, the size argument shouldn't be increased to cater for alignment, asthat would reflect back when a query is done~~ :checkered_flag: Fixed by PR #293 

* ~~Next, the alignment argument should be a power of two, and should reset to one 4kb page if anything smaller is specified~~ (All callers already do this.)

* ~~Then, alignment should be verified per potential range. If the alignment isn't correct at first, check if the range allows for a correction - meaning allocating more memory, and returning the correctly aligned address within that range. If neither is possible, try another range.~~ :checkered_flag: Fixed by PR #293 

Be aware of that offset when freeing previously aligned contiguous memory - the actual allocated range might be larger due to upward alignment, compared to the requested size, plus, it's starting address might be lower than the returned address

EDIT :  
* ~~The current implementation is lacking a feature, namely the ability to query within (but not at the start of) a pre-allocated memory block.~~ :checkered_flag: Fixed by PR #301 
* Support has to be added for the limiting arguments to return address lower and upper bound

PatrickvL:
A comment (https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/236#issuecomment-290345274) shows that one of those issues is hit by Rayman Arena (which makes this a good test-case)

LukeUsher:
Thanks, I've noticed a few other titles hitting it too.

Note that it could also be due to missing a call to MemoryManager::Allocate somewhere else too. Eg: I think there are a few places that allocate memory with `new int[n];` which should also be updated to use MemoryManager.

PatrickvL:
@LukeUsher With the latest bugfix in the memory manager, all the above mentioned issues (and more) is finished. Could you verify the last remark (regarding incorrectly using new, and possibly malloc, for Xbox allocations), and close this issue if that turns out to be all right now?

PatrickvL:
Since there's now a new memory manager, this issue can be closed.

PR #294, opened 2017-03-30
Quick-fix

PatrickvL:
This will only affect titles using palette's, like Battlestar Galactica


PatrickvL:
See https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/283#issuecomment-290355851

PR #301, opened 2017-03-31
MemoryManagerQueryFix

PatrickvL:


src/CxbxKrnl/MemoryManager.cpp
```diff
@@ -54,6 +54,25 @@ MemoryManager::~MemoryManager()
 	DeleteCriticalSection(&m_CriticalSection);
 }
 
+TypedMemoryBlock *MemoryManager::FindContainingTypedMemoryBlock(void* addr)
+{
+	if (m_MemoryBlockInfo.empty())
+		return nullptr;
+
+	// Find the first block whose address is not less than the requested address
+	std::map<void *, TypedMemoryBlock>::iterator low = m_MemoryBlockInfo.lower_bound(addr);
```
LukeUsher: I'd suggest using the auto keyword here to simplify the declaration of the iterator.
```C++
auto low = m_MemoryBlockInfo.lower_bound(addr)
```

src/CxbxKrnl/MemoryManager.cpp
```diff
@@ -54,6 +54,25 @@ MemoryManager::~MemoryManager()
...
+	if (low == m_MemoryBlockInfo.end())
+		// If there's no such block, it might fall into the last block
+		low = --m_MemoryBlockInfo.end();
+
+	// Check if the requested address lies inside this block
+	TypedMemoryBlock *info = &(low->second);
+	if ((size_t)addr >= ((size_t)info->block.addr + info->block.size))
```
LukeUsher: Could you please use braces, even around single statement blocks, this reduces the chance of bugs if the code is extended by somebody who neglects to add them.

src/CxbxKrnl/MemoryManager.cpp
PatrickvL: Quite right - I had that there, but made it more explicit during debugging. Well spotted.

src/CxbxKrnl/MemoryManager.cpp
PatrickvL: Where would you like to put them? The required braces are already in place...

PR #305, opened 2017-03-31
Speedup OOVPA Detection by only scanning Executable sections

LukeUsher:


LukeUsher:
While the idea is good, this is sadly not always the case. I have seen XAPI functions get linked in the .text section in the past. 

On the OOVPA note, this was just a quick optimisation to stop large (50mb) xbes from taking forever to scan. I'm working on a refactor of the scanning code anyway, so it's likely we'll see some improvement once I'm ready with that

Issue #288, opened 2017-03-28
Implement D3DDevice_GetLight


phire:
Needed by Star Wars: KotOR

~~Note: Currently Cxbx-Reloaded fails on a D3DDevice_SetLight call which it fails to detect. Here is a temporary patch which gets past that point:~~

Edit: Removed the SetLight patch, which is no-longer needed to reproduce this bug.

PatrickvL:
Did https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/291 or https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/292 fix this? If so, then please close this issue.

phire:
No, They both fix the detection of `SetLight`. This bug is about the corresponding `GetLight` function

PatrickvL:
Ah, you're right, thanks. Are you still looking into this issue?

LukeUsher:
We determined that it requires a new HLE Patch, we do not have an implementation for GetLight just yet, but it should be a quick fix

jarupxx:
I implemented PR #319 GetLight function and some missing OOVPAs  added in PR #318.
Testing with these two PRs gets a sound on a black screen. I think it was improved a little.

PatrickvL:
@phire Could you test PR #319 with KoTOR?

PatrickvL:
@jarupxx Thanks for this! Let's hope it works without causing any regressions.

phire:
Yep. KotOR gets a lot further now. Doesn't crash and plays some sound, though no graphics.

PR #319, opened 2017-04-03
Implemented D3DDevice_GetLight function

jarupxx:
I back-ported GetLight function. 
Star Wars: KotOR #288 no longer crashes, but still it is a black screen, I do not know whether it could be implemented correctly...


bhodg22:
Sometimes a black screen can be a result depending on what GPU you are using

PR #324, opened 2017-04-05
Revert D3DFormat Conversion Lookup Table

LukeUsher:
This was causing breakage in some titles. A decision has been made to revert this change until we can figure out exactly what is failing and why.


PatrickvL:
Let's give this one last try; it'd be worthwhile to keep this new code, if we could just get it to return identical output as before

PatrickvL:
Luckily, this isn't needed anymore

PR #326, opened 2017-04-06
Safe D3DFormat updates, plus component encoding info

PatrickvL:
For some D3D formats, this commit makes `EmuXB2PC_D3DFormat()` return not `D3DFMT_UNKNOWN`, but a better suited format. (Either formats that were already returned from other format-conversions, or identical native formats.)

Component encoding info can be used in the future to convert pixels into ARGB D3D colors, for all (color-based) formats not natively supported.


LukeUsher:
Good work as always Patrick!

PR #348, opened 2017-04-10
Conversion of unsupported native formats to ARGB

PatrickvL:


LukeUsher:
Potentially risky but important for future development.

I'll merge and we can deal with any breakage as separate issues.

PR #356, opened 2017-04-11
Alloc zeroed

LukeUsher:


PatrickvL:
Thanks (also @blueshogun96 for the suggestion in #356, which can be closed with this PR)

PR #357, opened 2017-04-11
Color component fixes and FormatInfo code cleanup

PatrickvL:


import/DirectX8/include/d3d8types.h
```diff
@@ -1240,6 +1240,14 @@ typedef enum _D3DFORMAT
     D3DFMT_A8                   = 28,
     D3DFMT_A8R3G3B2             = 29,
     D3DFMT_X4R4G4B4             = 30,
+#if 0 // Direct3D 9
```
LukeUsher: It would be best to include these definitions elsewhere, or `#include<d3d9types.h>` instead of D3D8Types.h. This is because the DirectX8 folder is unmodified Microsoft Code from the DX8 SDK, and it would be best to keep it that way.

d3d9types.h is is included with the standard Windows SDK installed with Visual Studio 2015, so it should be available to use.

PatrickvL: Oh, I didn't know that, thanks for the heads up! This commit should be removed from the PR then...

PR #364, opened 2017-04-12
Prevent infinite KeBugCheck loops by addng an Ignored flag

LukeUsher:


PatrickvL:
I merged this blindly - what test case did you need this for?

LukeUsher:
Some titles are able to progress after calling KeBugCheck, and they did before we (correctly) emulated this function by not returning/ending execution.

This is a hack allowing those titles to still progress, rather than be spammed with KeBugCheck dialogs, ignoring the first one will stop further dialogs appearing. One of the effected titles is Bufffy The Vampire Slayer.

PatrickvL:
Thanks for explaining that. One last question about this: Would an "ignore once" button still be useful?

PR #365, opened 2017-04-13
Stretch movies

PatrickvL:
Ported back of Dxbx. Includes libyuv PR, so merge that first.

PR #367, opened 2017-04-13
Use libyuv for YUY2ToARGB conversion

PatrickvL:


PatrickvL:
@bhodg22 Commented that this PR breaks video's in NamcoMuseum - see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/365#issuecomment-293865901 and https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/365#issuecomment-293903487

PatrickvL:
@bhodg22 What happens when you change this line :
>			YUY2ToARGB(pbSource, g_dwOverlayP, pbDest, BackBufferDesc.Width * 4, W, H);
into
>			YUY2ToARGB(pbSource, g_dwOverlayP, pbDest, LockedRectDest.Pitch, W, H);

src/CxbxKrnl/libyuv_extract.cpp
```diff
@@ -0,0 +1,226 @@
+// This is an open source non-commercial project. Dear PVS-Studio, please check it.
+// PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com
+
+#include "libyuv_extract.h"
+
+// The following code is an extract of libyuv to keep code size low (this may be revisited later).
+// Source : https://github.com/lemenkov/libyuv/commit/7e936044d154b9fe159a67f9562e10b1ef1cb590
```
JayFoxRox: Why is this part of CxbxKrnl then and not some common folder?
Also maybe just use a submodule?

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -331,7 +330,6 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
                     {
 						// Save D3D8 build version
 						g_BuildVersion = BuildVersion;
-						g_OrigBuildVersion = OrigBuildVersion;
```
JayFoxRox: Changes not related to what's mentioned in the PR?
(Other instances too)

PatrickvL: yes, indeed, but in a separate commit, which we don't squash into one when merging, so bisects will still see these changes as separate commits

src/CxbxKrnl/libyuv_extract.cpp
PatrickvL: I tried to add this in a small, quick commit. there's not much more we can use from libyuv anyway at the moment, so I opted for this solution. feel free to improve it

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6194,89 +6186,8 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
...
-							// Clip to backbuffer height :
-							if (dy >= H)
-								break;
-
-							dx = 0;
-						}
-
-					}
-				}
+				YUY2ToARGB(pbSource, g_dwOverlayP, pbDest, BackBufferDesc.Width * 4, W, H);
```
JayFoxRox: ```YUY2ToARGB(pbSource, g_dwOverlayP, pbDest, LockedRectDest.Pitch, W, H);```

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -6194,89 +6186,8 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_UpdateOverlay)
 				uint32 W = min(g_dwOverlayW, BackBufferDesc.Width);
 				uint32 H = min(g_dwOverlayH, BackBufferDesc.Height);
```
JayFoxRox: I know this is beyond the scope of the PR, but shouldn't this respect the size of the locked rect instead the backbuffer?

PatrickvL:
Well spotted! After getting back a report on garbled movie playback, I suggested the same thing this morning. I've yet to hear the results.

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: Well spotted! After getting back a report on garbled movie playback, I suggested the same thing this morning. I've yet to hear the results.

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: I think you're right. The current call to min() is/was a precaution to prevent buffer overruns. I've got another PR pending which stretches the video to the desired rectangle. That should prevent most of the overruns, but even then you're suggesting would still be wise as a precaution for corner cases

src/CxbxKrnl/EmuD3D8.cpp
bhodg22: Sorry i didnt reply earlier, it was midnight where I was, and I am currently away from my main desktop and can't find a copy of your build, do you think you could send one through to me?

bhodg22:
Sorry i didnt reply earlier, it was midnight where I was, and I am currently away from my main desktop and can't find a copy of your build, do you think you could send one through to me?

src/CxbxKrnl/EmuD3D8.cpp
JayFoxRox: gitgudatgithub

src/CxbxKrnl/EmuD3D8.cpp
JayFoxRox: @PatrickvL gitgudatgithub

(Sorry for nagging, but this really bothers me a lot - it's impossible to read the conversations in other clients and even on the GitHub discussion thread [where most review should happen])

PatrickvL: I've opened an issue for the Reply bug in OctoDroid here : https://github.com/slapperwan/gh4a/issues/582

PR #370, opened 2017-04-14
Verified the LoadVertexShaderProgram (4627-5344)

jarupxx:
I changed it with commit 108228e75a7c2b73654ad392581ed4aa4a6fb11d, but length was different between 5344 and 4627, so retn 8 was removed.
Source: Terminator - Dawn of Fate


blueshogun96:
Who's actually going to "implement" it?  Games like Azurik need this.

Shogun.

PR #374, opened 2017-04-17
Update EmuVerifyResourceIsRegistered to no longer rely on pData being 0

LukeUsher:
Unpatched Create Resource functions can set the pData value to non-zero values, causing EmuVerfiyResourceIsRegistered assuming the registration has already happened.

Should improve stability. (JSRF shows better results on my machine with this)

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -113,7 +113,7 @@ static int                          g_iWireframe    = 0;
 extern uint32						g_BuildVersion;
 
 // resource caching for _Register
-static XTL::X_D3DResource pCache[16] = {0};
+std::vector<DWORD> g_RegisteredResources;
 
 // current active index buffer
 static XTL::X_D3DIndexBuffer       *g_pIndexBuffer  = NULL; // current active index buffer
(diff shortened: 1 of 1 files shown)


src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -1226,34 +1226,15 @@ static void EmuVerifyResourceIsRegistered(XTL::X_D3DResource *pResource)
     if((pResource->Data & X_D3DRESOURCE_DATA_FLAG_SPECIAL) == X_D3DRESOURCE_DATA_FLAG_SPECIAL)
         return;
 
-    int v=0;
-
-    for(v=0;v<16;v++)
-    {
-        if(pCache[v].Data == pResource->Data && pResource->Data != 0)
-        {
-            pResource->EmuResource8 = pCache[v].EmuResource8;
```
PatrickvL: removing this assignment looses the caching functionality, was that intended?

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -1226,34 +1226,15 @@ static void EmuVerifyResourceIsRegistered(XTL::X_D3DResource *pResource)
...
-        for(v=0;v<16;v++)
-        {
-            if(pCache[v].Data == 0)
-            {
-                pCache[v].Data = pResource->Data;
-                pCache[v].EmuResource8 = pResource->EmuResource8;
-                break;
-            }
-
-            if(v == 16)
```
PatrickvL: removing this limitation could lead to uncontrolled growth of the registered resources vector - is that intended?

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -113,7 +113,7 @@ static int                          g_iWireframe    = 0;
 extern uint32						g_BuildVersion;
 
 // resource caching for _Register
-static XTL::X_D3DResource pCache[16] = {0};
+std::vector<DWORD> g_RegisteredResources;
```
PatrickvL: is this type threads safe, or does thread safe usage require a lock?

PatrickvL:
Let's address thread-safety later, this is already quite an improvement - thanks!

PR #389, opened 2017-04-21
Fix HLE Cache

LukeUsher:
DeferredRenderState and DeferredTextureState are now stored and processed from the cache 
g_BuildVersion is now saved/loaded from the cache

A menu option has been added to allow the cache to be cleared manually


src/Cxbx/WndMain.cpp
```diff
@@ -867,6 +867,28 @@ LRESULT CALLBACK WndMain::WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lP
                     ShowVideoConfig(hwnd);
                     break;
 
+				case ID_CACHE_CLEARHLECACHE:
+				{
+					std::string cacheDir = std::string(XTL::szFolder_CxbxReloadedData) + "\\HLECache\\";
+					std::string fullpath = cacheDir + "*.*";
```
LukeUsher: I figured the only contents of the HLECache folder will be HLE Cache related, which is why I clear the whole folder contents

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -77,10 +77,17 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
 	printf("*******************************************************************************\n");
 	printf("\n");
 
+	// Make sure the HLE Cache directory exists
+	std::string cachePath = std::string(szFolder_CxbxReloadedData) + "\\HLECache\\";
```
LukeUsher: Yeah that's a good idea

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -122,6 +121,28 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
 				}
 			}
 
+
+			// Fix up Render state and Texture States
+			if (g_HLECache.find("D3DDeferredRenderState") == g_HLECache.end()) {
```
LukeUsher: For now this serves the purpose of caching patch locations, in the future I have plans to extend this to a more global cache mechanism, including FS and media patches too.

Caching X-Refs is a possibility, I do agree we should have a full map of symbols in the cache, I left that until later (when I fully separate symbol scanning and patching)

It was an intentional decision to skip the scanning code if the cache is used. As the cache gets regenerated if the HLEDatabase changes, adding OOVPAs will cause the HLE Cache to be regenerated anyway so that shouldn't be a problem.

src/Cxbx/WndMain.cpp
PatrickvL: I've seen entire volumes been erased, losing weeks of work, because of constructs like this. I'd err on the safe side with deletes...

src/CxbxKrnl/HLEIntercept.cpp
PatrickvL: You shouldn't view the HLECache as a real cache, but rather as a symbol map, like you called it.
The actual caching happens when this map is written to file and read back again later on.

If the HLECache were renamed into SymbolMap, it would already be a step in this direction.
Following that, The Function SymbolMap could be used instead of the XRefDataBase.
Etc etc.

Anyway, do as you please. This was just a suggestion. Great work all around anyway!

src/Cxbx/WndMain.cpp
LukeUsher: Okay fair enough. 

This cannot happen so long as szFolder_CxbxReloadedData is set correctly, however there is the (very slim, but very real) possibility that a bug could wipe out that variable, so I'll amend it

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -77,22 +77,28 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
...
 
+	// Make sure the HLE Cache directory exists
+	std::string cachePath = std::string(szFolder_CxbxReloadedData) + "\\HLECache\\";
+	int result = SHCreateDirectoryEx(nullptr, cachePath.c_str(), nullptr);
+	if ((result != ERROR_SUCCESS) && (result != ERROR_ALREADY_EXISTS)) {
+		CxbxKrnlCleanup("Couldn't create Cxbx-Reloaded HLECache folder!");
+	}
+
 	// Hash the loaded XBE's header, use it as a filename
 	uint32_t uiHash = XXHash32::hash((void*)&CxbxKrnl_Xbe->m_Header, sizeof(Xbe::Header), 0);
```
donwayo: Could SHA-1 be used instead, kind of like how the XBE signature is calculated? 

`SHA1(length(HEADER) + HEADER)`

Where HEADER is actually all the data in the header except the magic XBEH and the signature (offset 260).

LukeUsher: Not a problem, I'm always open for suggestions and the ability to reuse code is always useful to have. I just went with XXHash32 because it was already present in the codebase, it is simple to use, and we already use it for hashing resource data.

donwayo: SHA-1 is also part of the codebase: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/master/src/CxbxKrnl/EmuSha.h

LukeUsher: I know, I implemented that ;)

LukeUsher:
I have done enough testing that I am happy to merge this, I will address any regressions that come up quickly, if there are any I missed.

This was the first step towards separating Symbol Scanning and HLE Patching, something I'll continue to work towards and create another pull request for when complete.

PR #390, opened 2017-04-21
Patrickvl improvements

LukeUsher:
Batch of improvements by PatrickvL

There's still more to do, but these are the ones I have testes so far.

At least JSRF behaves better now, the character models are much less likely to become distorted.


JayFoxRox:
At some point you should start caring about your history, or you'll have a very very hard time bisecting issues.
So by not taking time to split this kind of PR into smaller chunks, you are probably causing a ton of trouble in the future (This also goes for temporary hacks or inaccuracy - all of that should be documented)

PR #394, opened 2017-04-23
Implement AvGetSavedDataAddress & AvSetSavedDataAddress

LukeUsher:
The existing implementation was overkill, as all these functions do on Xbox is get/set a single pointer value. The actual surface creation is handled by D3D in D3DDevice_PersistDisplay.


blueshogun96:
This should be really helpful for Unreal Championship.

Shogun.

PR #405, opened 2017-04-28
Improved signatures

jarupxx:


PatrickvL:
Sorry this want merged quickly but we were busy with graphical improvements.

Thanks for your hard work!!

jarupxx:
No problem. Graphical improvement is a great work!

PatrickvL:
I did notice there are still a few occurrences of multiple OOVPA versions of the same symbol in the D3D HLE databases. That could lead to trouble, and would best be solved by a refactoring of the HLE scanning code, together with the merge of all OOVPA registrations into one big table.

At some point in time we'll address this, but for now we're focusing on better graphics

PR #419, opened 2017-05-02
Bug fixes for Panzer Dragoon games

blueshogun96:
Added a few necessary fixes for Panzer Dragoon ORTA and PD1.

- Fixed a bug in NtCreateEvent.  The XDK documentation says that the attributes struct isn't supported in the first place for CreateEvent(), so if this function fails, try creating it without the object attributes.
- Quick hack that converts X1R5G5B5 to R5G6B5 since this texture format doesn't seem to work on modern drivers.
- Ensure that WSAStartup is called using a version of Winsock that is guaranteed to be supported.  This fixes an issue with Xbox Dashboard 4920.


LukeUsher:
I'm happy to merge the NtCreateEvent bug and the WSAStartup patch.

Let's wait for Patrick to verify the X1R5G5B5 hack: I know he's been doing significant reworking of the texture code so I'd rather him check that part.

Great work & thanks!

src/CxbxKrnl/EmuKrnlNt.cpp
```diff
@@ -318,7 +318,21 @@ XBSYSAPI EXPORTNUM(189) xboxkrnl::NTSTATUS NTAPI xboxkrnl::NtCreateEvent
 	// that would require us to create the event's kernel object with the Ob* api's too!
 
 	if (FAILED(ret))
-		EmuWarning("NtCreateEvent Failed!");
+	{
+		EmuWarning("Trying fallback (without object attributes)...");
+
+		// If it fails, try again but without the object attributes stucture
```
PatrickvL: Apart from a little typo (stucture -> stRucture), it would be good to known *why* the call fails, as then we might be able to prevent the failure in the first place (that's what the conversion functions are for).

Can you elaborate on this, and/or add a log message stating the problem?

src/CxbxKrnl/EmuXOnline.cpp
```diff
@@ -66,7 +66,10 @@ int WINAPI XTL::EMUPATCH(WSAStartup)
 		LOG_FUNC_ARG(lpWSAData)
 		LOG_FUNC_END;
 
-    int ret = WSAStartup(wVersionRequested, lpWSAData);
+	// Prevent this function from failing by requesting a version of Winsock that
+	// we know for sure is actually supported on Windows.  This also fixes one error
+	// in the 4920 dashboard.
+    int ret = WSAStartup(MAKEWORD(2,2) /*wVersionRequested*/, lpWSAData);
```
PatrickvL: Good find. One thing though : https://msdn.microsoft.com/en-us/library/windows/desktop/ms742213(v=vs.85).aspx mentions a few supported versions (1.0 up to 2.2) - it might be wise to accept any of these if the Xbox requests them, and fallback to 2.2 for anything outside this set. What are your thoughts?

src/CxbxKrnl/EmuD3D8/Convert.cpp
```diff
@@ -271,7 +271,8 @@ BOOL XTL::EmuXBFormatIsSwizzled(X_D3DFORMAT Format)
 
 XTL::D3DFORMAT XTL::EmuXB2PC_D3DFormat(X_D3DFORMAT Format)
 {
-	if (Format <= X_D3DFMT_LIN_R8G8B8A8) {
+	if (Format <= X_D3DFMT_LIN_R8G8B8A8 && Format != -1 /*X_D3DFMT_UNKNOWN*/) // The last bit prevents crashing (Metal Slug 3)
```
PatrickvL: This precaution might not be the only thing we must do - we might need to look into where the erroneous comes from;
So, when Metal Slug 3 sends the -1 Format argument to EmuXB2PC_D3DFormat, where does it get this value from?

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -4720,6 +4728,15 @@ HRESULT WINAPI XTL::EMUPATCH(D3DResource_Register)
                 PCFormat = D3DFMT_R5G6B5;
             }
 
+			// TODO: HACK: Since I have trouble with this texture format on modern hardware,
```
PatrickvL: Cxbx already contains a few more of these hardcoded fallbacks, so it's okay to merge this in.
Ultimately though, we shouldn't fallback until we actually see CreateTexture failing.
If it fails, instead of choosing a somewhat-similair format, it'd be simpler to just fallback straight to ARGB (with an according conversion of course).

This can be done once Xbox resources are left untouched, and rendering is purely done using separate copies of these resources on the host. ( This is something I'm working on in https://github.com/PatrickvL/Cxbx-Reloaded/tree/texture_updates )

src/CxbxKrnl/EmuD3D8/Convert.cpp
blueshogun96: Well, -1 is not really an erroneous value.  Since X_D3DFMT_UNKNOWN's value does not match that of the actual Xbox enum, I simply just used -1 (or 0xFFFFFFFF) catch the literal value of that flag for Xbox.  If we don't, then causes an overflow in the array of texture format information and reads from invalid data.

Shogun

src/CxbxKrnl/EmuXOnline.cpp
blueshogun96: Well, we could do that.  At the time, I just didn't believe it really mattered.  Since I don't remember what the differences between Winsock versions are, that might be a better idea.  At the time, I simply wanted it to work :)

Shogun

src/CxbxKrnl/EmuKrnlNt.cpp
blueshogun96: Okay, I took another look into this, and the exact error is 0xC0000008 (STATUS_INVALID_HANDLE).  Looking at the debug log, the handle that's used for RootAccess is 0xFFFFFFFC, which appears to be an alias for something (I tried to google it but didn't find anything).

The reason I did things this way was because I assumed that this structure data was related to the LPSECURITY_ATTRIBUTES structure somehow.  On Xbox, this is unsupported, so I didn't bother looking further into it and just wrote a quick fallback.  Furthermore, this also seems to fix Celebrity Deathmatch, causing it to go ingame again.  But I could be wrong though.  It's good that you asked for an explanation on this.

Shogun

src/CxbxKrnl/EmuD3D8/Convert.cpp
PatrickvL: Yeah, the origin of the -1 is clear, but the question is: why is it returned? And from where in this particular title? Knowing that can help finding the underlying issue

LukeUsher:
@blueshogun96 You should probably work in other branches: Since you created this pull request against master, anything you commit to your master gets automatically appended to this pull request. I'm not sure that this is what you want?

PR #437, opened 2017-05-04
Added D3DResource_BlockUntilNotBusy to fix Turok

blueshogun96:
Since I was getting several complaints (directed at me) regarding the D3D::BlockOnTime error, I found 4 possible functions that call this:  Lock2D/3DSurface, D3DCubeTexture_LockRect (probably other resource locking functions too) and D3DResource_BlockUntilNotBusy.  The latter is what Turok needed.


PatrickvL:
This pull request contains merge conflicts, probably because your branch wasn't synched with master at the time you pushed it.

The conflict can easily be resolved, but before we do: why did you re-enabled the "int 3"? These seem to cause problems for our testers, so I'd say: leave them commented out, or only compile them in for debug builds. What do you think?

blueshogun96:
Yeah, I forgot to do that since I'm still new to pull requests and stuff.

Re-enabling the int 3 is for the best, simply because in debug builds, I can get the stack trace that way and find out exactly what's calling that function that should not be emulated in the first place.  Since it gives the fatal error afterwards, I say it's better to leave it in for debug builds at least.

Shogun

PR #453, opened 2017-05-08
Texture updates

PatrickvL:
This doesn't really change much, it's just a few small refactorings that cleanup a bit of texture-related code.


LukeUsher:
Tested with around 5 titles, saw no change in behavior, if any regressions do surface from this, please create an issue for them

PR #458, opened 2017-05-12
XAPI: Improved Controller response

jarupxx:
If the function succeeds, the return value is ERROR_SUCCESS.
I was tested Quantum Redshift and JSRF controller does work.


jarupxx:
Please check PR #460.
This PR be incomplete. This informs that the device has reconnected. In fact, the device must be connected to bootup.

PatrickvL:
You closed this issue yourself, without merging it. Does that mean the patches you enabled can be left DISABLED?

LukeUsher:
I fixed controller input with a different method in my latest pull-requests: this aused XGetDeviceChanges to report one new device inserted, where I altered XGetDevices to report one controller connected since boot. My method prevented the dashboard from detecting too many controllers

jarupxx:
I read Luke's pull-requests and I understood that this fix is tweak. My method can detect the controller by reconnecting. Just a wait. Turned on the real console with connecting the controller, so the device will not be changed. I think that this patch can be DISABLED until we support change devices,

PR #461, opened 2017-05-13
Re-enable D3DSurface_GetDesc patch

LukeUsher:
This should work unpatched but we are not quite there yet.
Patching this fixes missing graphics in X-Marbles and stops ZSNESBOX from crashing instantly.

Possibly solves other issues too!


PatrickvL:
Thanks, let's revisit this later, hopefully Lego benefits from this too, returning to it's previous state

PR #463, opened 2017-05-13
Enable AVX support

LukeUsher:
This should bring a significant speed increase in some titles.

NOTE: This reduces the number of processors that are compatible with Cxbx-Reloaded, see the list linked in README.md for details

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Cxbx-Reloaded is still pretty unstable, don't expect it to run much at this poin
 
 ## System Requirements
   * 64-bit Windows (7, 8 and 10 are known to work, 32-bit installations are not supported)
+  * Any CPU that supports AVS [View List](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#CPUs_with_AVX)
   * A graphics card that supports Direct3D 8.
 
 ## Automated Builds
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,6 +11,7 @@ Cxbx-Reloaded is still pretty unstable, don't expect it to run much at this poin
 
 ## System Requirements
   * 64-bit Windows (7, 8 and 10 are known to work, 32-bit installations are not supported)
+  * Any CPU that supports AVS [View List](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#CPUs_with_AVX)
```
JayFoxRox: *`AVX` ?

JayFoxRox:
I feel this change is kind of unnecessary. It limits the usability and it might even scare away potential devs at this point.

As you don't have any AVX based inline assembly, it should be made very clear that this is easily fixable by a recompile.
I'm not sure about the situation on Windows, but maybe the compilers you use also support runtime detection of these things or fat binaries (which ship non-AVX and AVX version in the same binary).
If so, this would be / is an acceptable change.

You also shouldn't use CPU features yet, but instead focus on algorithmic improvements.
That way you at least have a way to squeeze out more performance using CPU optimizations as last resort. But that's just my opinion.

*(That said, XQEMU will soon get an AVX based shader jit probably; probably with an interpreter or SSE fallback though)*

LukeUsher:
I didn't think requiring AVX would be such an issue: It's supported on most CPU's made after 2011, that's 6 years ago already. Generally, people interested in gaming and emulation have fairly modern computers. 

That said, I can roll-back this requirement, if the majority think it's best.

gandalfthewhite19890404:
I think AVX at current state of progress of emulator is unnessesary and useless.

PatrickvL:
Let's​ roll it back then

PR #469, opened 2017-05-15
Re-patch XGetDeviceChanges (4432-5849)

jarupxx:


PatrickvL:
Can you explain why this is needed? And how it's tested?

jarupxx:
This function has been verified as functioning properly up to XDK 4361 by PR #465.

I tested with the newer XDK titles Rayman Arena (4627) and Quantum Redshift (4721) Rayman 3 (5233) but I decided that it was necessary because the input was improved.

PatrickvL:
Thanks for the explanation, I've merged it in now. Great work!

PR #474, opened 2017-05-20
Clear HLE Cache when LLE options are toggled

LukeUsher:
This is required because enabling/disabling LLE options alters which OOVPA Databases are used.


PatrickvL:
Wouldn't it be better if each LLE flag could function independent from symbol detection, and thus cause no clear of the HLE cache?

LukeUsher:
It would, but that requires a much larger rework of the symbol scanning and hle caching system than this short term fix

PR #480, opened 2017-05-28
JSRF Hack: Prevent crash after XGetDeviceChanges

LukeUsher:
I hate game specific hacks, but I already wasted three whole weeks trying to solve this problem.

Even making XGetDeviceChanges run completely unpatched has the same issue, if it returns anything other than false.

This gets it working for now, without breaking other titles that rely on XGetDeviceChanges to be implemented.

I'll revisit this issue after the kernel emulation is more complete.

### DIFF
--- a/src/CxbxKrnl/EmuXapi.cpp
+++ b/src/CxbxKrnl/EmuXapi.cpp
@@ -193,6 +193,17 @@ BOOL WINAPI XTL::EMUPATCH(XGetDeviceChanges)
 
 	BOOL ret = FALSE;
 
+	// JSRF Hack: Always return no device changes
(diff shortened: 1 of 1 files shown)


src/CxbxKrnl/EmuXapi.cpp
```diff
@@ -193,6 +193,17 @@ BOOL WINAPI XTL::EMUPATCH(XGetDeviceChanges)
 
 	BOOL ret = FALSE;
 
+	// JSRF Hack: Always return no device changes
```
Sarkie: Worth doing?
https://blogs.msdn.microsoft.com/zainnab/2013/08/15/visual-studio-2013-todo-comments-and-custom-tokens-in-the-task-list/

PR #484, opened 2017-06-04
Fix and enable XInput controller support.

LukeUsher:
This is configurable in the Settings menu

If Use XInput is selected, XInput controllers will be used, otherwise the default controller mapping system will be used.

On an Xbox 360 controller, buttons are mapped 1:1 with Xbox buttons, with the exception of BLACK & WHITE.

Xbox 360 controllers lack BLACK/WHITE buttons, but instead have LB and RB.
WHITE is mapped to LB and BLACK is mapped to RB.


LukeUsher:
Merging this as it is behind a configuration option and has no impact outside of that.

src/Cxbx/WndMain.cpp
```diff
@@ -455,628 +461,633 @@ LRESULT CALLBACK WndMain::WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lP
 
         case WM_COMMAND:
         {
-            switch(LOWORD(wParam))
```
PatrickvL: pity about all these indent changes

LukeUsher: Visual Studio is a nightmare sometimes -> I didn't notice it had automatically re-indented everything

PatrickvL: yeah, that's why I keep using the GitHub client - it allows me to visually​ interact with the lines being committed, so I tend to leave indent-only changes out of the commit, then I revert those useless changes afterwards

Luffykun007:
The X-box 360 control D-pad is not yet configuring in the emulator settings panel.

LukeUsher:
It's automatically mapped, just tick the xinput box and the controller will be used - xinput overrides the configure controller dialogue: that dialogue only listens to DirectInput

src/Cxbx/WndMain.cpp
RadWolfie: I'll leave a hint even if you already knew. :wink:

Any changes you do in Visual Studio, it will show yellow, unsaved, or green, saved, highlight right next to scrollbar on a file. I had used it to undo any changes I did or know which changes I had done to figure out what went wrong afterward.

Luffykun007:
Yes it works thanks so much

PR #498, opened 2017-07-10
VBlank timing improvements.

LukeUsher:
A titles registered VBlank Callback is called after a (simulated) VBlank imterval has passed. This is better than current behavoir where it is incorrectly called multiple times per frame.

BlockUntilVerticalBlank has been updated to wait for the VBlank counter to increase, rather than wait a set amount of time. This is more accurate in cases where it was called later in the frame, it may not be correct to always wait a whole frames worth of time.

I haven't noticed much change in behavior from this, but it is (slightly) more correct. A 100% correct approach relies on at least partial LLE NV2A emulation, some of which will be addressed by PatrickvL's current work. After that is merged, I'll revisit the timing subject.

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -1424,8 +1424,6 @@ static DWORD WINAPI EmuUpdateTickCount(LPVOID)
 
     DbgPrintf("EmuD3D8: Timing thread is running.\n");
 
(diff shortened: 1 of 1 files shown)


LukeUsher:
Actually, this breaks Psyvariar. Until I figure out how to solve this, I'm closing.

LukeUsher:
Gah, it was just cache related, we're still good!

PR #505, opened 2017-07-18
Add partial implementation of KeLowerIrql

LukeUsher:
There should be much less occurrences of KeBugCheck being called now!

Verified with various pieces of homebrew + XOnline Dashboard


PatrickvL:
Looks good, thanks!

Sarkie:
Making lots of Homebrew booting now!

PR #507, opened 2017-07-19
Implement alternate Anti-Piracy patching

LukeUsher:
This eliminates the need to iterate through the Xbe contents to apply media patches.

This method is the same one used by various 'hacked' kernels to bypass security checks.


PatrickvL:
Good to see a bit of patching code go, well done. How can this be verified?

LukeUsher:
Just run various (unmodified/unpatched) games, BattleStar Galactica or any later. If they don't immediately abort and attempt to reboot to the dashboard, it's working.

PR #508, opened 2017-07-19
Check palette caps

PatrickvL:


PatrickvL:
This was on my TODO-list for a while, until I saw how easy it can be checked in https://github.com/crosire/d3d8to9/pull/38/files (thanks for that)

PR #510, opened 2017-07-20
Introduce flash handling

PatrickvL:


src/CxbxKrnl/EmuX86.cpp
```diff
@@ -128,6 +128,22 @@ void EmuX86_Mem_Write8(xbaddr addr, uint8_t value)
 	*(uint8_t*)addr = value;
 }
 
+uint32_t EmuFlash_Read32(xbaddr addr) // TODO : Move to EmuFlash.cpp
+{
+	uint32_t r;
+
+	switch (addr) {
+	case 0x78: // ROM_VERSION
+		r = 0x46; // DVT4
```
LukeUsher: Why DVT4? That's a Debug kit?

PatrickvL: It's the latest hardware revision, shown by some system-info utilities

src/CxbxKrnl/CxbxKrnl.h
```diff
@@ -84,7 +84,17 @@ typedef uint32 xbaddr;
...
+
+#define XBOX_WRITE_COMBINED_BASE 0xF0000000
+#define XBOX_WRITE_COMBINED_SIZE 0x08000000 // - 0xF7FFFFF
+
+#define XBOX_UNCACHED_BASE       0xF8000000
+#define XBOX_UNCACHED_SIZE       0x07F00000 // - 0xFFEFFFFF
+
+#define XBOX_NV2A_INIT_VECTOR    0xFF000008
+
+#define XBOX_FLASH_ROM_BASE      0xFFF00000
```
JayFoxRox: I believe the Flash is 16 MiB of physical memory -> 0xFF000000 (= the Flash will be mirrored 16x 1 MiB or 64 x 256 kiB times)
The 2BL only maps 4 MiB to virtual memory -> 0xFFC00000

The actual flash in hardware is only 1 MiB (this commit) or 256 kiB on later consoles.
All Microsoft kernels (to my knowledge) only use 256 kiB images (which are mirrored 4 times for the 1 MiB images)

PR #514, opened 2017-07-23
HLE Audio Improvement

RadWolfie:
Lots of audio improvements and XADPCM codec support.

A warning... if a certain game crash or corrupt the registry database for Cxbx-Reloaded, then attempt to open audio configuration settings will cause it to be non-responsive.


src/Common/XADPCM.h
```diff
@@ -0,0 +1,169 @@
...
+// *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+// *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// *  GNU General Public License for more details.
+// *
+// *  You should have recieved a copy of the GNU General Public License
+// *  along with this program; see the file COPYING.
+// *  If not, write to the Free Software Foundation, Inc.,
+// *  59 Temple Place - Suite 330, Bostom, MA 02111-1307, USA.
+// *
+// *  (c) 2002-2003 Aaron Robinson <caustik@caustik.com>
```
PatrickvL: caustic didn't edit this, so this line can be removed

PatrickvL:
Great work!! We'll test this before merging, please be patient.

LukeUsher:
Don't worry about that warning about registry corruption: The same happens for all of the other Cxbx-Reloaded settings too.

LukeUsher:
I have tested the majority of my game library, great work, I'm noticing significant audio improvements across the board.

Just one suggestion: In the Audio Configuration settings panel, the default option was only PCM enabled. Could all supported codecs be enabled by default? A developer can disable them if they require. Thanks!

src/Common/XADPCM.h
RadWolfie: Didn't see this, will fix it soon.

RadWolfie:
@LukeUsher  reason I didn't enable XADPCM by default is because of some titles make distort noise just because of delay decode in lock caller such as Turok Evolution did. If both of you two do suggest all to be ena, I'll update it.

PatrickvL:
Is the amount of distortion related to the speed of decoding perhaps?

PatrickvL:
In which case fast computers might run xadpcm better than slower machines

LukeUsher:
Don't worry about it then, let's leave it optional for now. Thanks!. Just one more suggestion? Could you add yourself to Contributors file in the root folder? That way nobody forgets to add you :p. Good work, I'll merge after that

RadWolfie:
@PatrickvL Turok Evolution play the background music with half buffer size provided each time it called lock function. Unlike what dashboard does, which constantly input smaller buffer instead of half buffer size. That's the conflict I'm having so far.

@LukeUsher I'll do that in 4 hours from now.

PatrickvL:
Okay, I see. Perhaps it's possible to detect overlap in the buffers passed into the conversion functions, so previous conversion results can be reused. Let's consider that later though, let's get this merged in first

PR #537, opened 2017-07-28
DSound logging improvements

PatrickvL:
This PR mainly replaces all DSound function logging with LOG_FUNC macro's, and adds a few slight cleanup changes.


RadWolfie:
Thanks for making my job to verify if it work as intended or not. :P I usually just rename DbgPrintf to printf only in dsound file plus disable the LOG_FUNC macros to know what's the expected step it should take.

PatrickvL:
@RadWolfie I'm sure you'll find another method to test with. I did this, so we're one (big) step closer to consistent logging throughout our entire codebase.

Issue #535, opened 2017-07-28
[PAL] Colin McRae Rally 2005 - crashes during startup


rexcaptain501:
Tested on Cxbx-Reloaded build a8033fee, Jul 27 2017

Creating startup crash issue on request of @RadWolfie.

Crash: ![cmr05](https://user-images.githubusercontent.com/10300935/28697723-064563c6-7392-11e7-9efd-9d55c977835b.PNG)

.xbe info: [Xbe.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1181652/Xbe.txt)

Kernel Debug file: [KrnlDebug.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1181653/KrnlDebug.txt)

RadWolfie:
It has regress for no audio after merge audio improvement.

After reviewing the message of CDirectSoundVoice_SetMixBinsVolume + 0xd6 at the crash. By going back 2 caller functions, I have discovered a missing patch for this title which I think is IDirectSoundBuffer_SetNotificationPositions. Since the passdown on the first call show these params.
1. pX_CDirectSoundBuffer
2. 0x00000001
3. appears to be a pointer of CDSBPOSITIONNOTIFY containing -1 and a handle.

By adding **`IDirectSoundBuffer_SetNotificationPositions=21eb15`** to HLE cache file, it start to play two intros which is currently show no video. Then show light gray color background.

Can one of the dev team verify this patch is indeed "IDirectSoundBuffer_SetNotificationPositions"?

jackchentwkh:
one should be able to compare the HLE cache file generated before and after the regression to see the differences.

PatrickvL:
If this is a regression, then yeah, the contents of the HLE cache file could be compared between "git bisect" steps, but I fear that might show many differences (especially when OOVPA updates have been merged in).
Anyway, on which commit did it still work?

RadWolfie:
[...]

PR #543, opened 2017-07-29
Missing Patch To Improve Colin McRae Rally 2005

RadWolfie:
Add missing both SetNotificationPositions patches with existing OOVPA signature from 4627 HLE
database.

ref #535 - This fixed and made Colin McRae Rally 2005 go a bit further,
yet still crash after 2 black intro screens.

It does not affect Lego Star Wars, Lego Star Wars 2, and Star Wars KotoR 2 since all 3 return no detection for both SetNotificationPositions OOPVAs.

### DIFF
--- a/src/CxbxKrnl/HLEDataBase/DSound.1.0.5344.inl
+++ b/src/CxbxKrnl/HLEDataBase/DSound.1.0.5344.inl
@@ -984,6 +984,8 @@ OOVPATable DSound_5344[] = {
 	REGISTER_OOVPA(CDirectSoundBuffer_SetConeOrientation, 4134, XREF),
 	REGISTER_OOVPA(IDirectSoundBuffer_SetConeOrientation, 4134, PATCH),
 	REGISTER_OOVPA(CDirectSoundStream_SetConeOrientation, 4134, PATCH),
+	REGISTER_OOVPA(CDirectSoundBuffer_SetNotificationPositions, 4627, XREF),
+	REGISTER_OOVPA(IDirectSoundBuffer_SetNotificationPositions, 4627, PATCH)
 };
 
 // ******************************************************************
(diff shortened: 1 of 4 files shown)


BenNottelling:
I believe this will help Futurama, it wouldn't boot and complained about SetNotificationsPositions, but I'll test in the morning

RadWolfie:
I had checked Futurama title, the info state it is 5233 xdk version. This pull request doesn't cover 5233 xdk database. However I had looked in commits to see if it was added. Turns out it had been added in commit https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/6d8b261219cea524549ef7952b7507d4d1090c9f about a month ago. Looking in HLE cache file and it state SetNotificationsPositions for both xref and patch are not detected. The message you're getting might be at the "nearest" fatal crash address which then is showing SetNotificationsPositions + 0x???? and such.

BenNottelling:
@RadWolfie you are right sorry! I thought it was the SetNotificationsPositions error, but it's not

jackchentwkh:
Is there any clue for 4134 xdk signatures for SetNotificationsPositions?

RadWolfie:
Gunvalkrye and Jet Set Radio Future are the two known title that are 4134 I know of so far. They both had not crashed since SetNotificationsPositions do make it crash to me and others at least. I don't have a clue about 4134 xdk signature for it. :(

Side note: I noticed HLE database is missing SetNotificationsPositions in...

- 4134
- 4361
- 4432

PatrickvL:
Any patch that's not in the HLE database of any version, should probably be added with the closest possible version

PatrickvL:
So yeah, if it's missing, please add it. If it turns out we need an OOVPA of another version, or a new OOVPA specifically for that database version, we can always do that

RadWolfie:
One problem, I do see bunch of version changes over xdk revised releases. Therefore, I am not conformable with add the missing signature without validation first.

PatrickvL:
Using older versions of an OOVPA again in a newer database is tricky indeed. Especially if there's another version in between; Should the newer version be used, or the older? In some cases the newer version should be used, in other cases the older version is required. I've seen a lot of OOVPA that could be replaced by one or two more accurate ones, which would cover all xdk versions which we have patterns for. But that doesn't necessarily mean these OOVPA would work for unknown xdk versions too. So you're absolutely right to verify any change before it goes in. But we should also start to build up patterns for those unknown xdk's by extracting the patterns from titles, and converting those into OOVPA, as that's the only way we can ever hope to support all xdk versions.

PR #558, opened 2017-07-31
Add prerequisites needed to run to README

CakeLancelot:
Without installing Visual C++ Redistributable, Cxbx-Reloaded does not run. Add this to the README for clarification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Cxbx-Reloaded is early in development and still pretty unstable, so don't expect
   * OS: Windows 7 64-bit or newer. 32-bit installations are not supported.
   * Video card: anything that supports Direct3D 8.
 ### Prerequisites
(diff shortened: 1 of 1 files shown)


LukeUsher:
Where do you acquire your builds? Official builds are generally compiled with Visual Studio 2017, and would need that run-time instead

CakeLancelot:
This is from the appveyor build.

LukeUsher:
Could you perhaps link to this page and specify that both 2015 and 2017 may be required?

https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads

CakeLancelot:
> Could you perhaps link to this page and specify that both 2015 and 2017 may be required?
> 
> https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads

Added

BenNottelling:
I believe I've caused a lot of little eating commits from people 😝

Issue #552, opened 2017-07-29
Remove duplicate Xbe.cpp file


PatrickvL:
Xbe.cpp appears twice in the src tree.
Refactor the code such, that one of these two can be removed.

Luca1991:
I always wanted to take part in this project, and I think that this issue is my best chance to jump in. Can I be assigned to this issue? I've a very busy life, but during weekends I've some free time.

LukeUsher:
Sure!

Luca1991:
@LukeUsher Thanks for the trust, I'll work on this ASAP.

ergo720:
I tried to build cxbx simply by removing this file and it builds fine but doesn't function properly: all the games fail to start with a black screen or an error regarding XInput not being initialized.

Luca1991:
@ergo720 ah, I've tried to compile cxbx without that file too, but I haven't tried to boot any games. Well I'll try to find any references then ;) Do you have any idea right now?

PatrickvL:
No need for code bloat, so anything not used could at least be ifdef-ed out. But indeed, do test no regressions occur because of it

ergo720:
The big difference between the two is that the core version has an overloaded constructor not present in the common version plus a lot of small differences in the other functions throughout the source.

Luca1991:
@ergo720 are you sure that building without \Common\Xbe.cpp CXBX will fail to start any game? I'm asking this because I've built it without this file and it still boots my games.
Right now I'm having a lot of issues booting games (the ones supposed to works), and the only title that goes ingame is "Frogger Beyond", but as far as I can tell, I haven't noticed any regression so far. 
I'll give it a closer look when I'll have some free time ;)

PatrickvL:
If I'm not mistaken, `Cxbx-Reloaded\src\Core\Xbe.cpp` can be deleted, while `Cxbx-Reloaded\src\Common\Xbe.cpp` should stay

Luca1991:
@PatrickvL yeah sorry, I was in rush and mistakenly written \Common instead of \Core. I've edited my previous message to avoid any confusion. As I've said, I tried to boot "Frogger Beyond" and it goes ingame without any regression. So, I'll give a closer look to \Core\Xbe.cpp to check if there are some useful functions to merge in \Common\Xbe.cpp and then I'll send a PR, ok?

PatrickvL:
[...]

Issue #568, opened 2017-08-01
Add "Drag and Drop" XBE-loading feature to GUI


PatrickvL:
Fo this issue, extend the Cxbx GUI such, that an Xbe can be loaded using a Drag and Drop operatiun.
This should only be accepted when emulation is not running.
If no XBE is loaded yet, load the dropped XBE. If an XBE is already loaded, replace it with the dropped XBE.
Be carefull to use correct paths.

Some pointers that might contain valuable information on this subject:
* https://msdn.microsoft.com/en-us/library/windows/desktop/bb776406(v=vs.85).aspx
* https://msdn.microsoft.com/en-us/library/windows/desktop/bb774303(v=vs.85).aspx
* http://www.dreamincode.net/forums/topic/293568-drag-and-drop-files-demo-in-c/
* https://helgeklein.com/blog/2010/03/how-to-enable-drag-and-drop-for-an-elevated-mfc-application-on-vistawindows-7/

ergo720:
I have implemented the feature as requested in my branch and tested with my games: they all work as before. I'll submit a new PR.

PatrickvL:
That's great! I didn't expect a PR for this so quickly!

PatrickvL:
Implemented in 4d30506

PR #575, opened 2017-08-01
Removed duplicate Xbe.cpp file

Luca1991:
This PR fixes the issue #552.  As said in the issue,  the main difference is this unused constructor: `Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail)`.

Many thanks.


PatrickvL:
Thanks for this. Are you up for another task?

PatrickvL:
#568 for example

Luca1991:
Yes, but it looks like @ergo720 already fixed it :P

PR #576, opened 2017-08-02
Drag and Drop implementation + UIPI bypass

ergo720:
Implements the feature of issue #568


PatrickvL:
Thanks for this!

LukeUsher:
Thanks!
I'm not sure about specifically having code to allow features to work with elevated privileges though.

As we run xbox code natively on the host CPU, running Cxbx-Reloaded with elevated privileges could be dangerous.

PR #577, opened 2017-08-02
Variables renaming...

ergo720:
Just renaming a couple of variables that I named with the m_ prefix. Because they aren't members of a class, it's a bit misleading to call them this way...


LukeUsher:
Thanks!

PR #672, opened 2017-08-17
Removed unused Error.cpp file and now empty Core directory

Luca1991:
Hi all, I'm back from the holiday!
Here is a little PR that remove an unused file and dir.


LukeUsher:
Thanks

PR #683, opened 2017-08-27
Fix configuration logging code.

darrena092:
These changes fix an issue where code that logs the configuration commited in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/48f6c1107acfad6bdd08fc0e23a100466dbceff3 was causing a memory access violation.


darrena092:
It seems to be that not using pointers here will cause a memory access violation:

For example, this will cause it:
```
XBVideo XBVideoConf;
g_EmuShared->GetXBVideo(&XBVideoConf);
```
In theory, that should work just fine, however it doesn't. Using a pointer as I've done here seems to solve the issue. I'm really not 100% sure why it doesn't work as is.

I'll add in a commit to delete those pointers, overlooked that sorry.

darrena092:
This needs a bit more investigation. It seems that the issue happens when the objects are deleted at the end of the function. It turns out that the only issue I solved by using pointers was that the objects were not automatically deleted at the end of their scope - so no access violation.

Once I actually manually deleted those pointers at the end of their scope, the same issue happened.

Issue #682, opened 2017-08-27
Crash on emulation start due to logging code.


darrena092:
### Description

Code added in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/48f6c1107acfad6bdd08fc0e23a100466dbceff3 is causing a memory access violation which means that emulation fails to start.

It seems that the access violation is when memcpy is called in the various accessors in EmuShared. 

I've tested this with two different games, both resulting in the same behaviour.

---

### Game info
Tested with both "Kingdom Under Fire: The Crusaders" and "Crash Twinsanity".

---

### Status
**Previous behavior:** Emulation starts successfully.

**Current behavior:** Emulation fails to start due to memory access violation.

---

darrena092:
Submitted a PR which solves this issue https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/683

PatrickvL:
> seems that the access violation is when memcpy is called in the various accessors in EmuShared

If that's the case (seemingly a bounds overflow), then it's only avoided by using a separately allocated structure (not actually fixed). It'd be good to know which memcpy call(s) is/are the actual culprit (s).

darrena092:
It turned out to be simpler than that - the XBVideo and XBAudio classes weren't being instanced, so the functions from EmuShared were just writing to unallocated memory.

darrena092:
Also, I'm an idiot. It's to do with the objects being deleted at the end of their scope. The only issue I solved in my PR was accidentally avoiding their deletion, so things continued along happily.

Will look into it further and see if I can come up with a solution - it would be helpful to see if anyone else can reproduce this issue.

PatrickvL:
So the fix should be calling their constructors?

darrena092:
Submitted a second PR https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/684

This solves the issue by using copy constructors. The reason that it hadn't become a problem before was that it seems there was nowhere that actually used the accessor methods in EmuShared and subsequently deleted the copied object until now.

PR #684, opened 2017-08-27
Fix for access violations when deleting objects accessed through EmuShared.

darrena092:
This fixes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/682

The issue was that because memcpy was being used to duplicate objects, strange things were happening when those objects were then deleted. I took out the memcpy calls and replaced them with copy constructors.


PatrickvL:
Much better, thanks!

PR #687, opened 2017-08-29
Registered NTSC-J demo version in JSRF Hack

jarupxx:
Fixed issue 157
https://github.com/Cxbx-Reloaded/game-compatibility/issues/157


LukeUsher:
Thanks!

PatrickvL:
Thanks indeed!

It's worrying though, that this hack in `XGetDeviceChanges` is all that's required to get an XBE from status 'booting' to 'playable'; Who knows how many other titles would be able to progress further, if only the input handling was improved?!?

LukeUsher:
I agree, although we still have no idea why this hack is needed. Unpatching this function prevents some titles from detecting controllers, patching this function fixes said controller issues but breaks JSRF during early startup.

After spending weeks on the issue, I settled with the hack and I plan to completely unpatch XINPUT and LLE emulate the USB devices later on, so then input shouldn't be a problem anymore. My experimental NV2A branch contains good prep-work for this (DPC routines, some form of interrupt handling), PCI emulation and OHCI emulation are the required next steps, after those, actual gamepad devices can be emulated by hooking up virtual USB devices to the OHCI bus.

PR #693, opened 2017-08-30
Interrupt Support, DPC fixes and additional EmuX86 instructions

LukeUsher:
Fix some bugs in DPC handling:
  - DPC Routines are xbox code, but were being called under the windows thread context, without the Xbox thread handling code being used
 - DPC routines were using the wrong calling convention

Implement more X86 Instructions:
 - CMP
 - IN

Implement support for interrupts:
 - Interrupts use a processing thread on the same thread affinity as xbox cores, this prevents other xbox code running at the same time as an interrupt is being processed
 - As interrupts run in their own thread, the execution of ISR routines may be delayed: It cannot execute until Windows causes a context switch to that thread, this is the best we can do without either full CPU emulation, or at least simulating xbox context switching, neither of which as possible with the current design of Cxbx-Reloaded.

Prevented RtlAssert from aborting emulation, still shows a warning.

- Implemented IO port 0x8008. This implementation is not correct, but good enough to please NXDK.

### DIFF
--- a/import/OpenXDK/include/xboxkrnl/xboxkrnl.h
+++ b/import/OpenXDK/include/xboxkrnl/xboxkrnl.h
@@ -1527,7 +1527,7 @@ struct _KDPC;
 // ******************************************************************
 // * PKDEFERRED_ROUTINE
 // ******************************************************************
-typedef VOID (*PKDEFERRED_ROUTINE)
+typedef VOID (__stdcall *PKDEFERRED_ROUTINE)
 (
     IN struct _KDPC *Dpc,
     IN PVOID         DeferredContext,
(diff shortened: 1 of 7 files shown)


LukeUsher:
This shouldn't have any negative effect on HLE, but I'd appreciate if it could be tested before being merged, who knows, the DPC fixes might be helpful

literalmente-game:
I've tested:
- Smashing Drive
- Burnout
- Gauntlet Dark Legacy
- Dolphin (Tech Demo)
- Arx Fatalis
- Battlestar Galactica
- Unreal Championship
- Taz - Wanted
- ZSNES BOX (SNES emulator)
No regressions found and no difference on the performance

LukeUsher:
That's good enough for me, thanks!

PR #694, opened 2017-08-31
Update README.md

literalmente-game:
Typo fix


LukeUsher:
A quick note: "An xbe" is the correct form rather than "A xbe" this may be counter-intuitive when you are taught that an goes before a vowel, eg: an egg, however that rule is determined by sound, not by alphabetical representation. 

In this example, "an xbe" is correct simply because please the x in "xbe" is pronounced as "ex". 

A similar example is "in an hour"  rather than "in a hour" and the opposite is "a euro", "an" is not used because euro starts with a consonant sound, even though it starts with a vowel. 

Not important, just trying to help (:

literalmente-game:
And today I've learned something new, thank you sir ;)

Issue #688, opened 2017-08-29
Remove hacks from `XGetDeviceChanges`


PatrickvL:
A recent Pull Request (#687) added a title-specific hack to `XGetDeviceChanges`, which resulted in `state-booting` changing to `state-working`.

Instead of add hacks to support more titles, more fundamental research most be done, so a real fix can be made, allowing the removal of these hacks. It's fairly probable more titles will benefit from this.

LukeUsher:
The key problem is that we still have no idea why this hack is needed. 

Unpatching this function prevents some titles from detecting controllers, patching this function fixes said controller issues but breaks JSRF during early startup. If we wish to HLE controller devices, this function must indeed be patched.

After spending weeks on the issue, I settled with the hack and I plan to completely unpatch XINPUT and LLE emulate the USB devices later on, so then input shouldn't be a problem anymore. My experimental NV2A branch contains good prep-work for this (DPC routines, some form of interrupt handling), PCI emulation and OHCI emulation are the required next steps, after those, actual gamepad devices can be emulated by hooking up virtual USB devices to the OHCI bus.

PatrickvL:
One reason `XGetDeviceChanges` doesn't work (well), is because of [this line](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/19f0bbe7aa61e6fc9432268fe0d0d027f7232eda/src/CxbxKrnl/EmuXapi.cpp#L208), which accesses `DeviceType->CurrentConnected` (`DeviceType` being a `PXPP_DEVICE_TYPE`). It's possible this pointer isn't initialized properly, resulting in accesses to random memory adresses...

LukeUsher:
Although this is possible (and should be checked for) in practice, I have never seen this happen, and the crash never happens within this function, but some time after it returns, much later in execution but still before anything displays onscreen.

jarupxx:
JSRF executes the following instruction..

.text:001669D7 call _XGetDeviceChanges@12 ; XGetDeviceChanges(x,x,x)
.text:001669DC test eax, eax
.text:001669DE jnz loc_166B0D

.text:00166B0D loc_166B0D:
.text:00166B0D mov eax, 80004005h
.text:00166B12 pop edi
.text:00166B13 pop esi
.text:00166B14 pop ebx
.text:00166B15 mov esp, ebp
.text:00166B17 pop ebp
.text:00166B18 retn 14h

[...]

PR #698, opened 2017-08-31
Potential fix for issue #688

ergo720:
JSRF goes ingame without the hack with this.  I tested all my other games and didn't see any regressions/improvements, but I don't have lego star wars so someone else must check it.

### DIFF
--- a/src/CxbxKrnl/EmuXapi.cpp
+++ b/src/CxbxKrnl/EmuXapi.cpp
@@ -191,26 +191,7 @@ BOOL WINAPI XTL::EMUPATCH(XGetDeviceChanges)
 	LOG_FUNC_END;
 
 	BOOL ret = FALSE;
-
-    if(!DeviceType->ChangeConnected)
-    {
-        *pdwInsertions = 0;
-        *pdwRemovals = 0;
-    }
-    else
-    {
-        *pdwInsertions = (DeviceType->CurrentConnected & ~DeviceType->PreviousConnected);
-        *pdwRemovals = (DeviceType->PreviousConnected & ~DeviceType->CurrentConnected);
-        ULONG RemoveInsert = DeviceType->ChangeConnected &
-            DeviceType->CurrentConnected &
-            DeviceType->PreviousConnected;
-        *pdwRemovals |= RemoveInsert;
-        *pdwInsertions |= RemoveInsert;
-        DeviceType->ChangeConnected = 0;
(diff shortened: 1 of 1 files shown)


PatrickvL:
Could you either this a little so there's only a single point of exit? (Mostly by keeping the original RETURN, and making sure all code paths flow there.)

PatrickvL:
Congratulations on removing the only title-specific hack we had left, by the way!

ergo720:
Better now? (there are also the usual indentation problems because I used spaces whereas the other code used tabs)

LukeUsher:
Note to all: Make sure this gets a thorough testing before merging, at a minimum, Lego Star Wars, Jet Set Radio Future, Turok Evolution and Battlestar Galactica need to be checked to ensure controller support still works.

PatrickvL:
Thanks, much better ;)

I can't test for a while, so hopefully someone else can.

ergo720:
I played the first level of jsrf so that works but I don't have the other three games. The input works also for Panzer Dragon Orta if that helps

PatrickvL:
If I remember correctly, we had a few titles with input related issues, like Quantum Red Shift, maybe it's a good idea to test a few of those too (look up issues labelled with 'issue-input' in the compatibility project)

LukeUsher:
It's specific games that have issues, as Patrick and I mentioned. The key one to test is Lego Star Wars: There's never been a situation where both input works in Lego Star Wars, and JSRF actually boots, without hacks.

ergo720:
I don't have any of those, with the exception of Halo which crashes with the same error of before

LukeUsher:
We can't merge until somebody can test those games and verify that it works

ergo720:
Ok, if it doesn't work, I'll try to buy that game as well

RadWolfie:
@ergo720 this is very similar to my XGetDeviceChanges-Fix branch. However your's is so much more accurate for the else portion. 🎉 👍 

Sad news, it did regress for JSRF crash and Lego Star Wars is no longer getting any controller input.

ergo720:
Pity, then we have to close this

RadWolfie:
IMO, we should just restore the hack to the top and keep this new change for accuracy reason.

ergo720:
You mean, comment out the modification and restore the hack so we can keep it for future reference?

ergo720:
Ok, as @RadWolfie requested my modification is commented out and the hack restored

PR #699, opened 2017-08-31
Fixes OOVPA for timeSetEvent (XDK 5849).

darrena092:
This fixes the OOVPA for timeSetEvent (XDK 5849).

I was testing CXBX with "Kingdom Under Fire: The Crusaders" and timeSetEvent at $00221F2E wasn't being patched. 

The new OOVPA still matches the bytes in "EA Sports Rugby 2005" which it was originally derived from, but now also matches with KUFTC.


LukeUsher:
Can confirm that this new OOVPA also matches in World Championship Snooker, fixing a KeWaitForMultipleObjects warning there too.

PR #703, opened 2017-09-01
NVNet and Xonline

LukeUsher:
Network connection does not work yet but it satisfies titles enough to stop them crashing.
Tetris Worlds gets to (very broken) gameplay with this
All titles that use Xbox Live, or even System Link are likely to see some improvements.

### DIFF
--- a/src/CxbxKrnl/EmuNVNet.cpp
+++ b/src/CxbxKrnl/EmuNVNet.cpp
@@ -40,8 +40,16 @@
 #define _CXBXKRNL_INTERNAL
 #define _XBOXKRNL_DEFEXTRN_
 
+// prevent name collisions
+
+namespace xboxkrnl
+{
+	#include <xboxkrnl/xboxkrnl.h> // For PKINTERRUPT, etc.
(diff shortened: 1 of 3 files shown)


RadWolfie:
I can confirm xonlinedash.xbe for dashboard build 4920 does not regress and did not crash. In fact after it stop the first animation, the logo is constantly showing the letter I with a glow fade in and out. Well, that's all it does on my side. However I am unable to verify dashboard build 5659 since xonlinedash.xbe is making a reboot to dashboard and does not show anything before it.

rkalz:
I just tried some of the Winsock code (WSAStartup, socket, connect, send, recv) and it appears to work.

Should it be uncommented?

LukeUsher:
It partially worked, but causes problems/ crashes in some Xbox live titles. Rather than put the effort into fixing the HLE, we took the approach to emulate the network hardware instead. We emulate enough to prevent titles from cashing but it doesn't fully work yet, but it's much more stable than the disabled winsock patches

PR #710, opened 2017-09-03
XOnlineDash Improvements

LukeUsher:


src/CxbxKrnl/EmuKrnl.cpp
```diff
@@ -356,8 +356,8 @@ XBSYSAPI EXPORTNUM(252) xboxkrnl::DWORD NTAPI xboxkrnl::PhyGetLinkState
 )
 {
 	LOG_FUNC_ONE_ARG(Mode);
-
-	return XNET_ETHERNET_LINK_ACTIVE | XNET_ETHERNET_LINK_100MBPS | XNET_ETHERNET_LINK_FULL_DUPLEX;
+	LOG_UNIMPLEMENTED();
+	return 0;
```
Sarkie: or create a XNET_NO_CONNECTION ?

Sarkie: @LukeUsher you get the notification? Should we have a new enum?

LukeUsher: No connection is reported by having no flags set, so this is an acceptable solution, even if it's not the best.

Perhaps we should add XNET_ETHERNET_NOT_CONNECTED, just for readability sake

Sarkie: yeah, it shows intent to me and keeps the code readable.

PR #711, opened 2017-09-04
Removes // (Offset,Value)-Pair # text

jarupxx:
Note: In order not to create conflict, work of DSound only in WIP_HLEDB_v2 branch.


PatrickvL:
Thanks, much cleaner!

PR #714, opened 2017-09-05
Threading : Use InitXboxThread for CxbxKrnlInterruptThread too

PatrickvL:


RadWolfie:
Still a standard anyway, merging now.

PR #715, opened 2017-09-05
Log TitleID, TitleName and Region (also as readable string)

PatrickvL:


RadWolfie:
Passed test of visual output onto console for several titles.

PR #721, opened 2017-09-08
Various Bug Fixes

LukeUsher:
Fix an issue where EEPROM.bin was generated with an invalid Factory AV Region.
NOTE: This requires users to delete their EEPROM.BIN, then a clean one will be created, an option to detect/fix invalid values could be added in the future.

Finally fix HalReturnToFirmware. Multiple XBE executables now work without patches, and rebooting to the dashboard or game from update.xbe also works as intended.


RadWolfie:
I made a test on this PR. I can confirm it is working correctly.

One thing I noticed, I am unable to close running emulation via menu option when it start to run a 2nd xbe file. Plus the title did not change. It is affected before this pull request. Merging now.

PR #726, opened 2017-09-11
Check device format

PatrickvL:
This PR could help on some drivers which return incorrect results if `CheckDeviceFormat` is called at the wrong time.

This is done by calling `CheckDeviceFormat` *after*, instead of before creating the device.
Also, all D3D formats are now checked once and remembered into an array.
Lastly, all available DirectDraw FourCC codes are dumped, and any conflict with Direct3D formats is logged.


RadWolfie:
Which titles has made an improvement with this pull request? I had checked several titles and I don't see the improvement except for the messages output onto the console.

PatrickvL:
It's driver dependent. Some user (can't remember who) reported his driver didn't return the correct results. This change is for those situations. Also, it's a preparation for other stuff I'm working on : generic texture conversion

PR #727, opened 2017-09-11
Cleanup of Certificate usage

PatrickvL:
This PR fixes two remaining issues regarding Certificates;
One occurence where the global certificate can already be used, and
one useless certificate title readout, which will be overwritten anyway.


RadWolfie:
Window title is not affected by this pull request. Merging now.

PR #729, opened 2017-09-13
Texture conversion

PatrickvL:
This adds the following :
* fixed loading of XPR logo's that use a non-default alignment
* added loading of DDS logo's
* beginnings of generic, libyuv-like row-based texture decoding
* support for decoding DXT1, DXT3 and DXT5 to ARGB
* various fixes


LukeUsher:
Good to see this coming in, I'll give it a test later!

PR #733, opened 2017-09-16
OOVPA lowered to 5120

jarupxx:
Source: Forza Motorsport (5849)


RadWolfie:
Is the function against 5849 and 5120 are exactly the same or only very little change except did not change the size of the function? Just want to verify if it is true before I go ahead merge this PR.

RadWolfie:
Also we need to split 5120 OOVPAs from 5233 into its own OOVPA version file.

jarupxx:
That's true. Not change the size of the function, only very little change.
Also separate different version of OOVPAs before merging this PR.

PR #735, opened 2017-09-18
Separated XAPI OOVPAs and Lower XAPI OOVPAs Support

jarupxx:
Moved XInputGetState 4928 to 4831
Moved XInputSetState 4928 to 4831
Moved XID_fCloseDevice 4928 to 4831
Moved timeSetEvent 5849 to 5455

Verified with
NBA Live 2002 (3925)
Tony Hawk's Pro Skater 3 (4039)
ESPN Winter X Games Snowboarding 2002 (4134)
NFL Blitz 2002 (4242)
NCAA Football 03 (4432)
Crazy taxi 3 (4531)
Innocent Tears (4627)
Quantum Redshift (4721)
Galaxy Angel (4831)
Medal Of Honor Frontline (4928)
Shikigami no Shiro Evolution (5028)
MVP Baseball 2003 (5120)
C.A.T. Cyber Attack Team (5233)
Starsky And Hutch (5344)
NCAA Football 2004 (5455)
XIII (5558)
Dance Dance Revolution Ultramix (5659)
Steel Battalion - Line of Contact (5788)
Kingdom Under Fire (5849)


src/CxbxKrnl/HLEDataBase/Xapi.1.0.5659.inl
```diff
@@ -0,0 +1,48 @@
...
+// *  (c) 2017 jarupxx
+// *
+// *  All rights reserved
+// *
+// ******************************************************************
+
+// ******************************************************************
+// * XapiInitProcess
+// ******************************************************************
+OOVPA_NO_XREF(XapiInitProcess, 5659, 8)
```
PatrickvL: This one isn't referenced in v2 registration. Is that intentional?

jarupxx: This one saved as a document this is not used from v1. Should I register  in order to get the function name in error logs?

PatrickvL: If this OOVPA is to be used, at least register it in v2.
If it's kept for documentation purposes only, don't register it AND disable it with `#if 0` .... `#endif`.

LukeUsher: We should never remove OOVPA: Even if they are not used for a patch or XRef, they are still useful to have: The locations of symbols is helpful when reverse engineering games or debugging crashes.

PatrickvL: I absolutely agree!
That's why this `XapiInitProcess` OOVPA should be registered in the HLE database (at least in v2).

But, I was thinking about another case :
What if there is a better-suitable (more generic) OOVPA for the same symbol?
For now, in that case, I would keep this 5659 version around in code, but disabled with `#if 0` ... `#endif`.
That way we can decide later what to do with it; Remove it entirely, reinstate it, or whatever else is logical.

jarupxx: I agree! We will register again in the HLE database v2.

RadWolfie: Correction: If it's kept for documentation purposes only, do register it AND use `DISABLED` declare in registration.

This way, we can knowledge OOVPA exists. If a title is outputting a crash error, then Cxbx will show this symbol name as nearest address found. Or else we will be digging around longer without title we may do not have for time being.

If we do not register the OOVPAs, then it will be forgotten within few months to years later. This apply to documentation purposes, patches, references, etc.

PatrickvL: Please excuse me, and bear with me, but it's important to all understand each other, and agree with best practices.

Thus far, we have used the `DISABLED` flag to indicate OOVPA that aren't supposed to have a patch.
Registering an OOVPA (regardless of its flags) will always lead to the use of that OOVPA during symbol-scanning.
The `DISABLED` flag will not skip the scan. The `DISABLED` flag only indicates that no patch should be applied.

I think we all agree that we *really* want to scan for *all* registered OOVPA.
Including the ones that won't be patched.
As you correctly explain above, one of the benefits of scanning for all OOVPA is, that it will make crash-reports more detailed.
Also, if we ever turn our symbol-detection code into a generic library, we want to detect as many symbols as possible.
(For a generic symbol-detection scanning engine, it won't matter how the resulting detection results are used anyway.)

But that's not what I meant when I was talking about keeping old OOVPA "for documentation purposes".
Let's give them a name. Let's call the OOVPA that are replaced by a more generic version : "obsolete OOVPA".

What I propose is, that these obsolete OOVPA may be kept around in code, by surrounding their declarations with `#if 0` and `#endif`.
Obsolete OOVPA *must not* be registered, as they are replaced by a more generic version.
The generic version will be registered.
The obsolete OOVPA is merely kept for reference sake.
We should always add an end-of-line comment after the `#if 0`, stating all information we have on the OOVPA.
(I'm happy to see that almost all occurrences already have these comments, which keeps all research intact.)
The `#if 0`-ed version can be kept until we decide later what to do with it.

Now, I may be interpreting your statement differently than what you mean, but;
Do you really want these 'obsolete' OOVPA versions to be registered?
Why would you want to do that?
Especially when there's already a more generic version available?

Please clear up my confusion :confused:

RadWolfie: > I think we all agree that we really want to scan for all registered OOVPA.
Including the ones that won't be patched.

Yes, that is correct.

Oh for obsolete OOVPA, `#if 0` and `#endif` indeed should be used. Yet we do need to include a reason right next to `#if 0`.

PatrickvL: Thanks for confirming :+1:.

Let's use the word "obsolete" from now on when talking about commented-out OOVPA's. 
(I'll try and be more specific in the future too :grin:)

jarupxx: Excuse me, but, DISABLED flag has been temporarily disabled?
The DISABLED flag seems to be attached to the patch which became obsolete by the improvement of emulation. It may be cleared by giving a better name.

PatrickvL: Well, maybe we could call it NO_PATCH or UNPATCHED, instead of DISABLED. Or better : SCAN_ONLY (or ONLY_SCAN or JUST_SCAN).
But whatever we choose, the effect will stay the same: It's used to indicate OOVPA that aren't supposed to be patched.
Also, this meaning is not temporary.
Ever since Luke introduced the FUNC_EXPORT macro as a mechanism to indicate patches, the DISABLED flag lost it's function as a "patch disabler".

This is for the better, because symbol scanning should be independent from whatever is done with its output (being a list of located symbol addresses).

Radwolfie explained he currently still needs the flags for the current OOVPA maintenance, but once that's finished, all flags can be removed. We probably won't use the distinction anymore. Symbol scanning will be just that: the scanning for symbols, nothing more, nothing less. Only the output matters. Flags will no longer serve a purpose, once all reviewing is completed.

RadWolfie: **Correction:** Radwolfie explained he currently still needs the flags for the current OOVPA maintenance, but once that's finished, **`Flags`** from OOVPATable structure can be **safely** removed.

In fact... scanning method indeed need to change a bit. The original plan is to use **`OOVPATableV2`** structure, in OOVPA.h file, to dramatically reduce comparing strings repeatedly. (What its missing is the counter of how many OOVPAs there are per group registrar.)

Macro for `XREF`, `PATCH`, and `DISABLED` (Or replace with another elegant, straight to the point, keyword) are still permanent. Macro flags are very effective central workflow to work with v2 database. Totally not very effective with v1 database when perform disable individually for each library's revision file. Then cause confusions for others for some has a patch over certain titles and some does not have a patch on other titles.

I think `UNPATCHED` might be better instead of `DISABLED` flag.

PR #737, opened 2017-09-19
NV2A/CreateDevice (The function, not the XDK sample) can now run Unpatched

LukeUsher:
This pull-request brings a partial port of XQEMU's NV2A emulation into Cxbx-Reloaded

This emulation is still largely incomplete, and doesn't function in LLE mode just yet, however, it is *good enough* to allow for many HLE patches to be removed completely. The code quality of the NV2A implementation is hacky in places to work around Cxbx-Reloaded limitations: This will be improved in the future, but for now, this will greatly help even the HLE emulation.

The following patches are no longer required:

- Direct3D_CreateDevice
- D3D_KickOffAndWaitForIdle
- D3D_KickOffAndWaitForIdle2
- D3DDevice_AddRef
- D3DDevice_Reset	
- D3DDevice_End
- D3DDevice_Release
- D3DDevice_KickOff
- D3DDevice_KickPushBuffer
- D3DDevice_EndPushBuffer
- XMETAL_StartPush
- D3D_MakeRequestedSpace
- D3DDevice_GetPushBufferOffset

Along with this NV2A work/hybrid HLE/LLE support, the following changes are also included

- Improved VBlank timing, due to using a higher precision clock source
- Kernel memory region has been made executable (no behavioral change yet, but opens the door for loading a real kernel in the distant future)
- EmuX86: Implemented DEC, SUB and INC opcodes
- EmuX86: Fixed a bug where the carry flag was being calculated incorrectly

I have tested over 40 titles so far, and not noticed any regressions, so I'm confident this can be merged without any major issues.

NOTE: This is a prerequisite for real NV2A LLE, for Patrick's Texture Conversion branch, and Patrick's improved pixel shader conversions, so it is important to get this merged rather quickly.

### DIFF
--- a/src/CxbxKrnl/EmuX86.cpp
+++ b/src/CxbxKrnl/EmuX86.cpp
@@ -812,7 +812,7 @@ bool  EmuX86_Opcode_CMP(LPEXCEPTION_POINTERS e, _DInst& info)
 		return false;
 
 	// SUB Destination with src (cmp internally is a discarded subtract)
-	uint32_t result = dest - src;
+	uint64_t result = dest - src;
 
 	EmuX86_SetFlag(e, EMUX86_EFLAG_CF, (result >> 32) > 0);
 	EmuX86_SetFlag(e, EMUX86_EFLAG_OF, (result >> 31) != (dest >> 31));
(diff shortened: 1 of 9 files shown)


src/CxbxKrnl/EmuX86.cpp
```diff
@@ -82,8 +81,17 @@ uint16_t EmuX86_IORead16(xbaddr addr)
 	return 0;
 }
 
+static int field_pin = 0;
```
PatrickvL: This should really be a threadvar

LukeUsher: How so? It it is intended to be global and not thread specific.

PatrickvL: It turns out this variable is mirroring hardware state, and thus should be global. Case closed ;)

PR #738, opened 2017-09-20
Fix JSRF Regression after NV2A merge

LukeUsher:
The cause of this regression was two fold: The NV2A VBlank thread and the HLE VBlank thread were both running, leading to timing issues as the games VBlank routine was being called too often.

This was fixed by only running the NV2A VBlank thread when in LLE mode.

The second cause was the unpatching of Device_End. This parses display commands and is required for HLE to function correctly.


LukeUsher:
Note that JSRF is still unstable and will crash in the same location sometimes, it always has been the case, so there can be crashes. However this brings the stability back to where it was before the NV2A work.

PR #740, opened 2017-09-20
Patch Removal

LukeUsher:
The unpatching of CreateDevice allows these patches to be removed with no negative effects.


x1nixmzeng:
Bit late, but why were these methods exported to the EXE in the first place @LukeUsher ?

PatrickvL:
Exports are used to publish patches, avoiding the need to build a registration mechanism.

x1nixmzeng:
Ah got it 👍

I wasn't expecting 374 exports from Cxbx.exe!

PR #741, opened 2017-09-20
Re-enable SurfaceDesc patches

LukeUsher:
This fixes black screens/missing textures in XMarbles
These *shouldn't* be required, but until Patrick's texture branch is merged,we need them anyway


PatrickvL:
Thanks! To explain for interested readers: Cxbx historically abuses the Lock field of D3D resources (putting a pointer to the host D3D counterpart in it). These two patches 'fool' callers, returning seemingly correct data. In truth however, this abuse of the Lock field is covered up. My pending work will stop this abuse, instead using an associative map to couple host to Xbox resources. Like Luke says, once that's merged in, these two patches can be unpatched for good.

PR #753, opened 2017-10-04
XAPI: Fix incorrect device detection & Remove JSRF hack

LukeUsher:
This fixes the issue of titles erroneously detecting a memory unit from being connected, among other device bugs.  

This is achieved by deriving the address of XDEVICETYPE_GAMEPAD and updating the CurrentConnected variable within XInputOpen

By the time XGetDevices/XGetDeviceChanges are ran, this stucture is initialized so the patches can work exactly as the real xbox code.

if we could get XInitDevices to run unpatched, both XGetDevices and XGetDeviceChanges would no longer need to be patched at all!
I'll look into it in the coming days.


ergo720:
Congratulations!!

ergo720:
I suppose that's not a problem for retail releases: XInputOpen is used to obtain the handle to an attached device so unless the title doesn't support any controllers, mice or keyboards, it has to be included. I can see that happening only in homebrews, for exxample, that just display some info on the screen

LukeUsher:
The GetTypeInformation structure is required to locate XDEVICE_TYPE_GAMEPAD in later XDKs, in earlier XDKs this symbol doesn't exist and XInputOpen must be used (In the later XDKs, XInputOpen can't be used as the structure has changed location to being referenced by GetTypeInformation instead.

Any title that does not have XInputOpen won't use input, so it won't be a problem.

Issue #730, opened 2017-09-13
Multi-XBE Titles don't want to stop


hotzenplotz3000:
I found a bug, where multi-xbe titles won't stop emulation when using Emulation->Stop or F6
After a game launches another xbe, Stop/F6 do nothing
In this case it's also posible to start a second child thread, while the first is still running 
for example Controller Test default.xbe
It also happens, when a game crashes to dashboard

I tracked it back to PR #712 
it didn't happen before this
only way to stop it, is to close the kernel debug console or cxbx-r

![capture20170913224047439](https://user-images.githubusercontent.com/30478365/30400128-3340b7b8-98d6-11e7-80fc-d3fc7263e065.png)

PatrickvL:
Thanks for reporting this regression. All help is welcome.

I suspect this issue is because restarts of the emulator don't communicate the new ProcessID back to the host (GUI) process.

PatrickvL:
Look for writes to, and reads from, m_hwndChild. My guess is, cascaded launches send their process ID to the parent processes, but aren't detected

PR #757, opened 2017-10-06
Fix for issue #730

ergo720:
Tested with Tenchu (multi-xbe) and Panzer ORTA (which for me reboots to the dashboard) and also other single-xbe games. No regressions observed and "Stop Emulation" now works as intended

### DIFF
--- a/src/Cxbx/WndMain.cpp
+++ b/src/Cxbx/WndMain.cpp
@@ -53,8 +53,6 @@
 #define STB_IMAGE_IMPLEMENTATION
 #include "stb_image.h"
 
-typedef BOOL (WINAPI *ChangeWindowMessageFilterType)(UINT, DWORD); // for GetProcAddress()
-
 static int gameLogoWidth, gameLogoHeight;
 
 void ClearHLECache()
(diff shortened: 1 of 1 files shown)


PatrickvL:
Thanks for taking this one up - it's an annoying issue indeed.

I wonder, why don't you just put `m_hwndChild` in EmuShader instead??? That could make this implementation quite a bit simpler....

ergo720:
Simply because I planned to use WM_COPYDATA in my input and LED branches anyway. For this case, though, is probably more simple to do what you say, so I'll need some time to update this

ergo720:
After more thorough testing I came up with a much more simple fix, just 3 lines of code. I'm sure Patrick will appreciate the new implementation. I tested it again with the games of above and it works

RadWolfie:
Honestly, I had some doubt with this new change _might_ modify our "Recent Xbe files" list. When I tested it out myself, it did not modify the "Recent Xbe files" for multiple xbe loads.

@ergo720 You have my approval. Congrats! 🎉

@PatrickvL Please merge this whenever you are satisfy with this change.

RadWolfie:
Side note: It did not resolve the window title's name change when certain title has reboot to dashboard. Yet it is a separate issue anyway.

Issue #723, opened 2017-09-09
TODO: HLE Database v2 Progress


RadWolfie:
The following list of OOVPA database we need work on to finish moving into HLE database v2 table.
- [ ] XAPILIB
  - [x] Copy all OOVPA list into v2 table.
  - [x] Fix all OOVPAs to match with same address found in v1 table.
  - [x] Fix OOVPA's XDK revision to lowest known match.
  - [ ] Add missing OOVPAs if any has occured.
- [ ] D3D8
  - [x] Copy all OOVPA list into v2 table.
  - [x] Fix all OOVPAs to match with same address found in v1 table.
  - [x] Fix OOVPA's XDK revision to lowest known match.
  - [ ] Add missing OOVPAs if any has occured.
- [ ] DSOUND
  - [x] Copy all OOVPA list into v2 table.
  - [x] Fix all OOVPAs to match with same address found in v1 table.
  - [x] Fix OOVPA's XDK revision to lowest known match.
  - [ ] Add missing OOVPAs if any has occured.
- [ ] XGRAPHC
  - [x] Copy all OOVPA list into v2 table.
  - [x] Fix all OOVPAs to match with same address found in v1 table.
  - [x] Fix OOVPA's XDK revision to lowest known match.
  - [ ] Add missing OOVPAs if any has occured.
- [ ] XNET **(Is found with XNET, XNETS, and XONLINES library header)**
  - [x] Copy all OOVPA list into v2 table.
  - [x] Fix all OOVPAs to match with same address found in v1 table.
  - [ ] Fix OOVPA's XDK revision to lowest known match.
  - [ ] Add missing OOVPAs if any has occured.
- [ ] XONLINE **(Is found with XONLINE and XONLINES library header)**
  - [x] Copy all OOVPA list into v2 table.
  - [x] Fix all OOVPAs to match with same address found in v1 table.
  - [x] Fix OOVPA's XDK revision to lowest known match.
  - [ ] Add missing OOVPAs if any has occured.
- [ ] XACTENG
  - [x] Copy all OOVPA list into v2 table.
  - [x] Fix all OOVPAs to match with same address found in v1 table.
  - [ ] Fix OOVPA's XDK revision to lowest known match. **(Last known revision found is 4928)**
  - [ ] Add missing OOVPAs if any has occured.

LukeUsher:
I would not merge XNET and XONLINE, simply because they are unrelated APIs.

[...]

PR #758, opened 2017-10-07
Introduce XNET, XONLINES database v2

jarupxx:
Known revision found is
　XNETS 3911, 4039, 4361, 4721
　XNETN 5344
　XONLINES 4831 to 5849
Checked other revisions by the Dxbx Patterns.


PatrickvL:
Well done again!

I do wish to ask a favor;

Please insert an empty line before `#if` and after `#endif`, otherwise the code gets cramped.
(Empty lines also result in better output in difference-compare-tools.)

Thanks in advance.

RadWolfie:
I'm providing a share research to know what's the difference between XOnline, XOnlines, etc.

XOnline Library Versions for Xbox Live

| Library | Description |
| ------- | ------------- |
| XOnline | Standard |
| XOnlined | Debug |
| XOnlines | Secure |
| XOnlinesd | Secure debug |

XNet Library Versions

| Library | Description |
| ------- | ------------- |
| XNet | Standard |
| XNetd | Debug |
| XNets | Secure |
| XNetsd | Secure debug |
| XNetn | "Neutered" (only for demos of Xbox Live titles) |
| XNetns.lib | Neutered secure debug(?) | 

Anyhow, I had researched some titles and found out as mentioned in #723, some libraries actually require yet is not mentioned in library versions array within the title's xbe such as below.
```
//{ { Lib_XONLINES , LIB_XONLINE }, Sec_XONLINE, XONLINE_OOVPAV2, XONLINE_OOVPA_SIZEV2 },

//{ { Lib_XONLINES, LIB_XNETS, LIB_XNET }, Sec_XNET, XNET_OOVPAV2, XNET_OOVPA_SIZEV2 },
```

By not using the above method will cause problem with most of the titles. Except, I did not implement support for it...

P.S. For myself and others, `{ Lib_XONLINES,{ Sec_text, Sec_XONLINE`_**`, Sec_XNET `**_`}, XONLINES_OOVPAV2, XONLINES_OOVPA_SIZEV2 },` is not the correct way to do it with Sec_XNET included since it is a different library.

src/CxbxKrnl/HLEDataBase.cpp
```diff
@@ -209,7 +204,7 @@ const HLEDataV2 HLEDataBaseV2[] = {
     //{ Lib_XONLINE,{ Sec_XONLINE }, XONLINES_OOVPAV2, XONLINES_OOVPA_SIZEV2 },
 
     // Fun fact, XONLINES are split into 2 header sections.
-    //{ Lib_XONLINES,{ Sec_XONLINE, Sec_XNET }, XONLINES_OOVPAV2, XONLINES_OOVPA_SIZEV2 },
+    { Lib_XONLINES,{ Sec_text, Sec_XONLINE, Sec_XNET }, XONLINES_OOVPAV2, XONLINES_OOVPA_SIZEV2 },
```
RadWolfie: Okay, now I'm asking what's the reason to have Sec_text included? Like which function is inline?

PatrickvL: I have no answer, but inlined functions are worrisome, because *if* we need to parish them, how can we ever be sure there's only one occurrence in the Xbe?

LukeUsher: Since XAPI functions exist in .text rather in an XAPI segment, it's very possible that other libraries add functions to .text too.

I don't think this necessarily means the functions are being inlined, as that would alter the signature.

PatrickvL: Okay, then we need another term;

Inline is when a function is embedded into the caller. Which means multiple callers result in multiple inclusions.

But right now we're taking about something else. Namely, when a function is linked into an Xbe (but not in it's own section), and not inlined, thus it only appears once. What shall we call these?
"Out-of-section functions" is descriptive and clear, but too long.

We could call them "embedded functions", "roaming functions" or "foreign functions". Maybe something else - what do you guys think?

RadWolfie: @PatrickvL TBD (To be determined), it still need more research and accuracy data before we can say "yeah, it is embedded/inline function". My wip research about this XOnline and XNet libraries are showing otherwise.

RadWolfie:
@LukeUsher, I remember you were saying XOnline and XNet are completely separate thing. My current wip research and committed in this PR, are showing it differently. Both XOnline and XNet indeed have:

(XNet OOVPA registers, yet is registers in XOnline)
- XnInit (introduced in later revision)
- WSAStartup
- XNetStartup
- XNetGetEthernetLinkStatus
- bind
- connect
- listen
- recv
- ioctlsocket
- socket

I'll investigate this with Unreal Championship title base on @jarupxx's note stating all are found.

src/CxbxKrnl/HLEDataBase.cpp
jarupxx: XNetGetEthernetLinkStatus function has it in Sec_text. Other functions seem has it in Sec_XONLINE or Sec_XNET.

jarupxx:
I inserted an empty line before #if and after #endif, it looks good!

jarupxx:
@RadWolfie ```{ Lib_XONLINES,{ Sec_text, Sec_XONLINE, Sec_XNET}, XONLINES_OOVPAV2, XONLINES_OOVPA_SIZEV2 },``` is this wrong? 
OOVPA refers to XNET, It is intended to be OOVPA of XONLINES exactly.
I do make generic OOVPA XONLINES plus XNET, because the signatures are similar.

src/CxbxKrnl/HLEDataBase.cpp
RadWolfie: Strange, I can confirm what jarupxx is saying about XNetGetEthernetLinkStatus is in Sec_text section. It is indeed match perfectly.

RadWolfie:
Here's what I discover by choosing specific section to scan.

text:
- XNetGetEthernetLinkStatus

xnet:
- WSAStartup
- XNetStartup
- XnInit
- bind
- connect
- ioctlsocket
- listen
- recv
- send
- socket

xonline:
- CXo_XOnlineLogon
- XOnlineLogon
- XoUpdateLaunchNewImageInternal

LukeUsher is pretty much right. XOnline for any xonline functionality (doesn't matter if it has bind, listen, recv, etc functions in it). XNet for any general networking usage.

I find it odd how Xbox dev team process this into xbe build a bit differently than I thought.

src/CxbxKrnl/HLEDataBase.cpp
RadWolfie: @PatrickvL turns out it is not inline function. Odd huh?

RadWolfie:
@jarupxx What I think is XOnline__ library is built right on top of XNet library. Aka a combo. Currently it is correct. Except... we need to separate XNet and XOnline identically. Something like this...

```c
//XOnline section
REGISTER_OOVPAS(XOnlineLogon, PATCH, 4361),
...

//XNet section
REGISTER_OOVPAS(bind, PATCH, 3911, 4627),
...
```

What do you guys think?

@LukeUsher I don't think we can avoid XNet OOVPAs registrations being include in XOnline OOVPAs registrations? Since it _might_ be randomly include of those 3 sections over time? Please provide feedback if you can to straighten this out by a solution.

jarupxx:
@RadWolfie I got it. Added a note for reader.

RadWolfie:
Merged since no other solution is available.

PR #761, opened 2017-10-08
Reinstate JSRF Hack

LukeUsher:
Annoyingly, this is still necessary.

Without this, we have a choice between JSRF working and some titles not detecting input
Or JSRF not working, but many titles detecting input correctly.


ladsoftware:
clearly, JSRF > ALL.

Issue #678, opened 2017-08-22
Remove remnant visual elements after emulation stops/crashes


PatrickvL:
Once emulation stops, all visual emulation output (like FPS counter and LED states) should not be drawn anymore.
(See https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/574#issuecomment-324013493 for more details).

If this is already working correctly when emulation is terminated manually, then check if emualtion crashes are handled correctly too. If not, fix those by somehow detecting the crash.

LukeUsher:
One possibility would be storing the process ID of the child emulation process we create, then checking it's status in a polling loop (based around some kind of timer so it doesn't waste resources). If the process no longer exists, we can assume it has either crashed, or emulation has been stopped through other means.

Note that this may cause problems for titles with multiple xbe's.. When the emulator needs to load a new Xbe, it kills the current process and starts a new one. This situation would need to be handled, perhaps by using something in EmuShared to enable/disable the process monitoring.

PatrickvL:
Good idea. To support XBE's launching others, we could put the ProcessID of the most recently launched XBE in EmuShared. That way the GUI executable can keep monitoring all of them.

PatrickvL:
To check if a process is still running, call [GetExitCodeProcess](https://msdn.microsoft.com/en-us/library/ms683189(VS.85).aspx), it'll return `STILL_RUNNING`

PatrickvL:
Fixed by @ergo720 in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/763

PR #763, opened 2017-10-10
 Crash detection implementation

ergo720:
This fixes issue #678. From my tests, this is able to detect all crashes that can occur in cxbx, included those already caught by cxbx (the cxbx crash popup is displayed), LTCG errors and Dr Watson crashes. The only case I couldn't test is when cxbx just stops emulating without displaying any error. Constantine used to do this and I tried to build a past version with my implementation but curiously I can't reproduce it now. Also note that this doesn't fix the dashboard issue that @RadWolfie was talking about in #757 and that is because the title is rebooting and not crashing. See the example below
dash


PatrickvL:
Well done, thank you!

Issue #746, opened 2017-09-24
Use more meaningful names for HLE cache files


ergo720:
---
Currently the HLE cache files are named with an alpha-numeric string. This is very inconvenient if the user wants to upload the cache file to report compatibility issues because as cxbx creates more and more files it will become quite time consuming to find the one corresponding to the game you are searching since you have to open them one by one to check for the name. I suggest to use the name of the game instead of the current string used, if possible.

---

### Screenshots [Errors, Graphics, Glitches, ECT]
![hlecache](https://user-images.githubusercontent.com/30000751/30783726-3a0108a8-a148-11e7-917e-05e00740247d.PNG)

---

RadWolfie:
I wish it is that simple, it's not. Some titles such as Innocent Tears, Splinter Cell: Pandora Tomorrow, etc are multi-xbe bootable with same title names. Here's alternate solutions for this workaround:
1. Show the HLE cache file name of loaded/running xbe in Cxbx's Window (somewhere visible enough).
2. Simplify add Title's name before the alpha-numeric string with a dash.
3. Or **really** simple solution, use your search bar with "file contents" checked from "Advanced options" list. (Windows 8 and below may have different keyword instead of "Advanced options". Use your favorite search engine to find out.)

ergo720:
I opt for the second solution, that way the cache files can also be sorted alphabetically according to the title's name in explorer. The third one is temporary workaround for the moment, though.

PR #764, opened 2017-10-11
Fix for issue #746

ergo720:
I tested games with UNICODE characters (The Return of the King™), with illegal characters (Tenchu: Return from Darkness has a colon) and multi-xbe titles (DEAD OR ALIVE ULTIMATE). The generated cache files are the same (except for the date which is normal) and the clear cache options work as well. 
I also added myself to the list of contributing devs: if you don't want, tell me and I will remove it
[cache-files.zip](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1376753/cache-files.zip)

### DIFF
--- a/src/Common/Xbe.cpp
+++ b/src/Common/Xbe.cpp
@@ -1004,3 +1004,12 @@ void *Xbe::FindSection(char *zsSectionName)
 
 	return NULL;
 }
+
+void Xbe::PurgeBadChar(std::string &s, const std::string &illegalChars)
+{
+	for (auto it = s.begin(); it < s.end(); ++it)
+	{
+		bool found = illegalChars.find(*it) != std::string::npos;
+		if (found) { *it = ' '; }
+	}
+}
--- a/src/Common/Xbe.cpp
+++ b/src/Common/Xbe.cpp
@@ -1010,6 +1010,6 @@ void Xbe::PurgeBadChar(std::string &s, const std::string &illegalChars)
 	for (auto it = s.begin(); it < s.end(); ++it)
 	{
 		bool found = illegalChars.find(*it) != std::string::npos;
-		if (found) { *it = ' '; }
+		if (found) { *it = '_'; }
 	}
 }
--- a/src/CxbxKrnl/HLEIntercept.cpp
+++ b/src/CxbxKrnl/HLEIntercept.cpp
@@ -144,7 +144,6 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
 	CxbxKrnl_Xbe->PurgeBadChar(szTitleName);
 	sstream << cachePath << szTitleName << "-" << std::hex << uiHash << ".ini";
 	std::string filename = sstream.str();
-	setlocale(LC_ALL, "");
 
 	if (PathFileExists(filename.c_str())) {
 		printf("Found HLE Cache File: %08X.ini\n", uiHash);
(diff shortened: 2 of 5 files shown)


src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -137,8 +137,14 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
...
 	std::stringstream sstream;
-	sstream << cachePath << std::hex << uiHash << ".ini";
+	char tAsciiTitle[40] = "Unknown";
+	setlocale(LC_ALL, "English");
+	wcstombs(tAsciiTitle, g_pCertificate->wszTitleName, sizeof(tAsciiTitle));
+	std::string szTitleName(tAsciiTitle);
+	CxbxKrnl_Xbe->PurgeBadChar(szTitleName);
+	sstream << cachePath << szTitleName << "-" << std::hex << uiHash << ".ini";
 	std::string filename = sstream.str();
+	setlocale(LC_ALL, "");
```
RadWolfie: I don't think this is a requirement? `setlocale` is also called below and does not make another call later on in the function for revert to empty string. Just wondering.

src/Common/Xbe.cpp
```diff
@@ -1004,3 +1004,12 @@ void *Xbe::FindSection(char *zsSectionName)
 
 	return NULL;
 }
+
+void Xbe::PurgeBadChar(std::string &s, const std::string &illegalChars)
+{
+	for (auto it = s.begin(); it < s.end(); ++it)
+	{
+		bool found = illegalChars.find(*it) != std::string::npos;
+		if (found) { *it = ' '; }
```
RadWolfie: Instead of space, I think it should be underline.

src/CxbxKrnl/HLEIntercept.cpp
ergo720: I wasn't sure, I put it just in case but probably yes, it can be removed

src/Common/Xbe.cpp
ergo720: But if I do that for Tenchu you will get Tenchu_ Return from Darkness and I think that's odd

RadWolfie: It may look odd, it also tells us this title has a invalid character in it. I think Windows change it to underline by default. I could be wrong... I had seen it happen several times, just don't remember where I saw it.

RadWolfie: Yeah, I can confirm the illegal character convert into underline with chrome browser by saving a webpage.

ergo720: But I think we can change all spaces to underline chars

src/CxbxKrnl/HLEIntercept.cpp
ergo720: Confirmed: it works as well

src/Common/Xbe.cpp
RadWolfie: Then it will look more odd. 😨

ergo720: I can't leave the illegal characters in the filename, if I do that WritePrivateProfileString will stop at the first illegal char. For example, for Tenchu you will get just Tenchu and the remaining part will be truncated

ergo720: Ok. i did like you said

RadWolfie: | With underline of illegal character | With your suggestion with all spaces |
| --- | --- |
| `Tenchu_ Return from Darkness` | `Tenchu__Return_from_Darkness` |

WritePrivateProfileString accept spaces...

It's best to replace only the illegal characters for this function.

RadWolfie:
LGTM

PR #772, opened 2017-10-20
Fix an issue where xbe logo images failed to load

LukeUsher:


PatrickvL:
Thanks! Also to @CakeLancelot for reporting the issue

PR #773, opened 2017-10-23
HLE v2 XACT library leave UNPATCHED

jarupxx:
HLE v2 XACT library is causing more regressions. See issue Cxbx-Reloaded/game-compatibility#410


LukeUsher:
We need to do more research into XACT to determine what patches are required, and which patches can be discarded, but for now, this is better than regressions.

PR #778, opened 2017-10-27
MemoryManager : Fixed block-lookup with non-base addresses, by using …

PatrickvL:
MemoryManager : Fixed block-lookup with non-base addresses, by using block.upper_bound as key.

This resolves crashes in Cartoon sample (and potentially many other titles).


RadWolfie:
Doesn't solve JSRF, Lego Star Wars, Mafia, and some other titles. JSRF seem to be "slightly" improved, I'm not sure. Lego Star Wars still crash as before.

PR #780, opened 2017-10-28
DSound Major Improvement + Cleanup

RadWolfie:
- Major improvement overall.
- Reduced duplicate as possible into one inline function each.
- SetLoopRegion and SetPlayRegion are now supported.
- And major clean up.

WaveBank Sample now play audio properly. Some titles do have improvements such as Gaunlet: Dark Legacy.

DS Buffer's StopEx needs more research to finalize its purpose base on flag pass down.

### DIFF
--- a/src/CxbxKrnl/EmuDSound.cpp
+++ b/src/CxbxKrnl/EmuDSound.cpp
@@ -1351,7 +1351,7 @@ HRESULT WINAPI XTL::EMUPATCH(IDirectSoundBuffer_Play)
         if (pThis->EmuDirectSound3DBuffer8 != nullptr) {
             DSound3DBufferCreate(pThis->EmuDirectSoundBuffer8, pThis->EmuDirectSound3DBuffer8Region);
         }
-        DSoundBufferTransfer(pThis->EmuDirectSoundBuffer8, pThis->EmuDirectSoundBuffer8Region,
+        DSoundBufferTransferSettings(pThis->EmuDirectSoundBuffer8, pThis->EmuDirectSoundBuffer8Region,
                              pThis->EmuDirectSound3DBuffer8, pThis->EmuDirectSound3DBuffer8Region);
 
         hRet = pThis->EmuDirectSoundBuffer8Region->Lock(0, 0, &pThis->EmuLockPtr1, &pThis->EmuLockBytes1, nullptr, nullptr, DSBLOCK_ENTIREBUFFER);
@@ -1453,7 +1453,7 @@ extern "C" HRESULT __stdcall XTL::EMUPATCH(IDirectSoundBuffer_StopEx)
                 if (pThis->EmuDirectSoundBuffer8Region != nullptr) {
                     pThis->EmuDirectSoundBuffer8Region->GetStatus(&dwStatus);
 
-                    DSoundBufferTransfer(pThis->EmuDirectSoundBuffer8Region, pThis->EmuDirectSoundBuffer8,
+                    DSoundBufferTransferSettings(pThis->EmuDirectSoundBuffer8Region, pThis->EmuDirectSoundBuffer8,
                                          pThis->EmuDirectSound3DBuffer8Region, pThis->EmuDirectSound3DBuffer8);
 
                     hRet = pThis->EmuDirectSoundBuffer8Region->Stop();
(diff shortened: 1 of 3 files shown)


Sarkie:
SetPlayRegion -; affected a few homebrew I was testing.

Great work!

RadWolfie:
Can you name one or two affected homebrew? Maybe I can fix it quicker. Sorry it broke for you.

Sarkie:
No sorry, laptop is packed away, I might have time on Tuesday to look for you. Cheers

mrcmunir:
Broken in the menu when init sound  jet set radio future is caused by Get Status First, Stop Buffer, Then CurrentPosition!?

captura de 2017-10-29 23-21-33

RadWolfie:
Strange, I'm not getting any issue with titles except for JRSF due to D3D8 crash. I'll have to find out what went wrong later once HLE database is more complete or at least playable on my side.

Sarkie:
That is running in Wine. Fyi.

mrcmunir:
Yes is running under wine Cxbx-reloaded is supported in Wine 2.18 and higher.
before this commit no crash Soundbuffer in that game.
Only happen some graphics glitches in some scenes/zones and UI black as known issue reports .

Sarkie:
@mrcmunir Yes understood, but please state that, Wine supports Cxbx-Reloaded, but Cxbx-Reloaded doesn't say it is supported in Wine.

CakeLancelot:
For reference I get the same crash on Windows, so it's not a wine-specific issue:
cxbx_2017-10-30_12-50-39

RadWolfie:
@CakeLancelot, thanks for confirming it has duplicate on Windows platform. Since you like to "dabble" with reverse engineering. Think you can make a breakpoint where "CreateSoundBuffer Failed" message popup inside EmuDSoundInline.hpp file ([here's the source to find](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/master/src/CxbxKrnl/EmuDSoundInline.hpp#L381)). Then run the JSRF again to find out which previous caller is the cause?

CakeLancelot:
> make a breakpoint where "CreateSoundBuffer Failed" message popup inside EmuDSoundInline.hpp file . Then run the JSRF again to find out which previous caller is the cause?

Sure, here's the result:
``Cxbx.exe!XTL::EmuPatch_IDirectSoundBuffer_Play(XTL::X_CDirectSoundBuffer * pThis, unsigned long dwReserved1, unsigned long dwReserved2, unsigned long dwFlags) Line 1367	C++	Symbols loaded.``

[Here is the line it appears to be referring to.](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/master/src/CxbxKrnl/EmuDSound.cpp#L1367)

Hopefully this can help you, again I just like to dabble/experiment at times. Let me know if I did something wrong and I'll try to correct it.

RadWolfie:
Thanks! So, it is cause by introduction of SetLoop/SetPlayRegion.

Just one things I would like to ask from you. Can you tell me what `pThis->EmuBufferToggle` and `byteLength` values are? I'm thinking byteLength is too small.

CakeLancelot:
> Just one things I would like to ask from you. Can you tell me what pThis->EmuBufferToggle and byteLength values are? I'm thinking byteLength is too small.

pThis->EmubufferToggle is X_DSB_TOGGLE_LOOP (2)
bytelength is 0

RadWolfie:
Thanks for the detail I requested. It actually help me get the concept of the problem. 😄

P.S. DirectSoundBuffer creation hates zero byte length. 😨 😉

PatrickvL:
What would be most interesting to know, it's where this zero length comes from; if we somehow influenced that, the issue resides mainly there, not exclusively at the receiving end.

RadWolfie:
It's not caused by Cxbx-R, it is base on xbox's SetLoopRegion. If we "give" the length of 0 to SetLoopRegion function, it will use the whole buffer in that range (including starting point of the offset).

mrcmunir:
screenshot_233

Working again JSRF no SoundBuffer crash with Fix Loop/Play Region Length Requirement  testing under Windows 10 .

RadWolfie:
Has the audio improve or is not playing audio properly like it should had before?

mrcmunir:
@RadWolfie  Well now can't hear the sound the jumps of the skates into on railing  before if I listened to them fine

Songs and voices work the same as before

PatrickvL:
I spotted something at https://github.com/RadWolfie/Cxbx-Reloaded/blob/40030599665b11ed89ae5213251ffb154c88d609/src/CxbxKrnl/EmuDSound.cpp#L1381 : EmuDirectSoundBuffer8 could be overwritten is if was already set.

RadWolfie:
@mrcmunir Alright, I'll add it to my todo list. Thanks for the report! 😃

Not necessary overwritten, just using wrong DS buffer. _Should_ be using DS's region buffer instead. Also found another code fault for same thing. Oops! I add another fix for it in the branch btw.

PR #781, opened 2017-10-29
XREF Cleanup

jarupxx:
- Split up the count XREF and Offset-Pair.
- Rename full-uppercase XREF id's.
- Remove unused XREF entry.


PatrickvL:
Thank you very much!

One remark though:

Officially, the term is `TextureStageState`, not `TextureState`.
It's no problem for now, everyone will understand what it means.
But if there's maintenance being done on D3D OOVPA again, this could be updated at that time.

PR #785, opened 2017-11-03
Another DSound Improvements

RadWolfie:
- Titles with 5455/5558 should be able progress further than before.
- Added almost all stubs.
- Fixes DSound from previous pull request.

Verified with:
- [3936] Silent Hill 2
- [4039] Nightcaster
- [4134] Blood Omen 2
- [4242] Gauntlet Dark Legacy
- [5233] Futurama
- [5455] Ghost Recon: Island Thunder
- [5558] Dino Crisis 3


jarupxx:
I approval the CHRTFSource_SetLightHRTF* / CHRTFSource_SetFullHRTF* OOVPA. Those 0xC7 are really be used.

RadWolfie:
It is correct, all it does is replace the virtual table with different function methods which serve same purpose, just differently. By using 0xC7 repeatly with ending of specific value like 1, 2,3 and 3. I do not need to make revision by revision for another xref.

PatrickvL:
Nice!

RadWolfie:
In fact, if one do find a title with all 4 of DirectSound____HRTF(4Channel) and all of DirectSoundCreate(______) detection. It is 90% more likely a complete library compiled. One thing I do regret, I had not implement 5233 and older for detecting all of DirectSound___HRTF(4Channel) since I had not find one has it yet.

PatrickvL:
Thanks, I've merged it as is, testers seem to be glad, so, good work as always!

PatrickvL:
As for finding titties for specific xdk versions, our compatibility issue tracker might give some suggestions, just filter on label xdk-5233 and you'll get a list of games for that. Don't know if they're useable though...

RadWolfie:
That's what I had been using to find suitable titles to work with. 😉

PR #787, opened 2017-11-07
Transfer Updates From WIP_LessVertexPatching

RadWolfie:
This should cover most of the transferable into master with no regression.

All of D3D changes (unrelative to D3D OOVPAs) are not included in this pull request.

Team, please verify your changes are correct before approve the review.


PatrickvL:
I've reviewed this, thus far it looks all right. But since I'm the one that did the original changes, I won't merge it. I've asked Luke to do a review too, with some luck this will be done within a day or two.

RadWolfie:
Okay, then please go in `files changed` tab. Click green "review" button at top right-ish corner. Choose approve then finish your review. That way we know if a review is approved or not. You do not have to go in a file just to make an approval.

PatrickvL:
Thanks for explaining the review process - normally I'm interacting with github via OctoDroid on my mobile, which doesn't have these buttons. I'll use the website for this. Cheers!

Sarkie:
Today I learnt

RadWolfie:
Off topic,

I notice OctoDroid now has pull request review support (partially?) since yesterday base on their master branch. Hopefully it will get into play store within a month or two for you able to use the new feature. 😉

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -616,13 +722,13 @@ const char *GameRegionToString(DWORD aGameRegion)
 	};
 
 	if ((aGameRegion & ~XBEIMAGE_GAME_REGION_KNOWN) > 0)
-		return "REGION ERROR";
+		DbgPrintf("REGION ERROR! (0x%X)\n", aGameRegion);
```
RadWolfie: This does not look right, it should return here. Not output a debug message then continue on?

PatrickvL: Actually, you're right. About a week ago we discovered that the code that restored a few required image header values, actually damaged the certificate exactly at the region indicator. For that I added the DbgPrintf to avoid erroring out too soon. But now that we use a copy of the certificate, this can once again become an error return.

RadWolfie:
I didn't leave anything behind from PatrickvL's code for D3D8 OOVPAs. Only performed a wrapper for unset offset. What I did was copy raw files right on top of master's files. Then perform some wrapper for D3D8 OOVPAs, couple fixes to kernel files, and carry over some HLEIntercept necessary to preserve newer offsets process to work with D3D8 OOVPAs detection.

Though, I'll make another check with D3D8 on 4920 dashboard and AvP title.

RadWolfie:
My finding for 4920 dashboard is missing:
- D3DDevice_SetTexture
- D3DDevice_SetRenderTarget

As for AvP title, I don't have a problem with it?

RadWolfie:
AvP seems to have a different problem. Even if I re-add sleep timer, it still do the same crash anyway. Right now I'm stumped with confusion.

RadWolfie:
After 3 bisects attempt, I cannot reproduce the issue. Somehow, there's a missing patch since it's calling to EmuX86_Write32Aligned function. (This pull request is not missing any detection in hle cache for AvP title by the way.)

I can rarely get in-game with this pull request. Little bit more able to enter in-game with current master. I believe this pull request is processing emulation little bit more faster than master branch. Plus is not relative to sleep timer issue.

Any thoughts?

PatrickvL:
From what Luke and you wrote, AvP not getting in-game anymore sounds much like a timing issue. Beats me where it comes from though.

Sarkie:
@RadWolfie 
Master should get in game and be playable.

Can you roll back and test after each commit to see which one or which combination you can't get in game?

Next is timings, maybe we need to figure out the speed of the hardware and create "slow" proxy versions if the system is running/reading too fast. Sleeps don't really scale.

RadWolfie:
@Sarkie 
Cxbx-R is in early stage of development. Titles will regress at any time. And I can reproduce it in master. Bisect isn't helping for me as mention in my previous comments.

@PatrickvL, yes the title is accurate. In if / else, the `if` does not have open and close bracket. Which is causing this issue in debug build. In fact, the old code is not part of coding standard too.

PatrickvL:
> I can reproduce it in master
Did you bisect master too?

RadWolfie:
For over 200 commits? No thanks. Beside, before HLE v2 database were merge and all the changes afterward does not resolve the issue. I'm going to do something differently since one of the asm code attempt access a pointer is zero'd when it happen inside rdata or text section (can't remember which one).

This PR can be merge for now since none of the retail titles, even Turok Evolution, has regress on my end.

PatrickvL:
Log2(200) is only 8 steps, but I might be overlooking an important fact

RadWolfie:
The only Log2 I found is in EmuDSoundInline.hpp file, however is not the cause of increase of regress. Where did you find Log2(200) from?

PatrickvL:
Oops. Miscommunication. You indicated bisect over 200 commits would be too cumbersome. But bisect only takes log2(number of commits) steps. So even bisecting 1024 commits would only take 10 steps

RadWolfie:
Understood, however I believe it is only happening to player choosing marine character. Alien and Predator characters does not cause any problem. I was choosing marine every time which rarely able to get in-game.

So the issue is not relative to this pull request. I'll make a comment for the AvP homebrew issue ticket.

LukeUsher:
I'm sorry for taking so long on this: I didn't get the notification that the dashboard had been fixed, I must've missed the email/it must've got caught by my spam filter..

I'm happy to merge in this state

PR #790, opened 2017-11-10
XGSetTextureHeader leave unpatched

jarupxx:
Fix regressions in Taz Wanted reported by @literalmente-game https://github.com/Cxbx-Reloaded/game-compatibility/issues/149#issuecomment-327531583


PatrickvL:
Before we merge this in, could you put a `#if 0 // Patch disabled` ... `#endif`-pair around the entire `XGSetTextureHeader` function - both in the .cpp *and* in the .h file?

jarupxx:
D3DDevice_CreateTexture calls XGSetTextureHeader function not ready disabled this function.

PatrickvL:
Ah, then it's good like this. Thanks!

RadWolfie:
@PatrickvL Your request is no longer necessary. All it needs is UNPATCHED in inside central OOVPA v2 database. Comment `FUNC_EXPORTS` out should not be done either. We need to allow ALL `FUNC_EXPORTS` usable in order for asking tester, ourselves to re-test for needing to restore the patch or for troubleshooting. Not re-compiling every time to find out which patch needs to restore or needs to be disable.

RadWolfie:
See issue #792

PR #802, opened 2017-11-20
CxbxCreatePartitionHeaderFile : Prevent NULL-pointer error

PatrickvL:
This PR is an attempt to fix a problem that seems occur only on Windows 7 (not Windows 10), somehow resulting in Cxbx no longer being able to load Xbe's...


PatrickvL:
Indeed, this change turns out to help, so I'm merging this.

It's reported that when starting emulation with an empty %appdata% folder, this will still result in a failure, but that disappears at following starts.

PatrickvL:
So, there's still an issue when starting emulation without any app data.
I believe it's got something to do with generating the "partition0.bin" file for the first time.

PR #814, opened 2017-11-29
Properly implement D3DDevice_GetDisplayFieldStatus patch

LukeUsher:
This also fixes an issue where the default eeprom was not created correctly

### DIFF
--- a/src/Common/EmuEEPROM.cpp
+++ b/src/Common/EmuEEPROM.cpp
@@ -127,6 +127,7 @@ xboxkrnl::XBOX_EEPROM *CxbxRestoreEEPROM(char *szFilePath_EEPROM_bin)
 	// resetting FactorySettings.AVRegion and setting user video flags to the default
 	// The user can then set their desired settings using the Xbox Dashboard
 	if (pEEPROM->FactorySettings.AVRegion == AV_STANDARD_NTSC_M) {
+		DbgPrintf("INIT: Repairing bad EEPROM (from previous Cxbx-Reloaded builds)\n");
 		pEEPROM->UserSettings.VideoFlags = 0;
 		pEEPROM->FactorySettings.AVRegion = AV_STANDARD_NTSC_M | AV_FLAGS_60Hz;
 	}
--- a/src/Common/EmuEEPROM.cpp
+++ b/src/Common/EmuEEPROM.cpp
@@ -121,18 +121,28 @@ xboxkrnl::XBOX_EEPROM *CxbxRestoreEEPROM(char *szFilePath_EEPROM_bin)
 
 	// TODO : Verify checksums
 
+	// Check for (and fix) invalid fields that were set by previous versions of Cxbx-Reloaded
+	// Without this, all users would have to delete their EEPROM.bin
+	// The issue was that the AV_FLAG_XXhz was set in the wrong field, we fix it by
+	// resetting FactorySettings.AVRegion and setting user video flags to the default
+	// The user can then set their desired settings using the Xbox Dashboard
+	if (pEEPROM->FactorySettings.AVRegion == AV_STANDARD_NTSC_M) {
+		pEEPROM->UserSettings.VideoFlags = 0;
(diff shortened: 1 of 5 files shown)


LukeUsher:
I'd be interested to see if this properly solves the Azurik regression reported here: https://github.com/Cxbx-Reloaded/game-compatibility/issues/444

blueshogun96:
I'm sorry, but for me, the issue persists, even with this new commit.

Shogun

src/Common/EmuEEPROM.cpp
```diff
@@ -121,18 +121,28 @@ xboxkrnl::XBOX_EEPROM *CxbxRestoreEEPROM(char *szFilePath_EEPROM_bin)
...
 	// TODO : Verify checksums
 
+	// Check for (and fix) invalid fields that were set by previous versions of Cxbx-Reloaded
+	// Without this, all users would have to delete their EEPROM.bin
+	// The issue was that the AV_FLAG_XXhz was set in the wrong field, we fix it by
+	// resetting FactorySettings.AVRegion and setting user video flags to the default
+	// The user can then set their desired settings using the Xbox Dashboard
+	if (pEEPROM->FactorySettings.AVRegion == AV_STANDARD_NTSC_M) {
+		pEEPROM->UserSettings.VideoFlags = 0;
+		pEEPROM->FactorySettings.AVRegion = AV_STANDARD_NTSC_M | AV_FLAGS_60Hz;
```
PatrickvL: Maybe add a log message this had happened

src/CxbxKrnl/EmuKrnlAv.cpp
```diff
@@ -63,16 +63,40 @@ PVOID g_pPersistedData = NULL;
...
+	result = xboxkrnl::ExQueryNonVolatileSetting(
+		xboxkrnl::XC_VIDEO,
+		&type,
+		&userSettings,
+		sizeof(ULONG),
+		&resultSize);
+
+	// If this failed, default to no user-options set
+	if (result != STATUS_SUCCESS || resultSize != sizeof(ULONG)) {
+		userSettings = 0;
```
PatrickvL: Maybe add a message here too

src/CxbxKrnl/EmuKrnlAv.cpp
```diff
@@ -63,16 +63,40 @@ PVOID g_pPersistedData = NULL;
...
+	NTSTATUS result = xboxkrnl::ExQueryNonVolatileSetting(
+		xboxkrnl::XC_FACTORY_AV_REGION,
+		&type,
+		&avRegion,
+		sizeof(ULONG),
+		&resultSize);
+
+	// If this failed, default to AV_STANDARD_NTSC_M | AV_FLAGS_60Hz
+	if (result != STATUS_SUCCESS || resultSize != sizeof(ULONG)) {
+		avRegion = AV_STANDARD_NTSC_M | AV_FLAGS_60Hz;
```
PatrickvL: And here

jarupxx:
Good for me, solves the Azurik regression.

src/CxbxKrnl/EmuKrnlAv.cpp
LukeUsher: I don't think this case (and the other avRegion case) need any logging as this is the same behavior as the real xbox, so it's a situation that can happen, even on real hardware.

PatrickvL:
I've merged this, based on jarpuxx test results

Sarkie:
Could we have a repo where we can dump our system info?

Like the fact it works for one person and not the other.
Could be easier to spot why.

LukeUsher:
I managed to get test results from Azurik NTSC: It seems to behave differently in both release and debug builds. In release builds, it seems to crash, while in Debug builds, it boots and is able to play audio.

This leads me to the belief that it there must be some sort of race-condition that isn't being hit on slower machines, and the debug build works for some users due to the lack of compiler optimization.

PatrickvL:
Another explanation could be different allocation patterns due to logging; Because currently Xbox and host allocations might still go the same heap, code that access buffer bounds could damage functionality

PR #819, opened 2017-12-07
Build chain improvements

x1nixmzeng:
The old build chain had a few issues:

* The pre-build event to generate a version header would touch `Version.h` to write the git hash. This was bad because any files which included it needed to be recompiled!

* We called `Export.bat` after the build, which would gather all the files and ZIP them up! This took a few seconds!

To solve this I've made sure `Version.h` is only updated if the git hash is updated. I've also made sure only Appveyor will package files - see `AppveyorPackager.ps1` (although this is untested).


PatrickvL:
This is a very welcome change! Only, why does the AppVeyor build fail?

PatrickvL:
Thank you - beautifully done! I have no further remarks. Please tell if you want to add anything else.
Once we know you're done, this will be merged ASAP.
Thanks again

LukeUsher:
Thanks for doing this!

x1nixmzeng:
No worries! All that's left is the artifact mask - instead of the export folder, only the zip files!

x1nixmzeng:
It's ready to be merged!

LukeUsher:
Great work

Issue #788, opened 2017-11-07
Implement actual memory page management


PatrickvL:
Currently our memory management is flawed; Our memory manager processes contiguous memory separately from other memory, allocates other memory via heap malloc, and doesn't offer any kind of statistics for the kernel Mm API's.

For these issues, the above problems must be fixed by implementing memory page management.

This should include page table entries (PTE's) and page directory entries (PDE's), alongside the separate read/write/execute watches, described in issue https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/748.

Page table entries should reside at 0xC0000000, one DWORD per page, containing the upper 20 bits of the virtual address a page is mapped to, protection flags, end-of-allocation-range marker, and possibly other flags.

Page directory entries should reside at 0xC030000. TODO : Add more detail

As issue https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/769 describes, all required memory ranges should already be reserved. Allocating memory will retrieve free page numbers from the directory/table. The memory behind these pages numbers can be taken from a page 'pool' representation, in the form of a memory mapped file.

Memory mapped files under Windows, can be mapped partially (per page for example) to specific virtual addresses. Thus, each page from this 'pool' can be seen as a 'physical' page, which can be mapped into the appropriate virtual memory range (which were reserved at launch - see the other issue describing that in more detail).
Mapping will be done using MapViewOfFileEx.

A (part of a) memory mapped file can be mapped to multiple virtual addresses, so even mirror pages can be implemented.

Feasibility of the above should first be investigated in a small test application.

If the above can indeed be made to work, the kernel Mm API's should be updated to make use of the new page manager code, so that these API's will work much more like the real thing too.

PatrickvL:
Please let me explain what I was thinking (maybe you already know everything I'm going to write here, if so, please forgive me)

The great thing about MapViewOfFileEx is that it allows to do two very special things:

It can map any page from a memory mapped file to a virtual memory address that can be specified freely (the address range has to be unoccupied of course)

PatrickvL

23:48

It can map any page from a memory mapped file multiple timesto a virtual memory address (each time to another unoccupied virtual memory address range of course).

PatrickvL

23:55

[...]

PR #821, opened 2017-12-19
Virtual memory manager + Kernel memory functions implementation

ergo720:
This implements a new memory manager that supports the virtual page management requested in issue #788. I also rewritten the implementation of almost all the kernel memory functions to make them use the new manager. During my tests, I didn’t see neither improvements nor regressions with this new commit, but you have to delete not only the hle cache files but also all files/folders in the AppData
Folder of Cxbx-Reloaded otherwise you will get crashes from almost all the games, at least, this is what has happened to me. One thing: I don’t have any Chihiro games so I don’t know if the Chihiro initialization is done correctly, someone else with Chihiro games must test those before this is merged.


src/CxbxKrnl/EmuXapi.cpp
```diff
@@ -1042,12 +1042,12 @@ DWORD WINAPI XTL::EMUPATCH(XLaunchNewImageA)
 
 	// Update the kernel's LaunchDataPage :
 	{
-		if (xboxkrnl::LaunchDataPage == &DefaultLaunchDataPage)
-			xboxkrnl::LaunchDataPage = NULL;
-
 		if (xboxkrnl::LaunchDataPage == NULL)
 			xboxkrnl::LaunchDataPage = (xboxkrnl::LAUNCH_DATA_PAGE *)xboxkrnl::MmAllocateContiguousMemory(sizeof(xboxkrnl::LAUNCH_DATA_PAGE));
 
+		if (xboxkrnl::LaunchDataPage != NULL)
```
PatrickvL: Check if LaunchDataPage is assigned only once it's been allocated - if allocation fails, return an error.

src/CxbxKrnl/EmuXapi.cpp
```diff
@@ -1042,12 +1042,12 @@ DWORD WINAPI XTL::EMUPATCH(XLaunchNewImageA)
...
+			xboxkrnl::MmPersistContiguousMemory((PVOID)xboxkrnl::LaunchDataPage, PAGE_SIZE, TRUE);
```
PatrickvL: It might be nicer to call this after all fields is set.

src/CxbxKrnl/EmuNV2A.cpp
```diff
@@ -2764,7 +2767,7 @@ DEVICE_READ32(PFB)
 		result = 3; // = NV_PFB_CFG0_PART_4
 		break;
 	case NV_PFB_CSTATUS:
-		result = CONTIGUOUS_MEMORY_SIZE;
+		result = contiguous_memory_size;
```
PatrickvL: Checking for g_bIsChihiro here, would remove the need for this global variable.

src/CxbxKrnl/EmuKrnlXe.cpp
```diff
@@ -123,9 +156,35 @@ XBSYSAPI EXPORTNUM(328) xboxkrnl::NTSTATUS NTAPI xboxkrnl::XeUnloadSection
...
+		// Free the section and the physical memory in use if necessary
 		if (Section->SectionReferenceCount == 0) {
 			memset(Section->VirtualAddress, 0, Section->VirtualSize);
+
+			VAddr BaseAddress = (VAddr)Section->VirtualAddress;
+			VAddr EndingAddress = (VAddr)Section->VirtualAddress + Section->VirtualSize;
+
+			// Decrement the head/tail page reference counters
+			(*Section->HeadReferenceCount)--;
+			(*Section->TailReferenceCount)--;
```
PatrickvL: Does the part below work correctly if head and tail pointers are equal? Or would that require a slightly different ordering of statements?

src/CxbxKrnl/EmuKrnlMm.cpp
```diff
@@ -133,7 +139,7 @@ XBSYSAPI EXPORTNUM(166) xboxkrnl::PVOID NTAPI xboxkrnl::MmAllocateContiguousMemo
 	if (pRet != NULL)
 	{
 		// TODO : Allocate differently if(ProtectionType & PAGE_WRITECOMBINE)
-		pRet = g_MemoryManager.AllocateContiguous(NumberOfBytes, Alignment);
+		pRet = (PVOID)g_VMManager.Allocate(NumberOfBytes, LowestAcceptableAddress, HighestAcceptableAddress, NULL, Alignment, ProtectionType);
```
PatrickvL: Will this still return addresses in the contiguous address range (0x80000000 and up)?

src/CxbxKrnl/EmuKrnlMm.cpp
```diff
@@ -516,9 +487,9 @@ XBSYSAPI EXPORTNUM(180) xboxkrnl::ULONG NTAPI xboxkrnl::MmQueryAllocationSize
 {
 	LOG_FUNC_ONE_ARG(BaseAddress);
 
-	LOG_INCOMPLETE(); // TODO : Free PAGE_WRITECOMBINE differently
+	LOG_INCOMPLETE(); // TODO : Free PAGE_WRITECOMBINE differently (?? this function just queries the size, it doesn't free anything)
```
PatrickvL: This comment was probably supposed to appear elsewhere, as indeed it makes no sense here (better move or delete it)

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -1014,43 +1026,33 @@ void CxbxInitFilePaths()
...
-	FILE* fp = fopen(szFilePath_LaunchDataPage_bin, "rb");
-	if (fp)
-	{
-		// Make sure LaunchDataPage is writeable  :
-		if (xboxkrnl::LaunchDataPage == &DefaultLaunchDataPage)
-			xboxkrnl::LaunchDataPage = NULL;
-
-		if (xboxkrnl::LaunchDataPage == NULL)
-			xboxkrnl::LaunchDataPage = (xboxkrnl::LAUNCH_DATA_PAGE *)xboxkrnl::MmAllocateContiguousMemory(sizeof(xboxkrnl::LAUNCH_DATA_PAGE));
+	PAddr LaunchDataPAddr;
```
PatrickvL: Don't we have a fully functional kernel API for this now? If so, just call that instead of reinventing the wheel here ;)

Sarkie:
I have access to Virtua Cop 3 somewhere. I'll try and find it again.

src/CxbxKrnl/EmuKrnlXe.cpp
ergo720: I tested this with Broken Sword which triggers the calls to Deallocate and DeallocateOverlapped (head/tail are 1/1 and 2/2 before being decremented) and it worked as expected

src/CxbxKrnl/EmuNV2A.cpp
ergo720: Bit if I do this then it will have to check for g_bIsChihiro every time the function is called, which is a waste since this global bool never changes during the emulation

src/CxbxKrnl/EmuKrnlMm.cpp
ergo720: I returns an address from inside the aliased view of the contiguous range. I think it's just a matter to change m_Base to 0x80000000 and remove VMAType::MemContiguous from m_Vma_map (otherwise CarveVMA will conflict since it expects the region to be VMAType::Free). Do you want this?

src/CxbxKrnl/EmuNV2A.cpp
PatrickvL: True, but if this code isn't called very often, IMHO such a check is better than having a global variable. Not that it's all that important right now - it's great to have this coming in!

src/CxbxKrnl/EmuKrnlXe.cpp
PatrickvL: Excuse me, I didn't mean "have equal values", I meant : "what if these pointers are pointing to the same address?"

ergo720: That happens every time for each pair of adjacent head/tail pointers. Look at the xbe dumps

src/CxbxKrnl/EmuKrnlMm.cpp
PatrickvL: Contiguous memory **MUST** reside in high memory (from 0x80000000 onwards), and be identity mapped (meaning: virtual address = 0x80000000 + physical address), because the NV2A expects it to be that way. D3D resource buffer addresses must be CPU-accessible in virtual memory (with the high bit set) and at the same time they must be GPU-accessible in physical memory (with the high bit cleared).

src/CxbxKrnl/EmuNV2A.cpp
```diff
@@ -3743,6 +3746,8 @@ void EmuNV2A_Init()
 	// Allocate PRAMIN Region
 	VirtualAlloc((void*)(NV2A_ADDR + NV_PRAMIN_ADDR), NV_PRAMIN_SIZE, MEM_RESERVE | MEM_COMMIT, PAGE_READWRITE);
 
+	if (g_bIsChihiro) { contiguous_memory_size = CONTIGUOUS_MEMORY_CHIHIRO_SIZE; }
```
RadWolfie: @PatrickvL, contiguous_memory_size is check, then set at initialization process. Base on ergo720 and your discussion above.

src/CxbxKrnl/EmuNV2A.cpp
RadWolfie: I have to agree with @ergo720's code method for this. Plus if an xbe is Chihiro, there's no reason to re-check boolean every single time if xbe do decide revert back to normal xbox mode. Which IMHO should never happen at all. Use global values preset for either normal xbox or Chihiro "mode" is the proper way to do. It will reduce heavy load on CPU cycling and optimize better.

src/CxbxKrnl/EmuKrnlMm.cpp
```diff
@@ -114,6 +116,10 @@ XBSYSAPI EXPORTNUM(166) xboxkrnl::PVOID NTAPI xboxkrnl::MmAllocateContiguousMemo
 
 	PVOID pRet = (PVOID)1; // Marker, never returned, overwritten with NULL on input error
 
+	// size must be > 0
+	if (NumberOfBytes == 0)
+		pRet = NULL;
```
RadWolfie: It should be pRet = xbnullptr. NULL =/= nullptr in C++.

ergo720: But this function returns NULL on failure, not nullptr (which is xbnullptr). If I remember right, we should use NULL as return type for Xbox APIs and xbnullptr for the rest

src/CxbxKrnl/EmuNV2A.cpp
ergo720: @PatrickvL then what do I do now? I already changed this. do I revert it back to what is was before?

jackchentwkh:
I tried this new PR, and it failed to run Rallisport Challenge with this error
default
[KrnlDebug.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1575710/KrnlDebug.txt)

ergo720:
@jackchentwkh I need a debug kernel log to understand the crash, the release log doesn't show the logs from the manager. You can create one by running the game in a debug build of cxbx

src/CxbxKrnl/EmuKrnlMm.cpp
RadWolfie: C++
* PVOID == nullptr == (void*)0
* int,short,long,etc == 0 == NULL

In C++, NULL is equivalent to 0. However in C, NULL is (void*)0. This is the big difference between NULL and nullptr for two different language. @LukeUsher had brought this up in the past with hard evidence source. (I don't have the source.)

Also this is a xbox side which is using the address. Therefore should be xbnullptr. If we are managing our own code without xbox editing it directly, then it will be nullptr.

jackchentwkh:
Ok, switch to Debug Build. here are the logs.
RalliSport: 
[KrnlDebug.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1575750/KrnlDebug.txt)

VCOP3
default
[KrnlDebug.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1575754/KrnlDebug.txt)

ergo720:
@jackchentwkh The crash in VC3 comes from XeUnloadSection

jackchentwkh:
but the previous build can reach menu in VC3

ergo720:
XeUnloadSection now calls my manager function Deallocate, it means it's attempting to deallocate a section it doesn't know about

LukeUsher:
You can see in the attached VC3 log, that XeUnloadSection is being called on section address 00010540, which is previously loaded earlier in execution with a call to XeLoadSection: This means that the memory manager SHOULD be aware of this section, pointing to a bug.

PatrickvL:
Actually, this is the first time I've heard of XeUnloadSection being called by a title, might be worth a comment in code, for anyone in need of a test-case for it...

src/CxbxKrnl/EmuNV2A.cpp
PatrickvL: It's your code my friend ;)

I'd say, if this code is a bottleneck, then reading a global variable might be faster than a comparison that chooses between two constants. If it's not a bottleneck, I'd keep it like this.

src/CxbxKrnl/EmuKrnlXe.cpp
```diff
@@ -123,9 +156,35 @@ XBSYSAPI EXPORTNUM(328) xboxkrnl::NTSTATUS NTAPI xboxkrnl::XeUnloadSection
 		// Decrement the reference count
 		Section->SectionReferenceCount -= 1;
 
-		// Free the section if necessary
+		// Free the section and the physical memory in use if necessary
 		if (Section->SectionReferenceCount == 0) {
 			memset(Section->VirtualAddress, 0, Section->VirtualSize);
+
+			VAddr BaseAddress = (VAddr)Section->VirtualAddress;
+			VAddr EndingAddress = (VAddr)Section->VirtualAddress + Section->VirtualSize;
```
PatrickvL: Same as above : `EndingAddress` (if meant as "last _included_ byte") has an off-by-one error (needs a `- 1` correction).

src/CxbxKrnl/EmuKrnlXe.cpp
```diff
@@ -89,12 +90,44 @@ XBSYSAPI EXPORTNUM(327) xboxkrnl::NTSTATUS NTAPI xboxkrnl::XeLoadSection
...
+
+			if ((*Section->TailReferenceCount) != 0)
+			{
+				EndingAddress &= ~PAGE_MASK;
+			}
+
+			if ((*Section->HeadReferenceCount) != 0)
+			{
+				BaseAddress = (BaseAddress + PAGE_SIZE) & ~PAGE_MASK;
```
PatrickvL: Just a tiny proposal : Would `ALIGN_UP` and `ALIGN_DOWN` macro's be useful here?

src/CxbxKrnl/EmuKrnlMm.cpp
```diff
@@ -211,39 +221,19 @@ XBSYSAPI EXPORTNUM(169) xboxkrnl::PVOID NTAPI xboxkrnl::MmCreateKernelStack
...
-
-	if (NumberOfBytes & 0xFF) {
-		// Validate NumberOfBytes for alignment with the page size
-		CxbxKrnlCleanup("Assertion: '(NumberOfBytes & (PAGE_SIZE -1)) == 0' in MmCreateKernelStack()");
-	}
+	VAddr addr = NULL;
 
 	/**
 	* Function at present does not:
-	* - Create an additional guard PAGE_SIZE after allocation,
```
PatrickvL: It's alright to delete this comment, but it would help to add a comment to `VMManager::AllocateStack`, explaining that the call to `ReprotectVMARange` turns the end of the stack into a guard-page.

LukeUsher:
Please note that the Lowest/Highest address inputs to MmAllocateContiguousMemoryEx are based on Physical Address, not virtual: So when specifying a lowest acceptable address of 0x1000 and a highest address of 0x7FFFFFFF, this makes sure that the physical memory is located above 0x1000 and below 0x80000000. 

This has no effect on the VIRTUAL address this memory is located at: It can, and does still return virtual address offsets above 0x80000000, even with the highest address set to lower than that, just these addresses are allocated in Physical memory within the suggested range.

ergo720:
I'm temporarily closing this since I'm going to take a break for the upcoming holidays. I will reopen it in January. FYI, I made some progress on the overlapped allocations but it's still not good enough.

ergo720:
I reopen this because I think it's good enough at this point.
@jackchentwkh Can you please upload an Xbe dump of VirtuaCop 3? I ask because in the kernel dump of above I didn't see the Chihiro initialization log from the manager, and that sounds strange. If your dump is the same as the one present in the compatibility repository, then you have a debug build, and that explains why.
@PatrickvL UnmapRange must succeed on LaunchDataPage since XGetLaunchInfo calls MmFreeContiguousMemory on it and sets the launch page to NULL. This function is unpatched now so the title will do this for us. The only two multi-xbe games I have are Tenchu and DOA Ultimate, but unfortunately they either crash before calling it or they don't call it at all. I think it's the latter since, looking at the launch page in memory, I just see the name of the new xbe to launch and nothing else, meaning they don't have any data to read after rebooting.

ergo720:
Better now? Base address can only be larger than ending address if a section is so small that it fits in a single page. Because base address is increased of a page whereas ending address is rounded down, then the base will become bigger than the section end, meaning there is nothing to decommit since the section resides entirely in the shared page
sections

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+{
+	// This reserves a large enough memory region to map the second physical memory file view, and aligns the start address.
+	// It must be page aligned otherwise the mapping/unmapping functions will produce incorrect results and on
+	// debug builds they will assert.
+	UINT_PTR addr = (UINT_PTR)VirtualAlloc(NULL, CHIHIRO_MEMORY_SIZE + PAGE_SIZE, MEM_RESERVE, PAGE_EXECUTE_READWRITE);
+	if (!addr)
+	{
+		CxbxKrnlCleanup("VMManager: VirtualAlloc could not find a suitable region to allocate the second physical memory view!");
+	}
+	VAddr aligned_start = addr & ~(UINT_PTR)PAGE_MASK;
```
PatrickvL: Since this clears low bits, the resulting aligned address is lower than what's been allocated. This lower address might be partially in use. To prevent initialization failure in that case, it would be best to allocate one page extra, and round the starting address upwards from the resulting base address.

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+	m_Base = (VAddr)MapViewOfFileEx(
+		file_view,
+		FILE_MAP_READ | FILE_MAP_WRITE | FILE_MAP_EXECUTE,
+		0,
+		0,
+		CHIHIRO_MEMORY_SIZE,
+		(void*)aligned_start);
+	if (m_Base != aligned_start)
+	{
+		CxbxKrnlCleanup("VMManager: MapViewOfFileEx could not map the second physical memory view!");
```
PatrickvL: A small nit-pick: If the call succeeded, but didn't result in the expected address, it would still need unmapping. (However, it's acceptable to ignore this here, because the process will be ended anyway, so the OS will do this cleanup for us.)

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+	VMAIter upper_mem_vma_handle = CarveVMA(CONTIGUOUS_MEMORY_BASE + m_MaxContiguousAddress, 32 * PAGE_SIZE);
+	VirtualMemoryArea& upper_mem_vma = upper_mem_vma_handle->second;
+	upper_mem_vma.type = VMAType::Allocated;
+	upper_mem_vma.permissions = PAGE_EXECUTE_READWRITE;
+	upper_mem_vma.backing_block = AllocatePhysicalMemoryRange(32 * PAGE_SIZE, m_MaxContiguousAddress, XBOX_MEMORY_SIZE);
+	UpdatePageTableForVMA(upper_mem_vma);
+	m_ImageMemoryInUse += 32 * PAGE_SIZE;
+
+	// Allocate memory for the dummy kernel
+	// NOTE: change PAGE_SIZE if the size of the dummy kernel increases!
```
PatrickvL: Or use a define called KERNEL_SIZE already in preparation...

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+	Allocate(sizeof(DUMMY_KERNEL), XBE_IMAGE_BASE, XBE_IMAGE_BASE + PAGE_SIZE - 1, PAGE_SIZE, PAGE_EXECUTE_READWRITE, false);
+
+	// Map the tiled memory
```
PatrickvL: All these mappings are repeating the same pattern over and over again; It might be good to move the steps into a function (or macro), so these mappings can be done from within a loop, or at least in a condensed few one-liners.

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+	bios_vma.type = VMAType::DeviceBIOS;
+	UpdatePageTableForVMA(bios_vma);
+	m_NonImageMemoryInUse += BIOS_CHIHIRO_SIZE;
+
+	// Map the mcpx
+	// NOTE: Again, I can't use MCPX_BASE and MCPX_SIZE because those are not aligned
+	VMAIter mcpx_vma_handle = CarveVMA(MAX_VIRTUAL_ADDRESS - PAGE_SIZE + 1, PAGE_SIZE);
+	VirtualMemoryArea& mcpx_vma = mcpx_vma_handle->second;
+	mcpx_vma.type = VMAType::DeviceMCPX;
+	MergeAdjacentVMA(mcpx_vma_handle);
```
PatrickvL: This merge is the only deviation from the other mappings above - why's it needed here? (A comment could explain that.)

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+
+	if (g_bIsChihiro) {
+		printf("Page table for Chihiro initialized!\n");
+	}
+	else { printf("Page table for Debug console initialized!\n"); }
+}
+
+void VMManager::MemoryStatistics(xboxkrnl::PMM_STATISTICS memory_statistics)
+{
+	memory_statistics->TotalPhysicalPages = m_MaxPhysicalMemory / PAGE_SIZE;
```
PatrickvL: Assert argument is assigned

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+		LOG_FUNC_ARG(bNonContiguous);
+	LOG_FUNC_END;
+
+	Lock();
+	size_t ReturnedSize = size;
+	VAddr v_addr = MapMemoryBlock(&ReturnedSize, low_addr, high_addr, Alignment, bNonContiguous);
+	if (v_addr)
+	{
+		ReprotectVMARange(v_addr, ReturnedSize, protect);
+		protect & (PAGE_EXECUTE | PAGE_EXECUTE_READ | PAGE_EXECUTE_READWRITE | PAGE_EXECUTE_WRITECOPY) ?
```
PatrickvL: Why a ternary expression here, instead of a normal if-else?

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+size_t VMManager::QuerySize(VAddr addr)
+{
+	LOG_FUNC_ONE_ARG(addr);
+
+	// The Xbox returns at least PAGE_SIZE even for invalid addresses
+	size_t size = PAGE_SIZE;
+
+	Lock();
+	auto it = m_Vma_map.lower_bound(addr);
+	if (it != m_Vma_map.end() && it->second.base == addr)
```
PatrickvL: If an address is queried that lies inside, but not at the exact base of an allocation, it's delta should be taken into account

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+		if (AlignedCapturedSize > m_MaxContiguousAddress - m_PhysicalMemoryInUse)
+		{
+			*addr = AlignedCapturedBase;
+			*size = 0;
+			ret = STATUS_NO_MEMORY;
+			Unlock();
+			RETURN(ret);
+		}
+		m_PhysicalMemoryInUse += AlignedCapturedSize;
+		protect & (PAGE_EXECUTE | PAGE_EXECUTE_READ | PAGE_EXECUTE_READWRITE | PAGE_EXECUTE_WRITECOPY) ?
```
PatrickvL: Ternary or if...

src/CxbxKrnl/VMManager.cpp
ergo720: It's impossible: VirtualAlloc always reserves memory at a 64K boundary, meaning the lower bits are already cleared. Actually, because 64K = 16 * 4K then this check is redundant and can be omitted. Also, 128MiB = 2048 * 64K so there's no need to even allocate an extra page

src/CxbxKrnl/VMManager.cpp
ergo720: Be aware that PAGE_SIZE is referring to the size of the virtual allocation, not the size of the kernel. If KERNEL_SIZE is not an aligned quantity, it will assert

src/CxbxKrnl/VMManager.cpp
ergo720: I think I've forgotten to delete that line

src/CxbxKrnl/VMManager.cpp
ergo720: Just a matter of taste

src/CxbxKrnl/VMManager.cpp
ergo720: You mean to check if PMM_STATISTICS memory_statistics is nullptr? If it is it will crash in MmQueryStatistics since there it's accessing MemoryStatistics->Length so the assert can be placed there instead

src/CxbxKrnl/VMManager.cpp
ergo720: This is by design. QuerySize implements MmQueryAllocationSize, which always expects the address to be queried to be the start of an allocation of MmAllocateContiguousMemory or MmAllocateSystemMemory. If it lies inside, then it either faults or returns an incorrect size of at least PAGE_SIZE. If that happens, then it's either a game bug or (more likely) a bug somewhere else...

src/CxbxKrnl/VMManager.cpp
```diff
@@ -0,0 +1,1044 @@
...
+			m_ImageMemoryInUse += AlignedCapturedSize : m_NonImageMemoryInUse += AlignedCapturedSize;
+
+		*addr = AlignedCapturedBase;
+		*size = AlignedCapturedSize;
+		Unlock();
+		RETURN(ret);
+	}
+	else
+	{
+		Unlock();
```
PatrickvL: If you want, you could convert this last bit of code into having a single point of exit, by doing the unlock and return at the end, after all if-else compound statements.

src/CxbxKrnl/EmuXapi.cpp
```diff
@@ -1042,21 +1042,27 @@ DWORD WINAPI XTL::EMUPATCH(XLaunchNewImageA)
 
 	// Update the kernel's LaunchDataPage :
```
PatrickvL: Are we at a point at which we could unpatch this function just as well?

If not, a description on why the psych is still needed can be helpful (haven't checked, so ignore this remark if the reason is already described).

src/CxbxKrnl/VMManager.cpp
PatrickvL: Thanks for clarifying that. An assert could communicate that fact, but you decide ;)

src/CxbxKrnl/VMManager.cpp
PatrickvL: Aligned it will be ;)

src/CxbxKrnl/VMManager.cpp
PatrickvL: True enough. I mainly use asserts to clearly communicate preconditions (and sometimes postconditions) in code.

src/CxbxKrnl/VMManager.cpp
PatrickvL: Queries on non-base addresses are currently done (however undesired) by our HLE D3D patches (especially on slices of volume textures, or faces of cube textures, or even mipmap levels of regular textures) and possibly also by unpatched D3D and/or DSOUND functions.

I'm not entirely sure, so go by your own observations, but IF it happens, the delta between the allocation's base address and the supplied address should be deducted from the result.

src/CxbxKrnl/VMManager.cpp
BiatuAutMiahn: "so the OS will do this cleanup for us." wouldn't it be better on the long run to perform the  unmapping anyway? in case of regression?

src/CxbxKrnl/EmuXapi.cpp
ergo720: Unfortunately we cannot: it depends on NtQuerySymbolicLinkObject, NtOpenSymbolicLinkObject (and others) which have some TODOs at the moment. There's already a comment for it

src/CxbxKrnl/VMManager.cpp
ergo720: Constantine and JSRF call this from some D3D functions, but they always use the base address. However, I implemented the feature requested just in case...

src/CxbxKrnl/VMManager.cpp
ergo720: I used DbgPrintf because assert.h is not included there and I didn't want to include it just for this

LukeUsher:
This PR broke Cxbx-Reloaded on Linux (using Wine)
It now hits this case: 

```
CxbxKrnlCleanup("CxbxRestoreContiguousMemory: Couldn't map contiguous memory.bin into tiled memory at 0xF0000000!");
```

After debugging, it is because this call now fails, with error code STATUS_INVALID_ADDRESS when it used to succeed. Perhaps something in the new Virtual Memory manager conflicts with this?

```
	void *tiled_memory = (void *)MapViewOfFileEx(
		hFileMapping,
		FILE_MAP_READ | FILE_MAP_WRITE | FILE_MAP_EXECUTE,
		/* dwFileOffsetHigh */0,
		/* dwFileOffsetLow */0,
		TILED_MEMORY_CHIHIRO_SIZE,
		(void *)TILED_MEMORY_BASE);
```

ergo720:
That's strange, the manager is not even initialized at that point (the initialization function is called immediately after that). I just increased the sized of the contiguous and tiled memory to always be 128 MiB, however, they don't overlap (contiguous end =  0x88000000 and tiled start = 0xF0000000). Have you tried deleting all the files in the AppData folder of wine? (if there is one) This is a know issue of this new manager

PatrickvL:
To avoid breaking on already reserved pages, I reserve memory per page my Loader branch. That requires freeing per page too of course, but it allows for fine grained memory management, and some resilience against OS-specific quirks.

PatrickvL:
See https://github.com/PatrickvL/Cxbx-Reloaded/blob/Loader/src/Common/XboxAddressRanges.cpp#L108

PatrickvL:
In the mean time, as a quick work around, tiled memory can safely be shrunk to 64 MiB, or even entirely ignored, as our HLE patches currently avoid it at all cost. (Registering this might require some serious thought once we're going to LLE the GPU though)

ergo720:
I solved the problem by simply not mapping the tiled memory. Here is a screenshot of Tenchu on Linux after rebooting and crashing with the same error seen on Windows
tenchu-linux
We can either not map the tiled memory on both Windows and Linux or we can detect wine with the following code and not doing the mapping selectively on Linux 
```
int main()
{
	typedef const char * (CDECL *wine_get_version)(void);
	HMODULE hntdll = GetModuleHandle("ntdll.dll");
	if (!hntdll)
	{
		puts("Not running on NT.");
		return 1;
	}

	wine_get_version pwine_get_version = (wine_get_version)GetProcAddress(hntdll, "wine_get_version");
	if (pwine_get_version)
	{
		printf("Running on Wine... %s\n", pwine_get_version());
	}
	else
	{
		puts("did not detect Wine.");
	}

	return 0;
}

```
When run, it detects the wine version if present
wine

jackchentwkh:
@ergo720  here is the XBE dump from the VCOP3.xbe I have on hand. seems it's release version, not debug version. hope it help.
[Xbe.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1596837/Xbe.txt)

LukeUsher:
Not mapping tiled memory at all is a very bad idea. Currently with HLE it has very little effect, as our version of the resource registration functions currently ignore the tiled flag, however, once these functions are unpatched, or LLE for the nv2a is used, games will break if they cannot access the same memory through these regions. This could go in as a short term hack to prevent outright crashing but would need to be solved properly fairly soon.

ergo720:
@jackchentwkh From the xbe dump you posted I can see your build will be detected as debug according to our code. Entry point add & 0xf0000000 == 0x90000000 instead of 0x40000000 -> not chihiro and Kernel Image thunk addr & 0x80000000 == 0x80000000 which is > 0 so it's debug. However, debug builds should crash outright since AFAIK cxbx doesn't support debug functions. It either means that our detection code is faulty (but in that case it should still crash because it would decode the entry point / kernel thunk incorrectly because the XOR keys are different) or your build has been tampered somehow or something else that I don't know
@LukeUsher I think I can try to map at least 64 MiB of tiled memory since that was working before on Linux.

PatrickvL:
May I repeat myself? It's advisable to reserve all memory ranges in incremental blocks of 64 KiB (being the alignment of VirtualAlloc), so any page already reserved (for example for MMIO) can be skipped, while still reserving the rest of the memory range. This turned out to be a requirement in my Loader branch, otherwise I hit the same issue described above (and a few more - see my code for details).

PatrickvL:
As a side note: ultimately, tiled memory will need read/write access handlers that swizzle the address, because as far as I understand it, the tiled memory range increases locality-of-reference by realigning a pixel-square (in 2d, or a cube in 3d) into incremental addresses (so it can be read in as little cache line fills as possible).

LukeUsher:
Manually deleting Memory.bin from the AppData folder solves this issue for me with Wine. 

As this could effect users of Windows too (you said AppData needs a clear) could you add a check for if Memory.bin != the memory size, and if not, delete and re-create the memory file? That would prevent this happening in the future. It's best if we avoid the user having to manually delete files.

ergo720:
There is one already: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/master/src/CxbxKrnl/CxbxKrnl.cpp#L319. However, I know from my tests that's not enough, I was still getting the issue despite the memory.bin file being increased on the fly to 128MiB by CreateFileMapping

PatrickvL:
This did the trick for me : https://github.com/PatrickvL/Cxbx-Reloaded/blob/Loader/src/CxbxKrnl/CxbxKrnl.cpp#L314

ergo720:
Ok, I think I can send a quick PR with that change if you agree

ergo720:
@PatrickvL I tested the fix you suggested but I'm still getting the same error Luke reported above

PatrickvL:
Hm. And what happens when you change the size of tiled memory to 64 KiB? (As that's the only difference I can see between my Loader branch and this....)

ergo720:
I'm having a deeper issue: even if I delete all the files in AppData folder and use the site's version, it doesn't solve the problem, the tiled memory mapping error still appears, unlike what happens in Windows

ergo720:
If I use 64MiB for tiled memory the fix works

ergo720:
wine-3.0-rc4 and it's 64 bit

Issue #64, opened 2016-12-25
Complete kernel thunk table


PatrickvL:
At the very least, we should have correct function signatures and logging stubs for all kernel API's.

TODO:
* search which kernel API's are still registered in the kernel thunk table using PANIC
* For those, create stub implementations with correct signatures
* Log all their arguments and mark each one with LOG_UNIMPLEMENTED
* Register them in the kernel thunk table
* Remove the PANIC symbol

PatrickvL:
Currently missing: most Io, Ke, Ob and Xc prefixed functions, and about half the Rtl functions (most of these can be redirected straight to NtDll for a start).

PatrickvL:
Most Rtl prefixed functions are now complete.

Still missing :
RtlCaptureContext
RtlCaptureStackBackTrace
RtlEnterCriticalSectionAndRegion
RtlGetCallersAddress
RtlLeaveCriticalSectionAndRegion
RtlRaiseException
RtlRaiseStatus
RtlUnwind
RtlWalkFrameChain

PatrickvL:
This is nearing completion; only a handfull of API's to go...

x1nixmzeng:
Unfortunately I haven't had the time to work on Cxbx-Reloaded recently.

The MmDbg* stubs are here if you want to merge just this change: https://github.com/x1nixmzeng/Cxbx-Reloaded/commit/52156da89db922d15318b2b4dc3ebad229cfd58e

The other cleanup changes in this branch are not relevant.

PatrickvL:
Could you prepare a Pull Request of the things you think are useful for the main line?

LukeUsher:
It;s been a while since this was looked at, I believe we now have a complete Kernel Thunk table, @PatrickvL  are you happy that this is the case, or do you think we should keep this open a little longer?

PatrickvL:
Currently, there are 42 API's that are still missing (marked with `PANIC` in `CxbxKrnl_KernelThunkTable`) :

- KeInitializeDeviceQueue
- KeInsertByKeyDeviceQueue
- KeInsertDeviceQueue
- KeInsertHeadQueue
- KeInsertQueue
- KeInsertQueueApc
- KePulseEvent
- KeQueryBasePriorityThread
- KeReleaseMutant
- KeReleaseSemaphore
- KeRemoveByKeyDeviceQueue
- KeRemoveDeviceQueue
- KeRemoveEntryDeviceQueue
[...]

PR #824, opened 2018-01-02
Kernel stuff

Luca1991:
Hi all dear friends,

here are a couple of kernel functions I've written some weeks ago (KeInitializeDeviceQueue and KeInsertDeviceQueue).

I've not tested them, but they should work because the implementation is greatly inspired by the ReactOS source code.

When merged, we should update this list: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/64

Happy New Year,
Luca


PatrickvL:
Thanks, we need more of these!

PatrickvL:
I've taken the opportunity to update issues #64 as well, we've only got 23 API's left to do! (And 5 of those are already done, there's a commit link in that issue, that could easily be pulled in.)

PatrickvL:
Btw, KeInitializeDeviceQueue is quite easy to test too

PR #826, opened 2018-01-03
Fix an issue with memory.bin in wine

ergo720:
I tested this with a tiled memory of 64 MiB and it worked, however, for some reason using 128 MiB always fails (cannot map tiled memory). Considering that for PatrickvL this seems to work in that case as well, I guess it's because I have a different version of wine?


ergo720:
Yeah, I thought that temp file was strange indeed

PatrickvL:
Oops, a small miscommunication : 64 MiB worked under Windows - I didn't test with Wine. Sorry about that.

ergo720:
I did, it worked

ergo720:
I suppose I can delete the entire temp folder, not just the file?

LukeUsher:
Looks good, thanks!

PR #835, opened 2018-01-05
Restore unknown vertex buffer/push buffer size checks 

LukeUsher:
These were broken by the new VMManager, as it expected QueryMemorySize to return -1 on failure, now it returns 0.


PatrickvL:
Good catch!

PR #839, opened 2018-01-07
Memory manager tweaks

PatrickvL:
These changes classify the types of memory pages being requested, so memory statistics can potentially become identical to real hardware. Also, page protection is updated to better reflect what a real kernel does.

EDIT : Alas, this code is not yet ready for testing. Even the simplest XDK samples crash. 

EDIT : Turns out, some regressions are not caused by this branch, but a regression in master (between Release and Debug builds - see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/847 for more details.)

For any regression, check if it also exists in master, as than it should be addressed there first.


src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -8527,7 +8527,7 @@ XTL::X_D3DPalette * WINAPI XTL::EMUPATCH(D3DDevice_CreatePalette2)
     X_D3DPalette *pPalette = EmuNewD3DPalette();
 
 	pPalette->Common |= (Size << X_D3DPALETTE_COMMON_PALETTESIZE_SHIFT);
-    pPalette->Data = (DWORD)g_VMManager.Allocate(XboxD3DPaletteSizeToBytes(Size), 0, (~((::ULONG_PTR)0)), PAGE_SIZE, PAGE_EXECUTE_READWRITE, false);
+    pPalette->Data = (DWORD)g_VMManager.Allocate(XboxD3DPaletteSizeToBytes(Size), 0, (~((::ULONG_PTR)0)), PAGE_SIZE, PAGE_READWRITE, true); // TODO : Can't we use MmAllocateContiguousMemory here?
     pPalette->Lock = X_D3DRESOURCE_LOCK_PALETTE; // emulated reference count for palettes
```
ergo720: Yes, you can use MmAllocateContiguousMemory  here.

src/CxbxKrnl/VMManager.cpp
```diff
@@ -638,7 +638,7 @@ VAddr VMManager::MapMemoryBlock(size_t* size, PAddr low_addr, PAddr high_addr, U
 
 			VirtualMemoryArea& final_vma = vma_handle->second;
 			final_vma.type = VMAType::Allocated;
-			final_vma.permissions = PAGE_EXECUTE_READWRITE;
+			final_vma.permissions = bContiguous ? PAGE_READWRITE : PAGE_EXECUTE_READWRITE;
 			final_vma.backing_block = offset;
```
ergo720: If you just do this then the page for contiguous allocation will still be executable. You could call ReprotectVMARange on the allocation or (better) drop the FILE_MAP_EXECUTE flag from the file view of MapViewOfFileEx in CxbxRestoreContiguousMemory

src/CxbxKrnl/VMManager.cpp
```diff
@@ -649,6 +649,13 @@ VAddr VMManager::MapMemoryBlock(size_t* size, PAddr low_addr, PAddr high_addr, U
 
 		case PMEMORY_ALLOCATE_FRAGMENTED:
 		{
+			if (bContiguous) {
+				EmuWarning("Warning: Cannot allocate contiguous memory due to fragmentation!");
+				// TODO : Prevent the preceding call to AllocateFragmented. For now, cleanup.
+				DeAllocateFragmented((VAddr)offset);
+				return NULL;
+			}
```
ergo720: That warning is already present in PhysicalMemory::AllocateFragmented, no need to repeat it here

src/CxbxKrnl/EmuD3D8.cpp
ergo720: The (~((::ULONG_PTR)0)) thing, is just the definition of MAXULONG_PTR. I couldn't use it here because in the namespace XTL ULONG_PTR is typedef as a pointer type instead of being an integer, so I had to use this trick. Maybe you can create a new typedef for it in XTL and see if that works?

src/CxbxKrnl/VMManager.cpp
PatrickvL: Well that's the thing, on a real machine, the original kernel (bios) resides in contiguous memory, which range would require execute rights. We can currently still get away with only a dummy kernel image header, pretending to be the most recent kernel, so no kernel patches will be applied. But what if a title decides to call directly into the kernel, bypassing our thunk table? For now we can't support such things, so removing the execute right from the entire contiguous memory range seems alright. But once we get closer and closer to how the real machine behaves, we'll want to allow execution in kernel space too.

src/CxbxKrnl/VMManager.cpp
PatrickvL: Oh okay. Except it shouldn't be called for contiguous memory requests at all. I'm afraid we'll have to pass along an enum indicating the requested memory type (for example virtual, physical, contiguous, write-combined, tiled)

src/CxbxKrnl/VMManager.cpp
```diff
@@ -602,14 +602,14 @@ xboxkrnl::NTSTATUS VMManager::XbFreeVirtualMemory(VAddr* addr, size_t* size, DWO
...
+VAddr VMManager::MapMemoryBlock(size_t* size, PAddr low_addr, PAddr high_addr, ULONG Alignment, bool bContiguous)
 {
 	// Find a free memory block for the allocation, if any
 	VAddr addr;
 	u32 offset;
 	size_t aligned_size = (*size + PAGE_MASK) & ~PAGE_MASK;
 
-	if (high_addr == MAXULONG_PTR)
+	if (high_addr == MAXULONG_PTR) // TODO : && low_addr == 0) || bContiguous)
 	{
```
ergo720: If a title doesn't care about a specific range that the parameters will always be 0 and MAXULONG_PTR so the && low_addr == 0 is fine. However, || bContiguous is wrong: if a specif range is suggested bContiguous will be true, meaning that AllocatePhysicalMemory will be called instead of AllocatePhysicalMemoryRange

src/CxbxKrnl/VMManager.cpp
ergo720: I looked more closely: ReprotectVMARange is already called after MapMemoryBlock and will use whatever protection is being passed to VMManager::Allocate. So for contiguous allocation is enough to just pass PAGE_READWRITE to Allocate. Actually, the function updates the permissions of the vma is being passed to so that means you don't even need the check of above but probably it's still better to keep it anyway...

src/CxbxKrnl/VMManager.cpp
ergo720: Maybe you can write in this warning that the allocation is refused because is contiguous but cannot be served thanks to the fragmentation

src/CxbxKrnl/VMManager.cpp
PatrickvL: You're right. Can't we just do away with `AllocatePhysicalMemory`, and instead let all go through `AllocatePhysicalMemoryRanged`? I'd be one less function to maintain...

src/CxbxKrnl/VMManager.cpp
PatrickvL: Yes, once the physical allocation discovers it can't service the request for contiguous memory, such a log message can be generated there. No need to duplicate it here, indeed.

PatrickvL:
I'll address the things we discussed first. Once @ergo720 reviewed that to satisfaction, I'll reopen this PR.

src/CxbxKrnl/VMManager.cpp
ergo720: AllocatePhysicalMemory uses the memory defrag I implemented whereas AllocatePhysicalMemoryRange always allocates starting from the the high_addr - down. I'd prefer to keep both

PatrickvL: Fair enough.

PatrickvL:
I've done a bit of work on this. Please review again. DONT merge until that, and some tests are done!

src/CxbxKrnl/VMManager.h
```diff
@@ -94,7 +94,9 @@ struct VirtualMemoryArea
 	// vma size
 	size_t size = 0;
 	// vma kind of memory
-	VMAType type = VMAType::Free;
+	VMAType vma_type = VMAType::Free;
+	// the page type of this memory area
+	PageType page_type = PageType::Unknown;
 	// vma permissions
```
ergo720: I'm not convinced that the page type should be inside the VMA, it's actually saved in the PFN database on a per-page basis instead. Considering that implementing the PFN also requires PTE/PDE and that's a lot of work, for now this is fine

src/CxbxKrnl/PhysicalMemory.cpp
```diff
@@ -300,6 +307,8 @@ void PhysicalMemory::ShrinkPhysicalAllocation(PAddr addr, size_t offset, bool bF
 	else
 	{
 		auto it = m_Mem_map.lower_bound(addr);
+		if (it == m_Mem_map.end()) { return; }
+
 		PAddr old_base = it->first;
```
ergo720: ShrinkPhysicalAllocation is called from ResizeVMA which in turn is called from XbFreeVirtualMemory. The iterator passed is std::prev(auto it = m_Vma_map.lower_bound(AlignedCapturedBase);), which couldonly become end() if lower_bound is called on an address > of MCPX_BASE, which is forbidden by the check if (CapturedBase > HIGHEST_USER_ADDRESS) { ret = STATUS_INVALID_PARAMETER; RETURN(ret); }. Also, ShrinkPhysicalAllocation is only called if the vma found is of type Allocated or Stack, meaning a corresponding physical allocation exists, and so it here can never be end()

src/CxbxKrnl/PhysicalMemory.cpp
```diff
@@ -314,22 +323,26 @@ void PhysicalMemory::ShrinkPhysicalAllocation(PAddr addr, size_t offset, bool bF
 	}
 }
 
-void PhysicalMemory::DeAllocatePhysicalMemory(PAddr addr)
+void PhysicalMemory::DeAllocatePhysicalMemory(const PAddr addr)
 {
 	auto it = m_Mem_map.lower_bound(addr);
+	if (it == m_Mem_map.end()) { EmuWarning("DeAllocatePhysicalMemory : addr unknown!");  return; }
+
 	m_PhysicalMemoryInUse -= it->second;
```
ergo720: DeAllocatePhysicalMemory is only called in DestructVMA, which checks if the vma type is Allocated or Stack and, if so, calls the function and so the allocation must exist

src/CxbxKrnl/PhysicalMemory.h
```diff
@@ -93,27 +109,27 @@ class PhysicalMemory
 		// destructor
 		~PhysicalMemory()
 		{
-			for (auto it = m_Fragmented_mem_map.begin(); it != m_Fragmented_mem_map.end(); ++it)
+			for (auto it = m_Fragmented_mem_map.begin(), end = m_Fragmented_mem_map.end(); it != end; ++it)
 			{
 				VirtualFree((void*)it->first, 0, MEM_RELEASE);
```
ergo720: It's the same as before I think?

src/CxbxKrnl/VMManager.cpp
```diff
@@ -208,46 +217,60 @@ void VMManager::InitializeChihiroDebug()
...
 	VirtualMemoryArea& vma = vma_handle->second;
-	vma.type = type;
+	vma.vma_type = vma_type;
+	vma.page_type = PageType::SystemMemory;
 	UpdatePageTableForVMA(vma);
-	m_NonImageMemoryInUse += size;
+	// Note : On a real Xbox, hardware MMIO address ranges aren't
+	// backed by physical memory, so we don't count pages either.
+	Unlock();
 }
```
ergo720: When MapHardwareDevice is called inside Initialize, cxbx is still single threaded so there's no need for locks here, unless you plan to call this function during the emulation

src/CxbxKrnl/VMManager.cpp
```diff
@@ -183,20 +189,23 @@ void VMManager::InitializeChihiroDebug()
...
 	UnmapRange(TILED_MEMORY_BASE);
 	MapHardwareDevice(TILED_MEMORY_BASE, TILED_MEMORY_CHIHIRO_SIZE, VMAType::MemTiled);
 
 	// NOTE: we cannot just call Unmap on the mcpx region because its base + size will overflow to 0x100000000
 	// which will trigger an assert in CarveVMARange
-	m_Vma_map.lower_bound(MAX_VIRTUAL_ADDRESS - PAGE_SIZE + 1)->second.type = VMAType::Free;
-	m_NonImageMemoryInUse -= PAGE_SIZE;
+	auto it = m_Vma_map.lower_bound(MAX_VIRTUAL_ADDRESS - PAGE_SIZE + 1);
+	assert(it != m_Vma_map.end());
+
```
ergo720: The Xbox initialization always maps the mcpx at MAX_VIRTUAL_ADDRESS - PAGE_SIZE + 1 so lower_bound will always succeed here

src/CxbxKrnl/VMManager.cpp
PatrickvL: Locking all public functions is more consistent than not.

Also, this may get used by MmMapIOSpace at some point

src/CxbxKrnl/PhysicalMemory.h
PatrickvL: Yes, but faster

src/CxbxKrnl/PhysicalMemory.cpp
PatrickvL: I'll change it an assert, with your rationale in a comment

src/CxbxKrnl/VMManager.cpp
```diff
@@ -306,7 +336,6 @@ void VMManager::DeallocateStack(VAddr addr)
 	LOG_FUNC_ONE_ARG(addr);
 
 	Lock();
-	ReprotectVMARange(addr, PAGE_SIZE, PAGE_EXECUTE_READWRITE);
 	UnmapRange(addr);
 	Unlock();
```
ergo720: This is incorrect: UnmapRange will only dellocate a single vma and merge it with the next and the previous if possible. This means that only the StackBottom page will be dellocated since the Stack VMA was split in two by ReprotecVMARange when the stack was created. For example, a stack allocation of 16384 bytes will lead to two vma: 1- base StackBottom, 4096, NO_ACCESS, 2- base StackBottom + PAGE_SIZE, 16384, READ_WRITE. ReprotectVMARange will merge them back to a single VMA that UnmpaRange will deallocate entirely. Tested with Dead or Alive 3, which calls this from CreateFiber

src/CxbxKrnl/VMManager.cpp
PatrickvL: Yeah that's what this assert communicates in code. But I'll add a comment too.

src/CxbxKrnl/VMManager.h
PatrickvL: I felt reluctant to add it, as it's indeed not the best place to put this enum. But compared to building an actual page table, this is simpler. Luckily you seem agree ;)

src/CxbxKrnl/PhysicalMemory.cpp
PatrickvL: That's quite a lot of steps you described there! With such complexity, mistakes could easily be made. It's probably wise to split up this code into smaller steps. But for now asserts and comments will help convey knowledge about how this all works together.

src/CxbxKrnl/VMManager.cpp
PatrickvL: Ah, now that's valuable info! Ill reinstate i this call then, and add your comment + test case for future reference. Thanks!

ergo720:
Can you also add the fix a suggested for Monster 4x4: World Circuit in https://github.com/Cxbx-Reloaded/game-compatibility/issues/511 please? Also, I see the debug build is failing to compile for some reason

src/CxbxKrnl/VMManager.cpp
```diff
@@ -622,23 +665,28 @@ VAddr VMManager::MapMemoryBlock(size_t* size, PAddr low_addr, PAddr high_addr, U
...
+			}
+
 			VMAIter vma_handle = CarveVMA(addr, aligned_size);
 
 			VirtualMemoryArea& final_vma = vma_handle->second;
-			final_vma.type = VMAType::Allocated;
-			final_vma.permissions = PAGE_EXECUTE_READWRITE;
+			final_vma.vma_type = VMAType::Allocated;
+			final_vma.page_type = page_type;
+			final_vma.permissions = (page_type == PageType::Contiguous) ? PAGE_READWRITE : PAGE_EXECUTE_READWRITE;
```
PatrickvL: EXECUTE rights should be given to PageType::Image pages only.

src/CxbxKrnl/VMManager.cpp
```diff
@@ -112,29 +120,30 @@ void VMManager::Initialize(HANDLE file_view)
...
+	upper_mem_vma.page_type = PageType::SystemMemory;
+	upper_mem_vma.permissions = PAGE_READWRITE;
+	upper_mem_vma.backing_block = AllocatePhysicalMemoryRange(32 * PAGE_SIZE, upper_mem_vma.page_type, m_MaxContiguousAddress, XBOX_MEMORY_SIZE);
 	UpdatePageTableForVMA(upper_mem_vma);
-	m_ImageMemoryInUse += 32 * PAGE_SIZE;
 
 	// Allocate memory for the dummy kernel
 	// NOTE: change PAGE_SIZE if the size of the dummy kernel increases!
-	Allocate(KERNEL_SIZE, XBE_IMAGE_BASE, XBE_IMAGE_BASE + PAGE_SIZE - 1, KERNEL_SIZE & ~PAGE_MASK, PAGE_EXECUTE_READWRITE, false);
+	Allocate(KERNEL_SIZE, PageType::Contiguous, XBE_IMAGE_BASE, XBE_IMAGE_BASE + PAGE_SIZE - 1, KERNEL_SIZE & ~PAGE_MASK, PAGE_READWRITE);
```
ergo720: Shouldn't the kernel pages be executable instead?

src/CxbxKrnl/VMManager.cpp
```diff
@@ -387,18 +437,20 @@ size_t VMManager::QuerySize(VAddr addr)
...
-				while (prev_it != m_Vma_map.begin() && prev_backing_block == prev_it->second.backing_block)
+				const auto it_begin = m_Vma_map.begin();
+				while (prev_it != it_begin && prev_backing_block == prev_it->second.backing_block)
 				{
 					--prev_it;
 				}
 				it = std::next(prev_it);
-				EmuWarning("VMManager: QuerySize : quering not the start address of an allocation\n");
+				EmuWarning(LOG_PREFIX ": QuerySize : quering not the start address of an allocation\n");
 			}
```
ergo720: Small thing; here there is type, it should read "querying" instead

src/CxbxKrnl/VMManager.cpp
```diff
@@ -348,6 +351,12 @@ void VMManager::Protect(VAddr target, size_t size, DWORD new_perms)
 		LOG_FUNC_ARG(new_perms);
 	LOG_FUNC_END;
 
+	if ((new_perms & 0xFF) == 0)
+	{
+		EmuWarning(LOG_PREFIX ": Protect : Skipping due to missing permission flags\n");
+		return;
+	}
+
```
ergo720: It's probably better to add the check inside ReprotectVMA since ReprotectVMARange is also called somewhere else.

PatrickvL: Actually, I've moved it into CheckMmProtectFlags() in EmuKrnlMm, without any log message (since a real kernel doesn't seem to log such circumstance either)

ergo720:
Seems fine to me now, hopefully it just needs some testing now

PatrickvL:
STATUS : Once regressions in master are fixed, this PR will be updated, tested, expanded upon, and reopened again afterwards.

PatrickvL:
This PR is replaced by #905

Issue #820, opened 2017-12-15
Remove _CXBXKRNL_INTERNAL define


PatrickvL:
The define `_CXBXKRNL_INTERNAL` that appears everywhere appears to be a legacy inheritance.

For this issue :
* Verify it doesn't alter any behavior in our current code base
* If the above succeeds, remove this define from our entire source tree

x1nixmzeng:
This is never `#ifdef` checked anywhere - only ever defined. Removing it will have no effect so this is safe to delete.

Another define is `_XBOXKRNL_DEFEXTRN_` which has the same issue.

PatrickvL:
That other one is used in openxdk, so that's not the same issue (it might need another approach)

PatrickvL:
Fixed by https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/840

PR #840, opened 2018-01-07
Remove _CXBXKRNL_INTERNAL define for issue #820

Fisherman166:
As stated in issue #820, this define is not used in any #ifdef statements. Build/run was not affected by removing this define.


PatrickvL:
Thank you!

Issue #686, opened 2017-08-29
Log XBE flags as readable string


LukeUsher:
It would be good to log region flags by name within an xbe dump. We currently log region flags by the hex value, however this is not clear to a user unless they understand the region flag encoding. 

I propose outputting the textual name of these flags in brackets next to the region flag in the xbe dump.

Note that any combination of the above flags is possible, so a region code of 0x00000005 would mean a title that runs on US/CANADA hardware, and the rest of the world except for JAPAN, while 0x00000007 is region free. 

It would also be good if at the same time, logging of the media type flag could be implemented, as above, we currently only log the hex values, but not the text representation.

For a list of supported flags that need to be logged, see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/223e20d1528ac7bf9a2d8013f7352af6c9f39bae/src/Common/Xbe.h#L277

PatrickvL:
In addition to this, all other flags in the XBE header could be dumped as string too. But perhaps it's best to create one issue per case for this.

PatrickvL:
Part of this is done in PR #715 (it logs TitleID, TitleName and Region as string)

PatrickvL:
Taking inspiration from Dxbx's XbeExplorer, here some more fields that could be rendered as string :
- [ ] Xbe::Header.dwTimeDate
- [ ] Xbe::Header.dwInitFlags
- [ ] Xbe::Header.dwPeTimeDate
- [ ] Xbe::Header.dwDebugPathnameAddr
- [ ] Xbe::Header.dwDebugFilenameAddr
- [ ] Xbe::Header.dwDebugUnicodeFilenameAddr
- [ ] Xbe::Certificate.dwTimeDate
- [ ] Xbe::Certificate.dwAllowedMedia
- [X] Xbe::Certificate.dwGameRegion - done by #715
- [ ] Xbe::Certificate.dwGameRatings
- [ ] Xbe::SectionHeader.dwFlags
- [ ] Xbe::LibraryVersion.dwFlags
- [ ] Xbe::TLS.dwCharacteristics

PatrickvL:
Also, see https://github.com/PatrickvL/Dxbx/blob/master/Source/Delphi/src/uXbe.pas#L243

Fisherman166:
[...]

PR #841, opened 2018-01-08
Print String Representations of GameRegion and AllowedMedia

Fisherman166:
Xbe fields GameRegion and AllowedMedia now print their string representation in the Xbe dump file for issue #686. A diff of a Xbe dump of RainbowSix 3 with new the feature:

xbe_dump_diff

While implementing the GameRegionToString function, I noticed that a duplicate implementation existed in CxbxKrnl.cpp. Heavily leveraging the code from that function, the function was moved into the Xbe class. Doing so allowed both the Kernal and Xbe::DumpInformation to use the same function. A diff of KrnlDebug on RainbowSix 3:

krnldebug_diff


PatrickvL:
Thank you

Issue #836, opened 2018-01-06
The HLE "last compile time" timestamp does not update as expected


x1nixmzeng:
The high-level emulation database cache files are used by Cxbx-Reloaded to prevent rescanning the XBE for known symbols.

However, if the cache file already exists (see the logic for **EmuHLEIntercept** in `HLEIntercept.cpp`) then this field is read from the cache file (which uses the INI format)

```
[Info]
HLEDatabaseVersion=BLAH
```

then checked against the value of `szHLELastCompileTime` - defined in `HLEDataBase.cpp` as:

```
extern "C" const char *szHLELastCompileTime = __TIMESTAMP__;
```

Following from [this issue](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/832), this will only 

> reflect the last modified time of HLEDataBase.cpp, [not] the modification of included OOVPA definitions.

From @LukeUsher :

> We need somewhere we can define that timestamp so that it is only updated when the HLE database actually changes: Perhaps it's worth just hashing the entire HLE database and storing that hash in the cache file. The purpose of this is so that we can re-use HLECache files across Cxbx-Reloaded revisions in which the HLE database did not change at all, while not needing to be manually cleared when the database is updated.

PatrickvL:
Fixed in #844

x1nixmzeng:
My changes were a workaround - in future the rescanning could be a lot smarter!

PatrickvL:
@x1nixmzeng Could you describe your ideas on smarter rescanning?

x1nixmzeng:
Each patched function could have an associated version/checksum so that rescanning can be selective.

The version/checksum could be something compariable (such as a unix timestamp) instead of a computed value.

PR #844, opened 2018-01-09
Hash the HLE database so OOVPA changes force a rescan

x1nixmzeng:
As per #836, the old timestamp check doesn't work if the cpp file isn't touched after editing OOVPA sigantures. This requires the cache to be cleaned - which defeats the point of auto-detecting it 😉 

This change introduces a new INI value `HLECacheHash` which stores the a FNV-1a checksum of the HLEDataTable.

The downside to this change is that **any** differences (between existing cache files and Cxbx build) will force a rescan. In future a contextual hash would be better suited! Phew!


PatrickvL:
Thanks! A nice addition this is

PR #845, opened 2018-01-10
Skeleton PCI Device Support

LukeUsher:
Implements a basic PCIBus/Device mechanism and ports EEPROM,SMBus and NVNET to it.

This shouldn't have any noticeable difference in execution (Other than the bugfixes to HalReadWritePCISpace & EmuX86), but provides a skeleton for things to come in the near future.


PatrickvL:
Yay! It's in! Thanks for the PR

PR #846, opened 2018-01-10
Implement Disable pixel shaders hack

LukeUsher:
This adds a new menu option in the Settings menu to disable pixel shaders, this is an often requested option and masks the White-Screen problem in Jet Set Radio Future and issues in other titles.

While not ideal, it works as an interim 'fix' to make gameplay more bearable in problematic titles. Hopefully this won't be needed for very long.

Note that this is a hack, and not a solution. I log if the Hack is On or Off in the KrnlDebug log, so we can reject rendering bug reports with the hack enabled. Non related bug reports will still be valid.

As part of the prep, I also added error handling to the Registry reads.

The hack works by simply returning X_PIXELSHADER_FAKE_HANDLE as the handle from CreatePixelShader, this causes SetPixelShader and other functions to use a dummy shader that simply renders the input texture to the output, with no changes.


LukeUsher:
Merging as this is an optional hack, behaviour doesn't change if disabled

PR #852, opened 2018-01-13
Add XRefSaveIndex into #ifndef WIP_LessVertexPatching

jarupxx:


jarupxx:
Because revision 4627 of D3DDevice_SetRenderTarget and D3DDevice_GetDepthStencilSurface are need those XREF.

PR #853, opened 2018-01-13
PsCreateSystemThreadEx : Waiting for thread-creation updated

PatrickvL:
so it waits for the correct handle, avoids access-violations, releases resources correctly and logs additional details on waiting errors.


PatrickvL:
This caused a regression -see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/857

LukeUsher:
I tested upwards of 10 titles before merging and saw no regressions, but obviously there is one that effects some titles and not others. I'll investigate

PR #858, opened 2018-01-14
Wait betweeen CreateThread and WaitForSingleObject

LukeUsher:
Change some DbgPrintf's to EmuWarning, so that they show up in Release build logs as well as Debug builds.

Inserted a Sleep between call to __beginthreadex and WaitForSingleObject, this seems to resolve a race condition where WaitForSingleObject is called too soon.

### DIFF
--- a/src/CxbxKrnl/EmuKrnlPs.cpp
+++ b/src/CxbxKrnl/EmuKrnlPs.cpp
@@ -132,6 +132,7 @@ static unsigned int WINAPI PCSTProxy
 		StartSuspended,
 		hStartedEvent);
 
+
 	// Do minimal thread initialization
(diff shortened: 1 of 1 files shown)


LukeUsher:
Confirmed to fix the recent regression, so merging

PatrickvL:
We should investigate why this sleep helps circumventing a hang, as it's not very clean to fix things by inserting sleeps here and there (we've all done it out course, but still)

LukeUsher:
I suspect it's a race condition in the running title, rather than a flaw with our implementation: The hang seems to happen if PsCreateSystemThreadEx returns too quickly, perhaps not giving the running title time to react in it's other threads.

Issue #861, opened 2018-01-14
restore LaunchDataPage Bug


hotzenplotz3000:
I've just found a bug while testing some games with current build.
This issue appears only with multi xbe titles.
In this case CxbxRestoreLaunchDataPage() restors the wrong address from the previous run.

the code looks like this:
```
xboxkrnl::LaunchDataPage = (xboxkrnl::LAUNCH_DATA_PAGE*)CONTIGUOUS_MEMORY_BASE + LaunchDataPAddr;
```

I think CONTIGUOUS_MEMORY_BASE and LaunchDataPAddr should be casted to get the proper address:
```
xboxkrnl::LaunchDataPage = (xboxkrnl::LAUNCH_DATA_PAGE*)(CONTIGUOUS_MEMORY_BASE + LaunchDataPAddr);
```
Please correct me if I'm wrong.

I've tested this with these games, which regressed since the MemoryManager update
 - Prisoner of War
 - Unreal 2
 - TopSpin

Prisoner of War and Unreal 2 also pass some arguments through LauchDate to another xbe that otherwise got lost

PatrickvL:
It looks like you're right

LukeUsher:
I noticed another potential issue where the launchdatapage is restored before the memory manager sets up and registers memory, I was going to work on this but I don't have any multi xbe titles on hand right now

PatrickvL:
Note : I've checked our entire codebase on the regular expression `=.*\*\)[^(].*\+` (an assignment with a pointer typecast without further brackets that does addition) and found no more cases that go wrong.

PatrickvL:
This issue is fixed in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/862 , closing.

hotzenplotz3000:
@PatrickvL thanks for commiting my suggestion 👍 

PS: I'm glad that this didn't break anything 🤣

PR #862, opened 2018-01-15
Move persistent memory restore to after contiguous memory is restored 

LukeUsher:
Potential fix for (at least part of) #861.

If some mutli-xbe titles could be tested and the results fed back to me before merging, that would be appreciated. 

Sorry about the merge-commit in here, just do a squash or rebase merge to fix it ;)


CakeLancelot:
This PR fixes the issue described in https://github.com/Cxbx-Reloaded/game-compatibility/issues/64#issuecomment-356819026 (to sum up, the game relaunches the executable when starting a game). Cel Damage can now boot into game from the menu, but graphics are still broken.

LukeUsher:
Awesome, this can me merged, then

PR #865, opened 2018-01-15
Fix an issue where Xbox Dashboard could not successfully launch XOnlineDash

LukeUsher:
This has been tested and confirmed working with Dashboard version 4920 (the latest available that doesn't use LTCG)

### DIFF
--- a/src/CxbxKrnl/CxbxKrnl.cpp
+++ b/src/CxbxKrnl/CxbxKrnl.cpp
@@ -83,7 +83,7 @@ Xbe::Header *CxbxKrnl_XbeHeader = NULL;
 
 HWND CxbxKrnl_hEmuParent = NULL;
 DebugMode CxbxKrnl_DebugMode = DebugMode::DM_NONE;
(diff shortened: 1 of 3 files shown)


src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -83,7 +83,7 @@ Xbe::Header *CxbxKrnl_XbeHeader = NULL;
 
 HWND CxbxKrnl_hEmuParent = NULL;
 DebugMode CxbxKrnl_DebugMode = DebugMode::DM_NONE;
-char* CxbxKrnl_DebugFileName = NULL;
+std::string CxbxKrnl_DebugFileName = "";
```
PatrickvL: For this to work, all uses of `CxbxKrnl_DebugFileName` (that are currently assuming it to be a `char *` ) should be extended with a `.c_str()` call

PatrickvL: Nevermind. You did that in the next commit

PR #867, opened 2018-01-19
New EMUPATCH(D3DDevice_GetTexture), forwarded to D3DDevice_GetTexture2.

PatrickvL:
Corrected all texture-related types to X_D3DBaseTexture.

Added a few safeguards against accessing null pointers (this part is the only functional change of this commit).


PatrickvL:
No-risk, self-merged.

PR #868, opened 2018-01-19
Various fixes

PatrickvL:
SMBus now actually forwards calls to SMC (initial connection used the wrong address).
HalWriteSMBusValue now actually returns success when a device responded.
No more EMUPATCH(D3DResource_AddRef) - it's entirely bypassed without discernable downsides.
Texture-conversion now uses new row-based (libyuv-like) color conversion callbacks.


PatrickvL:
Thanks for merging (and testing, everybody)

Issue #198, opened 2017-02-27
Complete EEPROM handling


PatrickvL:
Now https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/191 is merged, the following steps must still be taken :
* ~~move EEPROM related code (EEPROMInfo) into a self-contained source code file~~
* extend the GUI with EEPROM configuration options (see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/35 )
* ~~update checksums of all EEPROM sections when values are changed~~
* ~~read and write the EEPROM from a "EEPROM.bin" file~~ (using a [memory-mapped file](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366761(v=vs.85).aspx) would be simplest)
* ~~when "EEPROM.bin" is missing, call EmuInitializeDefaultEEPROM()~~
* ~~fix all TODO's in ExSaveNonVolatileSetting() :~~
  * ~~TODO : Only allow writing to factory section contents when running in DEVKIT mode, otherwise, nil the info pointer.~~
  * ~~TODO : Make EEPROM-access thread-safe~~
  * ~~TODO : When writing to the factory settings section (thus in DEVKIT mode), set XboxFactoryGameRegion to FactorySettings.GameRegion, so XC_FACTORY_GAME_REGION will reflect the factory settings.~~
  * ~~TODO : For each section being accessed, recalculate it's checksum~~

PatrickvL:
PR https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/215 fixes :
* read and write the EEPROM from a "EEPROM.bin" file
* when "EEPROM.bin" is missing, call EmuInitializeDefaultEEPROM()

PatrickvL:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/217/commits/8ceca49e6fa4b0a98a17de0595543962a3d44cb7 fixes :
* move EEPROM related code (EEPROMInfo) into a self-contained source code file

Fisherman166:
PR #925 adds:

- Calculate eeprom Section CRC checksum on write (Checksum2/3 in the User and Factory settings). Checksum CRCs are also re-calculated on boot.
- Run data checksum on eeprom read to ensure section data is correct.
- Add plumbing for thread safety in eeprom reads/writes (throws an exception right now due to the current thread returning null).

[...]

PR #869, opened 2018-01-19
Create Xbe Printer class for more modular Xbe Dumping

Fisherman166:
DumpInformation in the Xbe class was a bit of a mess. All of the functionality contained in that single function was moved into its own class. Each major block of information dumped was put into its own function. (Almost) All of the c style I/O and strings was replaced with C++ style I/O and boost library formatted strings. Doing all of this allowed me to clean up the GUI code as well to make the code clearer.

Bundled with this puil request is printing the string representation of the XBE game rating field for issue #686.

Finally, to confirm that I did not change functionality of the dump, I diffed an Xbe dump from before my changes to an Xbe dump after my changes and observed no diffs.


haxar:
Is a Boost dependency necessary?

Fisherman166:
Initially I was replacing fprintf with stringstreams but that required major code modification and the syntax for stringstream is awful. sprintf is another possibility but again that would be using a C stdlib function. Researching online, boost seemed to be the preferred formatting library for C++ strings, so I decided to use it. I only included the dependencies needed to use the format library (which is fairly large).

It also looks like I broke the debug build with this because it cannot find boost/format.hpp. This is my first time working with Visual Studio in years so I didn't know know that debug/release builds could have difference import paths.

PatrickvL:
Thank you for cleaning things up and adding additional details to the Xbe dumping - it's appreciated.

I'm feeling reluctant however to merge in a dependency on the boost library, when all it really offers is some syntactic sugar coating. Could you please revise it such that boost can be left out?

I'd appreciate that, as your solution (whatever it will be) would also help in cleaning up the pixel shader dumping code I recently back-ported from Dxbx.

PatrickvL:
PS: Once you've finished this, which other issue would you like to take on?

x1nixmzeng:
The debug build is failing with this message:

> .\..\src\Common\Xbe.cpp(41): fatal error C1083: Cannot open include file: 'boost/format.hpp': No such file or directory

However, using the Boost library for this task is overkill - and ideally any imported libraries should use submodules rather than direct copies of the sourcecode!

PatrickvL:
Thanks for the update! And indeed: watch out for your health

src/Common/Xbe.h
```diff
@@ -62,8 +62,8 @@ class Xbe : public Error
         // export to Xbe file
         void Export(const char *x_szXbeFilename);
 
-        // dump Xbe information to text file
-        void DumpInformation(FILE *x_file);
+        bool DumpInformationToFile(std::string);
+        void DumpInformationToConsole();
```
PatrickvL: Can I have this one last request? Replace these two "dump to" functions by one "dump" function that returns a string. Then, in WndMain send that string to either file or console. That way, the Xbe file doesn't decide what can be done with the dump output. It just returns it.

LukeUsher: I second this. After this is changed, I think we're good to go!

PatrickvL:
#55 (fix all compile hints and warning) would be very nice to have (but be carefull not to influence current functionality)
#198 (complete EEPROM handling) is also a welcome addition (but more complex)

src/Common/Xbe.h
PatrickvL: Later, the XbePrinter class could be moved towards its own file, to remove even more unrelated functionality from the Xbe class itself (which would make it simpler to reuse in other contexts). But that won't be high on my priority list.

Fisherman166: Fixed. DumpInformation returns a string and the caller decides how to output it.

PatrickvL:
Thanks!

PR #872, opened 2018-01-21
Azurik updates

blueshogun96:
These three fixes should be enough to get Azurik: Rize of Perathia working more acceptably (tested with NTSC version).


src/CxbxKrnl/EmuFile.cpp
```diff
@@ -423,6 +433,19 @@ NTSTATUS CxbxConvertFilePath(
 				HostPath.append(1, '\\');
 		}
 
+		// Lastly, remove any '\\' sequences in the string (this should fix the problem with Azurik game saves)
+		replace_all( RelativePath, "\\\\", "\\" );
+
+		if (g_bPrintfOn) {
```
PatrickvL: This looks like a duplicate of the code below?

PatrickvL:
Nice work, and thanks for this PR!

I may have found one small issue - it's a remark on one file. Please review

LukeUsher:
Sonic Heroes E3 Demo barely works at all anyway, it's not a big loss at this point. Fix what Patrick mentioned (the duplication) and we'll merge. Sonic Heroes E3 can be investigated as a separate incident.

src/CxbxKrnl/EmuFile.cpp
```diff
@@ -331,6 +331,16 @@ void copy_string_to_PSTRING_to(std::string const & src, const xboxkrnl::PSTRING
 	memcpy(dest->Buffer, src.c_str(), dest->Length);
 }
 
+void replace_all(std::string& str, const std::string& from, const std::string& to) {
```
PatrickvL: Credit this to https://stackoverflow.com/a/24315631 ?

src/CxbxKrnl/EmuFile.cpp
blueshogun96: Yeah, my mistake.  I'm fixing that now.

Shogun

PatrickvL:
Thanks!

src/CxbxKrnl/EmuFile.cpp
blueshogun96: Okay, taking your advice on this one.  I was lazy and did a copy/paste from a different stackoverflow question.

Shogun

Issue #192, opened 2017-02-23
Add LEDs to our GUI


PatrickvL:
The original Xbox has a multi-colored led surrounding the power button, which is capable to show 4 intensity levels of green, and 4 levels of red - for a total of 8 bits, and thus 256 combinations.

For this issue, come up with a graphical representation to show the color of the led somewhere in our GUI.

(Internally, it's being controlled via byte-writes to SMBus address 0x20, with the low nibble indicating the 4 green levels, and the high nibble indicating the 4 red states).

ergo720:
Alright, I am at a good point with this. I successfully extended the GUI of cxbx to display the LED on the menu bar and now I have to intercept the SMBus value but I have one question. I see the function of interest is HalWriteSMBusValue() (and its twin HalReadSMBusValue()) in EmuKrnlHal.cpp but in their current form they seem to only be able to handle the addresses 0xA8 and 0xA9 so I will also have to extend these functions to support the address 0x20 right?

PatrickvL:
This has to be investigated first. Any information found, should be documented somewhere (xboxdevwiki perhaps), so whoever reads the code can verify it to the documentation)

PatrickvL:
My compliments for picking this up btw!

ergo720:
You mean something like writing a custom xbe that calls those functions above and pokes the values in a real Xbox? If it takes only this then I think I can do this. BTW below there is a preview of my extended cxbx GUI. Is this design good enough?
![capture](https://user-images.githubusercontent.com/30000751/29030228-e1d0781a-7b79-11e7-8850-e78294697dad.PNG)

LukeUsher:
Keep in mind that although official Microsoft software will use HalWriteSMBusValue, homebrew could write to the registers directly, so you would also have to handle this in EmuX86_Out();

PatrickvL:
Actually, it would be best if smbus was emulated separately, like EPROM

PatrickvL:
Also, it should be done via io read write handlers, which should be called from the kernel functions

ergo720:
[...]

PR #874, opened 2018-01-21
LED implementation

ergo720:
This implements the LED feature requested in issue #192 Tested with Avalauch dashboard (which has an option to change the led color) and confirmed as working.

black
green
orange
red

### DIFF
--- a/resource/Cxbx.rc
+++ b/resource/Cxbx.rc
@@ -161,6 +161,7 @@ BEGIN
         MENUITEM SEPARATOR
         MENUITEM "&About",                      ID_HELP_ABOUT
     END
+	MENUITEM "LED",						ID_LED
 END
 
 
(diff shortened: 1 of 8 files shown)


LukeUsher:
Awesome thanks!

One suggestion: Would it be possible to default the LED to green when an xbox game is running, then when an xbox title is exited, return it back to no state? That way it gives the impression of it working as intended, even if a title does not explicitly set the state.

ergo720:
It does that already: when no xbox game is running, the LED is off/no state (black), when an xbe is launched, it defaults to green (unless the titles specifies otherwise) and when the xbe is stopped, the LED is black again

LukeUsher:
My bad, I just noticed that the LED was blank in the top screenshot, that's all

Issue #877, opened 2018-01-21
Make kernel "Power off" end emulation altogether


PatrickvL:
Our kernel currently doesn't end emulation when the "Power off" API (`HalInitiateShutdown` or something like that) is called.

Make it so, that emulation ends.

PatrickvL:
Implemented by https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/881

Lovely!

PR #881, opened 2018-01-21
HalInitiateShutDown implementation

ergo720:
This implements the kernel function HalInitiateShutDown  and fixes the issue #877 Tested with Avalaunch dashboard: the shut down option correctly stops the emulation now


src/CxbxKrnl/SMCDevice.h
```diff
@@ -74,7 +74,7 @@
 //
 //Command	Description
 //0x01	PIC version string counter reset
-//0x02	reset and power off control
+#define SMC_COMMAND_RESET 0x02	 //0x02	reset and power off control
```
x1nixmzeng: Would this work better as an enum instead of using a preprocessor macro?

ergo720: All commands of the SMC device are defined with macros, I just followed the pattern to keep things consistent....

LukeUsher:
Great job, and thanks for emulating this properly (via the SMC) rather than hacking it in the kernel.

PR #882, opened 2018-01-21
The NV2A is now registered as a hardware device, just like all other devices

PatrickvL:
Like the NVNet device, it's implementation is currently just a forward to the existing code.
Still, this removes a dependency from EmuX86 on the NV2A device - from now on it's reached via the generic device framework.

### DIFF
--- a/src/CxbxKrnl/devices/video/nv2a.cpp
+++ b/src/CxbxKrnl/devices/video/nv2a.cpp
@@ -1,0 +1,122 @@
+// This is an open source non-commercial project. Dear PVS-Studio, please check it.
+// PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com
+// ******************************************************************
+// *
+// *    .,-:::::    .,::      .::::::::.    .,::      .:
+// *  ,;;;'````'    `;;;,  .,;;  ;;;'';;'   `;;;,  .,;;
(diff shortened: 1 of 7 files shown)


src/CxbxKrnl/devices/video/nv2a.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+	PCIBarRegister r;
+
+	// Register Memory bar :
+	r.Raw.type = PCI_BAR_TYPE_MEMORY;
+	r.Memory.address = NV2A_ADDR >> 4;
+	RegisterBAR(0, NV2A_SIZE, r.value);
+
+	// Register physical memory on bar 1
+	r.Memory.address = 0;
```
LukeUsher: This is not right: I should have done a better review ;) The starting address here is the physical address in the Xbox memory the VRAM should be placed at, and should point to the TILED memory region ((0xF0000000 >> 4))

This maps VRAM address 0 to 0xF0000000. (this is what our TILED memory is: VRAM accessed via the GPU device)

PatrickvL: Good to know. Right now, this bar isn't accessed yet, but it'll have to be addressed soon enough. Thanks for the heads-up.

PR #888, opened 2018-01-22
Isolated XbePrinter into it's own file, so the Xbe.h/.cpp files don't need to know about it.

PatrickvL:


PatrickvL:
No-risk self-merge

PR #889, opened 2018-01-22
HalIsResetOrShutdownPending implementation

ergo720:
Very easy to implement, it just checks for a global variable. HalInitiateShutdown still works in Avalaunch, however, I don't have any titles that call HalIsResetOrShutdownPending so I don't know if there are problems. It's so simple that there shouldn't be though


ergo720:
Yes, but I'm not sure if the atomic is not needed. Since the variable is global and shared, data races between threads could happen for it

PatrickvL:
Thanks, great to have yet another kernel API down!

PR #891, opened 2018-01-23
D3DBegin/End Tweaks

LukeUsher:
This fixes a crash in Silent Hill 2 where it overflows the IVB table. This very likely effects other titles too.
It still doesn't reach gameplay, but no longer crashes within SetVertexData4f.

A better long term solution would be to have this table and the resulting vertex buffer by dynamically sized, so overflow is not possible, but this works well as a short-term workaround, and is intended as a quick fix until we solve the problem better (by unpatching these functions, allowing push buffers to be created and parse them instead)


PatrickvL:
Another, intermediate, solution could be to flush the buffer when it is in risk of overflowing. This does require detection of primitive ending vertices, so it's a bit more complex than just enlarging the buffer like you've done now.

PR #892, opened 2018-01-23
NV2A expansion

PatrickvL:
Better to start work on NV2A with all available symbols, improvements and tweaks

### DIFF
--- a/src/devices/video/EmuNV2A.cpp
+++ b/src/devices/video/EmuNV2A.cpp
@@ -351,7 +351,7 @@ struct {
 	uint32_t enabled_interrupts;
 	std::thread puller_thread;
 	Cache1State cache1;
-	uint32_t regs[NV_PFIFO_SIZE]; // TODO : union
+	uint32_t regs[_NV_PFIFO_SIZE]; // TODO : union
 } pfifo;
 
 struct {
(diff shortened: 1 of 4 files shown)


x1nixmzeng:
Neat! Glad your branch is finally seeing the light of day!

PatrickvL:
Thanks, but this part wasn't the reason my WIP branch was capable to go further. (That had to do with less HLE patching, using D3D global state variables, generic render state conversions, centralised texture conversion and other stuff - Luke is redoing that now in his experimental branch, in smaller steps with similar results)

PatrickvL:
The NV2A should be emulated at low level, that's what this work is slowly preparing for.
We've got enough resources now to get it further, it just takes time. (Resources like pbkit, openxdk, nxdk, xqemu, OpenXbox, mame and other Nvidia chipsets emulated)

ObiKKa:
Patrick, to your second comment up above, has Luke been reading/checking your WIP branch before starting on his version work?

3rd comment: That's an awful work of research and work!

LukeUsher:
@ObiKKa I refer to it at time to time, but my work mostly an independent implementation based on my own research and findings. What is great is we both came to the same conclusions and hit similar problems in the process, so we must be on the right track ;)

PatrickvL:
Also, remember that Radwolfie already copied over most, if not all, of the non-rendering code changes from my WIP branch. This includes my OOVPA extensions, hidden behind an ifdef

RadWolfie:
@PatrickvL 😛 Yup, since some of the references are not declared or functional at the time back then in master branch.

It is supposedly for preparation to have WIP_LessVertexPatching branch with DXD8 relative files readily available re-port on top of master branch.

Issue #878, opened 2018-01-21
Improve "Xbe not found" error message


PatrickvL:
Currently, when a title reboots the emulated system, but didn't specify a path to another Xbe to launch, Cxbx-Reloaded defaults to loading a dashboard from partition 2.

However, when the Xbe can't be found, an error message is generated which might confuse the user.

For this, append the path and Xbe file name to the error message, making it clear which Xbe couldn't be found.

PatrickvL:
Fixed by https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/895

PR #895, opened 2018-01-26
Xbox fatal error implementation

ergo720:
This implements the Xbox UEM (Universal Error Message). I tested it with Tenchu by changing (with the debugger) the routine in HalReturnToFirmware from ReturnFirmwareQuickReboot to ReturnFirmwareFatal to force a fatal error screen.
BTW, I needed to change the encoding of EmuD3D8.cpp from UTF-8 to UTF-8-BOM (byte order mark) to render the Japanese and unicode characters of the message correctly. When I did it, the compiler warned about a '\u0096' character (START OF GUARDED AREA) on line 397 (Warning	C4566). I re-wrote that string without the hidden character and that fixed it. I hope it doesn't cause any ill effects (I think it was introduced by accident anyway).

Tenchu Fatal Error Screen
cxbx_x264

EEPROM at byte 250 now stores the error code
eeprom


src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -1365,6 +1365,83 @@ static LRESULT WINAPI EmuMsgProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPar
         }
         break;
 
+		case WM_PAINT:
+		{
+			if (g_CxbxPrintUEM)
+			{
+				// Draw the universal error message (UEM)
```
PatrickvL: Could you put this entire statement-block into a separate function, please?

PatrickvL:
Besides a few remarks, it's great to see more low-level stuff being implemented! Keep it up!!

src/CxbxKrnl/EmuKrnlHal.cpp
```diff
@@ -495,8 +495,21 @@ XBSYSAPI EXPORTNUM(49) xboxkrnl::VOID DECLSPEC_NORETURN NTAPI xboxkrnl::HalRetur
 		break;
 
 	case ReturnFirmwareFatal:
-		CxbxPopupMessage("Emulated Xbox hit a fatal error (might be called by XapiBootToDash from within dashboard)");
+	{
+		// NOTE: the error code is displayed by ExDisplayFatalError by other code paths so we cannot display it without
```
LukeUsher: ExDisplayFatalError is not part of the running Xbe but is part of the Xbox Kernel itself (specifically the Init section) this means that it is not part of the running game/xbe, and can never be called after the an Xbox title has started running, as it is deleted from memory as soon as an Xbe starts.

So this means two things
1. It's impossible to make an OOVPA for, as it never exists while an Xbe is running
2. It's impossible for this function to be called after init completes, even within the kernel.

So early boot is the one and only place ExDisplayFatalError can be called, it's safe to say we don't need it/can't patch or run it anyway.

If we really want an authentic experience (showing the error screen when an Xbe fails to load), we would just have to render the error screen using the correct error code.

ergo720:
@LukeUsher I think I can use your suggestion to show a fatal error screen when the MS dashboard cannot be found in the partition2 instead of showing the confusion error "xbe not found". This would also solve issue #878. Unfortunately, the error is triggered from XeLoadImage which doesn't call XeLoadSection so I will have to do the check from somewhere else

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -1365,6 +1365,83 @@ static LRESULT WINAPI EmuMsgProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPar
...
+
+				HFONT hFont = CreateFont(nHeight, 0, 0, 0, FW_NORMAL, FALSE, FALSE, FALSE, DEFAULT_CHARSET, OUT_DEFAULT_PRECIS, CLIP_DEFAULT_PRECIS, ANTIALIASED_QUALITY, FF_ROMAN, "Verdana");
+
+				HGDIOBJ tmpObj = SelectObject(hMemDC, hFont);
+
+				SetBkColor(hMemDC, RGB(0, 0, 0));
+
+				SetTextColor(hMemDC, RGB(0, 204, 0));
+
+				std::wstring wstr(
```
x1nixmzeng: Can you move this to a resource string, instead of hardcoding it here.

See how the contributors text file is imported and linked for reference.

ergo720:
I'm temporarily closing this since implementing what I said above is more complicated than I thought

ergo720:
The following needs testing: in my initial attempt sometimes I was noticing that the caption menu and the led color where cleared after rebooting to the dasboard. It doesn't seem to happen anymore with the new implementation though. Anyway, if the users haven't installed the dashbaord in the partition2 they will now be greeted by the UEM instead (unless they launch it from the recent xbe list). It also solves issue #878 by displaying the faulty xbe name
cxbx
dash

PatrickvL:
Thank you very much!

PR #896, opened 2018-01-26
Centralized hardware devices

PatrickvL:
Centralized hardware devices around Xbox sytem, (revisions of) components chosen based on hardware revision.

Most is not yet implemented, but more will surely follow.
Moved SMBUS-address defines and global device variables towards Xbox.h

### DIFF
--- a/build/win32/Cxbx.vcxproj.filters
+++ b/build/win32/Cxbx.vcxproj.filters
@@ -234,7 +234,7 @@
     </ClCompile>
     <ClCompile Include="..\..\src\Common\Win32\Threads.cpp">
       <Filter>Shared</Filter>
-    </ClInclude>
+    </ClCompile>
     <ClCompile Include="..\..\src\devices\MCPXDevice.cpp">
       <Filter>Hardware</Filter>
     </ClCompile>
--- a/build/win32/Cxbx.vcxproj.filters
+++ b/build/win32/Cxbx.vcxproj.filters
@@ -234,6 +234,8 @@
     </ClCompile>
     <ClCompile Include="..\..\src\Common\Win32\Threads.cpp">
       <Filter>Shared</Filter>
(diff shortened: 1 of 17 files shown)


src/devices/SMCDevice.h
```diff
@@ -49,19 +49,17 @@
...
 // From https://web.archive.org/web/20100617022549/http://www.xbox-linux.org/wiki/PIC :
 // Command	Description
 #define SMC_COMMAND_VERSION 0x01	// PIC version string
 //0x03	tray state
 #define SMC_COMMAND_AV_PACK 0x04	// A / V Pack state
-//0x09	CPU temperature(�C)
-//0x0A	board temperature(�C)
+#define SMC_COMMAND_CPU_TEMP 0x09 // CPU temperature (�C)
+#define SMC_COMMAND_GPU_TEMP 0x0A // GPU (board?) temperature (�C)
 //0x0F	reads scratch register written with 0x0E
```
ergo720: Are you sure about GPU temperature? I actually think the command measures the motherboard temperature. UnleashX on my xbox reports this as M/B temperature as well

PatrickvL: I only copied over what was already in code, so you're probably right

build/win32/Cxbx.vcxproj.filters
```diff
@@ -234,6 +234,8 @@
     </ClCompile>
     <ClCompile Include="..\..\src\Common\Win32\Threads.cpp">
       <Filter>Shared</Filter>
+    <ClCompile Include="..\..\src\devices\MCPXDevice.cpp">
+      <Filter>Hardware</Filter>
```
LukeUsher: This (and the change below) corrupt the filters file and prevent Visual Studio from processing it.
It looks like an easy fix, just insert a </ClCompile> line above these new includes, since for some reason, the one below <Filter></Filter> got omitted.

PR #897, opened 2018-01-27
Xbeinfo

donwayo:


src/Common/XbePrinter.cpp
```diff
@@ -378,6 +379,21 @@ std::string XbePrinter::GenAlternateSignatureKeys()
     return text;
 }
 
+std::string XbePrinter::GenExtraInfo()
+{
+    std::stringstream text;
+    SSTREAM_SET_HEX(text);
+    if (Xbe_certificate->dwSize == 0x1EC)
```
PatrickvL: Better check greater or equal

src/Common/XbePrinter.cpp
```diff
@@ -378,6 +379,21 @@ std::string XbePrinter::GenAlternateSignatureKeys()
     return text;
 }
 
+std::string XbePrinter::GenExtraInfo()
+{
+    std::stringstream text;
+    SSTREAM_SET_HEX(text);
```
PatrickvL: If you're applying the other suggestion, you could at the same time move this line inside the if statement block

PatrickvL:
Thanks for your contribution! Hopefully more will follow?!?

PatrickvL:
Thanks!!

PatrickvL:
Oh, does this relate to any pending issue? If so, please mention it here (with a #hash sign, so it becomes a bi-directional link) and evaluate how much would still need to be done.

donwayo:
Hi Patrick! No, it doesn't relate with any pending issue, but having this information in the dump might allow one to verify the dump data with the signature key :)

PR #898, opened 2018-01-28
Various Fixes from patch-removal-experiment

LukeUsher:
This is just a few of the modifications/tweaks from my patch-removal-experiment branch that are low impact/are safe to merge into master right now.

Nothing that impressive, but with this titles that use XFileMediaObject functionality of DSOUND are much less likely to freeze, hang or crash.


PatrickvL:
Thanks!

PR #900, opened 2018-01-30
HalRegisterShutdownNotification implementation

ergo720:
This implements the kernel function HalRegisterShutdownNotification. Note that the shutdown routines can't be executed at the moment since not enough hardware is LLEd


PatrickvL:
Looking good, thanks!

PR #903, opened 2018-01-31
Thread names

x1nixmzeng:
This is to help debugging under Visual Studio.

Visual Studio uses a special interrupt to set the thread name, and auto-generates thread name using `__FUNCTION__` to set a reasonable default name. However for `std::thread` the inferred thread name is mangled due to anonymous functions which makes it hard to pick out threads from a glance.

Going forwards, naming thread manually like this allows other debuggers to use correct names without any symbol files 😆 

Please feel free to correct any names I've given!

### DIFF
--- a/build/win32/Cxbx.vcxproj
+++ b/build/win32/Cxbx.vcxproj
@@ -190,6 +190,7 @@
     </PreBuildEvent>
   </ItemDefinitionGroup>
   <ItemGroup>
+    <ClInclude Include="..\..\src\Common\Win32\Threads.h" />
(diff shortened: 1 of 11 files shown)


PatrickvL:
Nicely done! Just one small remark, if you have time to correct it. (Otherwise we could just merge it and correct it later...)

x1nixmzeng:
Thanks for the review!

The [usage on MDSN](https://msdn.microsoft.com/en-us/library/xcb2z8hs.aspx) passes -1 for the "current thread", but it doesn't hurt to be explicit!

PR #907, opened 2018-02-06
Implement 'Reset to Defaults' menu feature.

### LINKED ISSUE #791: Implement functionality for resetting to default settings
In some cases, the Cxbx-Reloaded registry becomes corrupted, this can manifest in various ways, ranging from input issues, errors, to outright crashes.

I propose we should add a menu option to reset all settings to defaults, to cover this situation.

Obviously, It would be better if this corruption didn't happen at all, so if that can be solved, great. 

That said, a reset to default is still a nice option to have.

Voxel9:
This fixes #791.

Basically, the way this works is: _Warn user -> if they click yes -> delete all values in the root of Cxbx-r registry tree -> Remove XBAudio/Video/Controller keys -> Prompt to exit -> Exit_.
When Cxbx is started up again, it creates the default values on startup as if it had never been launched before.

Another quick note: I had to introduce a new bool variable 'g_SaveOnExit' in some places, since Cxbx currently saves all its registry values on exit. (As Cxbx already has the values from when it booted up already loaded, so we don't want _those_ to write to the registry on exit when we just deleted them!)


src/Cxbx/ResCxbx.h
```diff
@@ -1,6 +1,5 @@
 //{{NO_DEPENDENCIES}}
 // Microsoft Visual C++ generated include file.
-// Used by C:\msys64\home\thesy\Cxbx-Reloaded\resource\Cxbx.rc
```
x1nixmzeng: This file is auto-generated. Removing this line is pointless because it's hardcoded into the binary file too!

src/Cxbx/WndMain.cpp
```diff
@@ -1082,6 +1084,36 @@ LRESULT CALLBACK WndMain::WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lP
...
 			break;
 
+			case ID_SETTINGS_INITIALIZE:
+			{
+				int ret = MessageBox(m_hwnd, "Warning: This will reset all Cxbx-Reloaded settings to their default values."
+					"\nAre you sure you want to proceed?", "Cxbx-Reloaded", MB_ICONEXCLAMATION | MB_YESNO);
+
+				if (ret == IDYES) {
+					HKEY hKey;
+					if (RegOpenKeyEx(HKEY_CURRENT_USER, "Software\\Cxbx-Reloaded", 0, KEY_SET_VALUE, &hKey) == ERROR_SUCCESS) {
```
x1nixmzeng: Could you move this logic into it's own function

src/Cxbx/ResCxbx.h
Voxel9: Then how are you supposed to prevent this from happening?

src/Cxbx/WndMain.cpp
Voxel9: Done.

src/Cxbx/ResCxbx.h
x1nixmzeng: It's unfortunately a full path to the local resource file it's linked to (you have have multiple resource files).

It's path changes whenever the binary file is modified - but leaks potentially personal data as you've found - just see the file history https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commits/master/src/Cxbx/ResCxbx.h

I recommend just using mapping a drive for your Github work. My sources are generic:

> // Used by S:\GitHub\Cxbx-Reloaded\resource\Cxbx.rc

PatrickvL:
Thank you!

PR #908, opened 2018-02-07
Less debug file clutter

CakeLancelot:
This changes the pixel and vertex shader file dumping to only occur on debug builds. I'm not sure whether this should be the default behavior, but it does reduce clutter in the working folder on release builds significantly. (after playing a few games I had ~200 PsDef.txt and ~20 failed.xvu files)

Unless these are enabled on release for getting useful information from release builds (for example when someone uploads a log on the game-compatibility repository) I think this is a reasonable change but I'm open to feedback as I'm not fully aware if there are any implications from removing these.

### DIFF
--- a/src/Cxbx.h
+++ b/src/Cxbx.h
@@ -72,6 +72,7 @@ typedef u32              xbaddr;
 /*! xbnull is the type of null address or value*/
 #define xbnull  0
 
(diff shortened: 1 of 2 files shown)


PatrickvL:
Thank you!

PR #909, opened 2018-02-07
Debugger

x1nixmzeng:
So this is a tool I started writing to help debugging Xbox executables inside of Cxbx-Reloaded:

Cxbx-Debugger screenshot

It's ongong, and doesn't resolve symbols yet (one of the primary goals) but I'm looking for feedback and shamelessly bumping for interest.

### Building

* Grab this branch
* Update your submodules (recursively)
* Build CxbxDebugger from Visual Studio (it isn't a build dependency)

### Running

* File -> Open XBE
* Emulation -> Start Debugger
* CxbxDebugger opens, hit "Start" (top left)

### Debugging

You can suspend / resume execution at any point. Beware that Cxbx will be checking HLE cache for the first few seconds.

### Breakpoints

There is only support for suspending on file access at the moment.

Under the "Breakpoints" tab, you can add add filter strings (ie ".bmp") and wait for the app to hit.

Use the "File Watch" tab for a list of file access events.


src/Common/CxbxDebugger.cpp
```diff
@@ -0,0 +1,255 @@
...
+			}
+		};
+
+		uint32 GetVirtualFrameBuffer()
+		{
+			const xbaddr FrameBufferAddr = TILED_MEMORY_BASE + 0x40000;
+
+			uint32 VirtualFrameBufferAddr;
+
+			__asm
```
PatrickvL: Why is this part written in assembly?

Sarkie:
Wow!!!

Great work

src/Common/CxbxDebugger.cpp
haxar: Looks like it can be written as:

`uint32 VirtualFrameBufferAddr = **(uint32 **)(TILED_MEMORY_BASE + 0x40000);`

PatrickvL:
I can't wait to start using this once it's fully functional!

src/Common/CxbxDebugger.cpp
PatrickvL: Indeed. Although to be honest, we don't yet put the framebuffer there

LukeUsher: The framebuffer can change location per Xbe too: AvSetDisplayMode kernel API can be used to set the framebuffer to any address the running title chooses.

x1nixmzeng: It was written in assembly because it throws an access violation, which is resolved by the virtual memory manager to get the final pointer - and as Patrick mentioned it's unused for now :)

I'll remove it as there isn't any need to show the rendering result

haxar: FYI, `AvSetDisplayMode()` sets the framebuffer at NV_PCRTC_START (0xfd600800) in NV2A.

LukeUsher:
@x1nixmzeng: As this does not alter core Cxbx-Reloaded code, I'm happy to merge. Is there anything you wish to change or update before then, or can it be merged right now?

x1nixmzeng:
I'm happy with the current state if you want to merge it now 🎉 

Also - just incase you haven't seen - this PR adds a new module which is on GitHub - [see here](https://github.com/x1nixmzeng/Cxbx-Reloaded/blob/cxbx-debugger/.gitmodules). Ideally it could be hosted on the Cxbx-Reloaded team page. Right now it's just to workaround support for binding Capstone to C#

PR #910, opened 2018-02-08
Implement RTLCriticalSection and KeCriticalRegion functions

Fisherman166:
Trying this again. Critical section functions use a map to implement the host OS critical sections while including the Xbox functionality.


src/CxbxKrnl/EmuKrnlRtl.cpp
```diff
@@ -56,47 +56,72 @@ namespace NtDll
...
-		}
-	}
-
-	if (FreeSection < 0)
-	{
-		EmuWarning("Too many critical sections in use!\n");
-	}
-
-	return FreeSection;
+    return TryEnterCriticalSection(find_critical_section(xbox_crit_secion));
```
LukeUsher: Could you also create the critical section here if it doesn't exist yet, like you did with EnterHostCriticalSection

PR #917, opened 2018-02-13
Fixed build configurations for ci

### LINKED ISSUE #916: AppVeyor issue
After merging the debugger commit, AppVeyor is no longer creating the exported artifacts, both debug and release builds contain the error "No artifacts found matching 'export\*.zip' path"

x1nixmzeng:
To fix #916


x1nixmzeng:
So the original artifacts are back, but the debugger files aren't collected - except for `Capstone.dll` :cry:

One fix would be to ignore the platform and just push files into a common zip file

PR #922, opened 2018-02-15
Cleanup of appveyor build scripts

x1nixmzeng:
This should run just twice (once per configuration), meaning all artifacts are uploaded together


x1nixmzeng:
So glad that worked first time 😀

Good to go!

PR #923, opened 2018-02-16
NV2A split

PatrickvL:
This PR splits up the NV2A code in the same way OpenXbox did, to allow easier maintainance and simpler comparisons.

Additionally, more code is ported over (the OpenGL-using parts are still disabled behind a `#ifdef COMPILE_OPENGL`).

HLE shouldn't be impacted, LLE still hangs at an infinite loop (debug XDK samples generate : `ERR[D3D]: Possible deadlock, blocked for more than 2 seconds` messages).

Regardless, the code is now in a state which would best be merged in.


src/devices/video/nv2a.h
```diff
@@ -26,21 +26,605 @@
...
+#define g_free(x) free(x) // Compatibility
+#define g_malloc(x) malloc(x) // Compatibility
+#define g_malloc0(x) calloc(1, x) // Compatibility
+#define g_realloc(x, y) realloc(x, y) // Compatibility
+
+#define USE_TEXTURE_CACHE
+
+// Public Domain ffs Implementation
+// See: http://snipplr.com/view/22147/stringsh-implementation/
+constexpr int ffs(int v)
```
x1nixmzeng: The use of this constexpr means VS2015 is no longer supported

PatrickvL: Good catch! Please create an issue for that (or a pull request if you want) to restore Vs 2015 compatibility while keeping Vs 2017 specific optimisations

PR #925, opened 2018-02-18
EEPROM improvements for issue #198

Fisherman166:
Calculate eeprom Section CRC checksum on write (Checksum2/3 in the User and Factory settings). Run data checksum on eeprom read to ensure section data is correct. Add plumbing for thread safety in eeprom reads/writes (throws an exception right now due to the current thread returning null).

### DIFF
--- a/src/CxbxKrnl/EmuKrnlEx.cpp
+++ b/src/CxbxKrnl/EmuKrnlEx.cpp
@@ -64,6 +64,82 @@ namespace NtDll
 #include <ntstatus.h> // For STATUS_BUFFER_TOO_SMALL
 #pragma warning(default:4005)
 
+static xboxkrnl::RTL_CRITICAL_SECTION eeprom_crit_section;
+
+static xboxkrnl::PRTL_CRITICAL_SECTION get_eeprom_crit_section()
+{
+    static xboxkrnl::PRTL_CRITICAL_SECTION crit_sec_ptr = NULL;
+    if(crit_sec_ptr == NULL) {
+        crit_sec_ptr = &eeprom_crit_section;
+        RtlInitializeCriticalSection(crit_sec_ptr);
+    }
+    return crit_sec_ptr;
+}
+
+static bool section_does_not_require_checksum(
+    xboxkrnl::XC_VALUE_INDEX index
+)
+{
+    switch(index) {
+        case xboxkrnl::XC_FACTORY_GAME_REGION:
+            return true;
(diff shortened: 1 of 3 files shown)


src/CxbxKrnl/EmuKrnlEx.cpp
```diff
@@ -319,51 +395,52 @@ XBSYSAPI EXPORTNUM(24) xboxkrnl::NTSTATUS NTAPI xboxkrnl::ExQueryNonVolatileSett
...
-			// 32-bit 1's complement sum, that must equal 0xFFFFFFFF)
-
-			// Set the output value type :
-			*Type = value_type;
-			// Clear the output value buffer :
-			memset(Value, 0, ValueLength);
-			// Copy the emulated EEPROM value into the output value buffer :
-			memcpy(Value, value_addr, result_length);
+		if ((int)ValueLength >= result_length) {
+			// FIXME - Entering the critical region causes an exception because
```
PatrickvL: Does this mean critical sections have this problem everywhere they're being used? Because that would be a pretty serious issue worth tracking down.

PatrickvL:
Well done! Could you add a description to this issue (and possibly the readme) about what to do when the EEPROM gets corrupted?

Also (and this is actually my fault) the current way we access EEPROM values isn't doing EEPROM hardware accesses like a real machine would do. Thus, this implementation wouldn't run on the real machine, nor in openxbox. Something to keep in mind

src/CxbxKrnl/EmuKrnlEx.cpp
Fisherman166: I am not sure. I could have sworn that I have seen a game enter a critical region when I was doing my Rtl code testing, but I may be wrong. More investigation is needed here.

Fisherman166:
Issue updated with recovery steps.

The Xbox Kernel code reads 0xFFFE (encrypted section) and 0xFFFF (XC_MAX_ALL) in the exact same way. It uses xboxkrnl_398 (a function that calls HalReadSMBusValue value 0x100 times to read the entire eeprom) to read the contents of the eeprom into a local buffer. Then the local buffer is copied into the user provided buffer. No data checksum is ran on either of these reads and I can see no decryption functions that are called. I can only assume that the Xbox returns encrypted data and it is on the caller to decrypt the data.

My code takes into account that real hardware accesses are not occurring. It is a good approximation of what the hardware is doing based on our current modeling of the EEPROM.

PR #928, opened 2018-02-19
Fix visual update for RCT title

RadWolfie:
Other titles does not appear to have any fps impact except for Dino
Crysis 3 (it goes away after couple seconds though).


RadWolfie:
This is not a permanent fix, only improve visual update necessary for general titles. fps will drop for only 1 or two seconds then back to normal.

PR #933, opened 2018-02-20
Kernel : Add KWAIT_BLOCK's to all callers of KeWaitForMultipleObjects

PatrickvL:


src/CxbxKrnl/EmuKrnlKe.cpp
```diff
@@ -1836,6 +1836,8 @@ XBSYSAPI EXPORTNUM(159) xboxkrnl::NTSTATUS NTAPI xboxkrnl::KeWaitForSingleObject
 {
 	LOG_FORWARD("KeWaitForMultipleObjects");
 
+	KWAIT_BLOCK WaitBlock;
```
RadWolfie: What's the real reason for adding this? KeWaitForMultipleObjects from xboxkrnl does not use it internally from Cxbx.

src/CxbxKrnl/EmuKrnlNt.cpp
```diff
@@ -1978,6 +1982,8 @@ XBSYSAPI EXPORTNUM(235) xboxkrnl::NTSTATUS NTAPI xboxkrnl::NtWaitForMultipleObje
 {
 	LOG_FORWARD("KeWaitForMultipleObjects");
 
+	KWAIT_BLOCK WaitBlockArray[9];
```
RadWolfie: Using fixed array for multiple object is not exactly wiser way to do this. The proper way would be check the total count to see if there are more than 9, then it should increase the size of array. However, WaitBlockArray is not being in used inside Cxbx's xboxkrnl::KeWaitForMultipleObjects function.

PatrickvL: The real kernel does the exact same thing (reserving a fixed amount on stack), just with more plumbing around it. With this I'm trying to avoid exceptions once more kernel functions are going to be implemented.

RadWolfie:
Appveyor is failing for `EmuKrnlKe.obj : error LNK2001: unresolved external symbol _NtResetEvent`

PatrickvL:
This became a mess. I'll retry later

PR #934, opened 2018-02-20
NV2A : Prevented incorrectly logged "Unknown Graphics Class/Method" messages

PatrickvL:
Also some slight optimizations (and one additional, ignored PGRAPH command : SET_FLAT_SHADE_OP)


JayFoxRox:
The ramht cleanup should have been upstreamed to XQEMU. Please do!

PR #935, opened 2018-02-21
Misc build warning fixes

x1nixmzeng:


PatrickvL:
Nice work! There's a few complaints from AppVeyor on this pr, though. Could you take a look?

x1nixmzeng:
I can't understand why the logger doesn't work. I assumed it just piped stdout/stderr but it looks like it creates a special config :confused:

PatrickvL:
I've asked @LukeUsher to check this out

x1nixmzeng:
After re-reviewing my changes I'm not sure what this PR solves :confused: 

I'm tempted to close this and resurface the alignment fix with my XBE loading work

x1nixmzeng:
Yeah, I'll do that.

Only thing to add is I'm not sure where the x64 config came from

PR #940, opened 2018-02-23
Kernel work on Ob*, Ke* and Nt* API's

PatrickvL:
### It's possible this PR causes regressions, so DON'T MERGE this without solid testing! 
(If any regression occurs, add a comment with enough details, so it can be investigated properly.)

This pull request stops using NtDll for Events, but instead, uses public Windows API's (`CreateEvent`, `PulseEvent`, etc).
This is also reflected in the various `WaitFor` functions and even `NtClose`.

Also, a lot of API's are copied over from OpenXbox and various symbol declarations are put in their own headers files.

One of the positive effects of this work is, that the https://github.com/Luca1991/xbox_kernel_test_suite can run to completion, passing all tests (even the ones added by https://github.com/Fisherman166/xbox_kernel_test_suite/commits/CriticalSection_tests).

Homebrew build using NXDK and/or pbkit will run further too (although we need proper LLE GPU to let these show anything).


PatrickvL:
`NtWaitForSingleObjectEx` on file handles crashes Turok, so I'm closing this until that's resolved.

PR #947, opened 2018-02-26
Introduce GetVertexBufferSize and use it to determine the size of buffers for patching

LukeUsher:
The end result is less missing polygons/better rendering in many titles.

With this, the dashboard runs a little slower, but renders much better.

Many other titles render better with no performance impact, and in some cases, perform much faster than before, due to better caching/reducing the amount of hashing required.


ergo720:
@LukeUsher Does this remove entirely the need to query allocation sizes starting at an offset instead of the beginning? I see there is still one QuerySize call to the VMManager though

LukeUsher:
The behavior of QuerySize should not change: it needs to keep the same behavior as on the Xbox

PR #949, opened 2018-02-27
General fixes and tweaks

LukeUsher:


src/CxbxKrnl/EmuDSound.cpp
```diff
@@ -1549,7 +1549,7 @@ HRESULT WINAPI XTL::EMUPATCH(IDirectSoundBuffer_StopEx)
                 // TODO: How to mock up "release phase"?
                 break;
             default:
-                CxbxKrnlCleanup("Unknown dwFlags from IDirectSoundBuffer_StopEx: %8X", dwFlags);
+                EmuWarning("Unknown dwFlags from IDirectSoundBuffer_StopEx: %8X", dwFlags);
```
RadWolfie: It should require to crash since anything other than defined values base on the official docs. By allowing it to continue will produce unstable result later on which might be cause by false HLE detection.

Since some titles actually combine the define values as a flag method. Switch case would not be the best way to do so.

PR #954, opened 2018-02-28
Prevent Halo from blowing up the stack during initialization

Ernegien:
Without these changes Halo crashes inside a __chkstk call during early game initialization. Will need to be tested against other games known to spawn multiple threads.

### DIFF
--- a/src/CxbxKrnl/EmuKrnlPs.cpp
+++ b/src/CxbxKrnl/EmuKrnlPs.cpp
@@ -287,13 +287,14 @@ XBSYSAPI EXPORTNUM(255) xboxkrnl::NTSTATUS NTAPI xboxkrnl::PsCreateSystemThreadE
 		LOG_FUNC_ARG(SystemRoutine)
 		LOG_FUNC_END;
 
-	// TODO : Arguments to use : KernelStackSize, TlsDataSize, DebuggerThread
+	// TODO : Arguments to use : TlsDataSize, DebuggerThread
 
-	// TODO : Fill KernelStackSize like this :
-	//	if (KernelStackSize < KERNEL_STACK_SIZE) 
-	//		KernelStackSize = KERNEL_STACK_SIZE;
-	//	else
-	//		KernelStackSize = round up;
+	// use default kernel stack size if none specified
+	if (KernelStackSize < KERNEL_STACK_SIZE)
+		KernelStackSize = KERNEL_STACK_SIZE;
+
+	// round up to the next page boundary if un-aligned
+	KernelStackSize = KernelStackSize + 0xFFF & 0xFFFFF000;
 
     static bool bFirstTime = false;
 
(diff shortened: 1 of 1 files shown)


src/CxbxKrnl/EmuKrnlPs.cpp
```diff
@@ -287,13 +287,14 @@ XBSYSAPI EXPORTNUM(255) xboxkrnl::NTSTATUS NTAPI xboxkrnl::PsCreateSystemThreadE
...
-	//	if (KernelStackSize < KERNEL_STACK_SIZE) 
-	//		KernelStackSize = KERNEL_STACK_SIZE;
-	//	else
-	//		KernelStackSize = round up;
+	// use default kernel stack size if none specified
+	if (KernelStackSize < KERNEL_STACK_SIZE)
+		KernelStackSize = KERNEL_STACK_SIZE;
+
+	// round up to the next page boundary if un-aligned
+	KernelStackSize = KernelStackSize + 0xFFF & 0xFFFFF000;
```
x1nixmzeng: Is there not a macro for rounding up?

src/CxbxKrnl/EmuKrnlPs.cpp
```diff
@@ -287,13 +287,14 @@ XBSYSAPI EXPORTNUM(255) xboxkrnl::NTSTATUS NTAPI xboxkrnl::PsCreateSystemThreadE
...
-	// TODO : Arguments to use : KernelStackSize, TlsDataSize, DebuggerThread
+	// TODO : Arguments to use : TlsDataSize, DebuggerThread
 
-	// TODO : Fill KernelStackSize like this :
-	//	if (KernelStackSize < KERNEL_STACK_SIZE) 
-	//		KernelStackSize = KERNEL_STACK_SIZE;
-	//	else
-	//		KernelStackSize = round up;
+	// use default kernel stack size if none specified
+	if (KernelStackSize < KERNEL_STACK_SIZE)
```
x1nixmzeng: The parameter is not optional, so the `KernelStackSize` is required. What values are you seeing with Halo?

src/CxbxKrnl/EmuKrnlPs.cpp
Ernegien: No macro currently, but there is a RoundUp utility method in CxbxUtil.h that could be used if preferred.

src/CxbxKrnl/EmuKrnlPs.cpp
Ernegien: Yeah, poor commenting on my part. Should probably be changed to "use default kernel stack size if lesser specified". Nothing too crazy with Halo; at least a few threads wanted to start with 0x4000 for the stack but were only given 0x3000 instead as specified by the KERNEL_STACK_SIZE constant in Cxbx. I didn't care to track down specifically which ones were starving.

LukeUsher:
Awesome, thanks for this!

Ernegien:
Thanks, I plan on it :)

PR #956, opened 2018-03-01
Pixel shader switchover

LukeUsher:


PatrickvL:
The Turok regression could be due to the removal of the StateBlock patches (which has nothing to do with the switchover to new pixel shader).

gandalfthewhite19890404:
That PR cause regress in Need for Speed Underground - graphics is completely missing, in previous PR954 game shows something, broken but something.

LukeUsher:
#1054 and #1297 fix some issues that were introduced with this PR.

Issue #966, opened 2018-03-05
Enable V-sync by default


PatrickvL:
Many users ask how to limit the framerate. To avoid that, enable V-sync by default, meaning : When reading the V-sync setting from registry, if the value is absent, use "enabled" as default value.

Also, verify all LLE settings are turned off by default.

Margen67:
If the primary monitor's refresh rate isn't 60hz it won't limit to 60fps.

LukeUsher:
It also won't have any effect for games that use alternate timing sources other than vblank.

BenNottelling:
@Margen67 True, but let's say a game is running at 250fps enabling V-sync would bring the framerate down to 144 as I have a 144Hz monitor which is at least closer to being playable, plus we could include instructions on how to change the monitor refresh rate to 60Hz. All in all it won't cause any harm. Oh and an idea: is there any part of the directX API to change the monitor refresh rate? Or in a windows API? If it's available to us we could have Cxbx set the monitor refresh rate temporarily to 60 and then enable V-sync. I have no idea if it's possible, but it's still an idea

slx7R4GDZM:
Wouldn't it be better to implement a frame-limiter with a modifiable limit?

LukeUsher:
The frame limiting/vsync only hides the real problem: Which is that Xbox games are not designed to run at modern processor speeds. The *real* fix would be to implement proper CPU emulation rather than using direct code execution, and through that, intercept timing instructions like rdtsc to report the same clock speeds as the real xbox, but that is a LOT of work, as Cxbx-Reloaded does not currently emulate the processor, it runs Xbox code native on the host.

slx7R4GDZM:
I know it's hiding the main issues, I just figured it'd be the quickest way to bandage the problem. Somebody with a 240hz monitor would still have a very hard time figuring out what's going on in a 30fps game even with vsync.

PR #971, opened 2018-03-06
Force VSync on by default

DiscoStarslayer:
As discussed in #966, this PR enables V-sync by default when there is no related registry entry, also did confirm LLE options are disabled with a clean registry


PatrickvL:
Thank you!

PR #973, opened 2018-03-06
Prevent frame-rate dependant games from running too fast

LukeUsher:
This fixes the issue where titles that run too fast because of differences between host and xbox vblank.
This works by timing the frames duration, and if it took <= xbox frame time, it waits until the right amount of time has passed.

- VSync is no longer the default (Sorry @DiscoStarslayer, but it's no longer a requirement)
- Introduced a hack in the Hacks menu, to enable the frame-rate to be uncapped. This restores previous behavior.
- VSync option still exists, and can be used to prevent tearing, but it is no longer the recommended option for capping the frame-rate.


src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -3982,37 +3982,48 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_Present)
...
 	DEBUG_D3DRESULT(hRet, "g_pD3DDevice8->Present");
 
+
+	if (!g_UncapFramerate) {
+		// If the last frame completed faster than the Xbox VBlank period, wait for it
+		// TODO: Read the frame rate target from the Xbox display mode
+		// See comments in GetNextVblankTime();
+		auto targetDuration = 16.6666666667ms;
+		while (std::chrono::high_resolution_clock::now() - frameStartTime < targetDuration) {
+			// We use an empty while loop because actually sleeping is too unstable
```
slx7R4GDZM: You could sleep with more accuracy if you used timeBeginPeriod to increase the Windows timer resolution. Not sure if that'd be bad because of the platform specificity. Another idea is to have an option to sleep half the required time and then busy wait the second half.

LukeUsher: Increasing the windows timer resolution is a very bad idea because it effects all processes running on the same system, not only the current application. Additionally, sleeping for half the time is an option, but there's still no guarantee it won't wake up too late. All sleep instructions guarantee a minimum time of Xms, there's no upper limit.

PatrickvL: `timeBeginPeriod` is bad - see https://randomascii.wordpress.com/2013/07/08/windows-timer-resolution-megawatts-wasted/

PatrickvL: But if the host is really fast, you do have a point that busy-looping isn't very efficient to get accurate timing - for some tips about doing that more efficiently, see http://blog.nuclex-games.com/2012/04/perfectly-accurate-game-timing/

DiscoStarslayer: @PatrickvL your first link sent over mentions how `timeBeginPeriod` is abused, but doesn't conclude that using it is bad all around. It even recommends using it for a more accurate sleep in games to limit a framerate instead of busy locking the CPU:

> For instance, some games have a power saving mode that throttles the game to 30 fps and this can only be done accurately if Sleep(1) returns in one millisecond, rather than 15-16 milliseconds. By enabling a lower frame rate without requiring busy waiting the higher timer frequency actually saves energy, in this case. For details on Sleep(n) and timer frequency read Sleep Variation Investigated.

> Raising the timer frequency isn’t (despite everything I’ve said) universally bad. It can be necessary. Many games (including those that I work on) raise the timer frequency in order to allow high frame rates (100+ fps). Having a high timer frequency means we can call Sleep(1) while waiting for the next frame, which means that we save power compared to busy waiting for the next frame! My complaint is with programs that raise the timer frequency and then leave it raised for days at a time, even when the program is just sitting idle in the background. That is what bothers me.

I feel like as long as we only raise this frequency during gameplay we should be ok. Also mentioned in the article is that the tick-less kernel in windows 8+ reduces the effect to just processes running on the same core as the process increasing the timer resolution.

Not saying this should be the solution, but I don't think we should immediately dismiss it

PatrickvL: Yeah, agreed. Would you detail this in an issue, please?

PR #979, opened 2018-03-09
Pixel shader updates

PatrickvL:
This removes `ifdef`'s from the pixel shader conversion code, allowing this code to be expanded to higher shader models later on.

It also contains some fixes that make the code somewhat cleaner and well-behaving.
(This code was originally a translation back from Delphi, which resulted in sub-par code quality. It's still not great, but ah well.)

Note, that right now pixel shader model 1.3 is still used, as 1.4 introduces two stages with a few limitations on the first stage which make it less attractive to use.

A better option might be to translate to HLSL instead of pixel shader assembly, but that would mean a rewrite of this converter, and wouldn't bring much benefit until we've ported the entire HLE D3D code to DirectX 9.

### DIFF
--- a/src/CxbxKrnl/EmuD3D8/PixelShader.cpp
+++ b/src/CxbxKrnl/EmuD3D8/PixelShader.cpp
@@ -2321,6 +2321,21 @@ std::string PSH_XBOX_SHADER::DecodedToString(XTL::X_D3DPIXELSHADERDEF *pPSDef)
   return Result;
 }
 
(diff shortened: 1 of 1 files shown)


ObiKKa:
> A better option might be to translate to HLSL instead of pixel shader assembly, but that would mean a rewrite of this converter, and wouldn't bring much benefit until we've ported the entire HLE D3D code to DirectX 9.

Do what you and Luke think are the best to do. We'll wait till then.

PR #983, opened 2018-03-12
IVB : Renames and refactorings, copied over from WIP_LessVertexPatching.

PatrickvL:
This stops the inline vertex buffer from overflowing, and improves the support for flexible vertex formats.

As always, this should be well tested before merging it in!

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -3551,18 +3551,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_Begin)
 
 	LOG_FUNC_ONE_ARG(PrimitiveType);
 
-    g_IVBPrimitiveType = PrimitiveType;
-
-    g_IVBTblOffs = 0;
-    g_IVBFVF = 0;
-
-    // default values
-    ZeroMemory(g_IVBTable, sizeof(XTL::_D3DIVB)*IVB_TABLE_SIZE);
-
-    if(g_pIVBVertexBuffer == nullptr)
-    {
-        g_pIVBVertexBuffer = (DWORD*)g_VMManager.Allocate(IVB_BUFFER_SIZE);
-    }
+    g_InlineVertexBuffer_PrimitiveType = PrimitiveType;
+    g_InlineVertexBuffer_TableOffset = 0;
+    g_InlineVertexBuffer_FVF = 0;
 }
 
 // ******************************************************************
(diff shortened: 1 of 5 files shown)


LukeUsher:
Thanks, I'll give this a test when I get back from work

LukeUsher:
Sorry, but this causes a regression in Phantasy Star Online: Splash screens/menus/gameplay all render as just a black screen with this PR

PatrickvL:
Thanks for testing - reviewing my own code again, I think one mistake is these assignments `g_InlineVertexBuffer_FVF = D3DFVF_`* (within if blocks) must use an or assignment : `|=`

LukeUsher:
PSO regresses because DrawContext.uiHostVertexStreamZeroStride and DrawContext.phostVertexStreamZeroData are never set. They remain 0.

As a hack, I replaced all uses of pHostVertexStreamZeroData/uiHostVertexStreamZeroStride with their Xbox counterparts, and rendering was restored, so that's where the problem lies.

NNX3x:
This version appears to break Terminator Dawn of Fate, which just got in menu using version 93cf2ab7. A number of other games, including Terminator, now throw this error:

Received Exception Code 0xC0000005 @ EIP := 0x0405C777
I can attach any logs if need be.

PatrickvL:
@NNX3x there's a mistake in here that I still need to correct so please wait for that first

PatrickvL:
@LukeUsher I've copied your remark over to my draw primitives branch, to which it applies

original-username1:
Shadows now work in JSRF, but are white:
shadows

Also something strange going on with their opacity:
shadows2

PatrickvL:
Hm. I haven't forwarded the fog component yet, so that might be it. But it might as well be due to incomplete render state transfers

PatrickvL:
@NNX3x the mistake I mentioned is fixed https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/934a854b07dcc6f6fedaa09c4f015c0041f9ae4b

original-username1:
After the rebase shadows are now very faint and slightly blue/purple:

shadows3

LukeUsher:
I'm happy to merge.
There are still some rendering issues with Shadows in JSRF, but they are better than they were before, so it's a step in the right direction.

NNX3x:
@PatrickvL  Thanks! I was guessing that the reason it failed was that mistake. Still learning I guess haha!
Games are now back to normal. Excellent work!

original-username1:
I see this has already been merged, but upon further testing it seems random from session to session as to whether I see white shadows, faint purple shadows, or no shadows at all.

PR #984, opened 2018-03-12
D3D : Added YUY2 and UYVY conversions, derived from libyuv. 

PatrickvL:
Also, added a fallback to ARGB when a first CreateTexture attempt doesn't work.

This might help user with nvidia cards that fail YUY2 and/or UYVY textures.
Users that didn't have problems with YUY2 and UYVY textures aren't affected


hotzenplotz3000:
This PR fixes crashes on FMVs for Mechassault, Die Hard, Hitman Contracts, Dynasty Warrior 3 and probably many more 👍

PR #992, opened 2018-03-15
Various tweaks and fixes

LukeUsher:
- Simplify deriving of D3DRS_CULLMODE, Fixes titles made with early XDK versions (greater than 3911 but less than 4361)
- Use unordered_map instead of map where we don't need it to be ordered (slight speed boost)
- Implement DirectSoundGetSampleTime
- Remove the old Vertex Buffer cache
- Cache Vertex buffer objects (but not their content): Avoids unnecessary allocations
- Use AGP Shared memory for Vertex Buffers (faster data uploads!)
- Enable High Performance Graphics by Default where possible.

A new vertex buffer cache may be added in the future, but that's outside the scope of this work.

### DIFF
--- a/build/win32/Cxbx.vcxproj
+++ b/build/win32/Cxbx.vcxproj
@@ -654,6 +654,7 @@
     <ClCompile Include="..\..\src\devices\video\nv2a_vsh.cpp" />
     <ClCompile Include="..\..\src\devices\video\swizzle.cpp" />
     <ClCompile Include="..\..\src\devices\Xbox.cpp" />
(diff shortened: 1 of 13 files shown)


JohnGodgames:
Master against https://ci.appveyor.com/project/SoullessSentinel/cxbx-reloaded/build/74e94eaed2f6861b12307f609b88b424f97efd89


## Digimon Rumble Arena 2 (regression)

**Master:** 
digi 2 master

**Various tweaks and fixes:**
digi 2 tweaks


## Gauntlet: Dark Legacy (slightly speed gain)

**Master:** 
gauntlet master


**Various tweaks and fixes:**
gauntel tweaks

PR #994, opened 2018-03-16
D3DTSS flags fix

jarupxx:


jarupxx:
I will be fix and rebased. apply to known XDK below 3948. Thank you for detailed explanation.

PR #995, opened 2018-03-17
Shikigami no Shiro Tweaks

jarupxx:
Fix no boot in Shikigami no Shiro https://github.com/Cxbx-Reloaded/game-compatibility/issues/468
Title wait for more than 10 ms to cause KeWaitForMultipleObjects failed! (STATUS_INVALID_HANDLE)

Also fix freezes on release build in Shikigami no Shiro Evolution https://github.com/Cxbx-Reloaded/game-compatibility/issues/473

### DIFF
--- a/src/CxbxKrnl/EmuKrnlPs.cpp
+++ b/src/CxbxKrnl/EmuKrnlPs.cpp
@@ -322,7 +322,8 @@ XBSYSAPI EXPORTNUM(255) xboxkrnl::NTSTATUS NTAPI xboxkrnl::PsCreateSystemThreadE
 		// Note : DO NOT use iPCSTProxyParam anymore, since ownership is transferred to the proxy (which frees it too)
 
 		// Give the thread chance to start
-		Sleep(100);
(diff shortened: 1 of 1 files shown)


Sarkie:
Are you sure this doesn't break other games?

Always wary of sleeps.

LukeUsher:
Yeah, this is just a hack. The real solution would be to properly fix kernel event handling: this is something Patrick and I are working on, but it requires a large amount of rework on many kernel APIs.

jarupxx:
Ah ok. Thank you. Although it looks a bit better than the sleep(100). This can be closed then if necessary.

@Sarkie I think that work without sleeps other games. So I changed sleep value to maximum allowable value.

PatrickvL:
The effect of the sleep duration very much depends on what's happens in the mean time, timing differences per title, performance of the host, etc. So, as Luke said, instead of sleeping for any amount of time should be replaced by better kernel implementation. Sorry jarupxx, but tweaking this work-around (the existence of the sleep) won't benefit emulation much, so I'll close it for now. Hopefully you'll find other ways to improve things?!?

Issue #913, opened 2018-02-10
Underflow in physical memory usage counter


x1nixmzeng:
After running latest, the function `PhysicalMemory::AllocatePhysicalMemoryRange` is asserting at:

> assert(m_MaxContiguousAddress >= m_PhysicalMemoryInUse);

Where it looks like the physical memory in use track has underflowed:

`m_MaxContiguousAddress == 0x07fd0000`
`m_PhysicalMemoryInUse == 0xffff3000`

Note the protection flags are 0x404 which means PAGE_WRITECOMBINE is requested (and there is already a comment that this isn't implemented) but this shouldn't be an issue.

Tested with WormsX (WormS_beta2.rar) from http://www.xbdev.net/tuts/tuts.php

x1nixmzeng:
Turns out removing the asserts will run the XBE without noticable issue (with Worms anyway).

Maybe the asserts could be moved to a macro and enabled in certain builds?

PatrickvL:
@Ergo720 is currently reworking page tables, let's wait if that works (in)directly fixes this. Otherwise, sure

ergo720:
I'm not sure but could this be a duplicate of #831?

PatrickvL:
This out of memory issue is the fallout from leaks in various places, but one thing is special about this issue : Underflow in the memory counters. That shouldn't ever happen, and is worth this separate issue. (I'll rename it to make this more clear)

ergo720:
This one is fixed by #1002 on my end, so it should probably be fine to close this now

x1nixmzeng:
Good call!, thanks @ergo720 :+1:

PR #1002, opened 2018-03-20
Page tables, pte, pde, pfn database + VMManager reworking

ergo720:
This PR implements real page table support to cxbx and the pfn database + it updates the existing VMManager implementation to make use of these new features. I took also the time to implement the remaining kernel debug Mm functions to support devkits. From my tests, this doesn’t seem to cause any regressions, but I couldn’t test a couple of functions because I didn’t find any games that call them (MmIsAddressValid, MmMapIoSpace, MmQueryAddressProtect, MmQueryAllocationSize (but QuerySize is called by cxbx callers) and MmUnmapIoSpace). Also, issue #913 is gone now, Worms doesn’t underflow any longer. It was probably caused by the previous implementation of NtAllocateVirtualMemory, which was tracking the memory incorrectly.


hotzenplotz3000:
@ergo720 I found SOULCALIBUR2 is throwing this warning in the Log:
```
WARN: PMEMOut of physical memory!
```
after calling KRNL: MmAllocateContiguousMemoryEx
which is not present in master. It still crashes like in the master at the end

**Debug Log**
[KrnlDebug_debug_mm.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1830930/KrnlDebug_debug_mm.txt)

ergo720:
From the log you can see that the first time that happens is when the title requested 0x04000000 = 64 MiB of memory, which is the entire Xbox memory so of course it will fail. From there on, the title requests progressively less memory until it succeeds and then it requests even smaller amounts. It's like the game is doing a scanning to figure out the max amount of contiguous memory it can allocate in a single chunk. This is normal behavior anyway, I wouldn't worry about it

src/CxbxKrnl/Emu.cpp
```diff
@@ -110,7 +110,7 @@ std::string EIPToString(xbaddr EIP)
 {
 	char buffer[256];
 	
-	if (EIP < XBOX_MEMORY_SIZE) {
+	if (EIP < (g_bIsRetail ? XBOX_MEMORY_SIZE : CHIHIRO_MEMORY_SIZE)) {
```
x1nixmzeng: How should this handle the debug flags?

x1nixmzeng:
There is a lot of conditional logic which does something like this:

 `g_bIsRetail ? XBOX : CHIHIRO`

Why not use the `g_bIsChihiro` flag? Could the memory manager have a method to query the "current" system memory limits?

src/CxbxKrnl/Emu.cpp
PatrickvL: It might be beneficial to replace all comparable constructs (which all determine the amount of physical memory to emulate) with one global variable for just that purpose.

ergo720: Only one of g_IsRetail, g_IsDebug and g_IsChihiro can be set. This means that if g_IsRetail is not set, then the running xbe must be chihiro or debug, and both system have 128 MiB (CHIHIRO_MEMORY_SIZE) so you don't need to check for g_IsDebug

ergo720: That was my first approach, but I thought it was bad to have another global just for this, so I dropped it, but I can reinstate it if you want to...

RadWolfie: IMHO, retail, debug, and chihiro for conditional check to pick a memory size shouldn't be happening in different places. Instead, memory size should be set as global variable and only do perform comparison operator. This way, less CPU cycle consumed and more direct optimization as possible. Like what @PatrickvL said recently.

PatrickvL: It's not extremely important, so you don't have to spend another chance on it now per se. It can always be done later. What's most important now is to get this tested and accepted for merge

PatrickvL:
For the sake of accuracy, let's keep this PR unmodified for now, and think about this some more later.

Testers, please give us your feedback on this!

ergo720:
I've found a bug that slipped: a chihiro xbe which persists a launch page and reboots to the dashboard will trigger an assert. I'm pretty sure this is because the address of launch page is in the upper 64 MiB but the dashboard is a retail xbe, which means that the launch page address will be out of range after reboot, hence the assert. If this is it then it should be an easy fix. Sorry @PatrickvL but I will have to change this PR! BTW, is it normal that a chihiro xbe reboots to the dashboard? I don't know if chihiro units have standard MS dashboards on them...

RadWolfie:
I think I had seen error codes on arcade in the past (many years ago), one of them was Ghost Squad.

PatrickvL:
According to https://github.com/kexplo/imame4all/blob/master/Reloaded/trunk/src/mame/drivers/chihiro.c, [Chihiro] "contains a media board with a TSOP48 where there is an xbox .xbe loader (this is the dashboard you see when you power the Chihiro)", so yes, it seems it has a dashboard too (although different than on retail Xboxes).

But, if this is the only issue or not easy to fix, then I'd vote to merge this already, and fix it later. I'll let @Echelon9 decide.

For now, testers haven't reported regressions

ergo720:
Well, it turned out to be not that easy indeed, the different memory layouts cause quite a few issues when restoring the memory after a quick reboot. The real problem is that this issue shouldn't happen in the first place. A real kernel will only use a single memory layout, whereas ours changes depending on the xbe type that has been detected at startup. I think the easiest solution is to select a layout when a new xbe is first launched and keep that during quick reboots even if the xbe type changes, and only change it again if the emulation is stopped and another different xbe is launched

ergo720:
Yeah, @PatrickvL , if your fine with this small bug for now, you can merge this.

PatrickvL:
Thank you very much for this work, I'm very glad you succeeded!

ergo720:
Just an update on the testing of this, Avalaunch calls MmMapIoSpace with address 0xFFF00078 (Flash ROM) and MmUnmapIoSpace on the same address, so these two in the simplest case work good. Also, nexgen dashboard before crashing with RtlAssert (debug-xdk) calls MmQueryAllocationSize and MmQueryAddressProtect, which are executed correctly as well.

PR #1003, opened 2018-03-23
Support for controller vibration

x1nixmzeng:
Requires "Use XInput" to be checked from the settings menu

I've also removed logic which updates `g_pXInputSetStateStatus`, but any info on why it needs to poll there is welcomed!

### DIFF
--- a/src/CxbxKrnl/EmuXInput.cpp
+++ b/src/CxbxKrnl/EmuXInput.cpp
@@ -49,7 +49,6 @@
 // * Static Variable(s)
 // ******************************************************************
 static XINPUT_STATE		g_Controller;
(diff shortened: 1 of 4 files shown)


PatrickvL:
Nice work! Please ask for some tests on Discord or gitter. IMHO merging depends on that feedback.

RadWolfie:
Works with Turok Evolution title for each damage taken, hitting the ground with axe, and slight vibration for shooting an arrow.

Voxel9:
Glad to see somebody got this done! :)

RadWolfie:
@x1nixmzeng I agree. Since hHandle is optional variable given to xbox, some titles may rely on this to be complete (via SetEvent). Otherwise it will hang.

PR #1004, opened 2018-03-23
Warn the user instead of hard crashing if rebooting to a different xbe

ergo720:
This prevents rebooting to a non-matching xbe type since the VMManager can't handle that case. I think the real kernel can't handle this case as well but, if I'm wrong and it can, then we should create an issue for this. As a bonus, I also replaced the conditional logic discussed in PR #1002


PatrickvL:
Thanks, also for the cleanup contained within (although it polluted the core fix a bit, so separate commits would have been better - ah well)

PR #1006, opened 2018-03-27
Nt memory functions

ergo720:
This extends the VMManager to support the kernel Nt memory functions (NtAllocateVirtualMemory, NtFreeVirtualMemory, NtProtectVirtualMemory, NtQueryVirtualMemory) and gets rid of the Ntdll calls previously present. One thing: I couldn't find any games/homebrews which call the last two or make use of the ZeroBits/MEM_TOP_DOWN parameters of NtAlloc, so testing for these cases is welcomed.

### DIFF
--- a/src/CxbxKrnl/EmuKrnlNt.cpp
+++ b/src/CxbxKrnl/EmuKrnlNt.cpp
@@ -1357,16 +1357,31 @@ XBSYSAPI EXPORTNUM(217) xboxkrnl::NTSTATUS NTAPI xboxkrnl::NtQueryVirtualMemory
 	LOG_FUNC_BEGIN
 		LOG_FUNC_ARG(BaseAddress)
 		LOG_FUNC_ARG_OUT(Buffer)
-		LOG_FUNC_END;
+	LOG_FUNC_END;
 
-	NTSTATUS ret = NtDll::NtQueryVirtualMemory(
-		/*ProcessHandle=*/g_CurrentProcessHandle,
-		BaseAddress,
-		(NtDll::MEMORY_INFORMATION_CLASS)NtDll::MemoryBasicInformation,
-		(NtDll::PMEMORY_BASIC_INFORMATION)Buffer,
-		/*Length=*/sizeof(MEMORY_BASIC_INFORMATION),
-		/*ResultLength=*/nullptr);
+	#ifdef _DEBUG_TRACE
+		if (!Buffer)
+		{
+			DbgPrintf("KNRL: NtQueryVirtualMemory : PMEMORY_BASIC_INFORMATION Buffer is nullptr!\n");
+			LOG_IGNORED();
+			RETURN(STATUS_INVALID_PARAMETER);
+		}
+	#endif
 
+	NTSTATUS ret = g_VMManager.VirtualMemoryStatistics((VAddr)BaseAddress, Buffer);
+
+	if (ret == STATUS_SUCCESS)
+	{
+		DbgPrintf("   Buffer->AllocationBase    = 0x%.08X\n", Buffer->AllocationBase);
+		DbgPrintf("   Buffer->AllocationProtect = 0x%.08X\n", Buffer->AllocationProtect);
+		DbgPrintf("   Buffer->BaseAddress       = 0x%.08X\n", Buffer->BaseAddress);
+		DbgPrintf("   Buffer->RegionSize        = 0x%.08X\n", Buffer->RegionSize);
+		DbgPrintf("   Buffer->State             = 0x%.08X\n", Buffer->State);
+		DbgPrintf("   Buffer->Protect           = 0x%.08X\n", Buffer->Protect);
+		DbgPrintf("   Buffer->Type              = 0x%.08X\n", Buffer->Type);
+	}
+
+	#if 0
 	if (FAILED(ret)) {
 		EmuWarning("NtQueryVirtualMemory failed (%s)!", NtStatusToString(ret));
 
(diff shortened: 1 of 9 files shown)


src/CxbxKrnl/EmuKrnlNt.cpp
```diff
@@ -1354,16 +1357,31 @@ XBSYSAPI EXPORTNUM(217) xboxkrnl::NTSTATUS NTAPI xboxkrnl::NtQueryVirtualMemory
...
+	LOG_FUNC_END;
 
-	NTSTATUS ret = NtDll::NtQueryVirtualMemory(
-		/*ProcessHandle=*/g_CurrentProcessHandle,
-		BaseAddress,
-		(NtDll::MEMORY_INFORMATION_CLASS)NtDll::MemoryBasicInformation,
-		(NtDll::PMEMORY_BASIC_INFORMATION)Buffer,
-		/*Length=*/sizeof(MEMORY_BASIC_INFORMATION),
-		/*ResultLength=*/nullptr);
+	#ifdef _DEBUG_TRACE
```
x1nixmzeng: Any reason this is behind a pragma? If the buffer is ever null this looks useful!

src/CxbxKrnl/EmuKrnlNt.cpp
```diff
@@ -1354,16 +1357,31 @@ XBSYSAPI EXPORTNUM(217) xboxkrnl::NTSTATUS NTAPI xboxkrnl::NtQueryVirtualMemory
...
+		DbgPrintf("   Buffer->AllocationBase    = 0x%.08X\n", Buffer->AllocationBase);
+		DbgPrintf("   Buffer->AllocationProtect = 0x%.08X\n", Buffer->AllocationProtect);
+		DbgPrintf("   Buffer->BaseAddress       = 0x%.08X\n", Buffer->BaseAddress);
+		DbgPrintf("   Buffer->RegionSize        = 0x%.08X\n", Buffer->RegionSize);
+		DbgPrintf("   Buffer->State             = 0x%.08X\n", Buffer->State);
+		DbgPrintf("   Buffer->Protect           = 0x%.08X\n", Buffer->Protect);
+		DbgPrintf("   Buffer->Type              = 0x%.08X\n", Buffer->Type);
+	}
 
+	#if 0
```
x1nixmzeng: Have you checked this with Forza Montorsport?

src/CxbxKrnl/EmuKrnlRtl.cpp
```diff
@@ -55,6 +56,7 @@ namespace NtDll
 
 #include "CxbxKrnl.h" // For CxbxKrnlCleanup()
 #include "Emu.h" // For EmuWarning()
+#include <assert.h>
```
x1nixmzeng: Cxbx really needs assert macros. They don't play nicely with the debugger!

src/CxbxKrnl/VMManager.cpp
```diff
@@ -191,7 +185,7 @@ void VMManager::DestroyMemoryRegions()
 	{
 		for (auto it = m_MemoryRegionArray[i].RegionMap.begin(); it != m_MemoryRegionArray[i].RegionMap.end(); ++it)
```
x1nixmzeng: Unrelated, but modern iteration syntax is a bit neater - `for(auto it : my->container) { }`

src/CxbxKrnl/VMManager.h
```diff
@@ -156,10 +158,13 @@ class VMManager : public PhysicalMemory
...
 		// xbox implementation of NtAllocateVirtualMemory
-		xboxkrnl::NTSTATUS XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBits, size_t* Size, DWORD AllocationType,
-			DWORD Protect);
+		xboxkrnl::NTSTATUS XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBits, size_t* Size, DWORD AllocationType, DWORD Protect);
 		// xbox implementation of NtFreeVirtualMemory
 		xboxkrnl::NTSTATUS XbFreeVirtualMemory(VAddr* addr, size_t* Size, DWORD FreeType);
+		// xbox implementation of NtProtectVirtualMemory
+		xboxkrnl::NTSTATUS XbVirtualProtect(VAddr* addr, size_t* Size, DWORD* Protect);
+		// xbox implementation of NtQueryVirtualMemory
+		xboxkrnl::NTSTATUS VirtualMemoryStatistics(VAddr addr, xboxkrnl::PMEMORY_BASIC_INFORMATION memory_statistics);
```
x1nixmzeng: Nitpick, but maybe keep either the Nt naming convention (ie `QueryVirtualStatistics`) or the xbox implementation (ie `XbVirtualMemortStatistics`)

x1nixmzeng:
Done an quick pass and it looks good.

A bit worried by the use of `goto` which I know has been discussed before! Ideally you would create a "scope class" which has logic in the constructor/destructor which may be a bit safer for the lock/unlock - but it looks like you caught them all.

How can this be tested?

src/CxbxKrnl/VMManager.h
ObiKKa: Just an accidental typo ('t' should be 'y') in XbVirtualMemortStatistics.
XbVirtualMemoryStatistics.

src/CxbxKrnl/EmuKrnlNt.cpp
ergo720: I can't, I don't have it. If someone has it they could run it on a debug build to check what the function is returning

PatrickvL: Is this now addressed by the Small update? Anything else that needs to be done?

ergo720: Well, I think that if there are any errors in those two untested functions then they will show up in the compatibility reports at some point so, if you are fine with this, you can do the merging

PatrickvL: Will do!

PatrickvL:
Excellent work, again! Thank you very much

PR #1007, opened 2018-03-27
Kernel thunks

ergo720:
This PR adds the remaining kernel thunks and fixes issue #64. These are only stubs though, all these functions still need an implementation.


PatrickvL:
Nice work! Alas it's based on your other pr instead of master

PatrickvL:
Maybe cherry pick it into a separate branch off upstream master, and PR that?

ergo720:
Oops, I created the kernel thunk branch by cloning it from the Nt branch and so it has brought the other un-merged commits with it. I'll see what I can do about it.

ergo720:
BTW, there is still one PANIC left, and it's the function/variable at offset 0, which is "undefined" so, afaik, that should be correct already

PatrickvL:
True. So it can just be replaced by null and then the PANIC define can be removed

PR #1016, opened 2018-03-31
Eeprom improvements

ergo720:
This improves issue #198 (it 's just missing the gui now). From my tests: this seems to work fine however, more testing is welcome. One thing: with the new hash that is calculated, if that fails an error is logged and the LED flashes. As an alternative to this, we could just delete the existing eeprom.bin and regenerate a new one, so let me know what you think.

### DIFF
--- a/import/OpenXDK/include/xboxkrnl/xboxkrnl.h
+++ b/import/OpenXDK/include/xboxkrnl/xboxkrnl.h
@@ -2106,6 +2106,14 @@ typedef enum _XC_VALUE_INDEX
 XC_VALUE_INDEX, *PXC_VALUE_INDEX;
 
 // ******************************************************************
+// * EEPROM game region codes
+// ******************************************************************
+#define XC_GAME_REGION_NA             0x00000001
+#define XC_GAME_REGION_JAPAN          0x00000002
+#define XC_GAME_REGION_RESTOFWORLD    0x00000004
+#define XC_GAME_REGION_MANUFACTURING  0x80000000
+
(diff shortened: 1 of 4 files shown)


jarupxx:
Great work!
Now, I specify XC_GAME_REGION_JAPAN for Game Region, title is displayed in Japanese.

Master
pd-master
Eeprom improvements
pd-eeprom

Test case: Panzer Dragoon Orta, Quantum Redshift

ergo720:
That's great, but how did you change the game region code, there's no option for doing it now afaik. I supposed you used a hex editor on the eeprom.bin file. However, the game region code is used to calculate the hash so, unless you changed it when it was still all zeros or you recalculated the hash yourself, if the LED didn't flash something is wrong. Another thing: I actually noticed that, before doing my changes, my eeprom has become almost all 1's for some reason. This is also evident by my screenshot in #895, where the first 14 bytes are all 1. This means that something is silently corrupting the eeprom, but I couldn't track down what that is, and so I expect that at some point we see the LED flashing, unless the current changes also fixed that problem.

PatrickvL:
Dashboard?

PatrickvL:
Regarding memory corruption: Cxbx-Reloaded's memory model allows memory corruption to happen from both game code and emulator code. Preferably, our emulator should behave nicely, but until we're sure of that, one safeguards could be to always keep a "last known good" copy of the EEPROM around, and restore that if the EEPROM every gets corrupted. (A popup could ask to report the title that caused the corruption)

Issue #974, opened 2018-03-07
Add 'boot to dashboard' option


CakeLancelot:
Add an option to the menu bar that will allow the user to boot an installed dashboard without having to navigate to the Appdata folder and open it from there.

fancymelonub:
How do you install a dashboard?

x1nixmzeng:
The dashboard is itself an XBE that would need to be run through the emulator. Cxbx would need to be pretty stable to support booting straight into anything.

At the moment can you use the F5 key to run the last XBE you started - from the "most recently used" list?

CakeLancelot:
> The dashboard is itself an XBE that would need to be run through the emulator. Cxbx would need to be pretty stable to support booting straight into anything.
> 
> At the moment can you use the F5 key to run the last XBE you started - from the "most recently used" list?

This is effectively a menu option that does just that - launches the dashboard XBE if one is found in the Emudisk.

PatrickvL:
This is implemented by @ergo720 via pull request #1017

PR #1017, opened 2018-04-01
Load dashboard option

ergo720:
This is intended to fix issue #974. It adds a new "Open dashboard" option that searches for the MS dashboard in the app folder and loads the xbe if found. Afterwards, emulation can be started with "Start" as usual.


LukeUsher:
This PR completely broke the Recent Xbe functionality. (Because it shifted the menu indexes)
I fixed this as part of #1018, but just thought I'd make a note here.

ergo720:
Oops, sorry for it, I didn't notice that...

PR #1018, opened 2018-04-01
Introduce functional NV2A LLE

LukeUsher:
As the title suggests, this implements *working* NV2A LLE (and some kernel tweaks to make it work) HLE is still the recommended option for running software for now.

Known Issues:
* Slow. (MMIO exception overhead + lack of surface cache), 2D titles such as Beats of Rage (homebrew) are reasonable, 3D titles are incredibly slow. 
* Titles that directly access the framebuffer don't render (NXDK, OpenXDK), this is because I only swapbuffers in the NV2A Flip event, easily solved in the future.
* FPS counter does not work in LLE. (It's implemented in the D3DDevice_Swap function and needs abstracting away to a generic function that can be called from either path)

Other than that, it should work okay.
These remaining issues will be addressed in the (near) future, but I want to deliver what I have so far.


LukeUsher:
This has no effects on HLE behavior from my testing, so is safe to merge as-is, once we confirm that it compiles correctly in AppVeyor

PR #1020, opened 2018-04-02
Fix an assert on VMManager in Fable

ergo720:
This fixes the assert in Fable in https://github.com/Cxbx-Reloaded/game-compatibility/issues/389
BTY, it will still assert shortly after but this ones comes from EmuX86. In particular, looking at the eip when the assert triggers I see the offending instruction is 0f 20 d8, which disassembles as mov    eax,cr3 

capture
capture2

### DIFF
--- a/src/CxbxKrnl/VMManager.cpp
+++ b/src/CxbxKrnl/VMManager.cpp
@@ -1660,16 +1660,15 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
 			AlignedCapturedBase = ROUND_DOWN(CapturedBase, X64KB);
 			AlignedCapturedSize = ROUND_UP_4K(CapturedSize);
 
-			// Only reserve memory if the base address is higher than XBE_IMAGE_BASE
-
-			if ((AlignedCapturedBase >= XBE_IMAGE_BASE + ROUND_UP_4K(CxbxKrnl_Xbe->m_Header.dwSizeofImage) && AlignedCapturedBase < XBE_MAX_VA) ||
-				(AlignedCapturedBase >= XBE_MAX_VA &&
-				(VAddr)VirtualAlloc((void*)AlignedCapturedBase, AlignedCapturedSize, MEM_RESERVE, PAGE_NOACCESS) != AlignedCapturedBase))
+			if ((VAddr)VirtualAlloc((void*)AlignedCapturedBase, AlignedCapturedSize, MEM_RESERVE, PAGE_NOACCESS) != AlignedCapturedBase)
 			{
-				// Something else is already mapped at the specified address, report an error and bail out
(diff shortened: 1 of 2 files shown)


src/CxbxKrnl/VMManager.cpp
```diff
@@ -1660,16 +1660,15 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
 			AlignedCapturedBase = ROUND_DOWN(CapturedBase, X64KB);
 			AlignedCapturedSize = ROUND_UP_4K(CapturedSize);
 
-			// Only reserve memory if the base address is higher than XBE_IMAGE_BASE
-
-			if ((AlignedCapturedBase >= XBE_IMAGE_BASE + ROUND_UP_4K(CxbxKrnl_Xbe->m_Header.dwSizeofImage) && AlignedCapturedBase < XBE_MAX_VA) ||
-				(AlignedCapturedBase >= XBE_MAX_VA &&
-				(VAddr)VirtualAlloc((void*)AlignedCapturedBase, AlignedCapturedSize, MEM_RESERVE, PAGE_NOACCESS) != AlignedCapturedBase))
+			if ((VAddr)VirtualAlloc((void*)AlignedCapturedBase, AlignedCapturedSize, MEM_RESERVE, PAGE_NOACCESS) != AlignedCapturedBase)
```
PatrickvL: Wouldn't this leak?

src/CxbxKrnl/VMManager.cpp
```diff
@@ -1660,16 +1660,15 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
...
 			{
-				// Something else is already mapped at the specified address, report an error and bail out
+				// VirtualAlloc failed, check to see if we are reserving the memory for the xbe sections
```
PatrickvL: It failed if it returned null, but any other address is still allocated and if not stored and used after this, it would be a leak, right?

ergo720: If lpAddress !=0 it means I'm requesting to specifically allocate/reserve the memory at the specified address suggested and nowhere else (for that you pass 0 as suggested address). If it fails it means it couldn't allocate/reserve the memory at lpAddress and so nothing else has been allocated/reserved

src/CxbxKrnl/VMManager.cpp
ergo720: If this call succeeds than the resulting address must AlignedCapturedBase and nothing else, since that is what I've requested to the function. Its value will be stored in the vma by ConstructVMA so we won't lose it

src/CxbxKrnl/VMManager.cpp
PatrickvL: Ah, okay, if it won't return anything else but the requested address, then there's no leak. Thanks for clarifying

PR #1023, opened 2018-04-02
Introduced D3D8LTCG database

jarupxx:
Now, lots of D3D8 patches have been unused. This database will be useful for the remaining PATCH. until improve GPU LLE over time.


RadWolfie:
Hello @jarupxx We had some discussion on discord about HLE database. We're planning to separate HLE Database which will become as "symbol table" library to allow other xbox emulator/tools use it in their own application for research purpose. All the patches will be move into its own patch table internally.

LukeUsher:
Amazing that this even works at all since LTCG can inline functions, change calling conversions etc as well as just change signatures... (I saw the screenshots on Discord, so I know it does at least somewhat work)

How do you plan to handle these cases? We may even need per-game patches for functions where calling conventions changed (hopefullly, none did :P)

LukeUsher:
Let's see how this goes...

jarupxx:
I have no plan for those functions... And who cares them ;)

PR #1025, opened 2018-04-03
Use real framebuffer size (NV2A LLE)

LukeUsher:
This reads the framebuffer size from the current display mode when blitting the NV2A framebuffer to the window.

Additionally, it prepares for YUV overlay support, which is currently not implemented in LLE.


src/devices/video/nv2a.cpp
```diff
@@ -526,27 +388,30 @@ void NV2ADevice::SwapBuffers(NV2AState *d)
...
 	static GLenum format = GL_RGBA;
 	static GLenum type = GL_UNSIGNED_INT_8_8_8_8;
+	static int framebufferWidth = 640;
+	static int framebufferHeight = 480;
 
 	static GLuint texture = -1;
 
 	// Convert AV Format to OpenGl format details & destroy the texture if format changed..
-	// This is requried for titles that use a non ARGB framebuffer, such was Beats of Rage
+	// This is required for titles that use a non ARGB framebuffer, such was Beats of Rage
```
Margen67: *such **as**

PR #1034, opened 2018-04-04
NVIDIA Experiment: Round 2

LukeUsher:
It's hard to debug this without an NVIDIA card myself, but we'll see how this goes..


LukeUsher:
It worked!

ergo720:
@LukeUsher are you sure that it works? I tested Halo with GPU LLE and if I use the integrated Intel GPU the game boots fine, but if I use the dedicated NVIDIA GPU I just get an infinite black screen

PR #1036, opened 2018-04-04
Fix Debug Build & Nullptr Dereference

LukeUsher:


src/CxbxKrnl/EmuDSoundInline.hpp
```diff
@@ -417,6 +417,10 @@ inline void DSoundBufferTransferSettings(
     LONG lVolume, lPan;
     DS3DBUFFER ds3dBuffer;
 
+	if (pDSBufferOld == nullptr) {
```
RadWolfie: Just a note, this is hack solution. Not a real solution.

PR #1039, opened 2018-04-05
Disable LLE Placeholder Action

RadWolfie:
LLE APU is not implemented yet.
LLE JIT does nothing.

This will help users not to report problems with LLE placeholder
actions.

**ATTENTION: If you have APU or JIT left on afterward. Please select reset to default action from menu or rollback then disable it. There are absolutely no reason for these options to be toggle for time being and giving false reports.**


PatrickvL:
Thank you!

PR #1040, opened 2018-04-06
This seems to fix the .bik hang

LukeUsher:


LukeUsher:
This also fixes dirty disc errors in some titles

x1nixmzeng:
I did double-take when I read "ToUnicornSize" :laughing:

LukeUsher:
Lol, wow, that's what I get for committing at 1am ;) the codes fine though!

ObiKKa:
Is that from an Electric Dream?

PR #1046, opened 2018-04-07
Fixed RtlEqualUnicodeString

PatrickvL:


PatrickvL:
Note : This fix satisfies the test in the kernel test suite but overall it still concludes failure on test_RtlEqualUnicodeString - I'll create an issue over there for this result.

Issue #1053, opened 2018-04-09
Invalid Pixel Shader in Turok


PatrickvL:
Turok when entering in-game (when the camera pans aside the first door) generates an invalid pixel shader:

```
[0x27A0] New decoding - Parse result :
[0x27A0] xps.1.1
xmma t0,discard,discard, t0,c0,0.000000f,0.000000f ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
+xmma r0.a,discard,discard, v0.a,t0.a,0.000000f,0.000000f ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
xmma discard,discard,t3.rgb, t3.rgb,v1.a,1.000000f,1-v1.a ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
+xmma discard,discard,t3.a, t3.a,v1.a,1.000000f,1-v1.a ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
xmma r1,discard,discard, r1,t3,0.000000f,0.000000f ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
xmma r0.rgb,discard,discard, t0.rgb,1.000000f,0.000000f,0.000000f ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
+xmma r0.a,discard,discard, c0.a,r0.a,0.000000f,0.000000f ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
xfc 1.000000f,r0.rgb,fog.rgb,0.000000f,0.000000f,0.000000f,r0.a ; r0.rgb=s0*s1+{1-s0}*s2+s3, r0.a=s6.a, prod=s4*s5, sum=r0+v1 

[0x27A0] New decoding - MoveRemovableParametersRight :
[0x27A0] xps.1.1
xmma t0,discard,discard, t0,c0,0.000000f,0.000000f ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
+xmma r0.a,discard,discard, v0.a,t0.a,0.000000f,0.000000f ; d0=s0*s1, d1=s2*s3, d2={s0*s1}+{s2*s3} 
[...]

PR #1054, opened 2018-04-09
pixel shader conversion enhancement

PatrickvL:
This quick PR improves pixel shader conversion, by repeating cleanup steps (removing useless writes and no-ops) until no more modifications are done.

This can lead to smaller shaders and less conversion failures, improving visuals for some titles.

Test-case : Turok, when entering in-game, when the camera starts panning sideways while walking towards the first door. (This PR fixes issue #1053)


PatrickvL:
Self-merging this, it's rather safe

PR #1063, opened 2018-04-14
HACK: Alternate method of handling Performance Counter

LukeUsher:
As discussed on Discord: This adds an optional hack/alternative method of handling the performance counter/rdtsc issue.

This approach is the one that is usually used on 1.0Ghz upgraded original Xbox consoles, so results may very than the default in interesting ways. It could fare better in some games, worse in others.

It'd be interesting to find out


RadWolfie:
Test cases:
* Jet Set Radio Future
  * Hangs at boot.
* Red Faction 2
  * Speed is too fast. Plus long loading time to start a mission.
* Mafia
  * Hang at FMV (after starting a game)
* Turok Evolution
  * Speed is too fast. (FMV appear to work, except it is also going too fast.)

I believe a lot of titles are going to have side-effect with this hack option. Might be likely cause by missing hack patch on certain kernel functions.

AzurikRiseOfPerathia:
I also have tester on my side the game Azurik 

1. Does not crash as much as before.

2. But the speed and accelerated. I estimate 0.50X

PatrickvL:
The way this works needs a bit more documentation - perhaps in the wiki?

AzurikRiseOfPerathia:
To better specify the crash of the style the texture that breaks happens 10 to 15 minutes after activating "Alternate method of Handling Performance Counter" compared to 5 minutes usual.

PatrickvL:
@AzurikRiseOfPerathia what do you mean with "the texture that breaks"?

PR #1065, opened 2018-04-15
This is not to be merged. this is only for testing purpose to test us…

jackchentwkh:
…ing exception handler to handle the patched rdtsc instructions. for me, the release build can reach 60 fps with MechAssault. although sometimes it still crashes, but the successful rate to get in game is much higher. need more tests to validate this idea.

### DIFF
--- a/src/CxbxKrnl/CxbxKrnl.cpp
+++ b/src/CxbxKrnl/CxbxKrnl.cpp
@@ -649,11 +649,24 @@ static unsigned int WINAPI CxbxKrnlInterruptThread(PVOID param)
 	return 0;
 }
 
+DWORD rdtsc_addr[100]; //arrays for each rdtsc instruction found
+int            rdtsc_count = 0;//counts for total rdtsc instruction found
+
+bool is_rdtsc_break(DWORD addr)
+{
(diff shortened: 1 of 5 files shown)


LukeUsher:
This is actually not a bad idea... Mind if I replace my recent rdtsc hack with this approach and clean it up a little?

PatrickvL:
This is what was proposed before already Luke ;)

jackchentwkh:
Luke, sure, why not. But after testing ExaSekleton, it's really a very special case that uses 313 rdtsc instructions. I might fix it a little bit. Or you can go ahead use it first, and I shall add my fix later.
the first thing is that the array to store the rdtsc address shall be enlarged to around 400. 
next, the pattern shall add more patterns such as:
0f 31 29
0f 31 f3
0f 31 E9
0f 31 2b
0f 31 50
0f 31 0f
0f 31 8d 
0f 31 d9
0f 31 57
0f 31 b9
0f 31 85
0f 31 83
0f 31 33
0f 31 f7
and most strange one is 
0f 31 0f 31, which repeat the rdtsc instruction twice, I don't know why it's doing this. 
this shall be treated in special care.

LukeUsher:
Replaced by #1066 
Thanks!

PR #1066, opened 2018-04-15
Patch rdtsc instructions

LukeUsher:
This replaces both the QueryPerformanceCounter patch and the PatchCpuFrequency hack with a universal solution of patching rdtsc instructions, using a method devised by @jackchentwkh.

As timing is a fundamental part of emulation, this requires extensive testing, but so far, the results are rather positive, it's be interesting to see if any of the more problematic/crashy titles improve.


jackchentwkh:
Luke, you didn't check whether the opcode after the 0F 31 is valid or not. this will introduce false positive.
and after patching the 0F 31, addr shall +2 instead of +3.

jackchentwkh:
from the krnldebug log for DOA 3
INIT: Searching for rdtsc in section .text
INIT: Patching rdtsc opcode at 0x000CA02D
INIT: Searching for rdtsc in section D3D
INIT: Patching rdtsc opcode at 0x001B585D

the one this PR found at 0x000CA02D is a false positive. which shall not be patched.
and there is another 0F 31 at 1810D2, which is rdtsc, but not detected. I don't know why.

jackchentwkh:
ok, the addr shall +1 only, because the for loop itself will increment with 1, that will result in increment with 2, just enough to skip the opcode. there is a case that rdtsc is call twice, the patter is 0f 31 0f 31.

and for the false positive detection, here is my quick suggestion:
1. skip all data section/XGRPH, etc, which you already did.
2. After the 0f 31 was found, check whether the opcode after it is valid or not.
if(find 0f 31){
   if (followed opcode is valid)
       if(followed opcode.length==1){
              if(opcode==0xc3 || opcode==0x57)){  //these two single byte opcode are only positive cases.
                     patch rdtsc;
              }else{
                      log unpatched address for possible missed detection.
              }
       }else{   //for any opcode followed, if the length is greater than 1, then it's positive match.
                 patch rdtsc;
       }
       
}

this algo matches all titles I have on hand. and it can log those unpatched address info so we can do a quick examination if we suspect there is any missed patch.

LukeUsher:
There's no point adding yet another dependency for something this small while we already have a library that can do this. (Two actually, now capstone was added as a dependency for the debugger...)
And your suggested option t hasn't been updated a single time in the last 5 years ether.

Also the "out" opcode is fine, because in our exception handler, we use a lookup to check if the opcode was caused by an rdtsc patch. Our options are limited because we must use a one-or-two byte opcode which triggers an exception, but isn't a breakpoint or interrupt (as that breaks debuggers). It's true that this might slightly slow down the processing of MMIO opcodes, but the impact won't be noticeable, it's a tiny fraction compared to the overhead of the exception context-switch.

PatrickvL:
Perhaps use "IN" it's bound to be less common than "OUT"

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -684,18 +701,28 @@ void PatchRdtscInstruction()
...
-				// that it makes Cxbx-Reloaded unusable
-				// A privilaged instruction (like OUT) does not suffer from this
-				opAddr[0] = 0xEF; // OUT DX, EAX
-				opAddr[1] = 0x90; // NOP
-				g_RdtscPatches.push_back(addr);
-				addr += 2;
+				// Found potential rdtsc instruction, check for validity by disassembling the following instruction
+				_DInst info;
+				if (!EmuX86_DecodeOpcode((uint8_t*)(addr + 2), info)) {
+					return;
```
jackchentwkh: shouldn't this return be continue?

jackchentwkh:
Patrick, in my logic, the priority of flowed opcode length is 1st. if the opcode length is only 1, then we might have false positive, so we have to pick up by checking the pattern. 

Luke, when the followed opcode is invalid, the loop shall be continued instead of quit (return).

PatrickvL:
@LukeUsher if you complete the x86 switch from distorm to unicorn, then how would we determine the byte count of an opcode? This lib could do the job fine, it only costs half a kilobyte, it's fast and doesn't need us to pass in any info struct or whatever.

PatrickvL:
As for the opcode to use : whichever we choose, the exception handler (or later unicorn hook) would still need to check if it's an actual occurrence of that instruction in the Xbe, or a rdtsc replacement. Because of this check, it's be wise to select an opcode that's used as little as possible in practice. (And as long as we run it natively, it needs to generate an exception - perhaps we can even discern the exception type : illegal instruction Vs access violation). Looking at x86asm.net I noticed "INS" (http://ref.x86asm.net/coder32.html#x6C Input from Port to String) which to my inexperienced eye looks like a prime candidate for our purpose

jackchentwkh:
@LukeUsher , sorry. it seems that my logic for determining the rdtsc instruction is not robust. there are other false positives after testing with more titles. let's hold this PR till we found a robust way to identify the instructions.

PR #1071, opened 2018-04-16
Prepare d3d9 port

PatrickvL:
This prepares our code-base to make porting to Direct3D9 a little bit easier;
* a new build configuration "DEBUG_Direct3D9" enables the Direct3D9 code paths (won't compile)
* all simple changes are #ifdef'ed, adding arguments, using new symbols, etc.

Meanwhile, the Direct3D8 (over d3dto9) code path stays functionally identical to what's currently in master, so this should not change *anything* for now.

I'd still be good to merge this now rather than later, so any HLE work still underway can be done with the Direct3D9 port in mind (the new `#ifdef CXBX_USE_D3D9` 's sprinkled throughout the code will make sure of that).


x1nixmzeng:
Neat!

Try adding `  Debug_Direct3D9` to these lines to get AppVeyor to build it:

https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/e2395e6a041ddc89e4f040bdb0e4b97886023267/appveyor.yml#L14-L17

PatrickvL:
@x1nixmzeng Thanks for the suggestion, but the Debug_Direct3D9 build configuration doesn't compile, so until then I'll keep AppVeyor in the dark ;)

Issue #675, opened 2017-08-18
Add signature validation to Xbe loading


PatrickvL:
Luke proposed the following idea:
"
Someone should add signature validation to Cxbx, it's not that hard. We already have all the crypto functions required
That way we can show in log files or xbe dumped if it's not a legit xbe
All homebrew will be marked as non legit, but all commercial games will be legit, unless pirated
Also important, most homebrew shares the same title id, usually 0
And most homebrew also has the xbe signature nulled out too
But sometimes signed with a debug key, and sometimes a completely new key used specifically for homebrew
"

PatrickvL:
@ergo720 fixed this in pull request https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1076

PR #1076, opened 2018-04-18
Xbe signature validation

ergo720:
This fixes issue #675 by validating the digital signature of the xbe and printing the result of the validation in the kernel and the xbe dumps. It also performs section integrity validation when a new xbe is launched and writes all the results in the kernel dump. This will make very easy to detect pirated or otherwise tampered xbe's by simply inspecting the dumps.


ergo720:
@x1nixmzeng didn't notice that, thanks for pointing it out. Also thanks to @PatrickvL for doing it in my place

PR #1081, opened 2018-04-20
Remove legacy audio hack

Margen67:
Note: The registry key for the setting will have to be deleted manually.


PatrickvL:
Thank you

Issue #1083, opened 2018-04-21
Xbox Kernel Test Suite Crashes Immediately


Fisherman166:
Last weekend the Xbox Kernel Test Suite ran fine. This weekend, it crashes immediately with no error messages in the logfiles. I was able to git bisect the issue back to this commit.

commit e8c1c7670a403923d39434d09db530c3d68f263d
Author: @ergo720 
Date:   Wed Apr 18 19:30:47 2018 +0200

    Use real RSA public key

Attached is the copy of the test suite xbe: [default.zip](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1934879/default.zip)

ergo720:
The crash comes from the popg function, which is called during the decryption of the signature header. Not sure yet why it causes a crash, this handles fine xbe's with invalid signatures as well. Perhaps xbe's generated by nxdk are slightly different than those generated by official XDK?

![capture](https://user-images.githubusercontent.com/30000751/39088123-e1083000-45ac-11e8-8467-463f31f2b6a1.PNG)

ergo720:
Fixed it: it seems that the digital signature generated by nxdk is all zeros, I modified the xbe by filling it with all 0xAA and now it works again. You will probably have to tell to nxdk to fill the digital signature with some bytes instead of all zeros.

PatrickvL:
It might have something to do with calling conventions

Fisherman166:
Nxdk does set the signature to 0 by default in tools/cxbe/Xbe.cpp:

        // nobody has the private key yet, so zero this out
        memset(m_Header.pbDigitalSignature, 0, 256);

Without submitting a pull request to NXDK I don't think this is possible to change in user code since the Xbe variable is created locally. Is it possible to modify Cxbx-R to handle a DigitalSignatue of 0 correctly?

ergo720:
Well, detecting if the signature is all zero with memcmp and skipping the check should be enough

Luca1991:
Wow I missed this issue. @ergo720 I've recently used (even today) various version of CxBx to test my newly implemented tests in KTS, and I haven't experienced any crash. So maybe the problem experienced by @Fisherman166 isn't caused by the digital signature?

I'm asking that because in the past (can't remember when) I had to manually remove the APPDATA cxbx folder to make cxbx to work again, and I know that @x1nixmzeng (and possibly others) had this problem too.

Luca1991:
[...]

PR #1084, opened 2018-04-21
Fix nxdk built xbe crash when the signature is all zeros

ergo720:
See #1083

### DIFF
--- a/src/Common/Xbe.cpp
+++ b/src/Common/Xbe.cpp
@@ -821,6 +821,15 @@ const wchar_t *Xbe::GetUnicodeFilenameAddr()
 
 bool Xbe::CheckXbeSignature()
 {
+	// Workaround for nxdk (and possibly oxdk?): xbe's built with nxdk have the digital signature set to all zeros, which will lead
+	// to a crash during its decryption in RSAdecrypt. Detect this condition and skip the check if true
+	{
+		UCHAR Dummy[256] = { 0 };
+		if (memcmp(m_Header.pbDigitalSignature, Dummy, 256) == 0) {
+			return false;
+		}
+	}
+
 	DWORD HeaderDigestSize = m_Header.dwSizeofHeaders - (sizeof(m_Header.dwMagic) + sizeof(m_Header.pbDigitalSignature));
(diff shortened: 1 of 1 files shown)


x1nixmzeng:
I have some XBEs from SmartXX - `reboot.xbe` is one which throws errors with the "thread local storage" header, but Cxbx doesn't handle the error properly.

Also I've just spotted `delete[] m_SignatureHeader;` should be `delete m_SignatureHeader;`

ergo720:
m_SignatureHeader  is ```m_SignatureHeader = new uint08[m_Header.dwSizeofHeaders - (sizeof(m_Header.dwMagic) + sizeof(m_Header.pbDigitalSignature))];```, which is an array allocated on the heap and so, afaik, to delete it I should use ```delete[] m_SignatureHeader;``` and not a plain ```delete```

ergo720:
Unable to reproduce, reboot.xbe is detected to have an invalid signature, it reboots to the dashboard as expected and then I close it, which doesn't cause any crash on my end, in both release and debug builds.

x1nixmzeng:
Yes sorry your use of delete is correct - I was wrong about that!

The error I get from my test XBE occurs before checking the signature. Initializing in `Xbe::ConstructorInit` will fix it.

I've attached my test file here: [reboot.zip](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/1935109/reboot.zip)

ergo720:
My findings: your attached xbe indeed crashes, but only on debug builds, because m_SignatureHeader is left uninitialized: the reading of the xbe fails because it hasn't a TLS section. Looking at it with xbe explorer, I see it also lack the library version section in addition to the TLS section. The reboot.xbe that I have works correctly instead, xbe explorer confirms it has the missing sections (library and TLS). My guess is that my version was built with an XDK, whereas the one you uploaded was built with openxdk

Release
capture
Debug
capture1
Your reboot.xbe
capture2
My reboot.xbe
capture3

x1nixmzeng:
@ergo720 thanks for investigating and for adding the fix. From the screenshots, your file looks like gamepad sample renamed

PatrickvL:
Thank you!

PR #1085, opened 2018-04-22
Clean memory.bin during boot and quick reboots

ergo720:
This clears the memory.bin file during normal boots and all the non-persisted pages during a quick reboot to prevent carrying over data remnants from the previous emulation. In particular, the aim is to clear the previous framebuffer which is otherwise persisted (but I'm not sure if the framebuffer is currently allocated in the contiguous memory). Seems to work good, but more testing is welcomed


AzurikRiseOfPerathia:
Strangely since the recent update Azurik takes longer to break its textures. Good news ^^

Issue #35, opened 2016-12-16
Allow the user to configure Xbox settings, such as connected AV Pack, Region, etc


LukeUsher:
Currently all the values returned by ExQueryNonVolatileSetting are hard coded.
The connected AV Pack is also hard coded to HDTV.

The user should be able to change these settings somewhere in the Cxbx-Reloaded UI, Additionally, settings changed via the Dashboard/ExSaveNonVolatileSetting should be preserved.

PatrickvL:
This has become a possibility, now that we have an array of EEPROMInfo structs - see ExQueryNonVolatileSetting and ExSaveNonVolatileSetting

PatrickvL:
Ergo720 implemented this in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1089

PR #1089, opened 2018-04-26
Eeprom viewer

ergo720:
This adds an eeprom editor menu to Cxbx-Reloaded and fixes issues #35 and #198. One thing: I noticed that if the regions Rest of the world or Manufacturing are selected, the dashboard (4920) no longer shows the parental control settings. I looked at the eeprom.bin file together to the values written and everything looks fine to me, so I guess the dashboard doesn’t show the parental controls for those two regions?

capture


LukeUsher:
Excellent work, I'll do some testing on this later

LukeUsher:
One comment so far: You interpret the Serial Number has a HEX string, while that field is actually just a plaintext/ASCII field.

ergo720:
@CakeLancelot There is one already, at line 1114 in WndMain.cpp
@LukeUsher Well, yes, the users will have to type it using the hex ascii codes 0x30 - 0x39 (decimal numbers 0 - 9). I suppose this is acceptable since they have to do it for all the other fields anyway?

ergo720:
I would prefer to let the user decide, if they don't want them for some reason they have the choice to do it

LukeUsher:
This PR is especially good because it allows for users to manage their settings without requiring a dashboard.

Most of these options (except the factory settings) were configurable with the MS Dash, but now that's no longer required.

PatrickvL:
> I would prefer to let the user decide, if they don't want them for some reason they have the choice to do it

How about replacing Reset with : 
* Clear (fill with zeros)
* Default (set values to what's normally best for emulation)
* Save
* Load

ergo720:
Clear: I'm not sure if setting the eeprom to all-zero is a good idea, some parameters (AV region, language, game region) have to be != zero, setting them to zero is an invalid value, but I didn't try it to see what happens.
Default: if you mean setting the eeprom to a default configuration, that's what reset does, otherwise I don't know what best values for the emulation to use.
Save/Load: the eeprom file is automatically loaded when the menu is opened, and it's only saved to disk if the user hits the "Accept" button to confirm their changes (if any). If they hit "Cancel", the changes are not saved and they are lost.

PatrickvL:
Okay, no clear, and keep Reset.

But Save/Load could show a file-selection dialog, so other dumps can be used, which would be a nice addition to this already great work

ObiKKa:
Those video settings and resolution options will help with screenshotting a game at the correct resolution and with no distorted ratio to their viewport/screen size, which is very important.

Could we leave these settings with the zeroes next to them like the Serial Number? Or do we have to manually input the digits?

RadWolfie:
@PatrickvL, in your case of using different eeprom.bin dump. Save/Load isn't the best wording for it. It would be Import/Export. That's my option.

PR #1090, opened 2018-04-26
D3D work

PatrickvL:
Description will follow, but code-reading is already welcome.


jackchentwkh:
this PR regressed ExaSkeleton with error "IDirect3D:createdevice failed"

PR #1093, opened 2018-04-26
D3D : Present overlay only when last present was more than 1 frame ago

PatrickvL:
This fixes the pause of the FPS counter during ongoing overlay playback when the software doesn't present itself (like in the XDK WMVCutScene sample). Turok movies still flicker, but less so.

Again, please await feedback from testers before merging this


RadWolfie:
I don't see the difference or performance impact with other titles except for Turok FMVs do indeed has much less flickering.

PR #1100, opened 2018-04-28
Fix VS 2015 errors

### LINKED ISSUE #1088: Visual Studio 2015 build failure
Visual Studio 2015 does not compile current master. You will get:
Some (more than 5) C2872 errors.
Two abc errors - missing brackets in FUNC_EXPORTS (where is MUST be bracket - I don't know, FUNC_EXPORTS also called before and later, but VS2015 doesn't find errors there).
I already removed VS2015 from my PC, so I can't say exactly which files and strigs are with errors. You can check it yourself.
So maybe you should remove Visual Studio 2015 from README (prerequisites)?

RadWolfie:
Now it will compile under VS 2015 for Cxbx-Reloaded project.

Resolve #1088


LukeUsher:
Thanks, we rely on the community for MSVC 2015 support, since many of us don't use it ourselves, this is appreciated.

PR #1105, opened 2018-04-29
Add fixup for pushbuffer.

jackchentwkh:
Using code from Dxbx to apply fixup data to pushbuffer.

### DIFF
--- a/src/CxbxKrnl/EmuD3D8/PushBuffer.cpp
+++ b/src/CxbxKrnl/EmuD3D8/PushBuffer.cpp
@@ -128,15 +128,50 @@ UINT XTL::DxbxFVFToVertexSizeInBytes(DWORD dwFVF, BOOL bIncludeTextures)
 	return Result;
 }
 
+//Apply fixup data to pushbuffer. Code comes from Dxbx
+void EmuFixupPushBuffer
+(
+	UINT8       *pPushBuffer,
+	UINT8            *pFixup
+)
+{
+	UINT32 SizeInBytes = 0;
+	UINT32 OffsetInBytes = 0;
+
+	while ((SizeInBytes = *(UINT32*)pFixup) != 0xFFFFFFFF) {
+
+		pFixup += 4;
+		OffsetInBytes = *(UINT32*)pFixup;
+		pFixup += 4;
+
+		memcpy(pPushBuffer + OffsetInBytes, pFixup, SizeInBytes);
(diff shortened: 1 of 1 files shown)


jackchentwkh:
Otogi is the only one encountered pushbuffer fixup in my titles.

jackchentwkh:
but unfortunately Otogi regressed with recent master, so I couldn't test it.

LukeUsher:
Otogi crashes when trying to get in-game with this change.

The cause is `memcpy(pPushBuffer + OffsetInBytes, pFixup, SizeInBytes);` , it causes a read from invalid memory. 

Are you sure SizeInBytes is always set/verified correctly?

src/CxbxKrnl/EmuD3D8/PushBuffer.cpp
```diff
@@ -128,15 +128,50 @@ UINT XTL::DxbxFVFToVertexSizeInBytes(DWORD dwFVF, BOOL bIncludeTextures)
...
+	UINT32 OffsetInBytes = 0;
+
+	while ((SizeInBytes = *(UINT32*)pFixup) != 0xFFFFFFFF) {
+
+		pFixup += 4;
+		OffsetInBytes = *(UINT32*)pFixup;
+		pFixup += 4;
+
+		memcpy(pPushBuffer + OffsetInBytes, pFixup, SizeInBytes);
+		pFixup += SizeInBytes;
```
PatrickvL: Given Otogi crashes with this, and reading the comment, it might be needed to round up this increment to a DWORD boundary, so that the while loop repeats on a-DWORD aligned pFixup buffer address.

jackchentwkh: Let me investigate with it.

jackchentwkh:
Wow, my bad. I mistook the pPushBuffer and pFixup as the target and source data address directly.
actually the address should be interpreted with the struct members.
the interpretation code is in Dxbx, but I missed it. sorry again.
I raised a new PR to fix it. 
With Otogi, the fixup data applied correctly. All data are within the pushbuffer data range, and the fixup data did end up with a size in 0xFFFFFFFF.
two new findings:
1. in Otogi, the fixup data seems to be applied to the same offset for many times, I don't know why, maybe it's generated by the compiler or something else.
2. The pushbuffer after Fixup was applied, introduce a error when being run. that part is beyond my knowledge.

PR #1110, opened 2018-04-29
Multi Controller Support with Xinput

jackchentwkh:
Add multi controller support with Xinput
tested with DOA3, two controllers
further test result awaiting.

### DIFF
--- a/src/CxbxKrnl/EmuXInput.cpp
+++ b/src/CxbxKrnl/EmuXInput.cpp
@@ -51,17 +51,38 @@
 static XINPUT_STATE		g_Controller;
 static BOOL				g_bXInputInitialized = FALSE;
 
+//
+//
+//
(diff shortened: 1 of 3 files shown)


AzurikRiseOfPerathia:
It should add the Xbox Original controller compatibility.

jackchentwkh:
Well, if you use xinput wrappers such as x360ce, then you should be able to use original Xbox controller.
Of course you need a usb adaptor for original Xbox controller.

CakeLancelot:
I don't have enough Xinput gamepads to test more than 2 players, and I must be missing something because I cannot for the life of me get x360ce to work. Regardless I tested a few games and they seem to work fine:
cxbx_2018-04-29_13-26-29
cxbx_2018-04-29_13-46-18
image
image

PR #1112, opened 2018-04-30
Data address interpretation fix for PushBuffer and Fixup Data

jackchentwkh:
Data address interpretation fix for PushBuffer and Fixup Data
I make a mistake by using the pFixup and pPushBuffer as data address directly.
now the code should be correct. 
I checked the Otogi with this code, the process of applying Fixup data seems to be correct. But after applying Fixup data, the following execution of pushbuffer went wrong. which is beyond my knowledge.
So we have to choose either to merge this fix, or to revert to not implement the fixup and escape the emu directly.

### DIFF
--- a/src/CxbxKrnl/EmuD3D8/PushBuffer.cpp
+++ b/src/CxbxKrnl/EmuD3D8/PushBuffer.cpp
@@ -128,49 +128,47 @@ UINT XTL::DxbxFVFToVertexSizeInBytes(DWORD dwFVF, BOOL bIncludeTextures)
 	return Result;
 }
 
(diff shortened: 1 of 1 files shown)


PatrickvL:
Having this in is more valuable than the missing implementation of push buffer commands

jackchentwkh:
Yeah, I kept the original code with passing the newly interpreted address first. But somehow VS compiler optimization made it difficult to trace the code. So I merged the code together, after debugging I kept the code as it was. 
I agree that this code is better being merged since the further issue remains in the pushbuffer process which shall be solved in the future.

PR #1113, opened 2018-05-01
RAW pushbuffer handling improve

jackchentwkh:
Try to improve the raw pushbuffer handling a little bit.
add support for NOP method and unknown methods by skpping the parameters indicated by dwCount.
add pushbuffer size check to prevent overflow.

Otogi now gets ingame, but redering is only wireframe. switch to wirefram mode actually looks better.

This improvement is already better than the original code. 
It can be merged if there is no regression (shouldn't have).
futher implementation or improvement might not come in near future since I am really not familiar with D3D stuff.
the handling of method 1818/1808/1800 shall be rewritten in the future, ther are really messed up.

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -2590,7 +2590,7 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_EndPush)(DWORD *pPush)
 		EmuWarning("D3DDevice_EndPush called without preceding D3DDevice_BeginPush?!");
 	else
 	{
-		EmuExecutePushBufferRaw(g_pPrimaryPB);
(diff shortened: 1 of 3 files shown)


jackchentwkh:
Please don't merge this now.
I need more test feedback.

Regressed titles:
hunter the reckoning redeemer

Progress titles:
Otogi.

jackchentwkh:
hoo, sorry again. 
with testing Hunter Redeemer, I found what I was missing.
The fix is in my last commit for method 1800.
If there is no regression, this could be merged.

zaykho:
For some reasons, I can get anything to be rendered in-game with Otogi (in both region, NTSC-US and PAL) and switching to wireframe (F11 key) does nothing in-game (screen filled with a plain blue color).

Is there any specific settings to make it work in wireframe needed ?

jackchentwkh:
you don't have to switch to wire frame mode
in normal mode it only renders as wire frame.

PR #1073, opened 2018-04-16
Dsound fixes part2

RadWolfie:
**ready to merge**
~~P.S. Wait until "Enable debug trace" commit is revert before merge.~~

~~It is still in WIP I believe. However in this current state, it is safe to test titles.~~
Currently is stable and improved enough to merge.

Currently improve title:
* JSRF (skating, jumping, people audio are no longer silent)
* Hunter Reckoning (? Need to verification)
* Hulk
* Burnout (Menu selection and ingame audio is audible since commit 70be0aa)
* Pocketbike Rider (Fixed music unplayble since commit d4f6514)
* The Simpsons: Road Rage (Fixed music and voiceover audios since commit 44a7045)
* Ghost Recon: Thunder Island (Fix ingame distort noises since commit 44a7045)
* SW: KotOR (Fixed first FMV video since commit 62c4493)

~~EDIT: Now only accepting regression titles.~~

EDIT2:
* Burnout - (Crashed when crash into a car at beginning of the race.)
  * Using GetCaps, can cause crash at any time.
  * **NOTE: This is a minor bug, the rest of the gameplay works fine.**
* NFL Blitz 2002 - (DSound buffer lock failed crash, not sure why this is a problem...)
  * **NOTE: It's happening in master branch anyway. Not a regress issue.**

PR #1114, opened 2018-05-01
DSound-fixes-part2.1 (performance update)

RadWolfie:
Both DS Buffer and Stream are now using C++ std vector instead of static
size array which is in C.

It should improve performance for lower-end computers. Before, high-end
computers will only notice very light performance impact. Now appear to
be no performance impact unless certain titles are preserving maximum
size.

P.S. Can be merge since there's no debug trace enable.


HyperScorpio:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1073#issuecomment-386751136 Only if people didn´t read yet.

PR #1123, opened 2018-05-03
Repropose RDTSC patch hack

jackchentwkh:
Repropose RDTSC patch hack
only patch and process exception whtn the hack option is selected.
using pattern match for rdtsc search
exclude certain false positive.


jackchentwkh:
Testers, if you encountered regression with the "Patch CPU Frequency (rdtsc)" hack enabled in this PR, please report back to me or the admins so we can check the real cause and refine this approach.

Currently with this PR I tested certain titles and the results are listed below.
need more test results.
but if the patch cpu frequency hack isn't selected, the emu shall behave like untouched.
If any regression is reported in the future, I shall revise the patter and false positive detection accordingly.

RDTSC patch test case:
******************************
Positive Section:
.text
D3D
BINK
WMADEC
******************************
False Positive:

******************************
Miss Detection:

******************************
Normal: number inside quotes is total rdtsc found
Blinx(3)
Burnout(1)
Burnout2(4)
Burnout3(5)
BurnoutR(4)
DOA3(1)
ExaSkeleton(306)
Futurama(2)
Godzilla Save the Earch (3)
Halo(7)
Hunter Redeemer(9)
Hunter The Reckoning(11)
JSRF(3)
Kung Fu Chaso (13)
Lego Star War(2)
MechAssault(21)
Otogi(2)
Panzer Dragoon ORTA(3)
    PD1.xbe(2)
PGR(9)
RalliSport Challenge(11)
RalliSport Challenge2(30)
Smashing Drive(4)
Sonic Rider(2)
Steel Batallion(9)
TAK2(16)
TestDrive(8)
VCOP3(2)
******************************
False Positive Section:

******************************

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -665,21 +717,67 @@ void PatchPerformanceFrequency()
...
+			|| std::string(CxbxKrnl_Xbe->m_szSectionName[sectionIndex]) == "XONLINE"
+			|| std::string(CxbxKrnl_Xbe->m_szSectionName[sectionIndex]) == "MDLPL") {
+			continue;
+		}
+
+		printf("INIT: Searching for rdtsc in section %s\n", CxbxKrnl_Xbe->m_szSectionName[sectionIndex]);
 		xbaddr startAddr = CxbxKrnl_Xbe->m_SectionHeader[sectionIndex].dwVirtualAddr;
-		xbaddr endAddr = startAddr + CxbxKrnl_Xbe->m_SectionHeader[sectionIndex].dwSizeofRaw;
-		for (xbaddr addr = startAddr; addr < endAddr; addr++)
+		//rdtsx is two byts instruction, it needs at least one op byte after it to finish a function, so the endAddr need to substract 3 bytes.
```
LukeUsher: rdtsx = rdtsc
byts = bytes
op byte = opcode byte

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -652,11 +653,62 @@ static unsigned int WINAPI CxbxKrnlInterruptThread(PVOID param)
...
+	printf("INIT: Patching rdtsc opcode at 0x%.8X\n", (DWORD)addr);
+	opAddr[0] = 0xEF; // OUT DX, EAX
+	opAddr[1] = 0x90; // NOP
+	g_RdtscPatches.push_back(addr);
+}
+
+const uint8_t rdtsc_pattern[] = {
+	0x89,//{0x0F,0x31,0x89 },
+	0xC3,//{ 0x0F,0x31,0xC3 },
+	0x8B,//{ 0x0F,0x31,0x8B },   //one false positive in Sonic Rider .text 88 5C 0F 31
```
LukeUsher: Thanks for documenting false positives, good information to have

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -665,21 +717,67 @@ void PatchPerformanceFrequency()
...
+						PatchRdtsc(addr);
+						//the first for loop already increment addr per loop. we only increment one more time so the addr will point to the byte next to the found rdtsc instruction. this is important since there is at least one case that two rdtsc instructions are next to each other.
+						addr += 1;
+						//if a match found, break the pattern matching loop and keep looping addr for next rdtsc.
+						break;
+					}
+				}
+				if (i>= sizeof_rdtsc_pattern)
+				{
+					//no patter matched, keep record for detections we treat as non-rdtsc for future debugging.
```
LukeUsher: patter = pattern

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -665,21 +717,67 @@ void PatchPerformanceFrequency()
...
+					printf("Skipped potential rdtsc: Unknown opcode pattern  0x%.2X, @ 0x%.8X\n", next_byte, (DWORD)addr);
+				}
 			}
 		}
 	}
 
-	printf("INIT: Done patching xbox performance frequency\n");
+	printf("INIT: Done patching rdtsc, total %d rdtsc patched\n", g_RdtscPatches.size());
```
LukeUsher: total %d rdtsc instructions patched

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -1264,9 +1362,8 @@ __declspec(noreturn) void CxbxKrnlInit
 	// See: https://multimedia.cx/eggs/xbox-sphinx-protocol/
 	ApplyMediaPatches();
 
-	if (g_PatchCpuFrequency) {
-		PatchPerformanceFrequency();
-	}
+	if(g_PatchCpuFrequency)
```
LukeUsher: Please keep braces around if statements, even for single line cases, it keeps code readable & prevents accidental introduction of errors when editing the code later.

jackchentwkh:
Sonic Rider got all rdtsc detected without false positive. But while the title gets in game without rdtsc patch, it crashed itself without any message when the hack is enabled. need further investigation.

src/CxbxKrnl/Emu.cpp
```diff
@@ -233,6 +234,20 @@ extern int EmuException(LPEXCEPTION_POINTERS e)
...
 	{
+		// Check if this exception came from rdtsc, but only whe g_PatchCpuFrequency hack is set.
+		if (e->ExceptionRecord->ExceptionCode == STATUS_PRIVILEGED_INSTRUCTION && g_PatchCpuFrequency )
+		{
+			if (IsRdtscInstruction(e->ContextRecord->Eip)) {
+				LARGE_INTEGER PerformanceCount;
+				PerformanceCount.QuadPart = xboxkrnl::KeQueryPerformanceCounter();
+				e->ContextRecord->Eax = PerformanceCount.LowPart;
+				e->ContextRecord->Edx = PerformanceCount.HighPart;
+				e->ContextRecord->Eip += 2;
```
LukeUsher: If you increment EIP by two here, there was no need to replace the 2nd byte of rdtsc with a nop, as it gets skipped anyway

src/CxbxKrnl/EmuKrnlKe.cpp
```diff
@@ -1096,11 +1096,6 @@ XBSYSAPI EXPORTNUM(126) xboxkrnl::ULONGLONG NTAPI xboxkrnl::KeQueryPerformanceCo
 	ULONGLONG ret;
 	::LARGE_INTEGER PerformanceCounter;
 
-	if (g_PatchCpuFrequency) {
-		RETURN((ULONGLONG)__rdtsc());
```
LukeUsher: Why not rdtsc? QueryPerformanceCounter and rdtsc were the same on Xbox, but on modern Windows, the performance counter works differently... It might be best to keep rdtsc when the hack is enabled

src/CxbxKrnl/EmuXapi.cpp
```diff
@@ -944,25 +944,6 @@ LPVOID WINAPI XTL::EMUPATCH(ConvertThreadToFiber)
...
-// ******************************************************************
-BOOL WINAPI XTL::EMUPATCH(QueryPerformanceCounter)
-(
-	LARGE_INTEGER *lpPerformanceCount
-)
-{
-	FUNC_EXPORTS;
-
-	if (g_PatchCpuFrequency) {
-		lpPerformanceCount->QuadPart = (LONGLONG)__rdtsc();
```
LukeUsher: Why unpatch this function completely? It is required for correct operation when RDTSC hack is disabled.

Unpatching in normal cases breaks timing in many titles, which is why it was patched with an if/else

src/CxbxKrnl/Emu.cpp
jackchentwkh: that's right. but since the rdtsc opcode is two bytes long, I have a habit to patch the rest unused bytes to NOP.

LukeUsher:
I attempted to fix the conflict using Github web UI, hopefully didn't break anything :P

jackchentwkh:
ha, I solved the conflict on my repo and proposed fix with new commits already. we're doing the same work in the same time. mine.....

Margen67:
Sneak King, and Big Bumpin go ingame again with the RDTSC hack enabled. (I confirmed it doesn't work on master with it on or off)

LukeUsher:
Thanks

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -653,11 +655,62 @@ static unsigned int WINAPI CxbxKrnlInterruptThread(PVOID param)
 	return 0;
 }
 
-void PatchPerformanceFrequency()
+std::vector<xbaddr> g_RdtscPatches;
```
PatrickvL: Fast lookup is important, so perhaps it's better to use a std::hash or std::set here?

LukeUsher: It's rare the array will be large enough for it to really make a difference, the overhead of hashing could potentially be slower in this case. Some titles only have two or three occurrences

Issue #1107, opened 2018-04-29
Fix regression "CreateDevice failed"


PatrickvL:
A few titles have regressed since PR https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1090 , like:
* Panzer Dragoon - https://github.com/Cxbx-Reloaded/game-compatibility/issues/141#issuecomment-385209989

jackchentwkh:
ExaSkeleton also shows the same regression

jackchentwkh:
one more title with this error
Steel Battalion

kennethedmonds:
The conker demo disc gives this error. Hopefully this is what you're looking for.
```---------------------------
Cxbx-Reloaded
---------------------------
Received Fatal Message:

* IDirect3D::CreateDevice failed

---------------------------
OK   
---------------------------
```

ObiKKa:
[Armed & Dangerous](https://github.com/Cxbx-Reloaded/game-compatibility/issues/764)

LukeUsher:
I ran into a case with the Dashboard where it tried to use a depth stencil format that is not supported by the GPU in the computer. We should check for format compatibility and fallback where required in CreateDevice too, not just on conversion.

gandalfthewhite19890404:
Which GPUs does not support such formats?

LukeUsher:
most, if they're xbox specific formats. It's going to be addressed by adding format detection and fallbacks, just like how we fixed textures

bryce13950:
I am receiving this issue on what seems like every single game I run. This is a bit of an older issue, so I don't know if what I am receiving is still relevant to this issue. Is there anything I can do to provide information on what may be going on?

LukeUsher:
@bryce13950 chances are your settings are not good: You could be using a display mode (resolution, etc) that your monitor and graphics card combination does not support.

If you're using 'hardware video mode', disable it.

Failing that, come join us on discord and we'll help diagnose the problem.

bryce13950:
@LukeUsher I have some more success when I use software mode, but I mostly get screens that tell me that the disc is corrupted. The resolution I am selecting is native, so I will try playing with that next time I mess around with this. Is there maybe a place in the code that is generating this error, which I could add some more detailed error messaging?

LukeUsher:
@bryce13950 Cxbx-Reloaded does not support software rendering: hardware rendering is *always* used.

[...]

PR #1127, opened 2018-05-03
Fix an issue where AutoDepthStencilFormat of D16_LOCKABLE caused CreateDevice to fail

LukeUsher:
Test case: Ultra Bust-A-Move, probably many others.


PatrickvL:
Is this a complete fix for https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1107 ?

RadWolfie:
The real question is, is it suppose to intend fail inside xbox title? I'm sure we can manually patch/trampoline the function and see the real result call directly return's value at LLE level since LLE gpu is supported. Unless it is alternated in LLE gpu as well.

If it xbox's function suppose to return as fail, then we shouldn't be passive about failures. Let it return as fail, then the title will correct it to produce a proper success.

PR #1131, opened 2018-05-04
Prevent (one) CreateTexture crash

PatrickvL:
Avoid a crash when creating a fallback texture instead of something that was supposed to become a DepthStencil.

This occurred when Panzer Dragoon ORTA goes in-game for example.


AzurikRiseOfPerathia:
Azurik also makes a weird crash, but without error message once entered the game. From the DSond Part 2

PatrickvL:
Please report that elsewhere - comments on pull requests should only concern the pull request itself (in comparison with master)

RadWolfie:
Since this is handled by xbox to host conversion. This is the way to go. 👍

Unfortunately, I don't have time to test it out at the moment.

PR #1144, opened 2018-05-09
D3D : Copy over from Standard functions to LTCG optimized functions

jarupxx:
due lost parameter in EAX.
- D3DDevice_LoadVertexShader_4
- D3DDevice_SetTextureState_BorderColor_4
- D3DDevice_SetStreamSource_8

Rewrite functions
- D3DDevice_SelectVertexShader_4

Also add functions
- D3DDevice_LoadVertexShader_0
- D3DDevice_SetTextureState_TexCoordIndex_0
- D3DDevice_SetTextureState_ColorKeyColor_4

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -2747,29 +2747,56 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_GetVisibilityTestResult)
 
     return D3D_OK;
 }
-
 // LTCG specific D3DDevice_LoadVertexShader function...
-// This uses a custom calling convention where parameter is passed in EAX
-// Test-case: Ninja Gaiden
-VOID __stdcall XTL::EMUPATCH(D3DDevice_LoadVertexShader_4)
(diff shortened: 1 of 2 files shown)


LukeUsher:
Keep in mind that because the Trampolines within the SetXXX functions do not work in LTCG titles that the GetXXX functions will need to be patched too. 

In non-LTCG XDKs it's not a problem, because the trampoline ensures internal state is updated properly, but with LTCG, subhook cannot make a working trampoline due to non-standard calling convention.

jarupxx:
Umm. LTCG titles does not contain any GetXXX functions...?

LukeUsher:
@jarupxx Functions such as GetStreamSource() exist in LTCG titles and non-LTCG titles, but we don't need to patch them because our trampolines make sure the internal state gets set, allowing the unpatched function to work as expected.

With LTCG titles, these trampolines don't work, so all of the Get functions (like GetStreamSource) need to be patched as well as the Set functions, where for non-LTCG, only the Set functions need to be patched.

jarupxx:
@LukeUsher That's interesting. We can add it later.

PR #1146, opened 2018-05-09
Fix RalliSpor crash by skip QUADLIST rendering with index buffer.

jackchentwkh:
a quick fix to skip QUADLIST rendering with index buffer.
the D3D shader will access memory out of vertex buffer range in this case.
Sorry, don't have time to fix the non index part since skip rendering in that part will make rendering result worse.


jackchentwkh:
well, it shouldn't take too much work since you already have the necessary code in your branch.

PatrickvL:
Here's how : https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1147

Issue #1151, opened 2018-05-11
X_D3DResource is released undetected.


jackchentwkh:
With RalliSport, if you tried carrier mode with series of races, the xbe encounters a exception which was introduced by accessing an X_D3DResource which was released somehow but still in our g_HostResources map.

the call stacks is as below:
```
>     Cxbx.exe!GetXboxCommonResourceType(const XTL::X_D3DResource *) Line 589 c++ symbol loaded
      Cxbx.exe!HostResourceRequiresUpdate(unsigned __int64 key, unsigned long) Line 859 c++ symbol loaded
     Cxbx.exe!EmuVerifyResourceIsRegistered(XTL::X_D3DResource * pResource, int iTextureStage, unsigned long dwSize) Line 2249 c++ symbol loaded
     Cxbx.exe!GetHostResource(XTL::X_D3DResource * pXboxResource, int iTextureStage) Line 796 c++ symbol loaded
      Cxbx.exe!GetHostBaseTexture(XTL::X_D3DResource *) Line 948 c++ symbol loaded
      Cxbx.exe!GetHostTexture(XTL::X_D3DResource *) Line 953 c++ symbol loaded
      Cxbx.exe!EmuUpdateActiveTextureStages() Line 6762     C++    c++ symbol loaded
      Cxbx.exe!CxbxUpdateNativeD3DResources() Line 6773     c++ symbol loaded
     Cxbx.exe!XTL::EmuPatch_D3DDevice_DrawVerticesUP(XTL::_X_D3DPRIMITIVETYPE PrimitiveType, unsigned int VertexCount, void * const pVertexStreamZeroData, unsigned int VertexStreamZeroStride) Line 6933 c++ symbol loaded

```
I add logs in functions in HostResourceRequiresUpdate, D3DResource_Release, and SetHostResource
logged list as below. 

```

D3DDevice_SetTexture pTexture=18b9ddc0, stage=0.
SetHostResource new X_D3DResource=18b9ddc0 Common=    40000
D3DDevice_SetTexture pTexture=18b9ddc0, stage=0.
GetXboxCommonResourceType X_D3DResouce =  18b9ddc0 Common=    40000 
D3DDevice_SetTexture pTexture=18b9ddc0, stage=0.
GetXboxCommonResourceType X_D3DResouce =  18b9ddc0 Common=    40000 
D3DDevice_SetTexture pTexture=18b9ddc0, stage=0.
[...]

PR #1161, opened 2018-05-12
Fix RallySport crashes in multi racing in series.

jackchentwkh:
Fix RallySport crashes in multi racing in series.
probably fix other titles in game crashes as well.
when a key is detected in the g_HostResource map, don't access the pXboxResouce's member directly, becasue the pXboxResource might be release by the title already.


jackchentwkh:
fixed https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1151

PR #1163, opened 2018-05-13
Hotfix for LLE support

RadWolfie:
In kernel log, first round will have actual memory address even if LLE is enabled. However, 2nd start up will show as zeroes if one of LLE is enabled. It is normal for time being.

Plus enforce show warning of D3D message after symbol scan.


RadWolfie:
Got approval from tester had an issue with LLE GPU enabled.

x1nixmzeng:
@RadWolfie why did you remove find()? Any keys queried like that will be added to the std::map with a default value, which is luckily 0 - but now it has a different meaning.

RadWolfie:
Due to first key value had been existed. The cache scan assume it is there with non-zero address. However, that's not the case every time. We still need warning messages remain even after 1st run of title boot.

PR #1168, opened 2018-05-14
Fixup stuff

RadWolfie:
Fixup for following things:
* Initialize nullptr for both D3D state variables, check if symbol is not exist in vector then check the value is 0.
* Pre-fixup for DSound's re-initialize DirectSoundCreate function. (Currently not active at the moment.)

### DIFF
--- a/src/CxbxKrnl/EmuDSound.cpp
+++ b/src/CxbxKrnl/EmuDSound.cpp
@@ -283,16 +283,16 @@ HRESULT WINAPI XTL::EMUPATCH(DirectSoundCreate)
         // clear sound buffer cache
         vector_ds_buffer::iterator ppDSBuffer = g_pDSoundBufferCache.begin();
         for (; ppDSBuffer != g_pDSoundBufferCache.end();) {
-            while (XTL::EMUPATCH(IDirectSoundBuffer_Release)((*ppDSBuffer))) {};
-            ppDSBuffer = g_pDSoundBufferCache.erase(ppDSBuffer);
+            while (XTL::EMUPATCH(IDirectSoundBuffer_Release)((*ppDSBuffer)) != 0) {};
+            ppDSBuffer = g_pDSoundBufferCache.begin();
         }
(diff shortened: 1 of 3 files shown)


RadWolfie:
@x1nixmzeng Does this meet your expectation? 😉

src/CxbxKrnl/EmuDSound.cpp
```diff
@@ -283,16 +283,16 @@ HRESULT WINAPI XTL::EMUPATCH(DirectSoundCreate)
...
         }
         g_pDSoundBufferCache.reserve(X_DIRECTSOUND_CACHE_MAX);
 
         // clear sound stream cache
         vector_ds_stream::iterator ppDSStream = g_pDSoundStreamCache.begin();
         for (; ppDSStream != g_pDSoundStreamCache.end();) {
-            while (XTL::EMUPATCH(CDirectSoundStream_Release)((*ppDSStream)));
-            ppDSStream = g_pDSoundStreamCache.erase(ppDSStream);
+            while (XTL::EMUPATCH(CDirectSoundStream_Release)((*ppDSStream)) != 0);
+            ppDSStream = g_pDSoundStreamCache.begin();
```
x1nixmzeng: So this has a side-effect of restarting your for loop, which will call `CDirectSoundStream_Release` again.

Also, no objects are removed from the stream - is that intentional?

x1nixmzeng:
:+1: for adding `find` back, but try not to overcomplicate your loop logic

src/CxbxKrnl/EmuDSound.cpp
RadWolfie: Release only return unsigned integer. PatrickvL was whining about checking the bits inside while. 😉

Objects are removed inside Release function. See here: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/270a821ad5d0d3728750830c5308e41003fa702c#r28802938 Calling erase twice isn't logical.

Reason there's a loop is due to AddRef is called directly to host's object. Don't worry about that, it is currently part of inactive code.

PR #1171, opened 2018-05-15
Fix GitHub tab style

RadWolfie:
For comparison between main repo vs this commit for tab size.

**Before: (default to 8 spaces)**
image

**After: (set to 4 spaces as default in config file)**
image

It also affect Visual Studio 2017 as integrated support. VS 2015 and below are not affected.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+root = true
+
+[*]
+indent_size = 4


LukeUsher:
.editorconfig can be loaded by Visual Studio too, it is natively supported in Visual Studio 2017. There is a menu option (Edit > Advanced > Format Document) which will apply the format defined in .editorconfig.

So this is a great addition!

RadWolfie:
So, what you are saying... Do not include indent size in Clang format file? Since .editorconfig only has very limited support. Correct?

LukeUsher:
@RadWolfie .editorconfig should be able to handle everything: All modern IDEs include native support, and the few that don't (such as Visual Studio 2015) almost all have third party plugins that can be installed to allow .editorconfig to work.

RadWolfie:
Hmm, I'm thinking it should be in src directory instead of root directory. This way it will not affect import directory for submodules and we can clone editorconfig file to specific third-party directory if we want or need to.

PR #1178, opened 2018-05-16
Update readme (again) and issue template

Margen67:
Added Discord link to issue template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,8 @@ We now boot more titles than ever before, with more and more reaching gameplay.
 ### Minimum
   * OS: Windows 7+ x64. 32-bit installations are not supported.
   * GPU: Anything that supports Direct3D 9 (HLE) and OpenGL 3.3 (LLE).
-  * CPU: x64 with SSE2.
 ### Prerequisites
-  * Visual C++ 2015 and 2017 redistributables may be needed. Download them [here](https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads).
+  * Visual C++ 2015 and 2017 redistributables may be required. Download them [here](https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads).
   
 ## Automated Builds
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,35 +16,36 @@ We now boot more titles than ever before, with more and more reaching gameplay.
 
 ## System Requirements
 ### Minimum
-  * OS: Windows 7 64-bit or newer. 32-bit installations are not supported.
-  * Video card: anything that supports Direct3D 9 (HLE) and OpenGL 3.3 (LLE).
+  * OS: Windows 7+ x64. 32-bit installations are not supported.
+  * GPU: Anything that supports Direct3D 9 (HLE) and OpenGL 3.3 (LLE).
+  * CPU: x64 with SSE2.
```
LukeUsher: It is redundant to specify this as '32-bit installations are not supported' implies that X64 support is required, and there are no x64 CPUs without SSE2, at all.

README.md
```diff
@@ -15,35 +16,36 @@ We now boot more titles than ever before, with more and more reaching gameplay.
...
 ### Prerequisites
-  * Visual C++ 2015 and 2017 redistributables may be required. Download them [here](https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads).
+  * Visual C++ 2015 and 2017 redistributables may be needed. Download them [here](https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads).
```
LukeUsher: Why this wording change? (required->needed)

README.md
Margen67: Should I replace it with a clock speed? (like 1.5GHz)

LukeUsher: There is currently no known minimum CPU requirements, and it's likely to change very regularly as we implement more functionality and optimize, probably best to leave CPU out completely

README.md
Margen67: They both mean the same thing so I'll just undo it.

Margen67:
@LukeUsher Is it good now?

PR #1179, opened 2018-05-16
Editorconfig second revision

RadWolfie:
* Apply only for src and resource directories.

With this setup, it should work as intended only for Cxbx-Reloaded's source code without affecting third-party projects.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,0 @@
-root = true
-
-[*]
-indent_size = 4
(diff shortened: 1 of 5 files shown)


RadWolfie:
@LukeUsher, are you sure there are more options for editorconfig file extension? I only see...
* root
* indent_size
* indent_style (not in VS2017?)
* tab_width (not in VS2017?)
* trim_trailing_whitespace
* end_of_line
* insert_final_newline

for standard, excluding CSharp list. I don't see any wording for how code format for brackets, spacing in-between, etc.

PR #1186, opened 2018-05-18
Fix Cube Mapping & Minor Tweaks

LukeUsher:
image

### DIFF
--- a/src/Common/EmuEEPROM.cpp
+++ b/src/Common/EmuEEPROM.cpp
@@ -177,21 +177,7 @@ xboxkrnl::XBOX_EEPROM *CxbxRestoreEEPROM(char *szFilePath_EEPROM_bin)
 
 	if (NeedsInitialization)
 	{
-		memset(pEEPROM, 0, EEPROM_SIZE);
-
-		// TODO: Make these configurable or autodetect of some sort :
-		pEEPROM->UserSettings.Language = XC_LANGUAGE_ENGLISH;  // = English
-		pEEPROM->UserSettings.VideoFlags = AV_FLAGS_NORMAL;   // = Use XDK defaults
-		pEEPROM->UserSettings.AudioFlags = XC_AUDIO_FLAGS_STEREO;   // = Stereo, no AC3, no DTS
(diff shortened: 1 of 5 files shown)


src/Common/EmuEEPROM.cpp
```diff
@@ -187,6 +187,7 @@ xboxkrnl::XBOX_EEPROM *CxbxRestoreEEPROM(char *szFilePath_EEPROM_bin)
 		pEEPROM->UserSettings.ParentalControlMovies = XC_PC_MAX; // = XC_PRTL_CRTL_MAX
 		pEEPROM->UserSettings.MiscFlags = 0;  // No automatic power down
 		pEEPROM->FactorySettings.AVRegion = AV_STANDARD_NTSC_M | AV_FLAGS_60Hz;
+		memcpy(&pEEPROM->FactorySettings.SerialNumber, "Cxbx-R      ", 12);
 		pEEPROM->EncryptedSettings.GameRegion = XC_GAME_REGION_NA;
```
ergo720: I would suggest to also add the above in DlgEepromConfig.cpp (look at IDC_EE_RESET). If you don't, then the reset functionality of the eeprom menu will still reset the serial number to a null value.

LukeUsher: Thanks! That begs the question of why eeprom data is reset in different places...  I'll clean it up later.

LukeUsher:
Appveyor failed to build the debug build, don't merge until this is solved.

Margen67:
The cubemap/specular on the car in RaceX is still missing.

LukeUsher:
@Margen67 probably a different issue not strictly related to this one.

LukeUsher:
Merging: It may be a little premature but this is such a small/simple fix in reality that it's minimal impact.

Margen67:
Crash Bandicoot: TWoC no longer gets past the Universal logo (confirmed regression):
```---------------------------
Cxbx-Reloaded
---------------------------
Received Fatal Message:

* CreateCubeTexture Failed!

Error: 
Desc: 

---------------------------
OK   
---------------------------
```

PR #1189, opened 2018-05-18
Duplicate handle BEFORE calling NtQueueApcThread!

blueshogun96:
As the comment states, before NtQueueApcThread or QueueUserAPC can be successfully called, it is necessary to duplicate the handle with the right permissions.  After banging my head on a wall trying to deal with Metal Slug 3, this was the only solution I could find.


RadWolfie:
@LukeUsher & @blueshogun96, only reason EmuXApi.cpp file had been formatted whole file is due to new lines are not crlf. Crlf are enforced by .editorconfig file since Xbox is made by Microsoft.

If we only need `src/CxbxKrnl/EmuKrnlNt.cpp` changes. @blueshogun96, you can perform soft reset to your commit then select only EmuKrnlNt.cpp file to be commit. I had done this before a lot and is easy to do.

PR #1191, opened 2018-05-19
Show correct icon to popup message dialog

RadWolfie:
Proper fix for message dialog's icon. This is to lead correct icon such as test case is information, warning is warning, error is error, etc.

### DIFF
--- a/src/CxbxKrnl/CxbxKrnl.cpp
+++ b/src/CxbxKrnl/CxbxKrnl.cpp
@@ -489,7 +489,7 @@ HANDLE CxbxRestorePageTablesMemory(char* szFilePath_page_tables)
 
 #pragma optimize("", off)
 
-void CxbxPopupMessage(CxbxPopupMsgIcon icon, const char *message, ...)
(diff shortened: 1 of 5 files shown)


src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -489,17 +489,38 @@ HANDLE CxbxRestorePageTablesMemory(char* szFilePath_page_tables)
...
+        }
+    }
 
 	va_start(argp, message);
 	vsprintf(Buffer, message, argp);
 	va_end(argp);
 
 	EmuWarning("Popup : %s\n", Buffer);
-	MessageBox(NULL, Buffer, TEXT("Cxbx-Reloaded"), MB_OK | MB_ICONEXCLAMATION | MB_TOPMOST | MB_SETFOREGROUND);
+	MessageBox(NULL, Buffer, TEXT("Cxbx-Reloaded"), uType);
```
RadWolfie: Just to add a note, EmuWarning function above had not been alternate to specific type of message is for. Though, feels like not high priority.

Issue #1160, opened 2018-05-12
Vibration crashes TechCertGame


PatrickvL:
TechCertGame: If vibration is enabled in options, the sample will throw an exception if you try to start a game. If it's turned off, however, the sample will run as normal. This applies even if an XInput controller with vibration is used.

If vibration is disabled in options, there's still a vibration-related crash when pressing START during in-game

ObiKKa:
How did you remember this issue was linked to that PR?

PatrickvL:
Because it says so right here : https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1194#issuecomment-390512673

PR #1194, opened 2018-05-20
Fix in SetState polling loop

jagotu:
Once the pFeedback->Header.hEvent event is set, the game is free to free the buffer that it was pointing to. In HP1 game, the pointer even lead to the stack. Because that stack was changing a lot, pFeedback->Header.dwStatus eventually resolved to a non-zero value and we started to overwrite some random memory.

Harry Potter and the Philosopher's Stone now gets in-game.

image


jagotu:
Turns out the previous patch stopped the crashing but caused some freezing.

These two commits also resolve #1160.

PR #1198, opened 2018-05-24
Xbdm

PatrickvL:
This pull request introduces loading of the xbdm debugging library - all functions are stubbed out right now, but at least with this we have something to build upon.

To get this working, thunk table loading and library importing have been extended. This shouldn't influence anything in regular emulation.


LukeUsher:
Cool idea, but we are still missing a lot of functionality that will allow XBDM to work: Things like breakpoints, network support, etc will cause problems

They'll trap into Visual Studio/Our exception handler rather than XBDM's traps. I'm not sure it's possible to make it work fully with Cxbx-Reloaded's direct code execution  approach.

That said, I'll accept this, anything is better than nothing

PR #1201, opened 2018-05-24
JaGoTu quickfix branch

RadWolfie:
**please do not merge yet**

Needs verification with Conker title (said to reported boot properly).

@ergo720, I believe a missing else statement cause the problem. Although this pull request clean it up a bit.


ergo720:
From Discord Conker does this:
```
[0x1720] KRNL: NtQueryVirtualMemory(
   BaseAddress          : 09530000
 OUT Buffer             : 0x092DB104
);
[0x1720]    Buffer->AllocationBase    = 0x00000000
[0x1720]    Buffer->AllocationProtect = 0x00000000
[0x1720]    Buffer->BaseAddress       = 0x09530000
[0x1720]    Buffer->RegionSize        = 0x02A4A000
[0x1720]    Buffer->State             = 0x00010000
[0x1720]    Buffer->Protect           = 0x00000001
[0x1720]    Buffer->Type              = 0x00000000
[0x1720] KRNL: NtQueryVirtualMemory returns 0
[0x1720] KRNL: NtAllocateVirtualMemory(
   BaseAddress          : (PULONG)0x092DB0E8 (*value: 0x09530000)
   ZeroBits             : 0
   AllocationSize       : (PULONG)0x092DB0EC (*value: 0x00070000)
   AllocationType       : (ALLOCATION_TYPE)0x00002000 = XBOX_MEM_RESERVE
   Protect              : (PROTECTION_TYPE)0x00000001 = XBOX_PAGE_NOACCESS
);
[0x1720] VMEM: XbAllocateVirtualMemory(
   *addr                : 0x09530000
   ZeroBits             : 0
   *Size                : 0x00070000
   AllocationType       : 2000
   Protect              : 1
);
[0x1720] VMEM: XbAllocateVirtualMemory requested range : 0x09530000 - 0x095A0000
[0x1720] VMEM: XbAllocateVirtualMemory returns C0000018
[0x1720] KRNL: NtAllocateVirtualMemory returns C0000018
[0x1720] KRNL: RtlNtStatusToDosError(
   Status               : -1073741800
);
[0x1720] KRNL: RtlNtStatusToDosError returns 1E7
```
I see that NtQueryVirtualMemory is reporting that the memory at location 0x09530000 is free (State is 0x10000 = XBOX_MEM_FREE) and the title tries then to allocate it with NtAllocateVirtualMemory, which fails. This can happen because the VMManager doesn't track the host allocation (and it can't known anything about those either) and they appear as free areas in the memory region map. This seems to suggest to me that the allocation is failing even though the vma is free because an host allocation is allocated there. If that is so, then sadly this problem is unsolvable with the direct execution model because we can't move the host allocations somewhere else and we have no control on those. The only way to solve this is to do real paging which requires a software MMU and in turn LLE CPU.

RadWolfie:
XbVirtualMemoryStatistics function isn't affecting this correct? If not, then it is a different issue.

ergo720:
Is the above from Splinter Cell or Conker? I don't have any of those so I can't test, I thought it was Conker. The above analysis still holds, though.

RadWolfie:
I believe it is Conker. Splinter Cell Double Agent only perform a new crash (it actually made small progress).

jagotu:
@ergo720 I think it could be solved by also tracking the host allocations and reporting them properly. That way the game will see some stuff it didn't allocate as not free, but at least properly coded games will work with our current HLE memory approach. What do you think?

PatrickvL:
But how to classify host allocations? There's the emulation executable, lots of DLL's and emulation data present in the emulator address space.
But Xbox allocations are known - so any query on pages that are already reserved or committed but which don't belong to any Xbox allocation, must be deemed a host allocation

jagotu:
Well the way I see it right now we have the choice between two incorrect behaviors:

1. Report host memory allocated by DLLs etc. as FREE. The game sees what it expects, but fails when attempting to Allocate it. We have no way of fixing that with HLE.
2. Report host memory allocated by DLLS etc. as it really is, that is sometimes FREE and sometimes occupied. The game might be surprised by all the occupied memory, but if it's correctly written, it will not try to allocate the occupied memory.

I tried implementing number 2 as a PoC and that gets Splinter Cell to boot. I will make a pull request so people can check for any regressions and we can continue the discussion there.

PR #1202, opened 2018-05-25
Hotfix EEPROM

### LINKED ISSUE #1199: Eeprom is corrupted if I'd set Xbox Region to Japan
I'd like to set Xbox Region to Japan but cannot manage to do it by myself.
I've accept Japan but eeprom is corrupted.

Configuration
config

Result
result

RadWolfie:
Found the source of cause, after this change. Attempted to reproduce the issue no longer return. Now everything works as intended.

Resolve #1199


PatrickvL:
Well done!

PR #1205, opened 2018-05-25
QueryMemory reports host allocated memory

jagotu:
**Do not merge yet**

Continuing from discussion on #1201.

With this fix QueryMemory reports the true (host) state of VirtualMemory when queried instead of marking it as FREE. A game might be surprised by all the occupied memory, but if it's correctly written, it will not try to allocate the occupied memory and therefore will not crash.

Case in point: With this fix, Splinter Cell - Double Agent boots.

Please, test this for regressions in other games and discuss.

### DIFF
--- a/src/CxbxKrnl/VMManager.cpp
+++ b/src/CxbxKrnl/VMManager.cpp
@@ -2105,6 +2105,28 @@ xboxkrnl::NTSTATUS VMManager::XbVirtualMemoryStatistics(VAddr addr, xboxkrnl::PM
 		return STATUS_INVALID_PARAMETER;
 	}
 
+	// If it's not in the XBE, report actual host allocations
+	// The game will see allocations it didn't make, but at least it has a chance to
+	// not try to allocate memory our emulator already occupied.
+	if (addr > XBE_IMAGE_BASE + ROUND_UP_4K(CxbxKrnl_Xbe->m_Header.dwSizeofImage))
+	{
+		MEMORY_BASIC_INFORMATION info;
+		if (VirtualQuery((void*)addr, &info, sizeof(info)))
+		{
+			memory_statistics->AllocationBase = info.AllocationBase;
+			memory_statistics->AllocationProtect = info.AllocationProtect;
+			memory_statistics->BaseAddress = info.BaseAddress;
+			memory_statistics->Protect = info.Protect;
+			memory_statistics->RegionSize = info.RegionSize;
+			memory_statistics->State = info.State;
+			memory_statistics->Type = info.Type;
+			return STATUS_SUCCESS;
+		}
+		else {
+			return STATUS_INVALID_PARAMETER;
+		}
+	}
+
 	Lock();
 
 	// Locate the vma containing the supplied address


jagotu:
@ergo720 As I said, I don't agree that what was there before is a proper implementation. We report memory as free that the game can't allocate and that is wrong. I think there is no proper implementation with direct execution. But this way of tackling it makes more sense to me.

LukeUsher:
I agree with @jagotu. If we had more control over the Xbox memory map (via CPU and MMU emulation) we could have everything work exactly as on Xbox, but with dynamic code execution, it's just not possible. Cxbx-Reloaded's code and data must sit *somewhere* in the 32-bit address space, all we can do is prevent Xbox titles from using the same memory range and hope it doesn't break anything (Which this fix takes care of)

Due to the direct code execution approach of Cxbx-Reloaded, there is NO separation between host and xbox memory, making a perfect solution impossible.

RadWolfie:
I agree with @ergo720. We must emulate real RAM limitation, using host's available memory is not an option.

Therefore it should be given as hack.

I wonder if this is being prevented by something allocated inside Cxbx's source code causing this problem. I remember attempted to use Xbox's allocated memory function with DirectSound for trying to resolve an old problem in the past. It didn't work out well. Until moment later, I remember DirectSound managed its own memory within DirectSound's section and APU hardware address too.

LukeUsher:
@RadWolfie we MUST make sure that Cxbx-Reloaded reports memory used by the host as allocated, so that Xbox titles do not try to allocate memory in that virtual address region. Making this a hack is pointless: It's necessary to make any title that iterates through memory to find a free page to work as expected.

This is a limitation of direct code execution.

The issue exists because both Host and Xbox have the full 32bit 4GB virtual address space available for usage (even though the Xbox only has 64MB of Physical Memory, it can still map virtual pages at ANY location within the 4GB address space): However, pages that are free on Xbox are NOT free in Cxbx-Reloaded, when they should be. This is unavoidable because our emulation code resides in the same virtual address space as the xbox memory.  As an example, a VirtualAlloc to 0x5000000 (or any other arbitrary address) will SUCCEED on an Xbox as long as the total memory allocated across all pages is less than the memory size (64/128MB depending on the console). On Cxbx-Reloaded, this allocation will most likely fail because the Cxbx-Reloaded code resides here. (The host Heap, Stack, and variables exist somewhere in this address space too but it is unknown where at runtime because Windows decides this)

This can NOT be fixed unless we stop using direct code execution and emulate the processor and MMU properly.

RadWolfie:
We need to get this https://github.com/PatrickvL/Cxbx-Reloaded/tree/Loader branch into master.

Memory manager is limited to 64MB. The other virtual space address are only used for mapping to internal hardware memory. [For details, use this link](http://xboxdevwiki.net/Memory).

LukeUsher:
@RadWolfie that's not true: The memory manager is limited to 64MB of Physical Memory, but any virtual address is available for usage. Memory Size and Virtual Address Space size are completely unrelated. That branch will NOT fix this issue, because Cxbx-Reloaded code will still be contained within the same 4GB virtual address space as Xbox code runs. There is real fix to this other than full CPU and MMU emulation

LukeUsher:
I'm not sure how to explain this better/more clearly: Would it be beneficial if I prepared a test case that works on real hardware but fails on Cxbx-Reloaded for this exact reason? (Using VirtualAlloc on an address higher than 64MB that succeeds on Xbox but fails on Cxbx-Reloaded due to a memory conflict)

src/CxbxKrnl/VMManager.cpp
```diff
@@ -2105,6 +2105,28 @@ xboxkrnl::NTSTATUS VMManager::XbVirtualMemoryStatistics(VAddr addr, xboxkrnl::PM
 		return STATUS_INVALID_PARAMETER;
 	}
 
+	// If it's not in the XBE, report actual host allocations
```
RadWolfie: Move lock above here to prevent race condition per ergo720 request.

src/CxbxKrnl/VMManager.cpp
```diff
@@ -2105,6 +2105,28 @@ xboxkrnl::NTSTATUS VMManager::XbVirtualMemoryStatistics(VAddr addr, xboxkrnl::PM
...
+		MEMORY_BASIC_INFORMATION info;
+		if (VirtualQuery((void*)addr, &info, sizeof(info)))
+		{
+			memory_statistics->AllocationBase = info.AllocationBase;
+			memory_statistics->AllocationProtect = info.AllocationProtect;
+			memory_statistics->BaseAddress = info.BaseAddress;
+			memory_statistics->Protect = info.Protect;
+			memory_statistics->RegionSize = info.RegionSize;
+			memory_statistics->State = info.State;
+			memory_statistics->Type = info.Type;
```
RadWolfie: Add unlock below here.

src/CxbxKrnl/VMManager.cpp
```diff
@@ -2105,6 +2105,28 @@ xboxkrnl::NTSTATUS VMManager::XbVirtualMemoryStatistics(VAddr addr, xboxkrnl::PM
...
+			return STATUS_SUCCESS;
+		}
+		else {
```
RadWolfie: Here as well.

PatrickvL:
I'm with Luke and jagotu on this one - sorry Radwolfie and ergo720, but they make a few good arguments that'd make me accept this approach.

Although I understand you could see this as a hack, it's really not; Xbox code should be prepared for memory queries to fail - if it doesn't, that'd be bad programming. (Perhaps we could create custom hacks for such cases, which would be easier with my loader branch : for certain titles we could reserve predefined ranges, so that the memory range for which a query failure is ignored, would still be dedicated for emulation purposes)

RadWolfie:
I agree with @LukeUsher since MMU emulation is what we really need to enable full support for host not directly touching these 4GB of 32 bit memory.

src/CxbxKrnl/VMManager.cpp
LukeUsher: There is no need: The Lock is to prevent multiple threads accessing our Page Table structure simultaniously, however, VirtualQuery is not effected by this.

The race condition Ergo720 mentioned is that another thread could allocate memory at any point after VirtualQuery is called, but there's nothing we can do about that at the moment.

LukeUsher:
Merged early: I really want this in the upcoming v0.1 stable release. This is going to go through extensive testing, so any issues will be ironed out pretty quickly.

src/CxbxKrnl/VMManager.cpp
RadWolfie: Alrightly then, I approve since pull request is already merged. 😛

Sarkie:
A lot of early games and homebrew have issues with lots of memory being available.

So this might fix a few more things

PR #1208, opened 2018-05-25
Tweak Compiler Optimization Settings

LukeUsher:
This should bring a slight to moderate performance increase.


build/win32/Cxbx.vcxproj
```diff
@@ -233,15 +233,17 @@
...
       <EnableEnhancedInstructionSet>StreamingSIMDExtensions2</EnableEnhancedInstructionSet>
       <MinimalRebuild>true</MinimalRebuild>
-      <ExceptionHandling>Sync</ExceptionHandling>
+      <ExceptionHandling>SyncCThrow</ExceptionHandling>
       <BufferSecurityCheck>false</BufferSecurityCheck>
       <ControlFlowGuard>false</ControlFlowGuard>
       <IntrinsicFunctions>true</IntrinsicFunctions>
       <FavorSizeOrSpeed>Speed</FavorSizeOrSpeed>
+      <FloatingPointModel>Precise</FloatingPointModel>
+      <WholeProgramOptimization>true</WholeProgramOptimization>
```
RadWolfie: When this is set to true. Then how come "WholeProgramOptimization" with condition set to false base on your other commit here below?

build/win32/Cxbx.vcxproj
```diff
@@ -438,6 +441,7 @@
       <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='Debug_Direct3D9|Win32'">%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <WholeProgramOptimization Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">false</WholeProgramOptimization>
```
RadWolfie: Okay, so CxbxKrnl.cpp source code cannot have optimization? By defining proper call such as stdcall, cdecl, etc. It shouldn't matter with optimization enabled.

I'm curious about this.

LukeUsher: Applying LTCG to CxbxKrnl.cpp breaks the virtual memory placeholder: it moves the variable away from the xbe base address. Other optimisation are still applies there, just not LTCG

RadWolfie: Ah, gotcha. Thanks for explanation. 😃

PR #1210, opened 2018-05-26
Keep track of thread handles duplicated for APCs

blueshogun96:
I should have thought about this before.  DuplicateHandle increases the reference count of objects it duplicates, so to ensure it gets closed successfully, we need to ensure that all duplicates are closed so that the reference count is zero.


src/CxbxKrnl/EmuKrnlNt.cpp
```diff
@@ -977,18 +991,38 @@ XBSYSAPI EXPORTNUM(206) xboxkrnl::NTSTATUS NTAPI xboxkrnl::NtQueueApcThread
...
+		if(!DuplicateHandle(g_CurrentProcessHandle, ThreadHandle, g_CurrentProcessHandle, &hApcThread, THREAD_SET_CONTEXT,FALSE,0))
+			EmuWarning("DuplicateHandle failed!");
+		else
+		{
+			g_DuplicateHandles[ThreadHandle] = hApcThread;	// Save this thread because we'll need to de-reference it later
+			DbgPrintf( "DuplicateHandle returned 0x%X (ThreadId)\n", hApcThread, GetThreadId( hApcThread ) );
+		}
+
+
+		ret = NtDll::NtQueueApcThread(
```
PatrickvL: What if the second handle can't do it either - would that lead to continuously duplicating?

build/win32/Cxbx.sln
```diff
@@ -128,9 +128,11 @@ Global
 		{B8D9AFC2-B38F-4714-846D-8A2754F076C9}.Debug_Direct3D9|Any CPU.Build.0 = Release|Win32
 		{B8D9AFC2-B38F-4714-846D-8A2754F076C9}.Debug_Direct3D9|Win32.ActiveCfg = Debug|Win32
 		{B8D9AFC2-B38F-4714-846D-8A2754F076C9}.Debug_Direct3D9|Win32.Build.0 = Debug|Win32
+		{B8D9AFC2-B38F-4714-846D-8A2754F076C9}.Debug_Direct3D9|x64.ActiveCfg = Debug|Win32
```
PatrickvL: Why these two additions?

LukeUsher: Visual studio adds them automatically because one of our dependencies has a 64bit target. I've just been manually excluding the file from the commit, but they come back with every open/save of the project.

src/CxbxKrnl/EmuKrnlNt.cpp
blueshogun96: Right, I forgot about that.  I should delete the duplicated thread handle upon failure.

build/win32/Cxbx.sln
RadWolfie: Yeah, we need to re-config sln's settings to exclude all of x64 support for Cxbx main project only. Once this fix goes into master, we should not have any further problem in the future for pull requests.

blueshogun96:
Okay, took PatrickvL's advice and did another update.

Shogun

PatrickvL:
Lgtm

PR #1214, opened 2018-05-29
Fix rc files

RadWolfie:
I has confirm rc files are now readable difference between two file changes on Cxbx.rc file (including CxbxDll.rc as well).

Also, Cxbx.rc file has been converted to UTF-8 encode.


jagotu:
I have a hunch that this might be incorrect.

From the comment you deleted from .gitattributes, it was added because "-# Microsoft Visual C++ generated resource script with forced Unicode". The way I understand that is the rc file is automatically generated by Visual Studio, and it will always generate it in UTF-16, and git expects it to be UTF-8.

So maybe you reformatted it to UTF-8 now, but next time the file changes because VS regenerates it in UTF-16, git might commit it as a text file with broken unicode.

LukeUsher:
That would make sense actually, I think I'll revert this for now

PR #1216, opened 2018-05-29
D3D : Fix pushbuffer based drawing 

PatrickvL:
This pull request fixes drawing errors in push-buffer based drawing, by calling CxbxUpdateNativeD3DResources from there too.

Test-case : BeginPush XDK sample


LukeUsher:
This requires testing with JSRF and other titles, as this PR is not limited to Push buffers and also effects titles rendering using SetVertexData4f/D3D_Begin/D3D_End.

PatrickvL:
JSRF works like a charm!

PR #1220, opened 2018-05-29
Revert "Revert rc files to binary (changed in 02141bd1c65)"

RadWolfie:
Reverts Cxbx-Reloaded/Cxbx-Reloaded#1218

Do not revert back to binary. Thanks.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,7 +17,3 @@
 *.PDF	 diff=astextplain
 *.rtf	 diff=astextplain
 *.RTF	 diff=astextplain
-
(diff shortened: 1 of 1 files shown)


jagotu:
I agree that it's nice to have diffs, but from what I've seen this change made it impossible for jackchen to compile the project. I reverted it mainly so we can discuss it before it's done and make sure we fix it in a proper way, as you self-merged even the first pull request without leaving any time for comments.

In the end the goal is to have them recognized as text files, if possible, but I think this quick solutions might have some downsides you didn't allow to discuss.

PR #1226, opened 2018-05-30
Fix rc encode issue

### LINKED ISSUE #1223: Visual Studio resource editor re-encode issue
**NOTE:** The current workaround is to save as UTF-8 encode before commit to master branch for Cxbx.rc file.

This is a discussion how to resolve prevent automation re-encode back to UTF-16 issue in Visual Studio.

I had read one of a sort of answered in stackoverflow with instruction to do this task.
`tools > options > environment > documents > "Save documents as Unicode when data cannot be saved in codepage"`

After editing and saving. It is remain in UTF-8. Maybe we could use compatible codepage for both git and Visual Studio to use?

Resources:
https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/3082310-support-rc-files-in-utf8-encoding
https://stackoverflow.com/questions/17594923/using-unicode-for-resource-file-with-visual-studio-2010

~~One possible solution is to use UTF-8 with BOM since I do not see it been re-encoded.~~ This method doesn't work since it is failing to open resource editer view.

RadWolfie:
Resolve #1223


RadWolfie:
@jackchentwkh Once this is merge, you should no longer have a problem with resource file re-encode back to utf-16 encode. It will remain as utf-8 encode without any conflict with resource editor.

jackchentwkh:
@RadWolfie currently I have to save cxbx.rc in unicode encoding which is supposed UTF16 to make VS work. If I save the rc file in unicode UTF8 it won't compile. 
If you can solve the issue and make it consistent to the project for all devs that would be great.
thanks a lot.

RadWolfie:
UTF8 (without BOM) and UTF16 works fine with Visual Studio's resource editor. I had confirmed it on my side few times to be sure. UTF8 (with BOM) cause a problem, also has been confirm as well.

Can you DM me for further assistance?

Or at least try pull `fix-rc-encode-issue` from my forked repo into your new branch (with master's commits only). Then try open the file through resource editor again.

PR #1230, opened 2018-05-31
Merge g_RegisteredResources and g_HostResources into a combined g_XboxDirect3DResources.

LukeUsher:
This cleans up the code significantly and prevents issues caused by inconsistencies between the two structures.

As a bonus, this pretty much halves the number of map/vector lookups that happen per call to GetHostResource, bringing a (slight) performance boost

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -162,7 +162,6 @@ static XTL::INDEX16                *pQuadToTriangleIndexBuffer = nullptr;
 static UINT                         QuadToTriangleIndexBuffer_Size = 0; // = NrOfQuadVertices
 
 static XTL::X_D3DSurface		   *g_XboxBackBufferSurface = NULL;
-static XTL::D3DFORMAT			   g_HostBackBufferFormat = XTL::D3DFMT_X8R8G8B8; // Most common format, will be overwritten at runtime if incorrect
 static XTL::X_D3DSurface           *g_pXboxRenderTarget = NULL;
 static XTL::X_D3DSurface           *g_pXboxDepthStencil = NULL;
 static bool                         g_bColorSpaceConvertYuvToRgb = false;
@@ -2061,9 +2060,6 @@ static DWORD WINAPI EmuCreateDeviceProxy(LPVOID)
 
                 if(FAILED(g_EmuCDPD.hRet))
                     CxbxKrnlCleanup("IDirect3D::CreateDevice failed");
-
-				// Store BackBuffer Format
-				g_HostBackBufferFormat = g_EmuCDPD.HostPresentationParameters.BackBufferFormat;
 
 				// Which texture formats does this device support?
 				memset(g_bSupportsFormatSurface, false, sizeof(g_bSupportsFormatSurface));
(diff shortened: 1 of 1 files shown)


RadWolfie:
Looks better to me, have not test for any regression btw.

PatrickvL:
Why separate `hasHostResource` from the actual `pHostResource` itself? Can't the latter be check for 'assigned-ness' for the former too?!?

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -4972,6 +4976,11 @@ void CreateHostResource(XTL::X_D3DResource *pResource, DWORD D3DUsage, int iText
 			if (pbSupportedFormats[X_Format]) {
 				// Then use matching host format
 				PCFormat = EmuXB2PC_D3DFormat(X_Format);
+
+				// If, and ONLY if this is the default backbuffer, make sure the format matches the host backbuffer
```
LukeUsher: This actually improved the Fire XDK sample on my test machine from 8FPS all the way through to 60FPS

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -729,6 +726,7 @@ void *GetDataFromXboxResource(XTL::X_D3DResource *pXboxResource)
 }
 
 typedef struct {
+	bool hasHostResource = false;
```
RadWolfie: Actually, would it be better to check if `pHostResource != nullptr` instead of using this variable?

LukeUsher: PatrickvL just said that :P but yeah, you're both right

src/CxbxKrnl/EmuD3D8.cpp
LukeUsher: Basically, if the host resource for the Xbox Backbuffer (our default render target) does not match the actual host backbuffer, the copy operation in D3DDevice_Swap ends up doing a VERY slow pixel conversion on the CPU, instead of a fast blit operation.

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -2305,7 +2303,7 @@ static void EmuVerifyResourceIsRegistered(XTL::X_D3DResource *pResource, DWORD D
 
         //check if the same key existed in the HostResource map already. if there is a old pXboxResource in the map with the same key but different resource address, it must be freed first.
         
-        if (it->second.hasHostResource && (it->second.pXboxResource != pResource)) {
+        if (it->second.pXboxResource != pResource) {
```
RadWolfie: I wonder if both are nullptr. Which will cause corruption if didn't fatal crash on its own. Would it be better to have a safety check for nullptr or not?

RadWolfie: If pResource never do return as nullptr, then above comment can be ignore.

LukeUsher: If they are both nullptr, the if statement will be false, the else case will be executed, calling HostResourceRequiresUpdate which will return false (since it detects pXboxResource is not set). So it's not a problem

RadWolfie: Okay, I did misread it then. 😋

PR #1232, opened 2018-05-31
Removed unused files

x1nixmzeng:
Not sure if this belongs in the master or dev branch?


LukeUsher:
Master's fine,  it will have on impact on compatibility so it's zero risk for the stable release.

PR #1235, opened 2018-06-01
Fix an issue where the ucSubType was derived incorrectly

LukeUsher:
This may fix missing input in some titles.

Lens Flare XDK sample still broken, needs further investigation


PatrickvL:
According to https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1244#issuecomment-394109774 this pull request fixes Sonic Riders

PR #1240, opened 2018-06-01
Call Xbox SetRenderTarget to update internal state

LukeUsher:
This *might* fix the recent regression.


LukeUsher:
I honestly thought this was already being done, which is why I went ahead with #1233.
But obviously, it wasn't :P

PR #1241, opened 2018-06-01
D3D: Fix SetRenderState_FillMode wrongly rendering as points

Voxel9:
This fixes many 5XXX XDK samples and Sonic Riders from erroneously rendering meshes as points instead of solid.

Tested with some other titles and samples, found no regressions whatsoever.

Before:
image
image

After:
image
image


LukeUsher:
Good find,

PR #1243, opened 2018-06-01
Pass iterator around to minimize map lookups:

LukeUsher:
Slight to Moderate performance boosts in titles using a lot of resources (GetHostResource call chain calls g_XboxDirect3DResources.find(key); less times now)


LukeUsher:
Bleh, this makes THPS 2X and 3 crash after intros, sorry!

PR #1244, opened 2018-06-02
Xinput fix and minor clean up

jackchentwkh:
fix non-responsive xinput for Otogi again.
code clean up
correct note for xinput subtype define.


PatrickvL:
Does this also fix lensflare?

jackchentwkh:
donno, I don’t have xdk sample built

Voxel9:
Sonic Riders input was fixed since PR #1235 
As for lensflare, this didn't help it unfortunately. Can't even toggle wireframe/points rendering; I think it just freezes.

PR #1257, opened 2018-06-07
D3D : Limit SetViewport to RenderTarget bounds

PatrickvL:
This could fix some cases where nothing is visible.


LukeUsher:
thanks!

PR #1258, opened 2018-06-07
EmuWarning stack fix + Rename _logPrefix

### LINKED ISSUE #1256: EmuWarning Stack Corruption
Currently, EmuWarning internally uses a static buffer of 1024 bytes in size. with no bounds checking. This is a piece of left-over code from original Cxbx that should have been fixed a LONG time ago

This means that if EmuWarning is called with a string > 1023 characters in length (+1 for null terminator), it WILL corrupt the stack. This is very bad, as stack corruption can cause instability or even be exploited for malicious purposes.

Either implement a bounds check in EmuWarning, refactor to use a dynamically sized buffer, or better yet, refactor so it doesn't need a buffer at all.  

Something like this should be possible, and avoids allocating a buffer (untested)
```C
va_list argp;
fprintf(stdout, "[0x%.4X] WARN: ", GetCurrentThreadId() ");
va_start(argp, fmt);
vfprintf(stdout, fmt, argp);
va_end(argp);
fprintf(stdout, "\n");
```

RadWolfie:
Test code: `EmuWarning("%s 123 - %d=321, %.4f=1234.5000", "Test", 321, 1234.5f);`
output: `[0x18A8C] WARN: Test 123 - 321=321, 1234.5000=1234.5000`

Resolve #1256

### DIFF
--- a/src/Common/Logging.cpp
+++ b/src/Common/Logging.cpp
@@ -40,7 +40,7 @@
 
 // For thread_local, see : http://en.cppreference.com/w/cpp/language/storage_duration
 // TODO : Use Boost.Format http://www.boost.org/doc/libs/1_53_0/libs/format/index.html
-thread_local std::string _logPrefix;
+thread_local std::string _logThreadPrefix;
 
 const bool needs_escape(const wint_t _char)
 {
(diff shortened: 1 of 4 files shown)


PatrickvL:
Lgtm

RadWolfie:
Added @LukeUsher as co-author since the idea came from him in the issue.

RadWolfie:
Verified over 1024 bytes length with no stack corruption. Merging now.

PR #1259, opened 2018-06-08
Fix compile warnings general

RadWolfie:
Resolve part of compile warning #55


RadWolfie:
Once this pull request is merge, `fix-compile-warnings-nv2a` branch for pull request will be created.

Since we are "stuck" with GetVersion function from CxbxKrnl.cpp file. Plus other solution I discover are using Windows' API. We are not able to treat warning as error.

PR #1262, opened 2018-06-08
Fix compile warnings nv2a

RadWolfie:
Resolve part of compile warning #55

Most warnings came from NV2A apparently.

### DIFF
--- a/src/devices/video/EmuNV2A_PGRAPH.cpp
+++ b/src/devices/video/EmuNV2A_PGRAPH.cpp
@@ -3102,10 +3102,11 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
         unsigned int reg = (i < 2) ? NV_PGRAPH_CSV1_A : NV_PGRAPH_CSV1_B;
         for (j = 0; j < 4; j++) {
             unsigned int masks[] = {
-                (i % 2) ? NV_PGRAPH_CSV1_A_T1_S : NV_PGRAPH_CSV1_A_T0_S,
-                (i % 2) ? NV_PGRAPH_CSV1_A_T1_T : NV_PGRAPH_CSV1_A_T0_T,
-                (i % 2) ? NV_PGRAPH_CSV1_A_T1_R : NV_PGRAPH_CSV1_A_T0_R,
-                (i % 2) ? NV_PGRAPH_CSV1_A_T1_Q : NV_PGRAPH_CSV1_A_T0_Q
+                // NOTE: For some reason, Visual Studio thinks NV_PGRAPH_xxxx is signed integer. (possible bug?)
+                (i % 2U) ? (unsigned int)NV_PGRAPH_CSV1_A_T1_S : (unsigned int)NV_PGRAPH_CSV1_A_T0_S,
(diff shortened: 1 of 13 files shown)


RadWolfie:
Release build's warning has drop from 53, master branch, to 15 warnings (as of commit dff0ef1). 🎉

All feedback of what's need to be change are welcome. Keep in mind, LLE GPU had not been tested on my side.

Also any correction you do know is right, feel free to make a commit directly to this branch/pull request.

Margen67:
@RadWolfie Maybe once all the warnings at this level are fixed it can be raised?

src/devices/video/EmuNV2A_PGRAPH.cpp
```diff
@@ -3013,7 +3013,7 @@ static void pgraph_update_shader_constants(PGRAPHState *pg,
     }
 
     if (binding->surface_size_loc != -1) {
-        glUniform2f(binding->surface_size_loc, pg->surface_shape.clip_width,
+        glUniform2ui(binding->surface_size_loc, pg->surface_shape.clip_width,
```
LukeUsher: Make sure this doesn't break anything: there might be a reason xqemu used the 2f call

RadWolfie:
@Margen67, there's no way to treat warning as error level to be raise. The option is either treat warning as error or warning. See "Treat Warnings as Errors" option.

Visual Studio's "Warning Level" is entirely different thing. It only verbose more warnings to alert the developers about something need to be fix before sending new release to public or private group.

PatrickvL:
LGTM

JayFoxRox:
Please upstream the changes to XQEMU.
Ideally we'd do changes upstream and they'd eventually land in Cxbx-R.

Doing them downstream first and not even informing upstream of these changes, or putting the burden to redo all changes on the upstream maintainers makes it very hard to maintain the codebase (which also means you won't benefit from future XQEMU changes).

JayFoxRox:
The major difference is that the XQEMU code in Cxbx-R is being compiled as C++, when it's really C code. On XQEMU we probably also run with less strict flags (which we could probably review) which we inherit from QEMU.

However, there's some changes which are good eitherway:

* https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1262/commits/03f672e58a727974f56597479546d1bb92895b5b should be done.
* https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1262/commits/87c15944322e18dd17d1c83d55b88f8f13a179a1 might also be worth an upstream issue.
* https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1262/commits/be049defa955f721c0418f4f1e1ce5969b3fd014#diff-9fd5d9f42f307703c6a4d87e30c687baR309 should be handled somehow (not sure if we have an `UNREACHABLE()` macro).
* ...


It shouldn't be the upstreams responsibility to cherry-pick such changes from downstream.
If the downstream makes such changes, it will only make it harder to accept future changes from upstream; and if the downstream is more popular than the upstream (which is the case here), it can easily turn into a hostile fork (which nobody wants).

So the motivation for the changes (silence errors) should not matter here in my opinion.
If the upstream is not warning-free or rejects changes, they can still be added downstream, or disabled using pragmas (as also done here).

PR #1265, opened 2018-06-08
DInput : Poll each device only once (instead of per object)

PatrickvL:
Before this change, for each input object (any button, trigger, axis, etc) the entire device state was polled.
This change polls each device only once per update.
This reduces the overhead of DirectInput polling.

### DIFF
--- a/src/Common/Win32/XBController.cpp
+++ b/src/Common/Win32/XBController.cpp
@@ -482,6 +482,7 @@ void XBController::ListenPoll(XTL::X_XINPUT_STATE *Controller)
         return;
 
     XTL::LPDIRECTINPUTDEVICE8 pDevice=NULL;
+    XTL::LPDIRECTINPUTDEVICE8 pPrevDevice=(XTL::LPDIRECTINPUTDEVICE8)-1;
 
     HRESULT hRet=0;
     DWORD dwFlags=0;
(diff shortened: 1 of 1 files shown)


RadWolfie:
These changes should not be in here as it is not optimize method. There are already statements for it.
https://github.com/PatrickvL/Cxbx-Reloaded/blob/ec846238686a1864d6b8d8b38bae389990a9dbf8/src/Common/Win32/XBController.cpp#L528-L543

The only changes I see should be in here is check previous device then poll. Plus your current method doesn't check if third device or later are the same device as first device.

src/Common/Win32/XBController.cpp
```diff
@@ -507,16 +512,36 @@ void XBController::ListenPoll(XTL::X_XINPUT_STATE *Controller)
...
+					hRet = pDevice->Acquire();
+			}
+
+			if (dwFlags & DEVICE_FLAG_JOYSTICK) {
+				JoyState = { 0 };
+				if (pDevice->GetDeviceState(sizeof(JoyState), &JoyState) != DI_OK)
+					continue;
+			}
+			else if (dwFlags & DEVICE_FLAG_KEYBOARD) {
+				memset(KeyboardState, 0, sizeof(KeyboardState));
```
RadWolfie: Why memset? `{ 0 }` should be fine.

RadWolfie: It's really strange structure accept it while array by itself doesn't.

src/Common/Win32/XBController.cpp
```diff
@@ -482,6 +482,7 @@ void XBController::ListenPoll(XTL::X_XINPUT_STATE *Controller)
         return;
 
     XTL::LPDIRECTINPUTDEVICE8 pDevice=NULL;
+    XTL::LPDIRECTINPUTDEVICE8 pPrevDevice=(XTL::LPDIRECTINPUTDEVICE8)-1;
```
RadWolfie: Also if checking pDevice for nullptr, then -1 is not need?

PatrickvL: Beter safe than sorry

RadWolfie:
lgtm

PR #1268, opened 2018-06-09
New Speedhack: Render Directly to Host BackBuffer

LukeUsher:
If enabled, this speedhack negates the impact of the new SetRenderTarget implementation WITHOUT breaking the functionality, at the expense of a little accuracy.

This could be useful for bumping up the frame-rate for some users, especially when resolution scaling hacks are used.

### DIFF
--- a/resource/Cxbx.rc
+++ b/resource/Cxbx.rc
@@ -482,9 +482,13 @@ BEGIN
         END
         POPUP "Hacks",                          65535,MFT_STRING,MFS_ENABLED
         BEGIN
+            POPUP "Speed Hacks",                    ID_HACKS_SPEEDHACKS,MFT_STRING,MFS_ENABLED
+            BEGIN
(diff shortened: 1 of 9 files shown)


LukeUsher:
A few notes:
This PR was zero risk, hence why it made it in even though there's a freeze: If the hack is disabled there's no impact.

Without the hack I only get 40FPS even on splash screens in JSRF, now they're 60FPS
Without the hack I get 30-40FPS in-game in JSRF (garage), with the hack, I get 50-60FPS

With the hack, resolution scaling works again in the Dashboard, maybe others with patches.
Without the hack, resolution scaling destroys performance (1080p becomes ~1fps)

The performance impact of the hack is very real for many users, it's a huge improvement.

The accuracy cost is very slim too: On xbox, YUV overlays are rendered ontop of the backbuffer by the GPU, as an overly, the actual backbuffer isn't affected. With this hack, the YUV overlays overwrite backbuffer content, but it won't matter for the majority of games.

PR #1269, opened 2018-06-09
Pass iterator around to minimize map lookups:

LukeUsher:
This brings a slight to moderate performance boost in titles using a lot of resources

Basically, this reduces the number of calls to .find on the unordered_map, and also reduces the calls to GetHostResourceKey by saving and passing around the iterator returned from the first find as much as possible.

Previously, a function would do a lookup, then call a function which would then do another lookup, and then call another function which would do another lookup, for each active resource!

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -1,4 +1,4 @@
-﻿// This is an open source non-commercial project. Dear PVS-Studio, please check it.
+// This is an open source non-commercial project. Dear PVS-Studio, please check it.
 // PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com
 // ******************************************************************
 // *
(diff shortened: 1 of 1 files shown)


LukeUsher:
Too many regressions

LukeUsher:
Reverted because too many regressions, see above comment.

Basically, there is a situation that causes crashes when an interator stops being valid but continues to be used. This needs fixing before this can be merged

gandalfthewhite19890404:
This PR cause crash after booting.
Here are two logs from release build (one with d3d8 wrapper, one - without it) and one log from tracebuild.
screen
[releaselogs.zip](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/2086849/releaselogs.zip)
[Uploading tracelog.zip…]()

PatrickvL:
Oops! I'm sorry - didn't see that

PR #1279, opened 2018-06-10
Various fixes (Mainly multi-xbe and LLE flags status)

### LINKED ISSUE #1061: Feature: Show HLE or LLE state option for GPU/APU during emulation
From discord discussion with @PatrickvL 

> Because people tend to forget mentioning HLE or LLE GPU, maybe we should put that before the FPS counter?

My idea is to put it after FPS counter due to the fact width of FPS status do change from time to time depending on user. Then @PatrickvL had an idea to use one letter H or L.

My final thought is to use G:L** / G:H** and A:L** / A:H** after Help menu or after FPS counter. It should only use L or H at start of emulation depending on user's selection.

Using this method will help both developers and testers to tell the difference for screenshots and videos quicker.

_** G(PU), A(PU), L(LE), H(LE)_

### LINKED ISSUE #1182: Multi-Xbe Launch bug
Sometimes it work as intended, sometimes it doesn't.

For example, Crimson Skies is using same xbe file as reboot. After several reboots, it can cause "Emulating title" state back to "Loaded title" as if there's a problem with launching another xbe. However, the gameplay had not changed.

Cxbx-Reloaded's GUI has a bug for multi-xbe launch. It needs investigation to find source of regression.

### LINKED ISSUE #1274: LLE flags text not being set correctly
The LLE flags text is supposed to change when:

1. Emulation is not running
2. One or more of the actual LLE flags are changed.

Instead, the text does not change when emulation is not running, but the text does change when it is modified while emulation is running.

RadWolfie:
The following fixes in this pull request contain:
* Resolve #1182 (Multi-xbe launch bug)
* Resolve #1061 (Real time status from kernel process)
  * Also resolve #1274 (changing LLE flags configure while emulating)
* Fix numeric ordering for IDs (there are some different IDs using same value)
  * For prevent future risk cause of corruption from manual editing if occur.
* Prevent second title to emulate while previous title is already emulating.

Please buy me a hot cocoa, thanks. 😩


RadWolfie:
commit 7bd449c should had say
> This will help **stop** producing false postive report and inform user to stop current title emulating before start new emulation.

Oh well. 🤷‍♂️

src/Cxbx/WndMain.cpp
```diff
@@ -2276,57 +2299,58 @@ void WndMain::StopEmulation()
...
 	return 0;
 }
 
 // monitor for crashes
 void WndMain::CrashMonitor()
 {
 	bool bQuickReboot;
-	HANDLE hCrashMutex = OpenMutex(MUTEX_ALL_ACCESS, FALSE, "CrashMutex");
+	DWORD dwProcessID_ExitCode = 0;
+	GetWindowThreadProcessId(m_hwndChild, &dwProcessID_ExitCode);
```
PatrickvL: If any of these win API's return success/error codes, you'd better check them, especially given the WaitForSingleObject below might cause trouble otherwise

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -905,6 +906,37 @@ void CxbxKrnlMain(int argc, char* argv[])
...
+    int waitCounter = 10;
+    bool isReady = false;
+
+    while (waitCounter > 0) {
+        g_EmuShared->GetIsReady(&isReady);
+        if (isReady) {
+            break;
+        }
+        waitCounter--;
+        Sleep(100);
```
PatrickvL: Show machines (that could be number-crunching and/or memory swapping for all we know) could take more than a second to start - perhaps a retry or abort dialog is in order here?

src/CxbxKrnl/EmuShared.h
```diff
@@ -194,20 +206,22 @@ class EmuShared : public Mutex
...
+		int          m_UncapFramerate;
+		int          m_UseAllCores;
+		int          m_SkipRdtscPatching;
+		float        m_MSpF;
 		float        m_FPS;
-		bool		 m_bMultiXbeFlag;
-		bool		 m_bDebugging;
+		bool         m_bMultiXbeFlag;
+		bool         m_bDebugging;
+		bool         m_bReady;
```
PatrickvL: Changing the order of members in here could be a problem for any independently compiled program that accesses EmuShared too, like the debugger (not sure if it does, but still)

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -324,6 +324,28 @@ void EmuD3D_Init_DeferredStates()
...
+void EmuUpdateLLEStatus()
+{
+    int FlagsLLE;
+    g_EmuShared->GetFlagsLLE(&FlagsLLE);
+
+    if ((FlagsLLE & LLE_GPU) == false
+        && (g_SymbolAddresses.find("Direct3D_CreateDevice") == g_SymbolAddresses.end()
+            || g_SymbolAddresses["Direct3D_CreateDevice"] == 0)) {
+        bLLE_GPU = true;
+        FlagsLLE ^= LLE_GPU;
```
PatrickvL: If you add a fallback safeguard like this, at least log when it happens! Both times.

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -324,6 +324,28 @@ void EmuD3D_Init_DeferredStates()
...
+    }
+
+    if ((FlagsLLE & LLE_APU) == false
+        && (g_SymbolAddresses.find("DirectSoundCreate") == g_SymbolAddresses.end()
+            || g_SymbolAddresses["DirectSoundCreate"] == 0)) {
+        bLLE_APU = true;
+        FlagsLLE ^= LLE_APU;
```
PatrickvL: Here too

src/CxbxKrnl/EmuShared.h
RadWolfie: Which do you rather have?
A) Unaligned data structure** can cause risk involve other coding language not to match unaligned structure.
B) Or keep the data structure clean and align properly.

4 bools will reserve all 4 bytes aligned. At least that's what I recall.

** unused space between first single bool and second int variable

src/CxbxKrnl/CxbxKrnl.cpp
RadWolfie: I agree.

src/CxbxKrnl/EmuShared.h
PatrickvL: I'm not entirely sure EmuShared *is* used elsewhere - I do know it was a problem before, can when we had the emulator in a DLL

RadWolfie: In the future, it is best to create reserved variables. If two bools are used and 2 bools are not. Then it will be:
```c
bool m_bReady;
bool m_InUsed;
bool m_Reserved1;
bool m_Reserved2;
```
Do the same thing for short too.

RadWolfie: It's also best to make sure all the ordering are align before v0.1 release. Once after the release, we should lock it down unless someone forget to make the reserve variables to maintenance the alignment in the future.

src/Cxbx/WndMain.cpp
RadWolfie: Done, see commit cdbd1aa verification process.

src/Cxbx/WndMain.cpp
```diff
@@ -515,14 +515,14 @@ LRESULT CALLBACK WndMain::WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lP
 						g_EmuShared->SetLedSequence(LedSequence);
 						SetTimer(hwnd, TIMERID_FPS, 1000, (TIMERPROC)NULL);
 						SetTimer(hwnd, TIMERID_LED, XBOX_LED_FLASH_PERIOD, (TIMERPROC)NULL);
-						m_hwndChild = GetWindow(hwnd, GW_CHILD); // (HWND)HIWORD(wParam) seems to be NULL
+						m_hwndChild = (HWND)lParam;
```
RadWolfie: Fun fact, during troubleshooting multi-xbe launch bug. I came across lParam has the exact same value as `GetWindow(hwnd, GW_CHILD)` return value. By using lParam value, run-time process uses less CPU cycle. 😉

RadWolfie: Although, MSDN doc says it is a pointer to CREATESTRUCT structure. 🤔
Source: https://msdn.microsoft.com/en-us/library/windows/desktop/ms632619.aspx

PatrickvL:
Thanks for all the fixes. I'm not too fond of all the goto's, but this looks good to me otherwise

RadWolfie:
Long as we know how to use goto's properly, we're good. 😉

RadWolfie:
Tester report Conker boot up with LLE GPU (LLE GPU is unchecked). HLE database detection didn't find `Direct3D_CreateDevice` symbol. Is this intention in xbox or do we need to use `D3DDEVICE` symbol instead 
 before rollback to LLE GPU?

PatrickvL:
As for determining the possibility of running HLE GPU : It's true that without a detected Direct3D_CreateDevice symbol, we can't show anything. But there's more to it than that; we'd also need Swap or Present patched, at least a Draw or BeginPush patched, and a few resource creations (like textures, vertex or index buffers and shaders). If nothing of the above is found, it's either a launcher-type Xbe, or an Xbe not build with an xdk. If the Xbe does draw anything, then doing a fallback to LLE GPU is the only option left indeed.

RadWolfie:
Luke suggest to use library check inside xbe header instead of looking up individual symbol. Since DXD8's LTCG has different variety of symbols to create a device.

And yes, you are right. I meant fallback. 😋

If any titles missing Direct3D_CreateDevice, we need to report it in XbSymbolDatabase repo.

Conker apparently is using `Direct3D_CreateDevice_16` and `Direct3D_CreateDevice_4` base on game compatibility repo.

2 things will be fix:
1. Use fallback wording instead of rollback.
2. Check library name instead of relying on HLE cache symbols.

src/Cxbx/WndMain.cpp
PatrickvL: On my end I haven't seen usable content in any param, so please add a fallback to the getwindow thingy just to be sure ...

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -353,6 +352,27 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
 {
     Xbe::LibraryVersion *pLibraryVersion = (Xbe::LibraryVersion*)pXbeHeader->dwLibraryVersionsAddr;
 
+    uint16 xdkVersion = 0;
```
PatrickvL: I see you're applying existing code to reach a secondary goal - nicely done!

src/Cxbx/WndMain.cpp
RadWolfie: Done

src/CxbxKrnl/HLEIntercept.cpp
RadWolfie: Yep, no reason to reprocess twice. 😉

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -324,11 +324,55 @@ void EmuD3D_Init_DeferredStates()
...
+    if ((FlagsLLE & LLE_GPU) == false
+        && !((XbLibScan & XbSymbolLib_D3D8) > 0
+            || (XbLibScan & XbSymbolLib_D3D8LTCG) > 0)) {
+        bLLE_GPU = true;
+        FlagsLLE ^= LLE_GPU;
+        EmuOutputMessage(XB_OUTPUT_MESSAGE_INFO, "Fallback to LLE GPU.");
+    }
+
+    if ((FlagsLLE & LLE_APU) == false
+        && (XbLibScan & XbSymbolLib_DSOUND) == 0) {
```
jarupxx: XDK 4039 does not have DSound library info, so XDK 4039 title uses unintended Fallback.

RadWolfie: Thank you for a reminder about that.

RadWolfie: Resolved in commit 1ed96cf. Though, we should check section headers in XbSymbolDatabase just like commit 1ed96cf does instead of enforce scan for DSOUND if library version doesn't exist.

RadWolfie:
Is there anything else needs a fix before merge?

@Cxbx-Reloaded/developers: Since this is a big update and improvement, please merge it once nothing needs any fix.

PatrickvL:
Question; Would this PR help the debugger too? I ask because of this report stating the debugger looses track once Cxbx restarts : https://github.com/Cxbx-Reloaded/game-compatibility/issues/181#issuecomment-396122413

RadWolfie:
I tried to search for EmuShared.h alike files or any of method variable inside debugger project. I came back with no result. It sounds like a different issue to me.

PR #1281, opened 2018-06-12
Temporary check HLE cache file for LLE flags used.

RadWolfie:
This pull request enforce check LLE flags in HLE cache file was used with. Since patches still haven't been separated from symbol detection list from Cxbx-Reloaded itself.

This is a temporary solution until we get more complete functional of what should be patch or not into their own list.


RadWolfie:
Ready to merge

PR #1282, opened 2018-06-12
Enable Host Backbuffer by default

LukeUsher:
Enable Direct Host Back Buffer Access hacks by default. This hacks improves performance and allow resolutions scaling to (kind-of) work. (When the Scale Viewport hack is also enabled)

Also: The hacks config in the Kernel Log now specifies if they're the default setting, or modified.

### DIFF
--- a/src/Cxbx/WndMain.cpp
+++ b/src/Cxbx/WndMain.cpp
@@ -222,7 +222,7 @@ WndMain::WndMain(HINSTANCE x_hInstance) :
 			dwType = REG_DWORD; dwSize = sizeof(DWORD);
 			result = RegQueryValueEx(hKey, "HackScaleViewport", NULL, &dwType, (PBYTE)&m_ScaleViewport, &dwSize);
 			if (result != ERROR_SUCCESS) {
-				m_ScaleViewport = 1;
+				m_ScaleViewport = 0;
 			}
 
 			dwType = REG_DWORD; dwSize = sizeof(DWORD);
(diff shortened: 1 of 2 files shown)


LukeUsher:
image

LukeUsher:
~20FPS speed boost in some titles for myself

RadWolfie:
JSRF is the only title I see big improvement for fps. As for other titles I tested doesn't seem to improve or lose fps. However, Crimson Skies is one of the title I can actually see visible ocean on whole screen. Although, it's only affected by enabling "Scale Xbox viewport to host (and back)" hack.

Without 2 hacks on:
2 of hack option off

With 2 hacks on (by default in this pull request):
2 of hack option on

I have not seen any half screen render issues so far.

PatrickvL:
Ideally, hacks should all be off, so perhaps we should invert their description somehow, instead of turning these two on by default?

PatrickvL:
Thanks for this. I don't quite follow the discussion about the scaling hack (I seem to miss some background information), but I take it you've thought it through.

RadWolfie:
Quoting from two testers:
> My opinion - Scale to host hack should be turned off by default because it can bring performance impact with graphic bugs. Direct render to host - can be turn on by default. That is only my opinion.

> Yeah I've experienced crashes and bugs with it before so IDK if it should be default.
It would introduce even more variables in emulation.

Not sure about if LukeUsher has the same issue currently. Though, I do agree with both testers since the scale to host hack can break things and even crash between pull requests (since it did happened before).

PR #1284, opened 2018-06-14
Fix false positive of title is still emulating after crash.

### LINKED ISSUE #1283: Emulation State Becomes Stuck After A Soft Crash
This issue happens with any game that will crash, but not cause the emulator itself to crash as well.
You load a game, but at some point it well crash. If you try to start emulation of the same game, or close the game and try to start emulation on any other game, including working ones, you will get an error message.
image
The only way to clear the emulation state is to exit the emulator and open it again. This did not happen in previous builds.

I tested this on CXBX-Reloaded build number b5db64924962fd31336409a48f495b79f903218f.
Some of the games I tested are:
Animaniacs - The Great Edgar Hunt
CRAZY TAXI 3 High Roller

RadWolfie:
Resolve #1283

Whoops, forgot about crash manager to set emulating to false. (I thought it does call to StopEmulation function, guess I was wrong.)

### DIFF
--- a/src/Cxbx/WndMain.cpp
+++ b/src/Cxbx/WndMain.cpp
@@ -2355,6 +2355,7 @@ void WndMain::CrashMonitor()
 	DrawLedBitmap(m_hwnd, true);
 	m_hwndChild = NULL;
 	m_bIsStarted = false;
+	g_EmuShared->SetIsEmulating(false);
 	UpdateCaption();
 	RefreshMenus();
 }


RadWolfie:
This pull request is low risk. Though I want to be sure there is no other location making any more false positive for is still emulating.

RadWolfie:
Base on OP's reported issue for only when a crash has occur then get stuck with the popup message until close emulator. Merging it since the pull request contain the fix inside CrashManager function. Plus there is no other crash manager I know of in the Cxbx-R's source code

PR #1285, opened 2018-06-14
Fix Title_ID parsing/rendering to use the correct format.

LukeUsher:
This also updates the rendering of the Version Flag.

All retail Xbox Games use a title_id of this format: XX-000

Where XX = Publisher code (MS, EA, etc) and 000 = Game Number by that publisher.

This can be used to properly uniquely identify games, and even cross-reference with the Redump database that uses the same Serial No. format.

Additionally, this PR also fixes Version field rendering, to be in the correct 1.XX format.

This is useful because the combination of Title_ID and Version No uniquely identifies games!
Even in regional varients where the title_id doesn't change, the version number does!

Some special XBEs (like Dashboard, updaters, XDK samples) use non-printable characters in the title_id/serial number field, so we fallback to Hex in this case

This NEEDS to be in v0.1 so that we can use the same game-database going forward for future builds.
Suddenly changing title_id presentation after release would be confusing.

### DIFF
--- a/src/Common/XbePrinter.cpp
+++ b/src/Common/XbePrinter.cpp
@@ -42,6 +42,8 @@
 #include <sstream> // For std::stringstream
 #include <iomanip> // For std::setfill, std::uppercase, std::hex
 
+extern std::string FormatTitleId(uint32_t title_id); // Exposed in Emu.cpp
+
 // better time
 static char *BetterTime(uint32 x_timeDate)
 {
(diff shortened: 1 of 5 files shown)


LukeUsher:
http://redump.org/disc/26952/
http://redump.org/disc/50471/

^example of a title that differs only in version number between regions.
Both versions of SSX3 have a title_id/serial number of EA-060

But the EU is version number 1.02, the US version is 1.01.

The same pattern exists with many games.

LukeUsher:
I know this messes up the compatibility database on Github but it's the right thing to do: Nobody can realistically visually recognize hex strings and cross-reference them with the serial no's in Redump
as an example

0x54410001 (current title_id render format) is much less readable than TA-001

But they're the exact same title_id, just rendered in the correct format.

These formats can also be cross referenced with serial numbers printing on Xbox Discs too!
For example Halo: Combat Evolved has this serial number/mastering code printed on the DVD: near the barcodes MS00402A.

That's MS-004 in the above title_id format, plus 1.02 version string, and A for region code (America)
So by printing title_ids in this new format, we can even properly tie mastering codes to game title_ids/to serial numbers. So we can actually identify which physical disc the game was from!

Sadly, there's no way to accurately derive the suffix (A) from an XBE dump, but we can match up the rest of the mastering codes.

PR #1286, opened 2018-06-14
Fix grammatical error

Voxel9:
This PR just addresses a grammatical problem in the message box that appears if a title is already currently emulating. (Should be "attempting to start again" instead of "attempt start again")


RadWolfie:
Thanks teacher. 😜

PR #1289, opened 2018-06-14
Build all branches

LukeUsher:
Just trying to get appveyor working properly for stable builds as well as master


Sarkie:
Can you do a better name for the .zips?

I have Debug(342342).zip on my box :+1:

PR #1291, opened 2018-06-15
Debugger fixes from UI side

RadWolfie:
The following fixes contains:
* Launch one Cxbx Debugger at a time. If one already exist, then it will close previous Cxbx Debugger and launch new debugger.
* Close Cxbx Debugger when Cxbx is closed.
  * Also close Cxbx Debugger when xbe is closed.
* Fixed another false positive for title is emulating when debugger is opened then close.
  * (Changed from no-risk to low-risk solution. Low-end computers may will have med-risk.)
    * This will only happen on start of emulating, during emulation has no risk for multi-xbe launches.
    * It depends on how user abuse it or choose to use some sort of bot to break it. (However, this is more of a temporary solution.)
* Use CreateProcess instead of ShellExecutable function.
* Replaced: Using std string instead of fixed char array to prevent buffer overflow.

Known issue:
* Cxbx Debugger is not aware of multi-xbe launch.

Plus universal argument to string for better maintenance in case plan to add/edit/remove arguments.

P.S. Cxbx Debugger now works for me (other than multi-xbe usage titles).

### DIFF
--- a/src/Cxbx/WndMain.cpp
+++ b/src/Cxbx/WndMain.cpp
@@ -42,6 +42,7 @@
 #include "DlgXboxControllerPortMapping.h"
 #include "Common/XbePrinter.h" // For DumpInformation
 #include "CxbxKrnl/EmuShared.h"
+#include "CxbxKrnl/CxbxKrnl.h" // For CxbxConvertArgToString
 #include "ResCxbx.h"
 #include "CxbxVersion.h"
 #include "Shlwapi.h"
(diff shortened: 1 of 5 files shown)


build/win32/Cxbx.vcxproj
```diff
@@ -376,7 +376,6 @@
     <ClInclude Include="..\..\src\devices\video\swizzle.h" />
     <ClInclude Include="..\..\src\devices\video\vga.h" />
     <ClInclude Include="..\..\src\devices\Xbox.h" />
-    <ClInclude Include="DlgXboxControllerPortMapping.h" />
```
RadWolfie: Correction: Removed old header file.

PatrickvL:
Lgtm

x1nixmzeng:
I'm happy to see work being done on this! Thanks @RadWolfie :+1:

PR #1296, opened 2018-06-19
Resolve some compile warnings

RadWolfie:
Fixes:
* Remove minimal rebuild option due to conflict with /MP option.
* Ignore libcmt instead of ignoring msvcrt, it is part of reason warning is output for two different version of same library.
  * **NOTE:** libcmt library is integrated into exe/dll. Whilst msvcrt is a module required to be installed via C++ redistribute installer.
* Removed warning C4996: 'GetVersion': was declared deprecated from DirectX8's multimon header file by using WinUser header (User32.lib).
  * It is possible to remove `#include <WinUser.h>` line. Though, I included it to know why it is there with a comment.
* MSB8004: Intermediate Directory does not end with a trailing slash.

Warnings for release build should drop down to 7.
* 1 is for GetVersion from CxbxKrnl.cpp file.
* 6 repeats are for not finding distorm.pdb file.

### DIFF
--- a/src/CxbxKrnl/Emu.cpp
+++ b/src/CxbxKrnl/Emu.cpp
@@ -42,7 +42,6 @@ namespace xboxkrnl
 };
 
 #include "CxbxKrnl.h"
-#define COMPILE_MULTIMON_STUBS
 #include "Emu.h"
 #include "EmuX86.h"
 #include "EmuFS.h"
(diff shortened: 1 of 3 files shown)


PatrickvL:
Lgtm!

src/CxbxKrnl/Emu.cpp
```diff
@@ -42,7 +42,6 @@ namespace xboxkrnl
 };
 
 #include "CxbxKrnl.h"
-#define COMPILE_MULTIMON_STUBS
```
RadWolfie: P.S. I'm not sure how this define is in Emu.cpp file. It should be in WndMain.cpp file back then. 🙄

PR #1297, opened 2018-06-19
Allow PixelShaders to be updated in Real Time via SetRenderState calls

LukeUsher:
This allows titles that use SetRenderState to dynamically modify shaders at runtime to function.

This has been tested and works well with the XDK Samples (including ModifyPixelShader which does exactly what this PR is meant to fix)

Please test this PR with games with Pixel Shader issues such as JSRF, Halo, etc.

Games with Vertex Shader issues (eg. T-pose models) are unaffected by this change.


LukeUsher:
This implementation is a temporary solution/work-around for the problem, which is relatively safe until we finally unpatch the RenderState functions like Dxbx/WIP_LessVertexPatching branch did.

LukeUsher:
Good test cases for this are titles that previously spat a large amount of SetRenderState_Simple errors/warnings to the console: As ALL pixel shader related RenderStates are now implemented.

PatrickvL:
Lgtm

PR #1299, opened 2018-06-20
Various Fixes

LukeUsher:
1. Fix a ton of conversion errors/issues with Inline Vertex Buffers (SetVertexData4f, etc): Our previous Dword to Float conversion didn't scale: It purely did casting and shifting, making it inaccurate (a byte value of 255 to ub varients should have became 1.0f, but previously it didn't, a short value of MAX_SHORT should have became 1.0f but it didn't, etc. The same issue existed with D3DDevice_SetVertexDataColor, which was setting color components with 0-255 as 0.0f - 255.0f rather than scaling between 0.0f-1.0f)

2. Unpatch GetViewPort: Calling the SetViewPort trampoline as well as fixing the issue where the viewport size should be capped to the Xbox render target dimensions mean that this is no longer necessary.

3. Replace uncap framerate hack with cap framerate: Not needed anymore for most titles + real xbox doesn't have a framerate cap either. The default state was a hack to work around timing issues. see
c1b9963 for a more thorough explanation


src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -4080,9 +4027,22 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData2s)
 
 	LOG_FORWARD("D3DDevice_SetVertexData4f");
 
-    DWORD dwA = a, dwB = b;
+	float fa = a / 32767.0f;
```
PatrickvL: Are 'a' and 'b' cast to a float before the division? Would it be better to do that explicitly?

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -4080,9 +4027,22 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData2s)
...
 
-    EMUPATCH(D3DDevice_SetVertexData4f)(Register, DWtoF(dwA), DWtoF(dwB), 0.0f, 1.0f);
+	// Special case: If the input register is a color, don't transform!
+	// Test case: Halo
+	switch (Register) {
+		case X_D3DVSDE_DIFFUSE:
+		case X_D3DVSDE_SPECULAR:
+		case X_D3DVSDE_BACKDIFFUSE:
+		case X_D3DVSDE_BACKSPECULAR:
+			fa = a;
```
PatrickvL: Perhaps do the division only in the default switch case?

src/CxbxKrnl/EmuD3D8.cpp
LukeUsher: That's what the .0.f suffix is for, it forces a float calculation

src/CxbxKrnl/EmuD3D8/PixelShader.cpp
```diff
@@ -3859,33 +3859,33 @@ bool PSH_XBOX_SHADER::FixMissingR1a()
...
-		// Make sure if we insert at all, it'll be after the DEF's :
-		if (R1aDefaultInsertPos < 0)
-			R1aDefaultInsertPos = i;
-
-		// First, check if r0.a is read by this opcode :
+		// First, check if r1.a is read by this opcode :
 		if (Cur->ReadsFromRegister(PARAM_R, 1, MASK_A))
 		{
+			// Make sure if we insert at all, it'll be after the DEF's :
+			if (R1aDefaultInsertPos < 0)
```
LukeUsher: Fixes a regression in Lego Star Wars, many others

PatrickvL: That's strange, given the next line does the exact same assignment...?

PatrickvL: Perhaps the second (old) assignment should be removed?

PR #1300, opened 2018-06-20
fix no input issue of Lego Star War.

jackchentwkh:
it requires the XGetDeviceChanges to return changes all the time, but no removal, only insertions.


jackchentwkh:
tested with Lego Star War, RalliSport, JSRF, all work fine.
waiting for further test results.

LukeUsher:
This goes against what *should* happen, but it's a valid workaround for this problematic title. I'll allow it. Basis that this will all be replaced with OHCI LLE eventually, it's not worth spending too much time on this.

PR #1303, opened 2018-06-21
Removing more unnecessary patches

LukeUsher:
These block/wait functions are not required because they don't access hardware directly, they use internal reference counts which still get updated when we call trampolines within our patches.

Get/Set back-material were unimplemented in their patches, and are safe to unpatch.


LukeUsher:
Has anybody tested this yet? I'm a little concerned the removal of blocking/waits may introduce hangs in some titles.

It *shouldn't* but if any of our patches are not updating reference counting properly, it will.

PR #1304, opened 2018-06-22
Verify User to allow run as Admin or not

RadWolfie:
When UAC is completely disabled. It enable elevated privilege all the time.

Source: https://msdn.microsoft.com/en-us/library/bb384608.aspx

Entirely not sure how exactly Chrome is able to do create process without elevated privilege.

### DIFF
--- a/src/Common/Win32/EmuShared.cpp
+++ b/src/Common/Win32/EmuShared.cpp
@@ -138,6 +138,8 @@ EmuShared::EmuShared()
 	Load();
 	m_bDebugging = false;
 	m_bMultiXbeFlag = false;
+	m_bEmulating = false;
+	m_bFirstLaunch = false;
 }
 
 // ******************************************************************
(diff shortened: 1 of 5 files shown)


PatrickvL:
LGTM

Eurose209:
Perhaps a better idea would be to just open Cxbx without admin rights and display an error telling the user that Cxbx cannot be run as admin, and that it has been opened without admin rights? Straight up making Cxbx not open when the user tries to open it just doesn't seem very intuitive to me, and it's not a behaviour I've seen before.

RadWolfie:
@Eurose209 If UAC is disabled. It is not possible directly open a program without admin rights. There is a registry edit to right-click's context menu has the ability to launch program without admin rights. Except it's not the real solution, more of a workaround solution.

Eurose209:
I see. I do like that new message box though, since it was really not being able to open Cxbx _at all_ with admin rights enabled that was strange to me.

PR #1310, opened 2018-06-24
Add new rdtsc pattern for Unreal Championship

jackchentwkh:
please don't merge it yet. 
this commit add rdtsc patterns for Unreal Championship. but I don't have time to check whether it introduces false positive in other titles or not. 
this PR is only for testers to check whether they have different behaviors with UC.


gandalfthewhite19890404:
Still crashing if you try to start match:
Received Exception Code 0xC0000005 @ EIP := 0x04070F2C

jackchentwkh:
Confirmed with PAL version of UC. 
bad dump is the root cause of crash. I can get into different play mode with or without this rdtsc patch update with a good dump of UC.
one thing interesting is that when I enter tutorial mode, and get back to menu, then start a new game, UC will crash then. but this is unrelated to rdtsc.

jarupxx:
Is 01 of WMVDEC section certainly rdtsc?
This chunk has lots of strange opcodes, perhaps this chunk is not executable, I think.

jackchentwkh:
yeah, I have similar feeling regarding to the possible rdtsc in WMVDEC. 
frankly speaking, I have no idea about it.
but whether patch it as rdtsc or not, seems the intro movies play fine.

RadWolfie:
Tester report https://github.com/Cxbx-Reloaded/game-compatibility/issues/427 for AIs are fixed. See [video comment](https://github.com/Cxbx-Reloaded/game-compatibility/issues/427#issuecomment-400027038) for reference before pull request.

Another reason we want to merge this pull request.

jarupxx:
@jackchentwkh Confirmed, Unreal 2 (U2_Retail.xbe) uses opcode pattern 01 in section .text.

PR #1317, opened 2018-06-28
More multi xbe fixes

RadWolfie:
!!!Ready to merge!!!

Contains:
* Trigger CrashMonitor thread earlier.
* Check boot flag before kernel reset from EmuShared.
  * It helps to stop load bad reboot data.
* Replace bMultiXbeFlags to BootFlags variable.
  * I can confirm bMultiXbeFlags is only used to check if xbe will reboot and BootFlags is set for this.
* More focus toward kernel mode handling. (Still incomplete anyway.)

### DIFF
--- a/src/Cxbx/WndMain.cpp
+++ b/src/Cxbx/WndMain.cpp
@@ -2349,6 +2349,9 @@ void WndMain::StartEmulation(HWND hwndParent, DebuggerState LocalDebuggerState /
         return;
     }
 
+    // Reset to default
+    g_EmuShared->Reset();
+
     // register xbe path with emulator process
     g_EmuShared->SetXbePath(m_Xbe->m_szPath);
 
--- a/src/Cxbx/WndMain.cpp
+++ b/src/Cxbx/WndMain.cpp
@@ -2463,7 +2463,7 @@ DWORD WINAPI WndMain::CrashMonitorWrapper(LPVOID lpParam)
 // monitor for crashes
 void WndMain::CrashMonitor(DWORD dwChildProcID)
 {
-	bool bQuickReboot;
+	int iBootFlags;
 	DWORD dwExitCode = 0;
 
 	// If we do receive valid process ID, let's do the next step.
@@ -2479,9 +2479,9 @@ void WndMain::CrashMonitor(DWORD dwChildProcID)
 	 		GetExitCodeProcess(hProcess, &dwExitCode);
 	 		CloseHandle(hProcess);
 
-	 		g_EmuShared->GetMultiXbeFlag(&bQuickReboot);
+	 		g_EmuShared->GetBootFlags(&iBootFlags);
 
-	 		if (!bQuickReboot) {
+	 		if (!iBootFlags) {
 	 			if (dwExitCode == EXIT_SUCCESS) {// StopEmulation
 	 				return;
 	 			}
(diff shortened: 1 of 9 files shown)


src/CxbxKrnl/EmuKrnlHal.cpp
```diff
@@ -603,8 +602,10 @@ XBSYSAPI EXPORTNUM(49) xboxkrnl::VOID DECLSPEC_NORETURN NTAPI xboxkrnl::HalRetur
 		xboxkrnl::HalWriteSMBusValue(SMBUS_ADDRESS_SYSTEM_MICRO_CONTROLLER, SMC_COMMAND_SCRATCH, 0, SMC_SCRATCH_DISPLAY_FATAL_ERROR);
 
 		char szWorkingDirectoy[MAX_PATH];
-		bool bMultiXbe = true;
-		g_EmuShared->SetMultiXbeFlag(&bMultiXbe);
+		int QuickReboot;
+		g_EmuShared->GetBootFlags(&QuickReboot);
+		QuickReboot |= BOOT_FATAL_ERROR;
```
RadWolfie: @ergo720 Since `ReturnFirmwareFatal` is a fatal error, it should be fine?

RadWolfie:
Whoops, this pull request is meant for v0.1 branch.

src/CxbxKrnl/EmuKrnlHal.cpp
ergo720: Hmm, not sure since the VMManager (which uses the quick reboot flag) is initialized before checking for the fatal error flag. Ideally, it wouldn't be necessary since the fatal error flag will stop the initialization and just display the error screen

RadWolfie: Actually, it is handled inside HalWriteSMBusValue function. So, we don't need it in here. 😄

RadWolfie:
@ergo720 this pull request resolve forced exception inside VMManager's Initialize function. Since CrashMonitor thread is created earlier than waiting for kernel mode to create a window for graphic purpose. Plus reset boot flags too (while preserved previous boot flag internally temporary).

src/Cxbx/WndMain.cpp
```diff
@@ -591,9 +589,15 @@ LRESULT CALLBACK WndMain::WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lP
...
-					 switch(lParam) {
+					 switch(HIWORD(wParam)) {
 						// NOTE: If anything need to set before kernel process start do anything, do it here.
 						case ID_KRNL_IS_READY: {
+							Crash_Manager_Data* pCMD = (Crash_Manager_Data*)malloc(sizeof(Crash_Manager_Data));
+							pCMD->pWndMain = this;
+							pCMD->dwChildProcID = lParam; // lParam is process ID.
+							HANDLE hThreadTemp = CreateThread(NULL, NULL, CrashMonitorWrapper, (void*)pCMD, NULL, NULL); // create the crash monitoring thread
+							CloseHandle(hThreadTemp);
+
```
ergo720: Just an idea to aid cross-platform: you can probably use ```std::thread(CrashMonitorWrapper, pCMD).detach();``` instead of the Windows-only ```CreateThread```

PR #1332, opened 2018-07-05
Preserve window position after stop emulation

### LINKED ISSUE #1330: Support set Main window position
Main window always appears in the middle of the desktop. Save the position last used, restore to same position as last time I hope.

RadWolfie:
Title says it all.

Only thing not tested are left and bottom monitors of main monitor. In theory, it should work as intended.

Resolved #1330

### DIFF
--- a/src/Cxbx/WndMain.cpp
+++ b/src/Cxbx/WndMain.cpp
@@ -2439,9 +2439,15 @@ void WndMain::StopEmulation()
     }
 
 	UpdateCaption();
-    RefreshMenus();
-	// Set the window size back to it's GUI dimensions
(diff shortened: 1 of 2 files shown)


RadWolfie:
Confirmed it works as intended for top, bottom, left, and right side of main monitor.

My only concern is if `scale to host and back` option isn't enable. It still re-position.

RadWolfie:
Commit ca64404 check the hack is enabled before re-position on stop emulation.

NOTE: m_ScaleViewport value can be change at any time during emulation.

PatrickvL:
Lgtm

Issue #1315, opened 2018-06-25
"Turok" No Input


Paradise12314:
Testing the latest version from "v0.1", games aren't accepting any Input. Controllers and Keyboard can be mapped in CXBX settings just fine. But once Emulation starts, nothing. "Develope" Branch is working fine though.

RadWolfie:
Our tester report it works fine by:
1. "Reset to defaults" from menu (re-open Cxbx-Reloaded afterward)
2. Then re-map the DirectInput's controller
3. Start emulation

Please try the method above then try again. Both develop and v0.1 branches are not in constant 1:1.

RadWolfie:
Hmm... I am seeing an old code still being used. XInputEnabled menu is currently removed. If we are supporting both XInput and DirectInput at same time. We should not have this option at all.

Kernel side:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/1df336fdaa938d07ede8e851538915ee2275bd54/src/CxbxKrnl/CxbxKrnl.cpp#L1278-L1283
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/1df336fdaa938d07ede8e851538915ee2275bd54/src/CxbxKrnl/EmuXapi.cpp#L292-L301
GUI side:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/1df336fdaa938d07ede8e851538915ee2275bd54/src/Cxbx/WndMain.cpp#L200-L204
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/1df336fdaa938d07ede8e851538915ee2275bd54/src/Cxbx/WndMain.cpp#L1467-L1470

Paradise12314:
I'm sorry, I was slightly wrong. It isn't a problem with CXBX, it's a problem with Turok only. Other games are accepting input. So it looks like Turok will need disecting.

Paradise12314:
Hope you don't mind, I changed the Title to reflect latest findings. I tried the latest V0.1 also. Turok still accepts no input. All other games are fine.

RadWolfie:
A tester from discord also report 2002 World Cup (FIFA?) title has the same issue too.

Paradise12314:
So we may need another Title? Also testers needed and a list of Games affected.

RadWolfie:
[...]

PR #1333, opened 2018-07-05
Convert Lego Star Wars fix into hack method

RadWolfie:
Titles tested as working:
* Crimson Skies
* FIFA 2002 World Cup (**fixed**)
* Gauntlet: Dark Legacy
* Ghost Recon
* Ghost Recon 2
* Ghost Recon: Island Thunder
* Lego Star Wars
* Lego Star Wars 2
* Mafia
* Red Faction 2
* Roller Coaster Tycoon
* Smashing Drive
* Turok Evolution (**fixed**)

### DIFF
--- a/src/CxbxKrnl/EmuXapi.cpp
+++ b/src/CxbxKrnl/EmuXapi.cpp
@@ -356,7 +356,7 @@ bool TitleIsLegoSW()
 
 	// Array of known Lego Star Wars title IDs, must be 0 terminated
 	DWORD titleIds[] = {
-		0x4553001D, // NTSC
+		0x4553001D, // v1.01 - PAL; v1.02 - NTSC
 		0
 	};
 
--- a/src/CxbxKrnl/EmuXapi.cpp
+++ b/src/CxbxKrnl/EmuXapi.cpp
@@ -344,6 +344,40 @@ bool TitleIsJSRF()
 	return result;
 }
 
+bool TitleIsLegoSW()
+{
+	static bool detected = false;
+	static bool result = false;
+
+	// Prevent running the check every time this function is called
+	if (detected) {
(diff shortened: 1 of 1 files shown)


src/CxbxKrnl/EmuXapi.cpp
```diff
@@ -344,6 +344,40 @@ bool TitleIsJSRF()
...
+	static bool result = false;
+
+	// Prevent running the check every time this function is called
+	if (detected) {
+		return result;
+	}
+
+	// Array of known Lego Star Wars title IDs, must be 0 terminated
+	DWORD titleIds[] = {
+		0x4553001D, // NTSC
```
PatrickvL: Is this another version that would need to be added to this list? : http://redump.org/disc/24495/

PatrickvL:
Looking good

src/CxbxKrnl/EmuXapi.cpp
RadWolfie: ES-029 is the same as 0x4553001D hex format. Although, v1.01 is region free?

gandalfthewhite19890404:
2002 World Cup - confirmed, works.
image

RadWolfie:
Resolved #1315

LukeUsher:
I don't like adding more hacks, but i'll include this: The HLE Input hopefully won't be around much longer with @ergo720's OHCI work going on

RadWolfie:
No input Lego Star Wars fix is a hack. What I do is convert the "hack" fix into a hack for specific title id (Lego Star Wars). Therefore, nothing is added as hack. 😋

Beside, the hack is simulating the way pull request #1300 does in one area instead of two areas. Always insertion, return true, and no removal.

PR #1334, opened 2018-07-05
LLE framebuffer drawing

PatrickvL:
This pull request changes the way the xbox framebuffer is shown : Instead of blitting it to an additional framebuffer, it's now copied into a texture, which is drawn to the host backbuffer via a normal draw.

This work shouldn't change anything but a significant speed improvement.

The code doing the framebuffer drawing is also refactored to be nearly the same as overlay drawing.
This also opens up the opportunity for another speedup (not done yet), using asynchronous DMA copies, which would improve the speed even further (at the cost of 1 frame of lag).

Happy testing!


gandalfthewhite19890404:
RAW2 - 
image

RadWolfie:
DolphinClassic

Before:
I'm not able to make a screenshot. It's basically 2 frames are flickering back and forward.

After:
Now it is keeping a copy of previous frame then apply new update on top of it.
image

Ale100000:
With this PR halo menu got a bit worse.
<details>
  <summary>Click to expand</summary>

### Before https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/5d79c0b66e58bf38d39ea28cb4de954209d1e8ad

DEV

### After https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1334/commits/8856749e47994bb4ab1f3508710b1e108d003696
PR

Eurose209:
Burnout 3 and Burnout Revenge now both only display single flickering blue/grey pixels in the very top left and near top-right, respectively, during the initial loading screen. Nothing is displayed after that.

I also observed this strange effect with Marvel vs. Capcom 2. It appears throughout gameplay, not just on the screen I've taken these screenshots on:
<details/>
  <summary>Marvel vs. Capcom 2</summary>
Before:

image

After:
image

</details>

PatrickvL:
@gandalfthewhite19890404 Could you test RAW2 again? It's color might show up correctly now.

RadWolfie:
The master's flickering (previous and current frames) vs this pull request are at same rate of flicker. If that's what you're asking.

New commit 48f3ae9 causes light gray and black background are flickering like stroke light. Other than that, same result.

gandalfthewhite19890404:
@PatrickvL tested RAW 2 on latest commit 48f3ae9 - same thing in FMVs:
image
image

PatrickvL:
Hm, I'll be needing a copy of RAW2 to debug this

gandalfthewhite19890404:
This is how it looks at current 0.1 LLE:
image

medievil1:
ok... pink video are fixed in this PR 
BTVS 
Pink
pink

And after this PR
not

gandalfthewhite19890404:
FIFA World Cup 2002 has missing graphics (during loading there should be a rotating ball with loading) and missing FMVs at this PR.

PatrickvL:
Two things to try for the missing graphics:
* Clear z buffer?
* Read display mode format from hardware register(s) somewhere?

src/CxbxKrnl/EmuKrnlAv.cpp
```diff
@@ -256,9 +256,6 @@ XBSYSAPI EXPORTNUM(2) VOID NTAPI xboxkrnl::AvSendTVEncoderOption
 	}
 }
 
-// Cached Display Mode format, used by NV2A to deermine framebuffer format
```
Margen67: *determine

PatrickvL: Thanks for the sharp eye 👍

Luckily, that typo is removed with this same commit.

LukeUsher:
Looking good for the most part, still a little slow but that's to be expected.

The new frame-buffer blit code isn't taking the frame-buffer and window size into account: The dashboard renders to a frame-buffer much larger than the output resolution, which then gets downscaled in order to implement FSAA

Old rendering path (blit):
image

New Rendering path (draw):
image

PR #1341, opened 2018-07-10
Vectored exception handling

PatrickvL:
This PR uses the Windows AddVectoredExceptionHandler() API to make sure our exception handler is handled first. This, so that the potential overhead any prior code out of our control is avoided.

Additionally, privileged instruction handling is improved; First, our RDTSC handling is slightly faster now (especially in case a privileged instruction is *not* our RDTSC patch), and second, our exception handler will also be used for privileged instructions other than RDTSC.

Compared to before, this PR : 
* shouldn't cause any regression
* shouldn't cause any slowdown
* might cause a speedup for MMIO-heavy workloads
* might improve accuracy for some privileged instructions

Happy testing!


RadWolfie:
I agree with @gandalfthewhite19890404, it brings a ton of regression than on develop/v0.1 branch. However, are the exception a real problem we need to fix or false positive?

Will need investigate more to find the source.

P.S. Only did a basic testing, not debugging the issues atm.

Cade-H:
Crazy Taxi 3 is also instantly crashing:
`Received Exception Code 0xE06D7363 @ EIP := 0x7433DB92`

src/CxbxKrnl/Emu.cpp
```diff
@@ -216,81 +216,71 @@ void EmuExceptionNonBreakpointUnhandledShow(LPEXCEPTION_POINTERS e)
...
+bool TryHandleException(EXCEPTION_POINTERS *e)
+{
+	// Make sure access-violations reach EmuX86_DecodeException() as soon as possible
+	if (e->ExceptionRecord->ExceptionCode != EXCEPTION_ACCESS_VIOLATION) {
+		switch (e->ExceptionRecord->ExceptionCode) {
+		case STATUS_PRIVILEGED_INSTRUCTION:
+			// When the g_SkipRdtscPatching hack is disabled
+			if (!g_SkipRdtscPatching) {
+				// Check if this exception came from rdtsc 
+				if (IsRdtscInstruction(e->ContextRecord->Eip)) {
```
PatrickvL: The check on g_SkipRdtscPatching should be moved into IsRdtscInstruction

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -683,8 +687,13 @@ std::vector<xbaddr> g_RdtscPatches;
 
 bool IsRdtscInstruction(xbaddr addr)
 {
-	if (std::find(g_RdtscPatches.begin(), g_RdtscPatches.end(), addr) != g_RdtscPatches.end()) {
-		return true;
+	// First the fastest check - is this a patch? (see PatchRdtsc)
+	uint8_t* opAddr = (uint8_t*)addr;
+	if ((opAddr[0] == 0xEF) // OUT DX, EAX
+	 && (opAddr[1] == 0x90)) { // NOP
```
PatrickvL: g_SkipRdtscPatching check isn't even needed here when there are no patches anyway

src/CxbxKrnl/Emu.cpp
```diff
@@ -216,81 +216,71 @@ void EmuExceptionNonBreakpointUnhandledShow(LPEXCEPTION_POINTERS e)
...
+		return true;
+	}
+
+	// Try to Report this exception to the debugger, which may allow handling of this exception
+	if (CxbxDebugger::CanReport()) {
+		bool DebuggerHandled = false;
+		CxbxDebugger::ReportAndHandleException(e->ExceptionRecord, DebuggerHandled);
+		if (DebuggerHandled) {
+			// Bypass exception
+			return false;
```
PatrickvL: Not related to this pull request, but, is this the correct return value?

src/CxbxKrnl/Emu.cpp
```diff
@@ -216,81 +216,71 @@ void EmuExceptionNonBreakpointUnhandledShow(LPEXCEPTION_POINTERS e)
...
-				EmuExceptionExitProcess();
-			}
-		}
-		else
-		{
-			// notify user
-			EmuExceptionNonBreakpointUnhandledShow(e);
-		}
+		// Kill the process immediately without the Cxbx notifier
+		EmuExceptionExitProcess();
```
PatrickvL: Add a marker below this call, indicating unreachable code. Then remove the 'else' below this line, and unindent it's statement block.

RadWolfie:
For the following titles:
* Smashing Drive title
* Crimson Skies
* Lego Star Wars
* Turok Evolution

Here's my summary result:
* Default (base on PR): Popup exception message: If click cancel, it continue on as if nothing is wrong.
* Exclude AddVectoredExceptionHandler (enable USE_SEH): No popup exception message.
---
My belief is `AddVectoredExceptionHandler` function's exception handler should only be used for EmuX86 (MMIO) read/write access violation and rdtsc patch instruction. While everything else should be done in normal exception handler from our own Cxbx code problem and breakpoints too.

Meaning split exceptions into 2 different function to handle emu trigger (by VEH) and Cxbx faults (by SEH). For optimization and less false positive report.

What do you think?

PatrickvL:
There's currently a check at the start of EmuX86_DecodeException, that does an early exit when the EIP lies outside the Xbox code range. I wonder what would happen if we move that code over to EmuException

mkultimate:
crash
image

RadWolfie:
Had been using this pull request for almost a week with my warm up reverse engineering at low level hardware registers to get audio samples to work. (Haven't gotten to work though, except Smashing Drive is able to run with small freeze up in middle of the intro. This is from my warm up hacks to gather more hardware registers information.)

- [X] shouldn't cause any regression
- [X] shouldn't cause any slowdown
- [X] might cause a speedup for MMIO-heavy workloads
- [X] might improve accuracy for some privileged instructions

src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -683,11 +683,15 @@ std::vector<xbaddr> g_RdtscPatches;
...
 {
-	if (std::find(g_RdtscPatches.begin(), g_RdtscPatches.end(), addr) != g_RdtscPatches.end()) {
-		return true;
-	}
-
-	return false;
+	// First the fastest check - is this a patch? (see PatchRdtsc)
+	uint8_t* opAddr = (uint8_t*)addr;
+	// Note : Check second opcode first, as that's most likely to fail fast
+	return (opAddr[1] == 0x90) // NOP
```
RadWolfie: I think checking second opcode first is fine. Since we don't usually see NOP often in middle of a function.

Although, there is a way to check both opcode with less cpu instructions by using `*(uint16_t*)` cast and 0x90EF compare check.

Unoptimized code real sample:
**This commit in asm code**
```
   681: 	uint8_t* opAddr = (uint8_t*)addr;
8B 45 08             mov         eax,dword ptr [addr]  
89 45 F0             mov         dword ptr [opAddr],eax  
   683: 	// Note : Check second opcode first, as that's most likely to fail fast
   684: 	return (opAddr[1] == 0x90) // NOP
B8 01 00 00 00       mov         eax,1  
C1 E0 00             shl         eax,0  
8B 4D F0             mov         ecx,dword ptr [opAddr]  
0F B6 14 01          movzx       edx,byte ptr [ecx+eax]  
81 FA 90 00 00 00    cmp         edx,90h  
0F 85 7B 01 00 00    jne         IsRdtscInstruction+1AFh (0419A24Fh)  
	685:	&& (opAddr[0] == 0xEF) // OUT DX, EAX
B8 01 00 00 00       mov         eax,1  
6B C8 00             imul        ecx,eax,0  
8B 55 F0             mov         edx,dword ptr [opAddr]  
0F B6 04 0A          movzx       eax,byte ptr [edx+ecx]  
3D EF 00 00 00       cmp         eax,0EFh  
...
```
**My proposal change** 😎
```
   680: 	// First the fastest check - is this a patch? (see PatchRdtsc)
   681: 	return (*(uint16_t*)addr == 0x90EF) // OUT DX, EAX; NOP
8B 45 08             mov         eax,dword ptr [addr]  
0F B7 08             movzx       ecx,word ptr [eax]  
81 F9 EF 90 00 00    cmp         ecx,90EFh  
0F 85 55 01 00 00    jne         IsRdtscInstruction+170h (0419A210h)
...
```

RadWolfie: For optimized asm code, it may squash 3 instructions into one (or two).

PR #1343, opened 2018-07-17
Run with limited privileges by default instead of asking evey time

ergo720:
Tested with Windows 7 and confirmed as working (the child process also inherits the limited privileges as well). Better also testing this with Windows 10 since, according to the original StackOverflow question, this could not work with that OS

### DIFF
--- a/src/Common/Win32/InlineFunc.cpp
+++ b/src/Common/Win32/InlineFunc.cpp
@@ -37,6 +37,8 @@
 #if defined(_WIN32) || defined(WIN32)
 
 #include "..\..\CxbxKrnl\CxbxKrnl.h"
+#include <WinSafer.h> // For the Safer API functions
+#include <sddl.h> // For ConvertStringSidToSid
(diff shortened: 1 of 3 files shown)


PatrickvL:
How does this guarantee the new launched process is no longer elevated? (Because if the new process would still be elevated, this change potentially introduces an endless restart loop...)

ergo720:
Ok, so because it doesn't work with Windows 10 then I think it's pointless to insist with this code. Instead, can we add an option in the GUI (stored in the registry) to allow the user to always ignore the warning if he/she doesn't care about running with admin rights instead of warning every time?

RadWolfie:
In my opinion, we should have both. Attempt to leave elevated process if possible and disable warning message of running on elevated process.

RadWolfie:
P.S. If one Cxbx-Reloaded is already running. You can launch other Cxbx-Reloaded without popup warning message.

PR #1345, opened 2018-07-24
Thread safe FS processing using spinlock

LukeUsher:
Addresses issue #986.

This will probably have no noticeable improvement, but could improve stability in some cases.

PR #1346, opened 2018-07-24
Revert "Thread safe FS processing using spinlock"

LukeUsher:
Reverts Cxbx-Reloaded/Cxbx-Reloaded#1345
Caused a hang in World Cup 2002, needs investigating.


LukeUsher:
Actually, it may have been a false alarm
@gandalfthewhite19890404  can this be reliably reproduced? if not, this PR can be closed

LukeUsher:
Been told this is a false alarm,

PR #1348, opened 2018-07-24
Unset DepthStencil flag on ARGB conversion

LukeUsher:
ARGB isn't a valid stencil format, this is the cause of CreateTexture errors in some titles.

With this fix, Virtua Cop 3 (Chihiro title) boots again, but sadly still crashes before going in-game.

This has potential to fix the many titles that currently show CreateTexture errors, please test and report back,


LukeUsher:
I noticed an argb creation failed and I know that shouldn't be possible as all modern GPUs support that format, so then I stepped through the CreateHostResource process, checking for any special flags that got set

PR #1350, opened 2018-07-25
Detect DepthFormat is supported and fallback if not

LukeUsher:
Fixes a recent regression preventing the dashboard from working, while retaining AutoDepthBuffer support (this feature was previously broken)


LukeUsher:
Confirmed working

PR #1351, opened 2018-07-26
Fix an issue with Vertex Shaders containing EXP instruction

LukeUsher:
Fix an issue where Vertex Shaders containing exp instructions without a swizzle failed to compile.

Can't remember why I didn't PR this sooner: Make sure to test for regressions and report back.


LukeUsher:
Update: It's fine, the reason I didn't merge earlier is because I was still investigating other (unrelated) vertex shader issues. This can be merged for now, unrelated shader issues can be addressed in later PRs

PR #1357, opened 2018-07-31
VS 2015 Compile fix

gandalfthewhite19890404:
Visual Studio 2015 compile fixes


gandalfthewhite19890404:
For some reasons github put too many edits to VertextShader.cpp - there should be only two edits.

LukeUsher:
Probably just a line-ending change,

I'll accept this, but keep in mind Visual Studio 2015 is not officially supported, we only support use of the latest toolchain (Visual Studio 2017 at time of writing).

It doesn't matter right now, but Visual Studio 2015 doesn't support the latest C++17 standard, and even has some functionality missing (constexpr) from C++14, if we wish to make use of these new language features, it would become impossible to support 2015.

RadWolfie:
I can confirm it is line-ending change. Github editor is converting LF to CRLF as expected by .editorconf's setting.

PR #1365, opened 2018-07-31
LLE USB - disabled

ergo720:
This PR adds an LLE USB implementation (OHCI, Hub and Xid Duke gamepad) in a **disabled** state, because it doesn’t work yet. I compared my ohci register values with those of XQEMU after the first reset and device attachment and they are identical, so I believe it’s kernel related at this point (but feel free to prove me wrong). Also, some games crash when xinput is unpatched (JSRF). So, instead of waiting for kernel improvements, I decided to PR this now. I didn’t touch the HLE input, so this will have no effect there.


src/CxbxKrnl/CxbxKrnl.cpp
```diff
@@ -648,7 +654,7 @@ void TriggerPendingConnectedInterrupts()
 {
 	for (int i = 0; i < MAX_BUS_INTERRUPT_LEVEL; i++) {
 		// If the interrupt is pending and connected, process it
-		if (HalSystemInterrupts[i].IsPending() && EmuInterruptList[i] && EmuInterruptList[i]->Connected) {
+		if (g_bEnableAllInterrupts && HalSystemInterrupts[i].IsPending() && EmuInterruptList[i] && EmuInterruptList[i]->Connected) {
```
LukeUsher: Why not use g_bEnableAllInterrupts in the outer loop? No point iterating through interrupts when none will possibly be called anyway

src/devices/EmuNVNet.h
```diff
@@ -35,13 +35,11 @@
...
 #include "PCIDevice.h" // For PCIDevice
 
-#define NVNET_ADDR  0xFEF00000 
-#define NVNET_SIZE  0x00000400
 
 class NVNetDevice : public PCIDevice {
 public:
 	// PCI Device functions
-	void Init();
+	void Init(unsigned int address);
```
LukeUsher: Putting the initial address here is possibly a bad idea: Device addresses can (and are) set at run-time by the real bootloader/kernel when that is used. It's also possible for software to write to BAR registers to change the address at any time: For example: The boot loader does this to remap the NV2A to a different memory address during early boot, when 0xF0000000 isn't accessible yet.

These defaults only matter for our kernel, where the bootloader never writes to the initial bar registers.

src/devices/EmuNVNet.cpp
```diff
@@ -514,14 +514,14 @@ void EmuNVNet_Write(xbaddr addr, uint32_t value, int size)
...
+void NVNetDevice::Init(unsigned int address)
 {
 	PCIBarRegister r;
 
 	// Register Memory bar :
 	r.Raw.type = PCI_BAR_TYPE_MEMORY;
-	r.Memory.address = NVNET_ADDR >> 4;
-	RegisterBAR(0, NVNET_SIZE, r.value);
+	r.Memory.address = address >> 4;
+	RegisterBAR(0, NVNet_SIZE, r.value);
```
LukeUsher: NVNET_SIZE was caps because it was a constant/define, not a variable.
Constants should always be caps to easily identify them as such

src/CxbxKrnl/CxbxKrnl.cpp
ergo720: Because if the thread is executing the for loop and the CLI instruction is executed, then enabled interrupts will still be fired

src/devices/EmuNVNet.h
ergo720: I added the address variable because I needed to pass two different MMIO addresses to the initialization code for USB0 and USB1, but now USB1 is gone so I can revert this

ergo720:
I removed the SDL2 submodule because it's only used by the LLE, which is disabled so it's useless at the moment. I completely commented out the SDL code so this will build fine

ergo720:
A word of caution about the InputDeviceManager: the current implementation has the button bindings hard-coded, which can lead to wrong buttons being signaled and asserts being triggered depending on the controller you use. This should be fixed with LLE-specific input window in the GUI which lets the users mapping the buttons to the buttons of the emulated xbox controller. The idea is that the sdl buttons pressed will be captured in the sdl event loop and then translated to the xbox buttons with the Axis/ButtonMap array in the SDL2Devices class, but all of that is not implemented at the moment.

RadWolfie:
Warning counts has increased from 11 to 26 in this pull request, see below.
```
c:\projects\cxbx-reloaded\src\cxbxkrnl\timer.cpp(69): warning C4244: 'argument': conversion from 'uint64_t' to 'uint32_t', possible loss of data [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\cxbxkrnl\vmmanager.cpp(1726): warning C4805: '!=': unsafe mix of type 'XbeType' and type 'bool' in operation [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\hub.cpp(381): warning C4244: '=': conversion from 'uint16_t' to 'uint8_t', possible loss of data [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\hub.cpp(383): warning C4244: '=': conversion from 'uint16_t' to 'uint8_t', possible loss of data [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\hub.cpp(525): warning C4018: '>': signed/unsigned mismatch [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\ohci.cpp(438): warning C4018: '>': signed/unsigned mismatch [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\ohci.cpp(467): warning C4018: '>': signed/unsigned mismatch [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\ohci.cpp(1279): warning C4244: 'argument': conversion from 'uint64_t' to 'uint32_t', possible loss of data [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\ohci.cpp(1764): warning C4018: '<=': signed/unsigned mismatch [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\usbdevice.cpp(204): warning C4244: '=': conversion from 'uint64_t' to 'uint32_t', possible loss of data [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\usbdevice.cpp(336): warning C4244: '=': conversion from 'uint64_t' to 'uint8_t', possible loss of data [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\usbdevice.cpp(453): warning C4018: '>': signed/unsigned mismatch [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\usbdevice.cpp(490): warning C4018: '>': signed/unsigned mismatch [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\usbdevice.cpp(1049): warning C4018: '>': signed/unsigned mismatch [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
c:\projects\cxbx-reloaded\src\devices\usb\usbdevice.cpp(1252): warning C4267: '=': conversion from 'size_t' to 'uint8_t', possible loss of data [C:\projects\cxbx-reloaded\build\win32\Cxbx.vcxproj]
```

ergo720:
I know, the 2nd one is actually an error that I'm already fixing (it should be xtRetail) and all the others are inherited from XQEMU. Most of those are because many functions return -1 to signal failure and so the returned variable needs to be a signed integer. However, that same variable is also used in the function to do comparison, hence the warnings of the compiler. They can probably be solved by using std::optional however at the moment I'm not sure if that doesn't break anything in the process.

s2r222:
This regressed Fatal Frame 1 from state booting to blackscreen: Game Compatibility thread: https://github.com/Cxbx-Reloaded/game-compatibility/issues/85

ergo720:
I found the root cause: it's because the game wants to commit at address 0x1000000 with NtAllocateVirtualMemory but it cannot since that address is occupied by our memory placeholder. The build before this erroneously allowed it (which means the game was writing in the memory placeholder) but in this PR I fixed the implementation of CheckConflictingVMA which is now detecting it and denying the allocation. This is another problem that requires LLE CPU to solve properly.

LukeUsher:
@ergo720 is there a reason you had to make that change? Xbox titles can and should be able to write to our virtual memory placeholder region: The whole point of that reason is to allocate the memory range between 0x10000 to the 64MB mark. 

Due to Windows limitations, we cannot reserve this memory region without a place-holder, but as far as the VMA is concerned, all regions that don't conflict with an XBE section VirtualAddress should NOT be shown as allocated in the VMA. Allocations in the PlaceHolder should be allowed, so long as there is not already a conflicting Xbox allocation at that virtual address (Xbe section, or previous allocation through VMA system)

Issue #1342, opened 2018-07-13
About - Contributors (Error) 


CakeLancelot:
**Originally reported by @Mored1984
Copied from Cxbx-Reloaded/game-compatibility#846 (miscategorization)**

### Description

The text is not showing right, see the picture below.

### Screenshots

![image](https://user-images.githubusercontent.com/5413459/42703665-e0815282-86d5-11e8-8060-4f9da177ad65.png)

Must be like the below picture

![image](https://user-images.githubusercontent.com/5413459/42703842-68bb66ba-86d6-11e8-8ecb-7fc27564cfb8.png)

### Additional Information

Why not show the names of patreon ?

CakeLancelot:
Why not show the names of patreon ?

They are shown under supporters, though the list has not been updated in some time. 

Hopefully the GUI will be reworked soon in v0.2, which as far as I've heard is soon to start development.

LukeUsher:
Feel free to submit a PR to address this: It's probably a line-ending issue (if the contributers file got saved as Unix line-endings instead of Windows, it won't render right by the dumb dialog).
Kinda shitty that it ended up going live like this, but it's not a big deal.

I'll probably do a 0.1.1 minor hotfix to address this when I get back from holiday. (Including making sure the list is accurate/up-to-date)

LukeUsher:
Rather than attempt to force appveyor to use a specific encoding, it would be better to update the GUI to not depend on encoding.

If you iterate through the string and convert LF to CRLF before rendering in the dialog, that will solve the problem.

Relevant code: 
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/fc908b615b95ec34890eba3c28d9da416e37bed4/src/Cxbx/DlgAbout.cpp#L108

PR #1369, opened 2018-08-02
Fix for issue #1342

ergo720:
It also applies the same fix to the copyright tab, since it was affected by the same problem (LF line endings instead of CRLF).

### DIFF
--- a/src/Cxbx/DlgAbout.cpp
+++ b/src/Cxbx/DlgAbout.cpp
@@ -39,6 +39,7 @@
 #include "CxbxVersion.h"
 #include "DlgAbout.h"
 #include "ResCxbx.h"
+#include "CxbxCommon.h"
 
 #include <commctrl.h>
 #include <string>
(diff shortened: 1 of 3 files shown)


src/Cxbx/DlgAbout.cpp
```diff
@@ -134,6 +151,23 @@ INT_PTR CALLBACK DlgAboutProc(HWND hWndDlg, UINT uMsg, WPARAM wParam, LPARAM lPa
 				SizeofResource(GetModuleHandle(NULL), rCopying)
 			);
 
+			position = 0;
+			while (true) {
```
PatrickvL: `while(position > copying.length())` !?

ergo720: I thought it was necessary because of the ```position += 2``` done at the end, which could mean that position could exceed the string length when the loop reaches the end of the string, but debugging I discovered that position becomes equal to the string length and then ```find``` returns ```npos``` and breaks the cylce, so we can remove this check

RadWolfie: Whenever a find function cannot find a character at the beginning or the offset after given position, it always return npos.

This is for whoever may counter the above comments' discussion to understand the situation.

PR #1371, opened 2018-08-03
Fix a crash in the dashboard

ergo720:
Fixes a crash on the dashboard caused by an assertion triggered only on debug builds. The problem was that the char passed to isalnum expects characters to be in the range -1 – 255, and so they must be unsigned, otherwise they can be out of range depending on the value being passed (0xFF and 0xFE in the case of the dashboard). BTW, this is also documented in the docs of the function: “Like all other functions from cctype, the behavior of std::isalnum is undefined if the argument's value is neither representable as unsigned char nor equal to EOF. To use these functions safely with plain chars (or signed chars), the argument should first be converted to unsigned char:” 

dash_assert
dash_crash

### DIFF
--- a/src/CxbxKrnl/Emu.cpp
+++ b/src/CxbxKrnl/Emu.cpp
@@ -84,8 +84,8 @@ std::string FormatTitleId(uint32_t title_id)
 	// EG: MS-001 for 1st tile published by MS, EA-002 for 2nd title by EA, etc
 	// Some special Xbes (Dashboard, XDK Samples) use non-alphanumeric serials
 	// We fall back to Hex for those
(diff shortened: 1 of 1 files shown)


LukeUsher:
Does the dashboard title render as expected with this (as hex) or weird unicode characters instead? The intent of this was to check for ASCII-printable characters only: It might be a better solution all-around to just check if the characters are in the range of 65 (Ascii A) to 90 (Ascii Z)

ergo720:
Yeah, it prints the unicode characters 254 and 255 (þ and ÿ) so I need to fix also that

After
latest-dev
Before
prev-dev

PR #1372, opened 2018-08-03
Fix unicode characters printed in the dashboard title

ergo720:
Fixes unicode characters being printed in the dashboard

dash
doa3


PatrickvL:
Thanks!

PR #1373, opened 2018-08-04
XcUpdateCrypto implementation + refactored kernel crypto functions

ergo720:
Tested with the dashboard and the downloader xbe of Splinter Cell (which were crashing because of an error in the 1st commit) and confirmed as working.


ergo720:
I tested all my games and they work like before. A couple use the SHA functions and Tenchu and Halo use the RC4 functions. We also use ourselves HMAC (in the eeprom and the title keys code) so I'm rather confident this doesn't break anything. Unfortunately, I couldn't find a single one which calls XcUpdateCrypto so that's untested.

PatrickvL:
Good job, as always. Thanks!

Issue #1361, opened 2018-07-31
D3D9 Port: Use Vertex Shader Model 2.0 or higher


LukeUsher:
This will allow the use of more instruction slots and registers, allowing Xbox vertex shaders to be more accurately converted to the host. 

This has potential to solve a massive amount of rendering issues, from broken polygons, missing animations, t-pose models to crashes.

As with Pixel Shaders, Shader Model 2.X is preferred over 2.0. 3.0 would be better, but is more work to implement.

**Note**:  This issue depends on the completion of #1359, which in-turn, depends on #1358 

Resources:
https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/dx9-graphics-reference-asm-vs

ObiKKa:
Is this new PR #1377 directly relevant to this milestone issue?

Also, thanks to all of the contributors here for helping to improve this emulator even while I was looking at the updates to many other emulators.

LukeUsher:
Yeah, PR #1377 addresses this issue by moving to Shader Model 2.0 for Vertex Shaders

PatrickvL:
Close this now?

Issue #1360, opened 2018-07-31
D3D9 Port: Use Pixel Shader Model 2.0 or higher


LukeUsher:
This will allow the use of more instruction slots and registers, allowing Xbox pixel shaders to be more accurately converted to the host. 

Shader model 2.X preferred over 2.0, 3.0 would be better, but is more work to implement.

**Note**: This issue is stand alone, it doesn't depend on any other issue in this project, and no other issue in this project depends on this one

Resources:
https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/dx9-graphics-reference-asm-ps

Edit PatrickvL : This is currently assigned to revel8n

PatrickvL:
Can we close this now?

LukeUsher:
No, only vertex shaders are using Shader Model 2.0 at the moment

bass3:
Can you finish implementation of Hardware TnL and Hardware VS? I can't seem to boot Unreal Championship at all anymore...

LukeUsher:
As of #1400 we have been using shader model 2.0 for Pixel Shaders too, so this issue can be closed

@bass3 not sure what you're asking? Vertex Shaders and TnL has ALWAYS been hardware accelerated, we've never had support for software rendering of any kind.

bass3:
Exactly what I am reading about D3D9 and TnL when its creating device, also XBOX is just weird when it comes to this... also here is full leaked code which will help you, cause it includes Unreal C 1 code including libraries for XBOX including nv23a, the site is not mine, so idgf:

http://talonbrave.info/files/software/games/unreal_tournament_2003/source/

LukeUsher:
@bass3 The game crashes because it is LTCG and we don't have a working patch for the LTCG version of D3DDevice_Create, that's all: LTCG games are problematic with HLE and there is nothing we can do to fix this.

Unreal Championship 2 will likely require LLE to be playable.

Also please refrain from posting illegal content. You put us and the project at serious risk by doing that.

PR #1375, opened 2018-08-08
Initial D3D9 port

LukeUsher:
This is the initial port of Cxbx-Reloaded to DirectX 9.

Notes:
1. The emulator still uses shader model 1.X at this point, migration to a newer shader model will be done in the next few weeks. (This is the main motivation for this port)
2. There will be very few improvements. Some Vertex Shaders that previously failed to compile now work, but the changes are minor at this point. We'll see major improvement once #1360 and #1361 are implemented.
3. There may be some regressions compared to develop: These will be ironed out as they are identified.

I'm going to merge this within the next few minutes: This isn't normal procedure, but we want this to be as well-tested as possible: The best way to do this is to make it easily available as an automated build.

Those who want stability should use the Release builds, rather than Dev builds, but those who want to test/experiment with the latest/greatest (like this feature!) can use Develop.


src/Common/Settings.cpp
```diff
@@ -192,24 +191,32 @@ bool Settings::Init()
...
 		else {
 			return false;
 		}
 #endif
+		saveFile.append(szSettings_settings_file);
 
-		bRet = Save(saveFile);
+		// If the config file does not exists, create a blank one
+		// We can't call Save here because that overrides default values with false!
+		if (!std::experimental::filesystem::exists(saveFile)) {
```
RadWolfie: This should not *have* experimental. There is a macro using experimental as std::filesystem.

src/Common/Settings.cpp
```diff
@@ -58,9 +59,8 @@ Settings* g_Settings = nullptr;
 
 #define szSettings_setup_error "ERROR: Either setup have a problem or do not have write permission to directory."
 #define szSettings_init_error "ERROR: Unable to initialize Settings class."
-#define szSettings_save_user_option_message "If you want to save your settings in current/portable directory,\nclick 'Yes'." \
-                                            "\n\nIf you want to store your settings in user profile directory,\nclick 'No'." \
-                                            "\n\nClicking cancel will abort Cxbx-Reloaded."
+#define szSettings_save_user_option_message "Use Cxbx-Reloaded in Portable Mode?" \
```
RadWolfie: The reason for depth explanation is so user from different country can understand it.

src/Common/Settings.cpp
```diff
@@ -192,24 +191,32 @@ bool Settings::Init()
...
 			return false;
 		}
 #endif
+		saveFile.append(szSettings_settings_file);
 
-		bRet = Save(saveFile);
+		// If the config file does not exists, create a blank one
+		// We can't call Save here because that overrides default values with false!
+		if (!std::experimental::filesystem::exists(saveFile)) {
+			std::ofstream ofs(saveFile, std::ofstream::out);
```
RadWolfie: Why ofstream?

src/Common/Settings.cpp
LukeUsher: hmm, well that macro didn't work here anyway: Once filesystem stops being experimental, we'll remove the experimental namespace.

src/Common/Settings.cpp
LukeUsher: fopen is a C function and unsafe, std::filesystem does not have "touch"/"create file" function, so we use an output filestream instead.

mkultimate:
image

src/Common/Settings.cpp
RadWolfie: Also, I know why everything is set to blank/false. I removed the default values from a structure for not to re-set everything twice. One for default constructor, and second time for set the values.

MarcoEstevez:
RalliSport Challenge - regresed

image

src/Common/Settings.cpp
```diff
@@ -192,24 +191,32 @@ bool Settings::Init()
...
+			ofs << "\n";
+			ofs.close();			
+		}
 
-		// Check if saving a file is a success.
+		// Call LoadUserConfig, this will load the config, applying defaults for any missing fields
+		bRet = LoadUserConfig();
+
+		// Check if loading the file is a success.
+		bRet = true;
```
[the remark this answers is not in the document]
RadWolfie: Ignore this, now it is checking if initialize default value is working correctly.

RadWolfie:
Burnout title has a regression on boot.
```
Unhandled exception at 0x646C5853 (d3d9.dll) in Cxbx.exe:
0xC0000005: Access violation reading location 0x00000024.
```
Locate in CxbxKrnl/EmuD3D8.cpp, line 7306
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/99fb143a04aa03182311be54483498101a552d07/src/CxbxKrnl/EmuD3D8.cpp#L7306

I'm not sure what 0x24 is. Here's the argument values:
* ? (inline function)
* 0x00000002 = DrawContext.dwHostPrimitiveCount
* 0x07420530 = DrawContext.pHostVertexStreamZeroData
* 0x0000001C = DrawContext.uiHostVertexStreamZeroStride

PR #1377, opened 2018-08-09
Vertex shader improvements

revel8n:
Updates vertex shader compilation to use version 2.x
- Maximum register and instruction counts increased

Fixes for vertex shader conversion
- Instructions restricted to single swizzle component use correct swizzle masks
- Write masks excluded for oPts register
- Checks to prevent buffer overruns and access violations during token traversal


LukeUsher:
Merging as this is at a relatively stable state. More tweaks and fixes can happen as additional PRs.
This shows improvements in most of the titles I tried, some titles are dramatically better!

PR #1384, opened 2018-08-12
D3D: Fix vertex diffuse colors not initializing properly

Voxel9:
When a title renders a triangle/quad list, sometimes the diffuse value of vertices after the first processed vertex may be completely uninitialized (unintentionally causing the diffuse color to be the default 0xCDCDCDCD value).
This PR issues a fix that causes every vertex to use the diffuse color from the last processed vertex (unless it's the first one in the list, or the first vertex after a newly-defined vertex color after calling SetVertexColor.

Before
=====
untitled

After
===
image

EDIT: This also fixes JSRF shadows too:

after

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -4075,7 +4075,7 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData4f)
 	LOG_FUNC_ARG(d)
 	LOG_FUNC_END;
 
-    HRESULT hRet = D3D_OK;
+	HRESULT hRet = D3D_OK;
 
 	// Grow g_InlineVertexBuffer_Table to contain at least current, and a potentially next vertex
 	if (g_InlineVertexBuffer_TableLength <= g_InlineVertexBuffer_TableOffset + 1) {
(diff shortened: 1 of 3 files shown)


RadWolfie:
Your proposal is nice. Except I don't think this is the correct method for two reasons.
1. Using global variable (you can use static variable inside the function without the need for global variable)
2. Toggling doesn't seem to be logical for me, since it doesn't contain default value.

In my opinion, the issue came from here: (since offset 1 doesn't have initialize values whilst offset 0 do)
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/46f900f88c83ea8a61198a90834f198f21bf6e56/src/CxbxKrnl/EmuD3D8.cpp#L4143-L4146

If we change it to...
```c
static UINT uiPreviousOffset = 0;
g_InlineVertexBuffer_Table[g_InlineVertexBuffer_TableOffset] = g_InlineVertexBuffer_Table[uiPreviousOffset];
```

Then copying from default (with any changes) will be functional. Base on line 4097: (unless had been changed by the title.)
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/46f900f88c83ea8a61198a90834f198f21bf6e56/src/CxbxKrnl/EmuD3D8.cpp#L4092-L4100

PatrickvL:
@Voxel9 good find, and a creative solution, so a warm thanks at first. I do agree with @RadWolfie though, that there's probably a better solution. But to find it, the problem has to be understood completely. One of the things we need to be aware of, is that we cannot know in advance which vertex attribute is set first at offset zero. (And btw, zero must be usable as a valid index value.) So if color is set before position, the initializing copy of the previous vertex over the current vertex would overwrite this color. What's probably best, is to assign default values to all vertex attributes at the same time when the vertex table index is initialized/reset to zero.

PatrickvL:
Oh ps: these persistent flags Radwolfie mentions might indicate that the default values for the corresponding attributes must be read from somewhere (not sure from where exactly)

RadWolfie:
From discord chat:
> RadWolfie: Voxel, I left a remark on your pull request. In theory, it should not affect all titles. Plus I did verify with JSRF with my change works. Except I'm not able to reproduce the cartoon sample.

>Voxel: I find it weird you cannot reproduce the cartoon sample. And I appreciate the remarks you made; I understand that the project is trying to go as global-free as possible, but I couldn’t think of any other way at the time.

>RadWolfie: Yeah, even in debug build too. :/
Can you try my change if it do work on cartoon sample?

>Voxel: I did, and it didn’t work

Apparently my proposal change didn't work for cartoon sample. Which is even more weird.

Voxel9:
Global variable removed.
As for the way I did things, feel free not to merge it if you feel that it doesn't tackle the exact root of the problem for now. I'd completely understand. I have basically gone down a bit of a hacky-ish route here anyway...

PatrickvL:
I've merged it, as at least it solves a problem without introducing too much confusion. Please create an issue to track the underlying problem with, so we won't forget. (And while you're at it, maybe you want to mention we currently also lack sorry for those persistent attribute flags?)

RadWolfie:
Cartoon sample might be resolve with commit https://github.com/RadWolfie/Cxbx-Reloaded/commit/6bc4b206eb71aa98014eb9ffa3dcd4ea1e3b77c3. JSRF still show as working shadow with mentioned commit link. I am currently looking for titles using 5xxxx build having problem with color/shader which may be resolve from linked commit.

PatrickvL:
Follow-up issue created here : https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1385

Issue #1385, opened 2018-08-12
Properly initialize vertex attributes


PatrickvL:
In https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1384 @Voxel9 fixed an issue with the diffuse color attribute (it wasn't initialized properly before, causing random color gradients in XDK Cartoon sample, and white shadows in JSRF).

Besides the diffuse color attribute, other vertex attributes also need correct initialization.
* Specular
* Back diffuse (not supported yet)
* Back specular (not supported yet)
(TODO : specify all others here too.)

Another thing we don't support yet, are the persistent vertex attribute usage flags:
* D3DUSAGE_PERSISTENTDIFFUSE
* D3DUSAGE_PERSISTENTSPECULAR
* D3DUSAGE_PERSISTENTBACKDIFFUSE
* D3DUSAGE_PERSISTENTBACKSPECULAR
The default values for the corresponding attributes must be read from somewhere (not sure from where exactly).

This issue will be used to keep track of these problems, to solve them, to gather a set of test-cases, and to keep communication about these problems centralised.

LukeUsher:
Fixed by #1386

PR #1386, opened 2018-08-13
Set vertex data fixes

PatrickvL:
This builds upon and extends RadWolfie's solution, fixing #1385.

Test case : BeginPush XDK sample.

(One path is untested, thus marked with LOG_TEST_CASE)

### DIFF
--- a/src/CxbxKrnl/EmuD3D8.cpp
+++ b/src/CxbxKrnl/EmuD3D8.cpp
@@ -4100,7 +4100,6 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData4f)
 	}
 
 	int o = g_InlineVertexBuffer_TableOffset;
-	static bool m_DiffuseSet = false;
 	uint FVFPosType = g_InlineVertexBuffer_FVF & D3DFVF_POSITION_MASK;
 
 	switch(Register)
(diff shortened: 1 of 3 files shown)


src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -4093,14 +4096,48 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData4f)
...
+			X_D3DVertexShader *pXboxVertexShader = VshHandleToXboxVertexShader(g_CurrentXboxVertexShaderHandle);
+			if (!(pXboxVertexShader->Flags & 0x10/*=X_VERTEXSHADER_PROGRAM*/)) {
+				ActiveVertexAttributeFlags = pXboxVertexShader->Flags;
+			}
+		}
+
+		// Handle persistent vertex attribute flags, by resetting non-persistent colors
+		// to their default value (and leaving the persistent colors alone - see the
+		// "Copy all attributes of the previous vertex" comment below) :
+		static const uint32_t ColorBlack = D3DCOLOR_ARGB(0, 0, 0, 0);
```
RadWolfie: Incorrect, it is suppose to be 255 (Alpha), 0 (Red), 0 (Green), 0 (Blue). That's the correct define as black.

P.S. If setting Alpha to 0, then any color is not visible.

src/CxbxKrnl/EmuD3D8.cpp
```diff
@@ -4093,10 +4096,45 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData4f)
...
-		// TODO : Handle D3DUSAGE_PERSISTENTDIFFUSE, D3DUSAGE_PERSISTENTSPECULAR,
-		// D3DUSAGE_PERSISTENTBACKDIFFUSE, and D3DUSAGE_PERSISTENTBACKSPECULAR 
+
+		// Get the vertex shader flags (if any is active) :
+		uint32_t ActiveVertexAttributeFlags = 0;
+		if (VshHandleIsVertexShader(g_CurrentXboxVertexShaderHandle)) {
+			LOG_TEST_CASE("D3DDevice_SetVertexData4f with active VertexShader");
+			X_D3DVertexShader *pXboxVertexShader = VshHandleToXboxVertexShader(g_CurrentXboxVertexShaderHandle);
+			if (!(pXboxVertexShader->Flags & 0x10/*=X_VERTEXSHADER_PROGRAM*/)) {
+				ActiveVertexAttributeFlags = pXboxVertexShader->Flags;
```
RadWolfie: 👍 I don't think I would ever find out how to use the current shader into a "X_D3DVertexShader" pointer. :wink:

src/CxbxKrnl/EmuD3D8.cpp
PatrickvL: True, except D3D seems to set zero too

RadWolfie:
Since it is getting accurate. I'm merging this pull request.

PR #1393, opened 2018-08-16
Persist memory layout across quick reboots

ergo720:
This fixes a crash that would occur when the rebooted xbe type is different than the one that performed the reboot. This works by choosing a memory layout when the xbe is first launched (based on its type) and keeping it across quick reboots, like the real kernel does (it doesn't switch memory layout like we were doing before). I tested this with my games and they work like before but it's better to test this a bit more (especially multi-xbe titles).


LukeUsher:
Thanks, I'll test this shortly

PatrickvL:
Lgtm

PR #1394, opened 2018-08-16
More performance tuning

LukeUsher:
None of this is significant, but it's a much-needed cleanup and saved 1-5fps for me in some titles.

### DIFF
--- a/src/CxbxKrnl/CxbxKrnl.cpp
+++ b/src/CxbxKrnl/CxbxKrnl.cpp
@@ -195,14 +195,7 @@ void SetupPerTitleKeys()
 
 void CxbxLaunchXbe(void(*Entry)())
 {
-	__try
-	{
-		Entry();
-	}
-	__except (EmuException(GetExceptionInformation()))
-	{
-		EmuLog(LOG_PREFIX, LOG_LEVEL::WARNING, "Problem with ExceptionFilter");
-	}
+	Entry();
 }
 
 // Entry point address XOR keys per Xbe type (Retail, Debug or Chihiro) :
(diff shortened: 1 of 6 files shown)


LukeUsher:
@PatrickvL it depends on the case: SwitchToThread is better for us in most cases, but in this loop, Sleep is good because this loop spends most of it's time idle anyway: No point burning cycles when we can properly yield back to the OS.

RadWolfie:
@LukeUsher Did you read my comment from #1341 for the reason it must have SEH intact for xbox emulation level? More specifically this comment line: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/f73114df61d943ba73b329a6e4bd6805254af243/src/CxbxKrnl/Emu.cpp#L338

By removing SEH, we do not get any popup message box at all if it's our own code fault or a fatal crash has occur during xbox emulation. This make it harder to reproduce the process while keeping SEH exception allow us to attach then start search for the fault without need to re-produce.

Also for VEH, read the comment line: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/f73114df61d943ba73b329a6e4bd6805254af243/src/CxbxKrnl/Emu.cpp#L329

Even if we catch all VEH exceptions, some are actually *false* positive since some drivers are handling their exception internally.

LukeUsher:
Are you sure? I’ve been testing this and everything is the same as before

RadWolfie:
Tested with Halo demo title (during in-game), confirmed there is no popup message box with commit 1ff481a with silent exit emulation. Before, it does.

LukeUsher:
@RadWolfie okay, I reverted that commit

PR #1395, opened 2018-08-16
BATCH: IPC improvement, features, and bug fixes

### LINKED ISSUE #1267: Implement logging off by default, GUI indication of logging
I suggest to turn off logging by default and turn on by pressing button (F8 at this time, if I remeber correctly). Why - to better control what need to log, for example you need to log during gameplay and you do not want log all before that.
Also need to make GUI indication - for example - create string "Logging" near FPS counter and make it inactive (grey) when logging disabled. When enabled - black.
If destination(Console or TextFile) of log is not set - grey, and pressing F8 does nothing.

RadWolfie:
What's included in this pull request:
* **NEW:** Log indication in GUI menu bar for both GUI and kernel log status.
   image
   * From above image, you will see G(UI) and/or K(ernel) letters. If you see a letter, then it is active.
* **NEW:** Append to existing log file when reboot has occurred for any reason.
* **NEW:** Record any active log filter and level to console/file.
* **IMPROVED:** Moving windows' SendMessage between kernel and gui processes into one function.
  * Will allow to work with other platform by using one function caller.
* **IMPROVED:** Preserve hotkey log enabled/disabled after reboot.
  * Require to be used in EmuShared.
* **MOVED:** LED, LLE flags, kernel status (waiting for gui to reply) to ipc_send_gui_update function.
  * Since status has been moved, EmuShared's values are changed to reserved.
* **FIXED:** Sometimes menu bar isn't updated properly, require to call DrawMenuBar function for refresh.
* **FIXED:** LLE flags was obtain from wrong getter. Since status flags was updated later.
  * Discovered during relocate LLE flags status send as message instead of EmuShared.
* **FIXED:** EmuLog function not showing whole log level info.
* **FIXED:** Ensure macro returns are checked for is printable or not.

Known issue(s):
* Using std::cout with active GUI's console is not outputting.

This fixes #1267


RadWolfie:
Tester from discord reported GUI's log indication isn't shown.

PatrickvL:
It'd be helpful to log these additional details:
* during INIT : the log-settings in-use 
* during emulation : the enabling/disabling (F8) of logging
* during emulation : changes in log-settings

RadWolfie:
Since I was thinking about second task to add in this pull request. Now you mention about first and third tasks, I'm sure it will be very helpful.

I am working on adding all three tasks, mentioned by @PatrickvL, in this pull request.

GXTX:
Super small suggestion with fa33272, adds an additional warning about those repeating `\/`'s. You could also minimize this be removing 2 of the lines and using `\n`.
Something like this stands out and doesn't cause any warnings.
```
"\n------REBOOT------REBOOT------REBOOT------REBOOT------REBOOT------\n"
```

RadWolfie:
Not sure why I had spaces in reboot message. However, std::endl is needed to flush into the file just in case  a crash may occur afterward. "\n" string doesn't flush to the file.

gandalfthewhite19890404:
For example (tested on latest version 65cd8786):
Run TES3: Morrowind or any multi-xbe title, While running one xbe turn off logging by F8. GUI indicates logging is off, OK. Load or start New Game, it runs another xbe. GUI still indicates that logging is off (beacause I didn't enable it back by F8). Stop emulation and open created logfile. You will see CXBXR   Enable log is 0 (its when I disabled logging by F8) and..... seems logging is works again - logfile is filled.

PR #1398, opened 2018-08-18
Replace use of EXP in Xbox Shaders with FRC

LukeUsher:
Xbox Implements EXP as "extract the fractional component" but PC
implements EXP as "Provides full precision exponential 2x."

This can be resolved by replacing the instruction with FRC, which is
implemented by D3D on PC as "Return the fractional portion of each
input component."

This fixes the broken vertices on the player model in THP2X:

Before:
image

After:
image


PatrickvL:
There's one downside to this : dumping the original unmodified Xbox shader won't show EXP anymore. It would have if it had been translated into a new FRC opcode.

PR #1404, opened 2018-08-22
General Tweaks/Improvements.

LukeUsher:
1. KeEvent functionality, including KeWaitForSingleObject and KeWaitForMultipleObjects no longer forward to Windows. This solves the fatal error that could occur when waiting for a critical section, as well as prevents several deadlock issues.
2. Fixes an issue where the SMC Version String was read incorrectly
3. Reads AVPack from SMC, instead of hard-coding in the kernel
4. Fix an issue where some titles would crash when accessing a CR register
5. Experimental (but disabled and VERY incomplete) support for Chihiro Media Board ROM(AKA Segaboot)

### DIFF
--- a/src/CxbxKrnl/EmuKrnlKe.cpp
+++ b/src/CxbxKrnl/EmuKrnlKe.cpp
@@ -1906,6 +1906,8 @@ xboxkrnl::PLARGE_INTEGER FASTCALL KiComputeWaitInterval(
 	}
 }
 
+xboxkrnl::LIST_ENTRY KiWaitInListHead;
+
 // ******************************************************************
 // * 0x009E - KeWaitForMultipleObjects()
 // ******************************************************************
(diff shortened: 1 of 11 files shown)


LukeUsher:
This PR should be relatively safe to merge, but PLEASE do some testing first: There's a very slim, but very real chance we introduced some new deadlocks/hangs that were not present before.

gandalfthewhite19890404:
WWE RAW2 thows this:
---------------------------
KeBugCheck
---------------------------
The running software triggered KeBugCheck with the following information
BugCheckCode: 0x0000000A
BugCheckParameter1: 0x00000000
BugCheckParameter2: 0x00000000
BugCheckParameter3: 0x00000000
BugCheckParameter4: 0x00000000

This is the Xbox equivalent to a BSOD and would cause the console to automatically reboot

Continue Execution (Not Recommended)?

---------------------------
Yes   No   
---------------------------


I may continue but this behaviour not normal.

LukeUsher:
Rebased on latest develop, TODO: Fix KeBugCheck (IRQL_NOT_LESS_OR_EQUAL)

src/CxbxKrnl/EmuKrnlKe.cpp
```diff
@@ -2018,7 +2020,7 @@ XBSYSAPI EXPORTNUM(158) xboxkrnl::NTSTATUS NTAPI xboxkrnl::KeWaitForMultipleObje
 
 			// Handle a Timeout if specified
 			if (Timeout != nullptr) {
-				// If the timeout is 0, do not wait
+				// If the timeout is 0, do not `
```
RadWolfie: is this intended? 🤔

LukeUsher: Nope, fixed it

LukeUsher:
Sonic Heroes regressed with this PR, investigating

PR #1405, opened 2018-08-23
Fix file path to use RootDirectory properly.

RadWolfie:
Turok Evolution can now finally be able to save the game progress.

`RootDirectory` handle is very important role when a title pass down with non-null handle in the structure. We should never ignore it at the end.

Ideally should not cause any regression.

Requesting @LukeUsher for review. Since I had moved his `BaseNamedObjects` check to last. Originally, it was at front. I'm not sure if it needs to be at front or not.

### DIFF
--- a/src/CxbxKrnl/EmuFile.cpp
+++ b/src/CxbxKrnl/EmuFile.cpp
@@ -344,9 +344,9 @@ void replace_all(std::string& str, const std::string& from, const std::string& t
 }
 
 NTSTATUS CxbxConvertFilePath(
-	std::string RelativeXboxPath, 
-	OUT std::wstring &RelativeHostPath, 
-	OUT NtDll::HANDLE *RootDirectory,
+	std::string RelativeXboxPath,
+	OUT std::wstring &RelativeHostPath,
+	IN OUT NtDll::HANDLE *RootDirectory,
 	std::string aFileAPIName,
 	bool partitionHeader)
 {
@@ -372,7 +372,8 @@ NTSTATUS CxbxConvertFilePath(
 				// If the remaining path starts with a ':', remove it (to prevent errors) :
 				if ((RelativePath.length() > 0) && (RelativePath[0] == ':'))
 					RelativePath.erase(0, 1);  // xbmp needs this, as it accesses 'e::\'
-			} else if (RelativePath[0] == '$') {
+			}
+            else if (RelativePath[0] == '$') {
 				if (RelativePath.compare(0, 5, "$HOME") == 0) // "xbmp" needs this
 				{
 					NtSymbolicLinkObject = FindNtSymbolicLinkObjectByRootHandle(g_hCurDir);
@@ -381,8 +382,7 @@ NTSTATUS CxbxConvertFilePath(
 					CxbxKrnlCleanup(LOG_PREFIX, ("Unsupported path macro : " + OriginalPath).c_str());
 			}
 			// Check if the path starts with a relative path indicator :
-			else if (RelativePath[0] == '.') // "4x4 Evo 2" needs this
-			{
+			else if (RelativePath[0] == '.') {// "4x4 Evo 2" needs this
 				NtSymbolicLinkObject = FindNtSymbolicLinkObjectByRootHandle(g_hCurDir);
 				RelativePath.erase(0, 1); // Remove the '.'
 			} else {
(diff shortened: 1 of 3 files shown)


src/CxbxKrnl/EmuFile.cpp
```diff
@@ -407,10 +406,27 @@ NTSTATUS CxbxConvertFilePath(
...
-				} else {
-					// Finally, Assume relative to Xbe path
+				}
+				// NOTE: RootDirectory cannot be ignored.
+				// Any special handling for it should be done below.
+				else if (RootDirectory == nullptr) {
+					// Assume relative to Xbe path
 					NtSymbolicLinkObject = FindNtSymbolicLinkObjectByRootHandle(g_hCurDir);
 				}
+				else if (RootDirectory == (NtDll::HANDLE)-4) {
```
PatrickvL: -4 might be the result of some registration function, registering multiple 'special roots'. If that's the case, other xdk or bios versions might register in a different order, resulting in another value for this one. At least describe this possibility in a comment, and/or use a named constant for the value -4. Otherwise a great find & clean fix!

LukeUsher: These handle constants are not caused by registrations, they are special handles used by the kernel to represent special paths or special objects. It won’t change per XDK because the kernel needs to remain compatible

ergo720:
I'm not totally sure but I report this so that it can be reviewed at least. Recently, DOAX has started to crash on debug builds only because of the below debug assertion (it didn't happen before). Considering the type of the error, it could be caused by this PR. However, I have to enable the all cores hack to get it because otherwise it just shows the dirty screen error. So either it's caused by the hack or it's a real error which is not shown without the hack because it stops before reaching the affected code.

crash
exception

RadWolfie:
Base on OriginalPath `"Ì:\\"` and HostPath being empty string, I find it odd. There might be some sort of corruption going on.

ergo720:
You're right, RootDirectory is 0xfffffffd. Now I was also able to get the crash without the all cores hack, so it's unrelated.

crash

RadWolfie:
Can you enable all logs (even set to debug level) then output into a file? It may provide better detail of how this happened.

PatrickvL:
Less patches and more complete kernel API implementations, will probably give the best results indeed!

PatrickvL:
The $CC marker value is probably because HostPath isn't set, which can happen if NtSymbolicLinkObject is null. Without the above mentioned patch on XMountAlternateTitleA, functionally correct kernel API's will go a long way in getting this to work correctly.

LukeUsher:
@RadWolfie pretty sure XMountAlternativeTitleA can be unpatched with no other code changes: XapiMapLetterToDirectory already works unpatched, because it is used internally to mount the cache partitions, we know this process to work.

ergo720:
I can confirm that DOAX has exactly the same problem: XMountAlternateTitleA, called right before the faulty NtOpenFile, has the pchDrive member set to all 0xCC.

```
[0x1300] XAPI    XMountAlternateTitleA(
   lpRootPath           : T:\
   dwAltTitleId         : 1297285165
   pchDrive             : (char *)0x0D99CE80 = "0xCC" [repeated many times]
);
[0x1300] WARN: XAPI    XMountAlternateTitleA unimplemented!
[0x1300] XAPI    XMountAlternateTitleA returns 0
[0x1300] RTL     RtlInitAnsiString(
 OUT DestinationString  : 0x0D99CE60
   SourceString         : (char *)0x0D99CE8C = "0xCC\\DOA3SAVE.DAT"
);
[0x1300] NT      NtOpenFile forwarding to "IoCreateFile"...
[0x1300] IO      IoCreateFile(
 OUT FileHandle         : 0x0D99CE70
   DesiredAccess        : 100001
   ObjectAttributes     : 0x0D99CE4C -> OBJECT_ATTRIBUTES* {
   .RootDirectory        : FFFFFFFD
   .ObjectName           : 0x0D99CE60 -> STRING* {
   .Length               : 3
   .MaximumLength        : 3
   .Buffer               : (char *)0x0D99CE8C = "0xCC\\DOA3SAVE.DAT"}
   .Attributes           : 40}
 OUT IoStatusBlock      : 0x0D99CE44
   AllocationSize       : 0x00000000
   FileAttributes       : 0
   ShareAccess          : 3
   Disposition          : (CREATE_DISPOSITION)0x00000001 = FILE_OPEN
   CreateOptions        : (CREATE_OPTION)0x00004021 = FILE_DIRECTORY_FILE|FILE_SYNCHRONOUS_IO_NONALERT|FILE_OPEN_FOR_BACKUP_INTENT
   Options              : 0
);
```

RadWolfie:
@LukeUsher actually, there is a fix necessary to NtQuerySymbolicLinkObject function. Since it's not performing the process right.

RadWolfie:
@ergo720, pchDrive is actually an output pointer for single char by the way.

PR #1408, opened 2018-08-23
Fix for vertex declaration failures

revel8n:
Initial fix for vertex declaration failures due to incorrectly handled CONST declaration Step size.
- Update step size calculation to return the correct number of DWORD values to skip


LukeUsher:
Nice catch!

ESPPIRAL:
bandicam 2018-08-24 15-11-46-429

PR #1412, opened 2018-08-25
Unpatch XMountAlternateTitleA and XMountAlternateTitleA

RadWolfie:
Plus fix NtQuerySymbolicLinkObject function.

Tested against:
* Lego Star Wars 2
* Dead or Alive Xtreme Beach Volleyball (possible resolve)


ergo720:
Confirmed: with this, DOAX doesn't crash any longer on debug builds now.

PR #1419, opened 2018-08-27
Report correctly free/committed placeholder memory

ergo720:
Test Conker before merging this since I don't have it. It shouldn't have caused regressions hopefully.


PatrickvL:
Why test Conker? Was there a report about it related to this (otherwise okay) fix?

LukeUsher:
Conker currently hangs, but for reasons unrelated to this commit. Turns out the KeWaitForMultipleObject implementation makes Conker get stuck in an infinite loop... manually allowing it to continue allows it to retain previous behavior, so this is okay to merge.

I'll investigate the conker hang as a separate issue.

ergo720:
@PatrickvL Conker is known to hang if NtQueryVirtualMemory doesn't report host allocations. Because I changed its behavior here, I didn't want that I accidentally broke something with the PR, that's why I asked to test it.

PR #1420, opened 2018-08-28
Split Symbol scanning and patching phases

LukeUsher:
This allows for both steps to be completely disconnected, easily allowing patches to be turned on or off based on a set of flags, as well as preventing the need to clear the HLE cache when switching from HLE->LLE.

This also allows patches to be seen/modified from a central location, no more searching through the codebase to determine if a function should be patched or not, and no more 'FUNC_EXPORTS/GetProcAddress' magic!

Currently, this is used for HLE only, but could really shine when extended to introduce optional detour based logging even when LLE is enabled.

For example, We could easily add a LLE_D3D_DETOUR flag, which when enabled, patches functions with a wrapper, which simply logs input and output, calling the original Xbox function via a trampoline.

This would be great for debugging, as we'd get a full call trace from the API level, even when not implementing HLE.

There's also the possibility of mixing in some patches even with LLE enabled: for a hybrid HLE/LLE solution of the same functionality, but there are no plans to implement that at this moment of time.

NOTE: This has been verified by comparing a list of exports with the patch table that all patches that had a FUNC_EXPORT are present. There will be ZERO compatibility impact for this change.

### DIFF
--- a/src/CxbxKrnl/HLEIntercept.cpp
+++ b/src/CxbxKrnl/HLEIntercept.cpp
@@ -48,19 +48,15 @@
 #pragma comment(lib, "XbSymbolDatabase.lib")
 #include "../../import/XbSymbolDatabase/XbSymbolDatabase.h"
 #include "HLEIntercept.h"
+#include "HLEPatches.h"
 #include "xxhash32.h"
 #include <Shlwapi.h>
-#include <subhook.h>
-
-inline void EmuInstallPatch(std::string FunctionName, xbaddr FunctionAddr, void *Patch);
-
 #include <shlobj.h>
 #include <unordered_map>
 #include <map>
 #include <sstream>
 
 std::map<std::string, xbaddr> g_SymbolAddresses;
-std::unordered_map<std::string, subhook::Hook> g_FunctionHooks;
 bool g_HLECacheUsed = false;
 
 // D3D build version
@@ -73,9 +69,10 @@ bool bLLE_JIT = false; // Set this to true for experimental JIT
 
 void* GetXboxFunctionPointer(std::string functionName)
 {
-    if (g_FunctionHooks.find(functionName) != g_FunctionHooks.end()) {
-        return g_FunctionHooks[functionName].GetTrampoline();
-    }
+	void* ptr = GetPatchedFunctionTrampoline(functionName);
+	if (ptr != nullptr) {
+		return ptr;
+	}
 
     // If we got here, the function wasn't patched, so we can just look it up the HLE cache
     // and return the correct offset
@@ -118,14 +115,6 @@ std::string GetDetectedSymbolName(xbaddr address, int *symbolOffset)
 
     *symbolOffset = 0;
     return "unknown";
-}
-
-// NOTE: GetEmuPatchAddr do not get to be in XbSymbolDatabase, perform patch check in Cxbx project only.
-void *GetEmuPatchAddr(std::string aFunctionName)
-{
-    std::string patchName = "XTL::EmuPatch_" + aFunctionName;
-    void* addr = GetProcAddress(GetModuleHandle(NULL), patchName.c_str());
-    return addr;
 }
 
 // NOTE: VerifySymbolAddressAgainstXRef do not get to be in XbSymbolDatabase, perform verification in Cxbx project only.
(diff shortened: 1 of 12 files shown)


src/CxbxKrnl/HLEPatches.cpp
```diff
@@ -0,0 +1,446 @@
...
+	PATCH_ENTRY("XGetDevices", XTL::EMUPATCH(XGetDevices), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInitDevices", XTL::EMUPATCH(XInitDevices), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInputClose", XTL::EMUPATCH(XInputClose), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInputGetCapabilities", XTL::EMUPATCH(XInputGetCapabilities), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInputGetDeviceDescription", XTL::EMUPATCH(XInputGetDeviceDescription), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInputGetState", XTL::EMUPATCH(XInputGetState), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInputOpen", XTL::EMUPATCH(XInputOpen), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInputPoll", XTL::EMUPATCH(XInputPoll), PATCH_HLE_OHCI),
+	PATCH_ENTRY("XInputSetState", XTL::EMUPATCH(XInputSetState), PATCH_HLE_OHCI),
+
```
ergo720: Here, I don't see the patch for XID_fCloseDevice. You have removed it on purpose, I think?

LukeUsher: There's no patch for XID_fCloseDevice within XAPI.cpp

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -277,35 +266,11 @@ void CDECL EmuRegisterSymbol(const char* library_str,
...
-    //       We need to create an array for symbol, patch, library type, etc structure.
-    //       Then we can replace checks below into permanent solution.
-    if (bLLE_APU && ((library_flag & XbSymbolLib_XACTENG) || (library_flag & XbSymbolLib_DSOUND) > 0)) {
-        // Do nothing if emulating LLE APU
-    } else if (bLLE_GPU && ((library_flag & XbSymbolLib_XGRAPHC) || (library_flag & XbSymbolLib_D3D8) || (library_flag & XbSymbolLib_D3D8LTCG) > 0)) {
-        // Do nothing if emulating LLE GPU
-	} else if (bLLE_USB && ((std::strcmp(symbol_str, "XInitDevices") == 0) || (std::strcmp(symbol_str, "XGetDevices") == 0) || (std::strcmp(symbol_str, "XGetDeviceChanges") == 0) ||
-		(std::strcmp(symbol_str, "XInputOpen") == 0) || (std::strcmp(symbol_str, "XInputClose") == 0) || (std::strcmp(symbol_str, "XInputPoll") == 0) ||
-		(std::strcmp(symbol_str, "XInputGetCapabilities") == 0) || (std::strcmp(symbol_str, "XInputGetState") == 0) || (std::strcmp(symbol_str, "XInputSetState") == 0) ||
-		(std::strcmp(symbol_str, "XGetDeviceEnumerationStatus") == 0) || (std::strcmp(symbol_str, "XInputGetDeviceDescription") == 0) || (std::strcmp(symbol_str, "XID_fCloseDevice") == 0))) {
```
PatrickvL: Ergo probably noticed it's mention here

PatrickvL: XID_fCloseDevice I mean

src/CxbxKrnl/HLEIntercept.cpp
```diff
@@ -479,28 +440,7 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
 
                     std::stringstream output;
                     output << "HLECache: 0x" << std::setfill('0') << std::setw(8) << std::hex << location
-                        << " -> " << functionName;
```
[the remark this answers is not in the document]
LukeUsher: This block no longer required or needed as the patch lookup works differently, and happens at a different stage of execution entirely, it no longer serves a purpose.

The address too low/too high patches could be moved to EmuInstallPatches, but the rest is redundant. We still log when functions are patched or skipped in the function that applies the patches (EmuInstallPatches) rather than at the symbol scanning stage.

PR #1421, opened 2018-08-30
Fix flickering window on move

RadWolfie:
Using RedrawWindow method perform very minor flicker when moving around on screen.

InvalidateRect & UpdateWindow method
old method

RedrawWindow method
redrawwindow method


RadWolfie:
Confirms it work better on Nvidia graphic card (1080). Untested on different graphic cards.

PatrickvL:
Thanks!

Issue #1416, opened 2018-08-27
Separate patches from forced patch in scanning process


RadWolfie:
There will be some files rearrangement for duration at least couple of weeks.

Known library using HLE patches:
- [ ] **DSound** (working on it)
- [ ] **Xapi** (haven't start on it)
- [ ] **D3D8** (haven't start on it)
  - On hold until no active updates from pull requests and forked repo. (I'm not ready for it anyway.)

If there is a library using HLE patches not listed above, please comment below or edit this post.
**IMPORTANT:** Kernel patches will not be work on since it is using a vtable patch method.

Recommendation not to make any changes:
* HLEIntercept.cpp (including HLEIntercept.h)
* Including relative files of known library listed above if ***is*** being work on.

Header files using C++ will be using hpp extension instead of h file extension.

Possible affecting Cxbx Debugger project once switch to SimpleIni and category all symbols to respective library group.

LukeUsher:
@RadWolfie can I please pick up the issue of separating HLE patches? I already started work on this behind the scenes to separate scanning and patching, and should be able to make a complete PR within a week-or-so

PR #1426, opened 2018-09-02
Move source files to their own folder part1.2

RadWolfie:
After relocate D3D8 and Direct3D9 files and fix linking. I did verified a rebase from either @PatrickvL's d3d_work and @LukeUsher's renderstate-from-nv2a-state branches has no conflict.

Low risk merge


RadWolfie:
refs #1416

PR #1427, opened 2018-09-04
Use SimpleIni class for symbol cache file

RadWolfie:
refs #1416 

The following changes in this pull request are:
* Replace Windows API to SimpleIni class usage.
* Remove some keys no longer going to be used.
* Store symbol cache files into SymbolCache directory instead of HLECache.
* Replace spaces to tabs indent in edited functions.
* Add missing copyright authors

Before: 
[Blood Wake-376abe0f.ini](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/2347009/Blood.Wake-376abe0f.ini.txt)
After:
[Blood Wake-376abe0f.ini](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/2347010/Blood.Wake-376abe0f.ini.txt)

### DIFF
--- a/resource/Cxbx.rc
+++ b/resource/Cxbx.rc
@@ -579,10 +579,10 @@ BEGIN
             MENUITEM "Store with Executable",       ID_SETTINGS_CONFIG_DLOCCURDIR,MFT_STRING,MFS_ENABLED
             MENUITEM "Custom",                      ID_SETTINGS_CONFIG_DLOCCUSTOM,MFT_STRING,MFS_ENABLED
         END
-        POPUP "&HLE Cache",                     65535,MFT_STRING,MFS_ENABLED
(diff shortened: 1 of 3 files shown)


gandalfthewhite19890404:
Add posibillity to clear Symbol Cache via menu entry as it was before with HLECache.

RadWolfie:
@PatrickvL Symbol cache is "symbol" cache, nothing more. We should not be adding/modify things to it except when XbSymbolDatabase library has an update, then it will perform a re-scan base on hashing version.

There's no reason include backward compatible with older build of Cxbx-Reloaded than this pull request since it contains modification to remain partial compatible for non-separated patching table.

@gandalfthewhite19890404 Thanks for bringing it up. 👍

PatrickvL:
LGTM!

PR #1436, opened 2018-09-11
Fixed the argc checks in winmain and CxbxKrnl.cpp

faha223:
WinMain.cpp and CxbxKrnl.cpp were checking the value of argc before using the value compared against to get a value out of argv. The value that was checked against was not the correct value though so these optional arguments were not being treated as optional and if left out the emulator would not start correctly in kernel mode.


PatrickvL:
Well spotted! Thanks!

PR #1437, opened 2018-09-12
nv2a: Ptimer fixes

PatrickvL:
Ported from https://github.com/xqemu/xqemu/pull/113 by dracc https://github.com/dracc

A few GPU timing related fixes
NV2A crystal frequency is currently set to debug unit speed of 13.5MHz.
Later retail xboxes had a 16.67MHz crystal according to @JayFoxRox's research(?).
Also, current ptimer clock value calculation is flipped around and on top of that always overflows.

With these fixes, running get_timers.py against xqemu reports values close to what real hardware gives.

Proposed changes
Set NV2A_CRYSTAL_FREQ to retail speed
Fix ptimer_get_clock() behaviour


PatrickvL:
Got this feedback :

> 
> Tested your PR with: 
> - Burnout
> - Turok
> - Panzer Dragoon Orta
> - Exa Skeleton
> 
> Couldn't see any difference in the behavior (using LLE rendering)

PR #1440, opened 2018-09-17
Interpret larger blocks during exception handler to reduce the number of exceptions per-frame

LukeUsher:
This is mostly preparation for things to come in the (near) future but has a little impact in its current form.

Some titles may see a 1-2 FPS increase using LLE, some titles may show 5+ FPS increase with HLE, but overall, the difference is minimal.

### DIFF
--- a/src/CxbxKrnl/EmuFS.cpp
+++ b/src/CxbxKrnl/EmuFS.cpp
@@ -127,7 +127,7 @@ NT_TIB *GetNtTib()
 }
 
 
-xboxkrnl::KPCR* KeGetPcr();
+xboxkrnl::KPCR* WINAPI KeGetPcr();
 
 uint32_t fs_lock = 0;
 
--- a/src/CxbxKrnl/EmuKrnl.cpp
+++ b/src/CxbxKrnl/EmuKrnl.cpp
@@ -131,7 +131,7 @@ xboxkrnl::PLIST_ENTRY RemoveTailList(xboxkrnl::PLIST_ENTRY pListHead)
 // * Declaring this in a header causes errors with xboxkrnl
 // * namespace, so we must declare it within any file that uses it
 // ******************************************************************
-xboxkrnl::KPCR* KeGetPcr();
+xboxkrnl::KPCR* WINAPI KeGetPcr();
 
 // Interrupts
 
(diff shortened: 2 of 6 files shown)


gandalfthewhite19890404:
Cause regression with LLE GPU enabled, tested and confirmed on WWE Raw 2.
At previous Previous 1437 it get to Unknown IV_IMAGE_BLIT Method and further to ingame.


Latest 1440 just hangs just before that - IoCreateObject Failed. Stopping emulation does not terminate process - you can see high CPU load after it.

PatrickvL:
Previously, we skipped unrecognised opcodes, but not anymore - perhaps that's causing the observed hanging?

LukeUsher:
thanks, I'll investigate WWE Raw 2

gandalfthewhite19890404:
But 1440 improves (you can see rotating ball much faster) 2002 FIFA World Cup with LLE GPU enabled. 
Just keep it in mind.

gandalfthewhite19890404:
Same thing with The Sims. In Pre-1440 its get to Unknown NV_IMAGE_BLIT Method.
PR1440 - just hanges just before that on DSOUND-related warning.
Btw - The Sims throws to log many unsupported instructions on PR1440 - 4343, 223, 117, 793, 188, 987.

PatrickvL:
Our current CPU emulator just stops interpreting once it hits an unsupported instruction, and resumes execution from that point onward on the host CPU. So from there on emulation behaves like usual. Interpretation is slower than native execution, but since it avoids our access violation handler for MMIO instructions, interpretation actually lowers overhead and thus improves overall performance.

PatrickvL:
Interpretation starts when an MMIO instruction is hit, and ends on call, jump, return and unsupported instructions, like the ones mentioned above.

PatrickvL:
That being said, by adding support for more instructions, interpretation could run longer pieces of hardware-accessing functions

PatrickvL:
As for the fee titles that hang: most likely that's due to an unsupported instruction accessing uncommitted memory; Since it's unsupported, the host CPU is resumed on that instruction, it'll access this uncommitted memory, such once again triggers our exception handler, which tries to emulate the instruction, but it turns out to be unsupported. With that, the above described steps repeats indefinitely...

LukeUsher:
@PatrickvL actually, I would be more inclined to believe the hangs to be caused by incorrect flag calculation, I need to go through every instruction and verify the behavior. Incorrect flag calculation could lead to infinite loops when conditional branches are used.

gandalfthewhite19890404:
Mafia also hangs on 1440.

PatrickvL:
There's a good chance a fix on this will alleviate the hang in all cases.

PatrickvL:
@gandalfthewhite19890404 would you like to test #1444 again? We've improved CPU interpretation a bit, which might help...

gandalfthewhite19890404:
WWE RAW2 is not fixed yet. Should I test other titles?

PR #1441, opened 2018-09-19
HLE : Extract common CreateDevice code to two separate functions.

PatrickvL:
...And call these two functions in all three versions of our DIrect3D_CreateDevice patch.


LukeUsher:
Or better yet, refactor to prevent the need to patch this function. Some settings can be read from the nv2a state or via the AvSetDisplayMode kernel call. A successful bypass of the requirement for a patch for this function could coax more LTCG titles into actually booting with HLE: Sonic Mega Collection is a possible test case

PatrickvL:
Yeah this was only small stuff, low hanging fruit

LukeUsher:
Sorry, I read this as an issue on my phone, didn't realise it was a PR, my bad, I'm happy for the contribution!

PR #1458, opened 2018-10-02
Various fixes/improvements to D3D HLE

LukeUsher:
The key change is that more of the D3D state is written back to the
internal Xbox D3D state, allowing GetXXX functions to return the correct
values, and that GetVertexShaderDeclaration now returns the correct size
to the running Xbox title.

Note that this needs testing with as many titles as possible: The
additional trampolines may cause regressions if they hit an
unimplemented instruction in subhook (eg. I had to update Subhook to add
xor, in order to be able to make this PR work at all)


gandalfthewhite19890404:
I'll keep update this post further.

The Elder Scrolls 3: Morrowind - no regressions, no visual improvements too (at least I didn't noticed)
WWE RAW 2 - no regressions, no visual improvements too (at least I didn't noticed)

PatrickvL:
This exception seems to happen in Cxbx itself (not in the emulated games)

LukeUsher:
Thanks for updating the list, I'll look into it

PatrickvL:
The above reported regressions are mostly solved (see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1463#issuecomment-427350791)

PR #1460, opened 2018-10-03
Detect and fix the situation when EmuD3DDeferrdRenderState is derived incorrectly

LukeUsher:
This may be re-worked in the future to drive D3D__RenderStates directly, without requiring extensive changes, potentially leading to the unpatching of all RenderState, and Pixel Shader functions!


PatrickvL:
Will you be adding more commits to this, or shall we merge it as-is now?

LukeUsher:
It’s fine as is

PR #1461, opened 2018-10-04
Nv2a plugin clear

PatrickvL:
This PR makes the clear operation part of the PGRAPH rendering plugin framework (besides some other minor work).
There's no functional difference in behavior.

Note : This PR will probably cause a merge conflict for #1457

### DIFF
--- a/src/devices/video/EmuNV2A_PGRAPH.cpp
+++ b/src/devices/video/EmuNV2A_PGRAPH.cpp
@@ -690,7 +690,6 @@ void OpenGL_draw_state_update(NV2AState *d)
 	PGRAPHState *pg = &d->pgraph;
 
 	assert(pg->opengl_enabled);
-	assert(pg->shader_binding);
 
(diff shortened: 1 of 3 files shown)


src/devices/video/EmuNV2A_PGRAPH.cpp
```diff
@@ -668,13 +681,18 @@ void OpenGL_draw_inline_elements(NV2AState *d)
...
+
+	OpenGL_draw_end(d);
 }
 
 void OpenGL_draw_state_update(NV2AState *d)
 {
 	PGRAPHState *pg = &d->pgraph;
 
 	assert(pg->opengl_enabled);
+	assert(pg->shader_binding);
```
PatrickvL: This assert is not present in the original code, do should be removed

JayFoxRox:
Some of this cleanup would be good for upstream in my opinion.
I think our switch-case is too huge.

PR #1462, opened 2018-10-05
NV2A : Fixed PGRAPH light handlers

PatrickvL:
The case statement wasn't covering the final few light slots, potentially causing black spots here and there (specifically in scenes that use all 8 lights).


JayFoxRox:
Is this an upstream issue? If so, please upstream.
We don't have the dev capacity to monitor downstream, especially if PRs are merged so quickly.

PatrickvL:
Xqemu uses a GCC extension, ranged case statements, which MSVC doesn't have. This has been circumvented using CASE macros in Cxbx-Reloaded. It's possible xqemu uses a too-small range too.

JayFoxRox:
Please find a good way to do this and do the change upstream: we also want MSVC compatibility in XQEMU (for the future).

PR #1466, opened 2018-10-06
Performance and bug fixes

LukeUsher:
Crash Bandicoot: tWoC is playable with this!

image


AzurikRiseOfPerathia:
Let's GO for Azurik: Rise Of Perathia

PR #1481, opened 2018-10-12
Deduplicate list code

PatrickvL:
This PR de-duplicates some list-management code written by @Ergo720
Triggered by external input.


gandalfthewhite19890404:
Its fix for memory manager? What kind of issue I should find? I am tested some games, didn't noticed any regressions.

PatrickvL:
There shouldn't be any regressions, as functionally there's no difference (only code-duplication is reduced somewhat)

PR #1489, opened 2018-10-24
Replace atomic boolean m_bFrameTime with a mutex m_FrameTimeMutex

PatrickvL:


LukeUsher:
Why this replacement? Was there an issue with the existing boolean, or is the mutex faster?

PatrickvL:
Here a discussion about this: https://www.reddit.com/r/cpp_questions/comments/3lbxpc/using_a_global_atomic_variable_instead_of_a_mutex/

PatrickvL:
I should change all occurrences of "while not try_lock Sleep" to just "lock"...

PatrickvL:
Please postpone merging this until I've done the lock-replacement

PatrickvL:
Okay, done. I've rebased this, and replaced all "while try_lock() Sleep" constructs with just a lock()

PR #1490, opened 2018-10-24
Move acknowledgments towards the top of each file

PatrickvL:


LukeUsher:
Anything that increases the visibility of sources is great to have

PR #1492, opened 2018-10-24
Fix an issue where RootDirectory wasn't being set properly when converting file paths

LukeUsher:


src/CxbxKrnl/EmuFile.cpp
```diff
@@ -416,15 +417,21 @@ NTSTATUS CxbxConvertFilePath(
 				}
 				// NOTE: RootDirectory cannot be ignored.
 				// Any special handling for it should be done below.
-				else if (RootDirectory == nullptr) {
+				else if (*RootDirectory == nullptr) {
```
RadWolfie: I'm 100% sure one or two of our own code has RootDirectory as nullptr. Hence the reason of RootDirectory pass down is checking for nullptr *FIRST*. Before cause an exception for invalid pointer address.

src/CxbxKrnl/EmuFile.cpp
```diff
@@ -416,15 +417,21 @@ NTSTATUS CxbxConvertFilePath(
...
+					// We can safely remove it and forward to the Xbe directory.
+					// Test case GTA3
+					NtSymbolicLinkObject = FindNtSymbolicLinkObjectByRootHandle(g_hCurDir);
+				}
+				else if (*RootDirectory == (NtDll::HANDLE)-4) {
 					// NOTE: A handle of -4 on the Xbox signifies the path should be in the BaseNamedObjects namespace.
 					// This handle doesn't exist on Windows, so we prefix the name instead. (note from LukeUsher)
 					// Handle special root directory constants
-					RootDirectory = NULL;
+					*RootDirectory = NULL;
```
RadWolfie: Oops

src/CxbxKrnl/EmuFile.cpp
```diff
@@ -320,7 +320,7 @@ std::wstring string_to_wstring(std::string const & src)
 
 std::wstring PUNICODE_STRING_to_wstring(NtDll::PUNICODE_STRING const & src)
 {
-	return std::wstring(src->Buffer, src->Length / sizeof(NtDll::WCHAR));
+return std::wstring(src->Buffer, src->Length / sizeof(NtDll::WCHAR));
```
RadWolfie: Um... no tab here?

PR #1494, opened 2018-10-28
Avoid LRP error message when compiling pixel shaders

PatrickvL:
This prevents "error X5765: Dest register for LRP cannot be the same as first or third source register" in WWE RAW2.

A little background information :

When compiling pixel shaders, to facilitate later optimizations, FixOverusedRegisters() inserts MOV opcodes reading from constant, vertex color or texture registers, and writing to a free R register slot.

In this case, this transformed :

````
lrp r0, c0,r2,c1
````

into

````
mov r0, c0
lrp r0, r0,r2,c1
````

Not only introduces this a completely useless additional MOV (to R0, which is overwritten by the next LRP opcode), but it also results in LRP reading from R0, which is the same register it's writing to, which seems to be forbidden.

This fix prevents that from happening, by not-inserting a MOV when this situation would arise.

It's possible there exist more such scenario's, but let's tackle them when they are reported.

As for the other change in this commit : Instead of repeating a nearly identical while-loop three times, this loop is now wrapped in a for-loop, that handles all three versions (for PARAM_C, PARAM_V and PARAM_T type registers).

### DIFF
--- a/src/core/HLE/D3D8/XbPixelShader.cpp
+++ b/src/core/HLE/D3D8/XbPixelShader.cpp
@@ -5805,63 +5805,35 @@ bool PSH_XBOX_SHADER::FixOverusedRegisters()
 
         int InsertPos = i;
 
-        int addressCount = 0;
-        int total = 0;
-        while (Intermediate[i].ReadsFromRegister(PARAM_C, -1, addressCount, total) && (addressCount > 1 || total > 2))
-        {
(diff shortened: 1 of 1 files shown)


gandalfthewhite19890404:
Game just hangs when it get to main menu

RadWolfie:
Revived pull request #1811

PR #1495, opened 2018-10-28
Fix Virtua Cop 3 (Chihiro)

LukeUsher:
A selection of tweaks an improvements to the vertex buffer and shader conversion code, resuling in Virtua Cop 3 reaching in-game!

Note that Virtua Cop 3 is an (encrypted) Sega Chihiro title, and as such, will require decryption using https://github.com/JayFoxRox/Chihiro-Tools

image

### DIFF
--- a/src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
@@ -3433,6 +3433,7 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_CreateVertexShader)
 
 	// Now, we can create the host vertex shader
 	CxbxVertexShader* hostVertexShader = (CxbxVertexShader*)malloc(sizeof(CxbxVertexShader));
+	memset(hostVertexShader, 0, sizeof(CxbxVertexShader));
 
     LPD3DXBUFFER pRecompiledBuffer = NULL;
     DWORD        *pRecompiledDeclaration = NULL;
--- a/src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
@@ -3414,6 +3414,7 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_CreateVertexShader)
 	XB_trampoline(HRESULT, WINAPI, D3DDevice_CreateVertexShader, (CONST DWORD*, CONST DWORD*, DWORD*, DWORD));
 	HRESULT hRet = XB_D3DDevice_CreateVertexShader(pDeclaration, pFunction, pHandle, Usage);
 	if (FAILED(hRet)) {
+		LOG_TEST_CASE("XB_D3DDevice_CreateVertexShader failed");
 		RETURN(hRet);
 	}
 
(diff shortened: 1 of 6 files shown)


src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -8100,36 +8093,39 @@ VOID __stdcall XTL::EMUPATCH(D3DDevice_DeleteVertexShader_0)
...
-			g_VMManager.Deallocate((VAddr)pVertexShader->pFunction);
+			free(pVertexShader->pFunction);
 		}
 
 		FreeVertexDynamicPatch(pVertexShader);
 
-		g_VMManager.Deallocate((VAddr)pVertexShader);
-		g_VMManager.Deallocate((VAddr)pD3DVertexShader);
+		free(pVertexShader);
+		SetCxbxVertexShader(Handle, nullptr);
```
PatrickvL: It's probably safer to do this nulling right after GetCxbxVertexShader()

src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3544,43 +3538,42 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_CreateVertexShader)
...
+			hostVertexShader->pFunction = NULL;
+			hostVertexShader->FunctionSize = 0;
         }
-        pVertexShader->Handle = Handle;
+		hostVertexShader->Handle = Handle;
     }
     else
     {
-        pVertexShader->Handle = D3DFVF_XYZ | D3DFVF_TEX0;
+        hostVertexShader->Handle = D3DFVF_XYZ | D3DFVF_TEX0;
```
PatrickvL: This looks like a corner case, perhaps a LOG_TEST_CASE can help keeping a trace on this

src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3410,37 +3410,31 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_CreateVertexShader)
...
+		RETURN(hRet);
+	}
+
 	if (g_pD3DDevice == nullptr) {
 		LOG_TEST_CASE("D3DDevice_CreateVertexShader called before Direct3D_CreateDevice");
 		// We lie to allow the game to continue for now, but it probably won't work well
 		return STATUS_SUCCESS;
 	}
 
-	// Allocates an Xbox VertexShader struct
```
PatrickvL: It'd like to keep this piece of documentation

src/core/HLE/D3D8/XbVertexShader.cpp
```diff
@@ -2608,3 +2608,27 @@ XTL::CxbxVertexShaderInfo *GetCxbxVertexShaderInfo(DWORD Handle)
...
+	}
+
+	return nullptr;
+}
+
+void XTL::SetCxbxVertexShader(DWORD Handle, XTL::CxbxVertexShader* shader)
+{
+	auto it = g_CxbxVertexShaders.find(Handle);
+	if (it != g_CxbxVertexShaders.end() && it->second != nullptr && shader != nullptr) {
+		LOG_TEST_CASE("Overwriting existing Vertex Shader");
```
PatrickvL: Should the previous host shader be freed here?

LukeUsher: That's why we log a test case, this is a situation that should never happen. If it happens, something is wrong elsewhere

src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: The source code is not the  best place for documenting what original XDK functions do: The code comments should describe what our own code does, 

Documentation on what XDK functions do should be documented somewhere more appropriate, , like a Wiki, because it's useful for non-Cxbx projects

src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: agreed

src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Did you move this somewhere then?

src/core/HLE/D3D8/XbVertexShader.cpp
PatrickvL: True.

PR #1499, opened 2018-10-30
Fix a crash within LTCG SetBorderColor, fix off-by-one error in mip-map level test

LukeUsher:


src/core/HLE/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5734,7 +5734,9 @@ VOID XTL::EMUPATCH(D3DDevice_SetTextureState_BorderColor_0)
 	}
 
 
-	return EMUPATCH(D3DDevice_SetTextureState_BorderColor)(Stage, Value);
+	HRESULT hRet;
+	hRet = g_pD3DDevice->SetSamplerState(Stage, D3DSAMP_BORDERCOLOR, Value);
```
LukeUsher: This change was made because trampolines do not work, and cannot be made to work for non-standard LTCG calling conventions

PR #1509, opened 2018-11-25
Fix THPS2X Regression after PDO fix

LukeUsher:
Please don't merge until this has been verified to not break Panzer Dragoon Orta


literalmente-game:
I'll run the game again with this PR to be sure

literalmente-game:
LGTM :heavy_check_mark:

LukeUsher:
@literalmente-game @StrikerX3 just got a hang in this branch, so we'll hold off merging a little while longer

PatrickvL:
Any notification function that accesses thread local storage is at risk of accessing the wrong thread's storage - so either notifications need to be done within the new thread's context, or in the creating thread context.

LukeUsher:
@PatrickvL I checked with a disassembly of the real kernel: The thread notification routine is called by the function creating the thread, not within the new thread. Calling the notification routine in the new thread, is what had caused THPS2X to hang. Turns out, this is something we've had wrong since forever.

StrikerX3:
I confirm that this does fix the hangs. Great job Luke!

PR #1520, opened 2018-12-31
Add implementation of RtlRaiseStatus

Fisherman166:
Reverse engineered from a kernel dump of my personal Xbox. Requires RtlRaiseException to be implemented to truly work but this gets us one step closer to a complete kernel.


PatrickvL:
Clean and simple, well done!

PR #1521, opened 2019-01-02
Restore upscaling hack after broken by D3D9 port

LukeUsher:


RadWolfie:
Report of regression:
* Jet Set Radio Future - Before, was able to see during boost dash. Now, no longer able to see anything.
image

PR #1523, opened 2019-01-06
Check XBE signature on Open

GXTX:
Check XBE signature on open & allow users to disable this pop-up using the setting `allowAdminPrivilege`.

I think this will produce better overall compatibility reports of games as it incentivizes users to use unmodified games.

### DIFF
--- a/src/gui/WndMain.cpp
+++ b/src/gui/WndMain.cpp
@@ -1991,9 +1991,11 @@ void WndMain::OpenXbe(const char *x_filename)
         return;
     }
 	
-	if (!m_Xbe->CheckXbeSignature())
+	if (!m_Xbe->CheckXbeSignature() && !g_Settings->m_core.allowAdminPrivilege)
 	{
-		int ret = MessageBox(m_hwnd, "XBE signature check failed! It's possible that hackers have injected malicious code into this game.\n\nAre you sure you wish to continue?", "Cxbx-Reloaded", MB_ICONEXCLAMATION | MB_YESNO);
+		int ret = MessageBox(m_hwnd, "XBE signature check failed!\n"
+			"\nThis is dangerous, as maliciously modified Xbox titles could take control of your system.\n"
+			"\nAre you sure you want to continue?", "Cxbx-Reloaded", MB_ICONEXCLAMATION | MB_YESNO);
 		if (ret != IDYES)
 			return;
 	}
--- a/src/gui/WndMain.cpp
+++ b/src/gui/WndMain.cpp
@@ -1990,7 +1990,14 @@ void WndMain::OpenXbe(const char *x_filename)
 
         return;
     }
-
+	
+	if (!m_Xbe->CheckXbeSignature())
+	{
+		int ret = MessageBox(m_hwnd, "XBE signature check failed! It's possible that hackers have injected malicious code into this game.\n\nAre you sure you wish to continue?", "Cxbx-Reloaded", MB_ICONEXCLAMATION | MB_YESNO);
+		if (ret != IDYES)
+			return;
+	}
+	
     // save this xbe to the list of recent xbe files
     if(m_XbeFilename[0] != '\0') {
         bool found = false;
--- a/src/gui/WndMain.cpp
+++ b/src/gui/WndMain.cpp
@@ -1991,13 +1991,21 @@ void WndMain::OpenXbe(const char *x_filename)
         return;
     }
 	
-	if (!m_Xbe->CheckXbeSignature() && !g_Settings->m_core.allowAdminPrivilege)
+	if (!g_Settings->m_core.allowAdminPrivilege && !m_Xbe->CheckXbeSignature())
 	{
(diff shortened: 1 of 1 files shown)


src/gui/WndMain.cpp
```diff
@@ -1990,7 +1990,16 @@ void WndMain::OpenXbe(const char *x_filename)
 
         return;
     }
-
+	
+	if (!m_Xbe->CheckXbeSignature() && !g_Settings->m_core.allowAdminPrivilege)
```
JayFoxRox: Checks should be done in different order: Only run check if `!g_Settings->m_core.allowAdminPrivilege`.

Currently, you always check, but only decide if you want a messagebox if there's no admin privileges (possibly discarding the check result).

src/gui/WndMain.cpp
```diff
@@ -1990,7 +1990,16 @@ void WndMain::OpenXbe(const char *x_filename)
...
     }
-
+	
+	if (!m_Xbe->CheckXbeSignature() && !g_Settings->m_core.allowAdminPrivilege)
+	{
+		int ret = MessageBox(m_hwnd, "XBE signature check failed!\n"
+			"\nThis is dangerous, as maliciously modified Xbox titles could take control of your system.\n"
+			"\nAre you sure you want to continue?", "Cxbx-Reloaded", MB_ICONEXCLAMATION | MB_YESNO);
+		if (ret != IDYES)
+			return;
```
JayFoxRox: What happens to allocations like `m_Xbe`?

See error handling in the `if(m_Xbe->HasError())` case, for what's necessary to abort.

JayFoxRox:
While a good idea on paper, this feels like a bad idea in practice.

At least `CheckXbeSignature` should check all valid / common signatures: MS retail, MS debug, MS chihiro.
I believe some Xbox Live betas also had custom signatures.

Currently, the warning is a no-op and gives a false sense of security. Because it allows zero-signatures from homebrew, which *can* be malicious, because it wasn't verified by a trusted source.
I believe there's hardly any XBE which does not succeed the check in the current form, and if it does, it's probably not malicious, but just corrupted.
Meanwhile, even malicious XBEs are allowed.

So the homebrew signatures, like habibi or a zero signature, would have to be allowed, but warned.

I'm also not sure about the signature specifics, but if it's also checking header fields like "allowed-media-flags", then almost any XBE in existence will now be reported (affecting usability, and scaring people away). Many dumping solutions will modify these flags. To my knowledge, some tools also patched code to work around security checks.

src/gui/WndMain.cpp
GXTX: Fixed

src/gui/WndMain.cpp
GXTX: Fixed

GXTX:
Went ahead and fixed your concerns. As to the last paragraph there, like we chatted about sometime last night the biggest reason for this PR is users are currently for the most part using non-original versions of games and I believe this pollutes the compatibility tracker immensely by not only having different versions of games but also different flags for each game. There also comes a risk in using these modified versions such as if a bad actor were to inject malicious code after a user uses closed source tools to create personal copies.

If the user doesn't know how to set these tools not to modify game code I think that's a separate issue we need to address, and I think runs back into your concerns with the redump project -- they aren't saving the games correctly with ALL relevant information.

src/gui/WndMain.cpp
```diff
@@ -1991,13 +1991,21 @@ void WndMain::OpenXbe(const char *x_filename)
...
+	if (!g_Settings->m_core.allowAdminPrivilege && !m_Xbe->CheckXbeSignature())
 	{
 		int ret = MessageBox(m_hwnd, "XBE signature check failed!\n"
 			"\nThis is dangerous, as maliciously modified Xbox titles could take control of your system.\n"
 			"\nAre you sure you want to continue?", "Cxbx-Reloaded", MB_ICONEXCLAMATION | MB_YESNO);
 		if (ret != IDYES)
+		{
+			delete m_Xbe; m_Xbe = nullptr;
+			
+			RedrawWindow(m_hwnd, NULL, NULL, RDW_INVALIDATE);
```
JayFoxRox: nit: order in other case frees and redraws first, then shows message-box.

RadWolfie:
`CheckXbeSignature` is using XePublicKeyData which is already provided in the source code. I'm not sure if debug kit (most likely yes) and chihiro are using the same public key. If not, then someone will have to get the RSA public key from the hardware.

> I believe some Xbox Live betas also had custom signatures.

I'm not sure about this since I don't have access to any of Xbox Live games. Plus it should produce a failure on real hardware if it was using a different RSA public key.

RadWolfie:
Other than that, looks good to me since it will popup only on GUI's end when open an xbe file.

LukeUsher:
Thanks!

Currently this shows for Chihiro executables, but that's because we don't have the Chihiro xbe key in Cxbx-R yet. I'll address that in a later commit.

Issue #1154, opened 2018-05-11
Speed up walking through index buffer


PatrickvL:
Cxbx-Reloaded contains code that walks through the data in index buffers, searching for the highest (and in some cases also the lowest) value.

This walking happens at each draw, making this already inefficient code slow.

For this issue, try to move this walking over to when the index buffer is registered as a host index buffer, and store the outcome for later use.

More importantly, create optimized versions of this walking function using compiler intrinsics for parallel machine code instructions (either SSE, SSE2, AVX, AVX2, or wherever else is possible).
Select the fastest one based on host CPU capabilities.

PatrickvL:
Status : #1527 addresses the latter part (use a host CPU-optimized version).

The first part (only do index-walk when registration or alterations happen) is still open.

Issue #1277, opened 2018-06-10
Implement vertex buffer caching


PatrickvL:
Currently, Cxbx-Reloaded converts vertex buffers at each draw, even when nothing changed in the contents of this buffer.
This causes massive slowdowns that could be improved by caching the result of vertex buffer conversion.
Like with textures and index buffers, make sure conversion is redone when anything changed in the contents of a vertex buffer (or the declaration of the vertex stream / Flexible Vertex Format for a vertex buffer).

LukeUsher:
Tagged as High Priority due to the massive performance impact not having such a cache causes. 2D titles like Metal Slug would benefit greatly, as would any title using Quads

LukeUsher:
We did used to have a Vertex Buffer cache so old versions of Cxbx-R could be looked at for inspiration.
It was removed because it was extremely buggy, and caused more issues than it solves.

A vertex buffer cache needs thorough testing to avoid this situation again!

PR #1527, opened 2019-01-17
Add SSE WalkIndexBuffer implementation

NZJenkins:
Partially address #1154 

- SSE4.1 WalkIndexBuffer implementation
- SSE level detection
- Auto-select WalkIndexBuffer implementation

I'm learning C++ so I think the code may be quite poor quality (esp CPUID.h)
The function runs faster, but doesn't seem to be a bottleneck so there is no overall perf benefit (only tested JSRF, dashboard)

SSE4.1 was most straightforward and widely supported, but lower levels are possible
I don't have an >AVX hardware

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/WalkIndexBuffer.cpp
+++ b/src/core/hle/D3D8/Direct3D9/WalkIndexBuffer.cpp
@@ -1,0 +1,92 @@
+//#include <mmintrin.h> // - MMX
+//#include <xmmintrin.h> // SSE
+//#include <emmintrin.h> // SSE2
+//#include <pmmintrin.h> // SSE3
+#include <smmintrin.h> // SSE4.1
+//#include <nmmintrin.h> // SSE4.2
+//#include <immintrin.h> // AVX
+
+#include "core\kernel\support\Emu.h"
+#include "core\kernel\support\EmuXTL.h"
+
+#include "common\util\CPUID.h"
+#include "WalkIndexBuffer.h"
+
+void WalkIndexBuffer_SSE41(XTL::INDEX16 & LowIndex, XTL::INDEX16 & HighIndex, XTL::INDEX16 * pIndexData, DWORD dwIndexCount);
+
+void Init_SIMD
+(
+	XTL::INDEX16 &LowIndex,
+	XTL::INDEX16 &HighIndex,
+	XTL::INDEX16 *pIndexData,
+	DWORD dwIndexCount
+)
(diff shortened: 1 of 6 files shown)


LukeUsher:
Thanks, this isn't the main bottleneck but every little thing helps. 
Good job on the feature detection too, that'l be useful going forward for other processor-specific optimizations.

For those following, currently the largest performance increase could be gained by #1277

src/core/hle/D3D8/Direct3D9/WalkIndexBuffer.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "core\kernel\support\Emu.h"
+#include "core\kernel\support\EmuXTL.h"
+
+#include "common\util\CPUID.h"
+#include "WalkIndexBuffer.h"
+
+void WalkIndexBuffer_SSE41(XTL::INDEX16 & LowIndex, XTL::INDEX16 & HighIndex, XTL::INDEX16 * pIndexData, DWORD dwIndexCount);
+
+void Init_SIMD
```
PatrickvL: Could more accurately be called `Bootstrap_WalkIndexBuffer`.

LukeUsher: I thought similar, but then I was thinking it could later be moved/extended to be a generic SIMD capability checking function, that can be used for other areas too.

PatrickvL: Generic SIMD capability detection would collide in function signatures, so there's still one bootstrap callback needed per function anyway.

Also, it's probably best to do some more renaming:
* `WalkIndexBuffer_SIMD` should be renamed back to the original `WalkIndexBuffer`
* the new (dumb) `WalkIndexBuffer` should then be renamed `WalkIndexBuffer_NoSIMD`

With that, a pattern would be established that, when followed for other functions, doesn't require modification to any caller.

The steps for any function would be:
* create a small bootstrap for it, prefixed with `Bootstrap_`
* set the callback to initially point to the bootstrap
* rename the existing implementation to have a suffix `_NoSIMD`
* change the function declaration to a callback function pointer, with the same name as the original function
* implement one or more optimized versions for this function
* extend `Init_SIMD` to set the new function pointer to best implementation (based on host features)
* in the bootstrap, call `Init_SIMD` and forward the call to the (now optimal) function pointer again

PR #1529, opened 2019-01-18
Fix executable directory path function

### LINKED ISSUE #1528: "Store with Executable" incorrectly storing files with XBE location
Appears to be a longstanding issue with changing storage location to current directory.

The issue is when on first launch you change the setting to be current directory the actual directory set by `GenerateCurrentDirectoryStr()` will be that of the XBE location. I think the function checks what the last open location in Windows was NOT where the executable is launched from.
unknown

On 2nd launch open XBE and attempt to launch the `DataStorageToggle` will remain unchanged thus skipping the 2nd check of `GenerateCurrentDirectoryStr()` and will have the correct location.
reboot

RadWolfie:
Fixes #1528

Instead of improper naming in code base. A workaround solution is convert argv[0] into as global variable. Since argv passdown are not declared globally.

Also, Windows require WinMain to able use GUI instead of using main function.

Plus, I added a TODO note for all `__argv[x]` and `__argc` need to be replace in order to support cross-platform since they are Microsoft extension.

Tested and confirmed working for drag & drop XBE file on top of cxbx.exe in file explorer.


GXTX:
Issue appears to have been fixed with this. Thanks :)

PR #1530, opened 2019-01-21
Update XBSymbolDatabase version

LukeUsher:


PatrickvL:
LGTM

RadWolfie:
Confirmed XbSymbolDatabase.lib is able to compile base on release's log. Plus added `false` is the correct method since Cxbx-R only loaded emulated sections into xbox virtual address.

Merging now.

PR #1531, opened 2019-01-23
Chihiro work

LukeUsher:
With this, the SegaBoot portion of Chihiro firmware can boot in Cxbx-Reloaded.

Notes:
 - Chihiro media board rom (fpr21042_m29w160et.bin) must be located in EmuDisk\Chihiro
 - The media board rom contains two copies of SEGABOOT. if you open the media board rom as an Xbe, you'll get the old version. To try the newer revision of SEGABOOT, you will need to make a copy and trim off the first 1MB of data.
 - Currently it boots, but fails to detect (unemulated) Chihiro specific hardware. Work on this will continue/is on-going.

image

image

image

image

image


PatrickvL:
Thanks! Could you also copy the Chihiro ROM guidelines to our wiki (and/or in the readme)?

PR #1532, opened 2019-01-23
Fixup for HLE DSound's GetCaps

### LINKED ISSUE #1132: REQUEST: GetCaps' structure values from retail hardware.
I need IDirectSound GetCaps' DSCAPS information in order to emulate it correctly before start working on DSound fixes part3.

2DBuffers = ?
3DBuffers = ?
BufferSGEs = should be 2047 (Is already emulated)
MemoryAllocated = should be 0 by default. (Is already emulated)

I have some information and theory for 2D and 3D Buffers value. However it is not enough info and really need hardware's info. Can someone please help me with this simple task? All it need to do is create directsound then call it. That's it.

BONUS: It would be nice to know what values all 4 variables above has after calling DirectSoundUseFullHRTF and DirectSoundUseLightHRTF separately to prove my theory.

RadWolfie:
**!!DO NOT MERGE PLEASE!!**
**NOTE: 3 commits will be removed since being move into #1534**

The following fixes contains:
* DirectSound's GetCaps function for accurate dwFree2DBuffers and dwFree3DBuffers counters.
  * Research from experiment lle audio turns out GetCaps' values are from software end, not from hardware directly.
  * NOTE: SGE counter and memAlloc are not accurate. Hopefully titles are not relying on this information.

Partial fixed #1132

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
@@ -179,7 +179,7 @@ XTL::X_XFileMediaObject::_vtbl XTL::X_XFileMediaObject::vtbl =
  */
 
 // size of DirectSound cache max size
-#define X_DIRECTSOUND_CACHE_MAX 0x800
+#define X_DIRECTSOUND_CACHE_MAX 256 // Maximum size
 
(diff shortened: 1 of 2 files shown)


RadWolfie:
Titles not tested: (I'm testing it now)
* Mafia 
* Burnout

RadWolfie:
Temporary close for time being.

PR #1533, opened 2019-01-23
Fix SetVertexData implementations that have arguments of type short (int16_t)

PatrickvL:
D3D : SetVertexData[24]s always map their arguments to a floats in the range [-32768.0, 32767.0] (at least, that's what @mborgerson discovered in XQemu)

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -3882,21 +3882,12 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData2s)
 
 	float fa, fb;
 	
-	switch (Register) {
-		// Special case: If the input register is a color, don't transform!
-		// Test case: Halo
-		case X_D3DVSDE_DIFFUSE:
-		case X_D3DVSDE_SPECULAR:
-		case X_D3DVSDE_BACKDIFFUSE:
-		case X_D3DVSDE_BACKSPECULAR:
-			fa = a;
-			fb = b;
-			break;
-		default:
-			fa = a / 32767.0f;
-			fb = b / 32767.0f;
-			break;
-	}
+	// Test case: Halo
+	// Note : XQEMU recently verified that the int16_t arguments
+	// must be mapped to floats in the range [-32768.0, 32767.0]
+	// (See https://github.com/xqemu/xqemu/pull/176)
+	fa = (float)a;
+	fb = (float)b;
 
     EMUPATCH(D3DDevice_SetVertexData4f)(Register, fa, fb, 0.0f, 1.0f);
 }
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4237,25 +4228,12 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData4s)
...
-			break;
-		default:
-			fa = a / 32767.0f;
-			fb = b / 32767.0f;
-			fc = c / 32767.0f;
-			fd = d / 32767.0f;
-			break;
-	}
+	// See comment note in D3DDevice_SetVertexData2s
+	fa = (float)a;
```
RadWolfie: No C-style cast please.

Real example from my own casting mistake.
image

PatrickvL: Our codebase is full of these type of casts. If we're going to address that, than certainly in a separate pr

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3882,21 +3882,12 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData2s)
...
-			fb = b;
-			break;
-		default:
-			fa = a / 32767.0f;
-			fb = b / 32767.0f;
-			break;
-	}
+	// Test case: Halo
+	// Note : XQEMU recently verified that the int16_t arguments
+	// must be mapped to floats in the range [-32768.0, 32767.0]
```
JayFoxRox: It's not verified. It's just a very strong assumption.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3882,21 +3882,12 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexData2s)
...
-		case X_D3DVSDE_BACKDIFFUSE:
-		case X_D3DVSDE_BACKSPECULAR:
-			fa = a;
-			fb = b;
-			break;
-		default:
-			fa = a / 32767.0f;
-			fb = b / 32767.0f;
-			break;
-	}
```
JayFoxRox: This old logic makes more sense than the new one. Except that the conditions are inverted.

I still hope that someone actually runs a test on Xbox.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie:
```asm
; 91   : 	short ss = -123;
	mov	edx, -123				; ffffff85H
	mov	WORD PTR _ss$[ebp], dx
; 92   : 	float f = (float)ss;
	movsx	eax, WORD PTR _ss$[ebp]
	cvtsi2ss xmm0, eax
	movss	DWORD PTR _f$[ebp], xmm0
; 93   : 	f = 10.5f;
	movss	xmm0, DWORD PTR __real@41280000
	movss	DWORD PTR _f$[ebp], xmm0
```

I can confirm without cast is converted correctly.

PatrickvL:
@RadWolfie reports these titles contain SetVertexData#s :
"
@PatrickvL, letting you know it's really hard to find titles with 2s and 4s functions for SetVertexData function names.
These three titles are what I found has them.
Halo
Airforce Delta Storm
Steel Battalion - Line of Contact

However, none of these three are able to progress to menu. (Beside Halo 1 just show black screen)
"

JayFoxRox:
If you are looking for affected games, you can just use `reset && ./xqemu-compatibility-inspector.py  | grep VERTEX_DATA2S -C 10`

It currently returns: `["Baldur's Gate: Dark Alliance II", 'Steel Battalion: Line of Contact']` but by looking in the history you can probably also find the other games you have listed (as we removed many test results in the past to focus on the official repository).

Script is here: https://github.com/JayFoxRox/xbox-game-databases
Patches to support looking at sheet history and to make it more robust are welcome.

There's a lot more games that were known to use the 2S variant too (Conker being one of those games as far as I can remember). We never hit any assert for the 4S variant in the latest revisions of XQEMU, so it's possible that it's largely unused or used in engines which also trigger other issues first.

Unit tests would be good. [Even nv2a-trace could be hacked to do these](https://github.com/JayFoxRox/nv2a-trace/pull/1).

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: I'm dismissing this request for now, for 2 reasons:
1. like stated above, our code-base is currently full of C-style casts - changing that is better done in a separate PR
2. having an explicit cast to `float` here, shows the intent more clearly than without it.

PR #1534, opened 2019-01-25
Hotfix: Audio non-stop loop bug

RadWolfie:
Mainly fix toward audio non-stop loop bug when it should be stopped or "released" by EG's input. Affected titles using Envelope Generator, aka EG, which is part of DSP. However, it does not include fade-out or any special EG values throughout the audio. Think of it as a hotfix.

The following fixes contains:
* DirectSoundBuffer's StopEx
  * Implement envelope's release stage timer, however does not have a fade out volume.
* DirectSoundStream_FlushEx
  * Include envelope's release stage timer, also does not include fade out volume.

This pull request is to separate #1532 due to longer ongoing investigation for GetCaps.


RadWolfie:
Last known titles has non-stop looping audio are:
* Virtua Cop 3 (Chihiro)
* Max Payne 1

PR #1535, opened 2019-01-25
Pgraph plugin hle state

PatrickvL:
This a only a small PR, adding mostly comments to our code that could be valuable to others.
The only real thing this does is also quite insignificant : implement `D3DDevice_SetRenderState_EdgeAntiAlias`.

I'd suggest to do a quick review and decide if this is worth to keep or just throw away.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -5999,11 +5999,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetRenderState_EdgeAntiAlias)
 	XB_trampoline(VOID, WINAPI, D3DDevice_SetRenderState_EdgeAntiAlias, (DWORD));
 	XB_D3DDevice_SetRenderState_EdgeAntiAlias(Value);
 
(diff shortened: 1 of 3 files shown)


LukeUsher:
Please note that reading RenderStates directly from NV2A constants will not work properly (or at all, in some cases) until the following is true:

1. We call unpatched Draw functions and handle drawing entirely based on NV2A Command Streams  
OR: 2. We call unpatched SetStateVB and SetStateUP during draw functions to flush the pending state.

Point 2 is a problem right now, since the SetStateVB/SetStateUp functions are members of the D3DDevice object, and we do not currently have a mechanism to properly preserve the pThis pointer during the call chain. Additionally, our existing symbols for these functions are not sufficient: I spent some time on this in the past, and this is the key reason why it has not been delivered.

LukeUsher:
That's my point: The pgraph commands are not queued, they're not even generated until the SetStateVb/SetStateVb calls: These are the flush functions .They read from the D3D Lazy State structures and generate the pgraph commands from them,

PatrickvL:
Ah, like that. But even so, there's still a distinction between making sure `SetState[VB|Up]` are called, and making sure all push-buffer contents up until `HLE_draw_state_update` is called, are handled by our GPU emulation code.

PR #1537, opened 2019-01-25
Remove x64 target from MSVC Solution

LukeUsher:
Cxbx-Reloaded currently is not able to be compiled as a 64-bit binary,
in part due to its method of direct code execution.

To avoid confusion, this commit removes the x64 build target from the
solution.


LukeUsher:
Low risk/minimal impact change, so merging

x1nixmzeng:
This is a lot cleaner :+1:

PR #1538, opened 2019-01-28
Functional LLE USB

ergo720:
This fixes the last bugs in LLE USB so that it finally works now (tested with JSRF and the dashboard). However, it’s still disabled for now because it requires a new input GUI to be able to select and configure the LLE devices attached to the virtual xbox. I’m doing it, but it’s taking far too much time than I first anticipated and I don’t want to delay this any longer.


LukeUsher:
Thanks, I'll merge this now as since as long as it's disabled, it's low risk.
We can work on making it available to users over time.

PatrickvL:
LOL, I was about to reply "LGTM", but Luke beat me to it with a merge!

jthecodebuilder:
I have updated the wiki's FAQ to compensate for this pull request. Your welcome! 😁

Issue #1064, opened 2018-04-15
CreateTexture error in Farcry


PatrickvL:
See subject and https://github.com/Cxbx-Reloaded/game-compatibility/issues/62#issuecomment-381385105 fit details.

PatrickvL:
This sheds some light on this issue : https://www.gamedev.net/forums/topic/471600-depth-stencil-buffer-textures-only-supported-on-nvidia-cards/

It might be due to a missing D3DUSAGE_DEPTHSTENCIL flag, or the GPU driver doesn't support this format (ATI seems to have specific four-cc formats that contain depth stencil data, so that could be a solution for that specific vendor)

kennethedmonds:
I found that Tony Hawk Underground 2 displays 
```
---------------------------
Cxbx-Reloaded
---------------------------
Received Fatal Message:

* CreateTexture Failed!

Error: 0x8876086C
Format: 75
Dimensions: 256x256

---------------------------
OK   
---------------------------
```
And I have a 1080ti if that helps confirm anything for you.

JohnGodgames:
**Finding Nemo**
```

---------------------------
Cxbx-Reloaded
---------------------------
Received Fatal Message:

* CreateTexture Failed!

Error: 0x8876086C
Format: 75
Dimensions: 128x128

---------------------------
OK   
---------------------------
```

gandalfthewhite19890404:
WWE RAW2

Received Fatal Message:

* CreateTexture Failed!

Error: 0x8876086C
Format: 75
Dimensions: 512x512

kennethedmonds:
BLiNX: the time sweeper
```
---------------------------
Cxbx-Reloaded
---------------------------
Received Fatal Message:

* CreateTexture Failed!

Error: 0x8876086C
Format: 75
Dimensions: 640x480

---------------------------
OK   
---------------------------
```

ObiKKa:
Persistent error in [Crimson Skies](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1169#issuecomment-389338503).

NZJenkins:
If anyone feels like some testing #1541 may help with some of these

real-guilty:
Dead or Alive Ultimate

> Received Fatal Message:
> 
> * CreateTexture Failed!
> 
> Error: 0x8876017C
> Format: 21
> Dimensions: 1024x1024

[...]

PR #1541, opened 2019-02-01
Check surface parent format matches surface format before creating it

NZJenkins:
Outrun 2006 has an issue where both surfaces passed to SetRenderTarget have the same parent
And one surface matches the parent format, and one doesn't
Might help with some of #1064 but needs testing


PatrickvL:
Testers needed!

gandalfthewhite19890404:
What games are needed?

PR #1544, opened 2019-02-05
nv2a: Fix A8Y8 texture formats

PatrickvL:
Cherry-picked from https://github.com/xqemu/xqemu/pull/194, courtesy JayFoxRox (well done!)

I had a choice : Either do this now, or later (in which case we'd probably forget about it, re-discover it, and would have to do a lot of manual labor to compare if we're on-par with xqemu's code).

Ideally, we'd devise a solution (using sub-modules for example) that would automatically keep both sides in-sync. Alas, we're not there yet, so we'll have to live with these type of cherry-picks.

### DIFF
--- a/src/devices/video/EmuNV2A_PGRAPH.cpp
+++ b/src/devices/video/EmuNV2A_PGRAPH.cpp
@@ -191,7 +191,6 @@ static GLint gl_swizzle_mask_0RG1[4] = { GL_ZERO, GL_RED, GL_GREEN, GL_ONE };
 static GLint gl_swizzle_mask_111R[4] = { GL_ONE, GL_ONE, GL_ONE, GL_RED };
 static GLint gl_swizzle_mask_ARGB[4] = { GL_ALPHA, GL_RED, GL_GREEN, GL_BLUE };
 static GLint gl_swizzle_mask_BGRA[4] = { GL_BLUE, GL_GREEN, GL_RED, GL_ALPHA };
-static GLint gl_swizzle_mask_GGGR[4] = { GL_GREEN, GL_GREEN, GL_GREEN, GL_RED };
 static GLint gl_swizzle_mask_R0G1[4] = { GL_RED, GL_ZERO, GL_GREEN, GL_ONE };
 static GLint gl_swizzle_mask_RRR1[4] = { GL_RED, GL_RED, GL_RED, GL_ONE };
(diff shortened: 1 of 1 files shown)


JayFoxRox:
> I had a choice : Either do this now, or later (in which case we'd probably forget about it, re-discover it, and would have to do a lot of manual labor to compare if we're on-par with xqemu's code).

Sorry, but I strongly disagree.

There's `git history` per directory and `git blame`: It's not exactly easy to "forget about it".

It also isn't true because Cxbx-R isn't "on-par" as-is: Cxbx-R is ahead. It's ahead with own features, and also unreviewed PRs. Code changes in review, so accepting it ahead of time in Cxbx-R makes it impossible to be "on-par" in any way imaginable.

It's stupid that Cxbx-R as a downstream is always ahead commit wise. XQEMU has no room to grow, if development happens like this. 

This is such a hostile-fork by now; it's really making me furious.

I know that our license allows this - but that's a legal thing.
Ethically, this is just very bad.

> Ideally, we'd devise a solution (using sub-modules for example) that would automatically keep both sides in-sync. Alas, we're not there yet, so we'll have to live with these type of cherry-picks.

Right.. But a good temporary alternative has been suggested in the past:

* To cherry-pick from master after merges (not from PRs), and maybe with a delay (to not put pressure on the upstream review).
* Otherwise, refuse new ontributions to your copy of the code (as small as they might be).
* Send existing changes upstream (which was promised a couple of times, but never done).
* Actively direct people upstream, so that contributions flow one direction. Such as informing about the upstream in the README (with a call for participation), when this is [probably relatively large part of the origin, compared with other projects mentioned](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded#special-thanks) (also XQEMU is still active, so it's much more important to reference it).

*(I'm not sure how "copy upstream PR & hit merge ASAP" is still an acceptable option)*

LukeUsher:
Yeah I didn't really think this one through,

LukeUsher:
To clarify: Going forward, we will no longer accept pull-requests consisting of things copied from XQEMU. 

The LLE nv2a code in Cxbx-R doesn't really work as is, and porting changes to Cxbx-R poses no real benefit to us, but actively harms/discourages the XQEMU developers from contributing to any project.
 (What's the point contributing to XQEMU if the changes get approved here before in XQEMU itself, it's not good)

PR #1548, opened 2019-02-09
Include Verbose HLE Audio Log

RadWolfie:
Small work for implement Xbox DSound and Windows' DirectSound toward better verbose log. Rest of the work will be perform later.


gandalfthewhite19890404:
What log level is - Debug?

PatrickvL:
How about using AppVeyor/TravisCI on your own fork so this WIP doesn't pollute the PR list?

PR #1549, opened 2019-02-10
Minor DSound create fix

RadWolfie:
Only a minor fix up plus attempted to fill in the blanks for now. It is considered as a workaround hack fix in order for the host's DirectSound allocate the class.


RadWolfie:
User from discord reported the Hobbit has progress further, yet do have different issue.

However, commit c9db98d isn't a perfect solution. I'm thinking lpwfxFormat isn't used even though titles are passing allocated format. X_BufferSizeRequest argument aka `dwBufferBytes` can have zero size at buffer class creation.

PatrickvL:
How about using AppVeyor/TravisCI on your own fork so this WIP doesn't pollute the PR list?

Issue #769, opened 2017-10-19
Reserving memory ranges


PatrickvL:
As long as Cxbx-Reloaded runs code directly, it'll be needed to gain access to certain memory address ranges. This issue goes into detail on this subject.

Our prerequisite is, we must be running as a 32 bit process under 64 bit Windows (thus, relying on wow64).

To be able to run Xbox code directly, memory ranges that must be available are:
0x00010000 : 64 mb, for loading Xbox xbe's (128 mb for Chihiro)
0x80000000 : 64 mb, for kernel stub and contiguous memory shared with other devices
? : Tiled memory
0xF8000000 : 64 mb for various mmio ranges (TODO: split up per device in more detail)

Additionally, titles might request specific address ranges, some of those might assume these requests succeed.

The emulation itself, the DLL's used for it, and memory allocations during run time occupy address ranges too.

Previously, Cxbx was composed of 2 parts : the GUI and an emulation DLL.
The GUI was loaded at 0x00010000, the DLL somewhere else.
When running emulation, the GUI would be launched in "emulation-mode", handing over control to the emulation DLL. Xbe's were loaded by overwriting the GUI code.
This led to unstable and hard to debug code.

Nowadays, Cxbx-Reloaded is a single executable, stripped from relocation information, set to a fixed load address of 0x00010000. The first 128 mb of the executable is occupied by a memory placeholder. Following that, both the GUI and the emulating code are linked into this one big single executable.
This setup makes it possible to load xbe's at their designated addres, while keeping all code together.
The downside of this setup is, linking times are high, the executable contains 128 mb of otherwise useless space, and both the GUI as the emulation environment contain eachothers code, even though they don't use eachother.

(TODO: Move and expand the above into the appropriate wiki article, just link to that here).

Going forward, we envision a setup that would allow better control over available memory, and removes most of the problems we have right now. It works like this:

We'll split up the GUI from the emulation again, and add a loader into the mix.

The emulation code will be put in a DLL once again.
The GUI will be a separate executable, sharing only the configuration data with the emulation DLL.
The loader will be a tiny executable, loaded at address 0x00010000.

The GUI will stay the main executable.
The loader executable will arrange memory reservation and load the emulation DLL.
The emulation DLL will bootstrap our kernel and run emulation.

Since the tiny loader will reside at 0x00010000, the few pages occupied by it are already reserved.
[...]

PR #1552, opened 2019-02-12
Introducing near absolute control over memory ranges (WIP)

PatrickvL:
This introduces the basics for #769 by introducing 3 new projects (a loader executable, an emulation DLL and an empty GUI executable) without actually making use of any of it.

This setup will allow Cxbx-Reloaded to have near-absolute control over memory ranges under Windows (and probably Wine nearly as well).

This work is not being used just yet, but the code is now at a point that it'd be helpfull to merge it into develop, so others can start contributing (like the required changes to the memory manager classes, for example). 
(EDIT : Per @LukeUsher request, this PR is meanwhile updated to target a separate 'loader' branch)

Note, that existing functionality should NOT be changed, so if the existing Cxbx.exe still works, it'll be alright to merge.

Reviews & remarks welcome!


LukeUsher:
Since this is currently unusable (and adds multiple projects) it could be confusing to new contributors. 

Rather than merge into 'develop', may I suggest maintaining a 'loader' branch here on upstream Github?

That way work can continue, it can be kept in-sync with develop, and once finalized, it can be merged as a whole completed piece.

I'll do a full code-review later: I'm pretty busy for the rest of today.

build/win32/Cxbx/Cxbx.vcxproj
```diff
@@ -1,571 +1,575 @@
...
-    </ProjectReference>
-  </ItemGroup>
-  <ItemGroup>
-    <Image Include="..\..\resource\Cxbx-R.ico" />
-    <Image Include="..\..\resource\Logo-License-CC4.bmp" />
-  </ItemGroup>
-  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-  <ImportGroup Label="ExtensionTargets">
-  </ImportGroup>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
LukeUsher: What happened to the line-endings in here? It might be a minor thing, but this will introduce merge conflicts with other, ongoing work (such as the Chihiro work).

If the line-endings did not change, then Git would be able to figure out the merges easily.

build/win32/Cxbx/Cxbx.vcxproj.filters
```diff
@@ -1,836 +1,848 @@
...
-      <Filter>Emulator</Filter>
-    </ClInclude>
-    <ClInclude Include="..\..\src\core\kernel\exports\EmuKrnlKe.h">
-      <Filter>Kernel</Filter>
-    </ClInclude>
-    <ClInclude Include="..\..\src\common\Timer.h">
-      <Filter>Emulator</Filter>
-    </ClInclude>
-  </ItemGroup>
+<?xml version="1.0" encoding="utf-8"?>
```
LukeUsher: Same here, re: line-endings

src/common/AddressRanges.h
```diff
@@ -0,0 +1,116 @@
...
+	RANGE_ENTRY(0xFE800000, 0xFE87FFFF, KB(512), PROT_NAC, SYSTEM_ALL              , "DeviceAPU"),
+	RANGE_ENTRY(0xFEC00000, 0xFEC00FFF, KB(  4), PROT_NAC, SYSTEM_ALL              , "DeviceAC97    (ACI)"),
+	RANGE_ENTRY(0xFED00000, 0xFED00FFF, KB(  4), PROT_NAC, SYSTEM_ALL              , "DeviceUSB"),
+	RANGE_ENTRY(0xFEF00000, 0xFEF003FF, KB(  1), PROT_NAC, SYSTEM_ALL              , "DeviceNVNet"),
+	RANGE_ENTRY(0xFF000000, 0xFF3FFFFF, MB(  4), PROT_NAC, SYSTEM_ALL              , "DeviceFlash_a (Flash mirror 1)"),
+	RANGE_ENTRY(0xFF400000, 0xFF7FFFFF, MB(  4), PROT_NAC, SYSTEM_ALL              , "DeviceFlash_b (Flash mirror 2)"),
+	RANGE_ENTRY(0xFF800000, 0xFFBFFFFF, MB(  4), PROT_NAC, SYSTEM_ALL              , "DeviceFlash_c (Flash mirror 3)"),
+	RANGE_ENTRY(0xFFC00000, 0xFFFFFFFF, MB(  4), PROT_NAC, SYSTEM_ALL    | MAY_FAIL, "DeviceFlash_d (Flash mirror 4) Optional (will probably fail reservation, which is acceptable - the 3 other mirrors work just fine"),
+	RANGE_ENTRY(0xFFFFFE00, 0xFFFFFFFF,    512 , PROT_NAC, SYSTEM_RETAIL | MAY_FAIL, "DeviceMCPX    (not Chihiro, Xbox - if enabled) Optional (can safely be ignored)"),
```
LukeUsher: This region is only relevant if we were running the original Xbox boot sequence (including MCPX), so it's completely redundant to use it: By the time the Kernel has started execution, this region is disabled and cannot be re-enabled. Xbox software (and the kernel) have no access to this region whatsoever once 2BL has completed.

src/common/ReserveAddressRanges.cpp
```diff
@@ -0,0 +1,85 @@
...
+// *  All rights reserved
+// *
+// ******************************************************************
+
+#include "AddressRanges.h"
+
+// Reserve an address range up to the extend of what the host allows.
+bool ReserveMemoryRange(int index)
+{
+	unsigned __int32 Start = XboxAddressRanges[index].Start;
```
LukeUsher: <cstdint> is a header-only include that has no external library or runtime requirements, so you could include that and avoid using Microsoft compiler specific types.

src/common/win32/EmuShared.cpp
```diff
@@ -106,7 +108,7 @@ void EmuShared::Init(DWORD guiProcessID)
         );
 
         if(g_EmuShared == nullptr)
-            CxbxKrnlCleanupEx(CXBXR_MODULE::INIT, "Could not map view of shared memory!");
+            return false; // CxbxKrnlCleanupEx(CXBXR_MODULE::INIT, "Could not map view of shared memory!");
```
LukeUsher: Couldn't this derive from the Error class in order to retain specific error messages using GetError()?

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -1626,7 +1626,7 @@ DWORD WINAPI XTL::EMUPATCH(XLaunchNewImageA)
 
 			if (PathFileExists(szDashboardPath))
 			{
-				MessageBox(CxbxKrnl_hEmuParent, "The title is rebooting to dashboard", "Cxbx-Reloaded", 0);
+				(void)CxbxMessageBox("The title is rebooting to dashboard", MB_OK, CxbxKrnl_hEmuParent);
```
LukeUsher: Why the cast to void?

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -499,6 +504,18 @@ HANDLE CxbxRestorePageTablesMemory(char* szFilePath_page_tables)
 
 #pragma optimize("", off)
```
LukeUsher: Did you forget to re-enable optimization?

src/loader/cxbxr-ldr.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+#define EMULATION_DLL "cxbxr-emu.dll"
+
+DWORD CALLBACK rawMain()
+{
+	(void)virtual_memory_placeholder; // prevent optimization removing this data
+
+	// First detect if we are running on WoW64, if not, prevent Cxbx-Reloaded from starting
+	// Cxbx-Relaoded needs access to high memory, only exposed to WoW64.
```
LukeUsher: Cxbx-Relaoded? ;)

LukeUsher:
image

Regarding those few files that are interpreted as binary: On inspection, those files start with a unicode BOM character, and they appear to have been saved as utf-16 rather than utf-8 like the rest of our codebase. You can clearly see the Byte-Order-Mark (FF EE) as well as each character taking up two bytes, rather than one.

Resaving these files in utf-8 (or ascii) encoding should resolve the problem.

src/core/hle/XAPI/Xapi.cpp
PatrickvL: To be explicit that the return value is ignored. (Validators might complain otherwise)

src/core/kernel/init/CxbxKrnl.cpp
PatrickvL: Never turned it off in this branch...

src/loader/cxbxr-ldr.cpp
PatrickvL: Good catch, I'll search&replace all

src/core/kernel/init/CxbxKrnl.cpp
LukeUsher: Sorry, if this is required it should be commented as to why... best leave as is for now.

build/win32/Cxbx/Cxbx.vcxproj
PatrickvL: Argh! Unintentional differences in line-endings, and binary encoding ... why all of a sudden?!?

src/common/AddressRanges.h
PatrickvL: Good point, I'll disable it with your remark as a comment.

src/common/ReserveAddressRanges.cpp
PatrickvL: I'll use that

src/common/win32/EmuShared.cpp
PatrickvL: I wouldn't bother, as current callers already show similair error messages, plus recent discussions seem to indicate EmuShared will be reworked anyway.

build/win32/Cxbx/Cxbx.vcxproj
PatrickvL: Fixed binary files by saving them as UTF-8.

Line-ending differences however, appear not the cause for the big diff . I did try unix2dos and dos2unix, saving without UTF-8 BOM, none of which solved it. Maybe it's because of the move?

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -354,13 +354,18 @@ HANDLE CxbxRestoreContiguousMemory(char *szFilePath_memory_bin)
...
+#endif
+
 	// Map memory.bin contents into memory :
 	void *memory = (void *)MapViewOfFileEx(
 		hFileMapping,
 		FILE_MAP_READ | FILE_MAP_WRITE | FILE_MAP_EXECUTE,
 		/* dwFileOffsetHigh */0,
 		/* dwFileOffsetLow */0,
-		CONTIGUOUS_MEMORY_CHIHIRO_SIZE,
+		CHIHIRO_CONTIGUOUS_MEMORY_SIZE,
```
RadWolfie: For what reason this needs to be rename?
```
CONTIGUOUS_MEMORY_RETAIL_SIZE
CONTIGUOUS_MEMORY_DEBUGKIT_SIZE
CONTIGUOUS_MEMORY_CHIHIRO_SIZE
```
The above does make thing easier to find as an example. Or put retail / debugkit / chihiro after _SIZE.

RadWolfie:
I'm thinking... maybe perform a squash & merge method once review and fixes are complete. Then the rebase wouldn't have a problem with previous commits, from this PR, produce merge/rebase conflict.

src/core/kernel/init/CxbxKrnl.cpp
PatrickvL: Mainly to conform to all other occurences of `CHIHIRO_` as a prefix in \#defined symbols

PatrickvL:
Status : Awaiting more review

RadWolfie:
Recommendation by discord user if didn't see his message:
> Line endings are a non-issue if the repo contains a properly configured .gitattributes file

x1nixmzeng:
I've read over the review without testing and have a few points

* The CI script needs updating (it won't include the new binaries)
* The win64 config should be removed from the project
* Confusing number of new macros - KB()/MB() as functions are and ONE_KB/ONE_MB values
* Mixture of types - cstdint is included somewhere, yet most typenames are using `int` or even `__int`

PatrickvL:
@x1nixmzeng Good points, if possible I'll try to conform the new code more to what's being used in existing code.

As for testing: don't even bother to try, as there's additional work needed to get the existing memory classes to play nice with the now-reserved address ranges. (I'm hoping @ergo720 will join in at some point for that)

build/win32/Cxbx/Cxbx.vcxproj
PatrickvL: I checked again using Beyond Compare;

OLD :
Path : Cxbx-Reloaded\build\win32\Cxbx.vcxproj 
Encoding : UTF-8 BOM
Newline : CR+LF

NEW
Path : Cxbx-Reloaded\build\win32\Cxbx\Cxbx.vcxproj 
Encoding : UTF-8 BOM
Newline : CR+LF

Text Compare
312 same line(s)
4 important right orphan line(s)
259 important difference line(s)

31 difference section(s)

Conclusion : Github shows a huge amount of differences, not because of encoding or newine style changes, but because of the move into a sub-directory, all relative file paths got an additional "../" inserted.

PatrickvL:
On 2nd thought, using a 3rd branch (instead of just merging this into `develop`) might not be wise, given that all active deverlopers won't be confused by what's going on here...

LukeUsher:
The whole idea of adding another branch is to make sure that this work won't be merged into develop until it is 100% tested, and working, with no known regressions, and that we don't want conflicts with currently in-progress work.

After the new branch for the loader is stable, tested, and confirmed good, then it can be merged into develop as one piece.

PatrickvL:
This intend of this branch is to have no functional impact to the existing Cxbx project. Not now on it's introduction, nor anytime during development continues on the loader-approach. This will lead to a moment in time where the three-projects setup reaches or surpasses the functionality of the single-executants setup. At that point, users can switch over. We could keep the single-executable setup around for some time after that, then mark it deprecated and finally remove it.

Would all this be done in a separate branch, I fear it would be harder to keep the two branches in-sync.
Also, having it in a separate branch wouldn't motivate contributors to get this work finished as soon as possible.

RadWolfie:
To be honest, moving the existing project files does cause problem with develop and wip branches.

I believe the fixup commits should be append to broken commits in order to reduce amount of mistakes.

In my opinion, this pull request should be close then create a new pull request after the commits are corrected.

PatrickvL:
A squash merge would avoid the proposed fixing of broken (how?) commits.

The moving of the Cxbx project file needs to happen (to avoid projects illegally sharing compiled object file folders), so it might be best to take the pain once, and sooner rather than later.
It's still possible to wait for the few work in progress branches to be delivered (to my knowledge, there's currently just one being worked on, by ergo720)

RadWolfie:
Sent a PM about how to fix the broken commits.

If added the eol fix to `.gitattribute` file along with the file's eol correction. Then squash merge will resolve eol issue. However, it doesn't help with the massive changes in existing project files unfortunately.

If we are using cmake, then we wouldn't have to worry about generated project files at all.

x1nixmzeng:
OT: If you're interested in moving to generated project files I'll see what I can do

PatrickvL:
This work is awaiting cmake refactoring

PatrickvL:
Replaced by PR #1565

RadWolfie:
cmake progress is in near complete give or take within 2 days. I'm having two different issues at the moment which I'm not sure what went wrong. Though, I'll publish the branch some time after 8 hours of this comment.

PR #1553, opened 2019-02-14
Fix when titles use pushbuffers rather than SetVertexShaderConstant to update shader registers 

LukeUsher:
This improves Conflict Desert Storm 2, Headhunter: Redemption and Red Faction 2.


RadWolfie:
Even Star Wars Lego II start to show corrupted result then no characters visible. However, if I move, these visible remain at the same coordinate in the screenshot below.
image

PatrickvL:
This is somewhat timing sensitive, because since we don't actively flush pending push buffer commands yet, this code might be reading data that's about to be overwritten.

literalmente-game:
> This is somewhat timing sensitive, because since we don't actively flush pending push buffer commands yet, this code might be reading data that's about to be overwritten.

It is, you can take ``Scarface`` as a good test case, under the right conditions the entire screen will render perfectly.

RadWolfie:
I'm not sure if this PR is a solution or simply a workaround hack fix base on previous comments above. If it is not a workaround hack fix, then feel free to merge it.

Also, I had not seen any bad side effect with some titles by the way.

PatrickvL:
Merging as it's already an improvement, the forced push buffer command queue flush will have to come later

Issue #1559, opened 2019-03-04
Legal issues and broken logic in EmuRSA.cpp


JayFoxRox:
I recently noticed some breaking bugs in https://github.com/XboxDev/xbedump/issues/9 which affect giant.c; a variation of that code has found its way into src/common/crypto/EmuRSA.cpp (initially contributed to Cxbx-R by @ergo720).

@ergo720 took the code without the license text or copyright mentions. The license, however, is rather restrictive: it affects monetary payout rules, so it's risky actually.
The code only claims that it came from XQEMU and that it's GPLv2 - which is false. The repository was only hosted / archived on the XQEMU GitHub, but we never (to my knowledge) put a GPL on it.

Although the source project itself doesn't have that explicit copyright header on the files, [it has a generic README](https://github.com/rudimeier/mprime/blob/master/gwnum/readme.txt#L76):

> Copyright (c) 1996-2014, Mersenne Research, Inc.  All rights reserved. 
> 
> Redistribution and use in source and binary forms, with or without
> modification, are permitted provided that the following conditions are 
> met: 
> 
> (1) Redistributing source code must contain this copyright notice,
> limitations, and disclaimer. 
> (2) If this software is used to find Mersenne Prime numbers, then
> GIMPS will be considered the discoverer of any prime numbers found
> and the prize rules at http://mersenne.org/prize.htm will apply.
> 
> THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
> "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
> LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
> A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
> OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
> SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
> LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
> DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
[...]

PR #1560, opened 2019-03-04
Fix issue #1559

ergo720:


JayFoxRox:
I think this PR is a bad workaround for this problem.

The EmuRSA code has been there for a while, so a day or two don't hurt much either.
This PR also removes *a lot* of non-affected code.

Instead, I'd suggest to find a solution which is also compatible with the approach proposed in #1561, just by replacing the RSA code / MPI functions. This should happen *soon*, but does not have to happen within 24 hours either. I think if we can find an alternative by the end of this week, that's fine.

*There's also other minor copyright violations, where code was simply taken with a "Credit" line (ethics) - that's not how copyright works (law); this should be addressed. A link to a project is also pretty useless when the overall project license might differ from the license of individual files.*

PatrickvL:
Although very admirable how quickly you responded to this (and indeed, something should to be done to avoid any potential legal issues), it might be better to take a little more time to investigate if any replacements can be found. This problem has been around a few months already, so adding a few days doesn't really matter much (until proven otherwise, of course).

PR #1561, opened 2019-03-04
Implement `XcModExp`

JayFoxRox:
**This conflicts with #1560, however, I still think this should be accepted first.**

---

XcModExp will be required for system link. This PR also implements a re-usable design of the RSA MPI modular exponentiation for this function.

This implementation has been confirmed using system-link by connecting Cxbx-Reloaded to XQEMU and a real Xbox. I have also passed the tests (and more advanced ones I've created myself) in the xbox_kernel_test_suite.

---

First, I added `const` to relevant pointers, which acts as a hint to how these function work, and also increases code safety.

I then moved the `RSAdecrypt` into a custom `ModExp` function, which can be re-used by `XcModExp` emulation.

I was unsure about the return value, but chose to use 1, which is what the test suite expects.

*(My line endings might be bad - this is because my MSVS might not be configured correctly. Please tell me which configuration I should use.)*

### DIFF
--- a/src/common/crypto/EmuRsa.cpp
+++ b/src/common/crypto/EmuRsa.cpp
@@ -18,6 +18,7 @@
 // *  59 Temple Place - Suite 330, Bostom, MA 02111-1307, USA.
 // *
 // *  (c) 2018 ergo720
+// *  (c) 2019 Jannik Vogel
 // *
 // *  All rights reserved
 // *
@@ -236,11 +237,6 @@ giant newgiant(int numshorts)
 	return(thegiant);
 }
 
-// ergo720: there's a bug in the original implementation of gigimport that prevents the rsa algorithm from working correctly on Windows (but
-// for some reason it works on Ubuntu). This replacement implementation is taken from
-// http://xbox-linux-devel.narkive.com/Qw6o31DP/xbedump-fix-for-array-out-of-bounds-access#post1
-// and it has been improved based on JayFoxRox suggestions. See the link below for the details
-// https://github.com/xqemu/xbedump/pull/5
 void gigimport(giant g, const unsigned char *buff, int len) {
 
 	// copy buffered 'number' into giant's number buffer
@@ -248,7 +244,13 @@ void gigimport(giant g, const unsigned char *buff, int len) {
 
 	assert((len % 2) == 0);
 
+	// Get number of shorts
 	g->sign = len / 2;
+
+	// Only count used shorts
+	while((g->sign >= 1) && (g->n[g->sign - 1] == 0)) {
+		g->sign -= 1;
+	}
 
 	assert(g->sign != 0);
 }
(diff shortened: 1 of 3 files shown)


src/common/crypto/EmuRsa.cpp
```diff
@@ -151,27 +152,27 @@ void iaddg(int i, giant g); /* g += i, with i non-negative and < 2^16. */
...
-	/* x := x^n (mod z). */
-	powermodg(sig, e, n);
+	memcpy(a_number, b->n, d_len);
+}
 
-	memset(cryptbuffer, 0x00, 256);
-	memcpy(cryptbuffer, sig->n, 256);
+void RSAdecrypt(const unsigned char* c_number, unsigned char* cryptbuffer, RSA_PUBLIC_KEY key)
+{
+	ModExp(cryptbuffer, c_number, 256, key.KeyData.Exponent, 4, key.KeyData.Modulus, 256);
```
JayFoxRox: I'm not sure if I verified the functionality of this. Please review.

src/common/crypto/EmuRsa.cpp
```diff
@@ -151,27 +152,27 @@ void iaddg(int i, giant g); /* g += i, with i non-negative and < 2^16. */
...
+void ModExp(unsigned char* a_number, const unsigned char* b_number, unsigned int b_len, const unsigned char* c_number, unsigned int c_len, const unsigned char* d_number, unsigned int d_len)
 {
-	giant n = newgiant(GIANT_INFINITY);
-	giant e = newgiant(GIANT_INFINITY);
-	giant sig = newgiant(GIANT_INFINITY);
-
-	gigimport(sig, c_number, 256);
+	giant b = newgiant(GIANT_INFINITY);
+	giant c = newgiant(GIANT_INFINITY);
+	giant d = newgiant(GIANT_INFINITY);
```
JayFoxRox: I believe we could theoretically reduce the number of shorts to `_len / 2`, however, I decided to not do that, to keep the code simple. The existing code was also just using `GIANT_INFINITY`.

src/core/kernel/exports/EmuKrnlXc.cpp
```diff
@@ -212,9 +214,10 @@ xboxkrnl::ULONG NTAPI JumpedModExp
 	xboxkrnl::ULONG dwN
 )
 {
-	ULONG ret = 0;
-
-	LOG_UNIMPLEMENTED();
+	ULONG ret = 1;
+
+	unsigned int len = dwN * 4;
```
JayFoxRox: Alternative: `dwN * sizeof(DWORD)`

src/common/crypto/EmuRsa.cpp
LukeUsher: Seems to work fine: Cxbx-R would shout about invalid Xbe signatures if RSA verification failed otherwise.

LukeUsher:
Thanks!

PR #1563, opened 2019-03-05
Update copyright texts

ergo720:


src/core/kernel/exports/EmuKrnlKi.cpp
```diff
@@ -39,356 +39,10 @@
...
-Public License instead of this License.
-
-
-ReactOS may be used, runtime linked, and distributed with non-free software
-(meaning that such software has no obligations to open-source, or render
-free, their non-free code) such as commercial device drivers and commercial
-applications. This exception does not alter any other responsibilities of
-the licensee under the GPL (meaning that such software must still obey
-the GPL for the free ("open-sourced") code that has been integrated into
-the said software).
```
JayFoxRox: Couldn't this paragraph be relevant?

src/devices/usb/OHCI.cpp
```diff
@@ -25,8 +25,34 @@
...
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with this library; if not, see <http://www.gnu.org/licenses/>.
+*
+* TODO:
+*  o Isochronous transfers
+*  o Allocate bandwidth in frames properly
+*  o Disable timers when nothing needs to be done, or remove timer usage
+*    all together.
+*  o BIOS work to boot from USB storage
```
JayFoxRox: This TODO list seems unncessary for Cxbx-Reloaded / should be moved into issues.

What *should* be kept, is a commit revision of QEMU so people can find the history of this file + find out what needs to be updated in the future.

src/core/kernel/exports/EmuKrnlKe.cpp
```diff
@@ -27,7 +27,20 @@
 // ******************************************************************
 
 // Acknowledgment (timer functions): ReactOS (GPLv2)
-// https://github.com/reactos/reactos
+
+/*
+* PROJECT:         ReactOS Kernel
+* LICENSE:         GPL - See COPYING in the top level directory
```
JayFoxRox: Does not apply here probably?

src/common/util/CxbxUtil.cpp
```diff
@@ -25,8 +25,30 @@
...
+//	License(and no later version) are only accepted for the following files
+//	or directories: bsd - user / , linux - user / , hw / misc / vfio.c, hw / xen / xen_pt * .
+//
+//	3) The Tiny Code Generator(TCG) is released under the BSD license
+//	(see license headers in files).
+//
+//	4) QEMU is a trademark of Fabrice Bellard.
+//
+//	Fabrice Bellard and the QEMU team
+
```
JayFoxRox: Why is this here? This seems like it should be part of docs - not some random source file.

Instead, if you took copyrighted code, the headers of that files should be here and it should be clear which copyright / license applies to which parts.

src/common/crypto/EmuSha.cpp
```diff
@@ -43,8 +43,7 @@ A million repetitions of "a"
 /* #define LITTLE_ENDIAN * This should be #define'd already, if true. */
 /* #define SHA1HANDSOFF * Copies data before messing with it. */
 
-// Acknowledgment: see above
-// https://github.com/clibs/sha1
```
JayFoxRox: Why remove the link? This file is in the public-domain, so this old header is fine.
Having the source URL helps in finding the state of it + finding updates.

src/core/kernel/exports/EmuKrnlKe.cpp
ergo720: It applies, see new commit

src/devices/usb/OHCI.cpp
ergo720: 1.	Isochronous transfers are already implemented.
2.	Here, I think they are referring to section 3.4.2 (Bandwidth Allocation) of the OHCI standard. To my knowledge, indeed, the existing code doesn’t take into account any of this, it just transfers all packets until there are none left (or an error occurs). In practice, I’m not sure if it’s worth implementing since it seems to work fine even without.
3.	The only timer used in the code is the one that’s used to periodically run ```OHCI_FrameBoundaryWorker```, which is responsible to process the packets, so it’s required and cannot be removed. It’s only ever disabled during a reset  or suspend event or if an unrecoverable error interrupt is raised (aka a fatal error in OHCI terms), so this is already covered I’d say.
4.	This doesn’t apply to Cxbx-R.

PatrickvL:
Is discussion finished on this, so it can be merged?

JayFoxRox:
I never did a full review on this - I only skimmed over it and called out the biggest issues (turning them into examples, to fix a class of issues). I believe there's more stuff that's not fixed yet (in this PR, and possibly the codebase).

So even if the discussion is finished, it doesn't mean that the changes are correct or complete.
It should be the maintainers responsibility to make sure that copyright notices stay intact and remain correct (across the codebase).

PatrickvL:
Then maybe it's best to merge this and create an issue for the follow-up work

ergo720:
I don't think I have anything else to commit to this, so it could be merged now

RadWolfie:
It would be great to have all 3 fixup commits merge into "Update copyright texts" commit through `git rebase -i`.

Comparing to develop branch, looks better now.

src/core/kernel/exports/EmuKrnlKe.cpp
JayFoxRox: Why is there a copy of the COPYING file here then, if this claims people should look in `COPYING in the top level directory`: Which one counts? This paragraph (which is not in the top level) or the file in the top level?

I think both are kind of bad, because the COPYING file chosen by Cxbx-R might have to be changed eventually, whereas this code can't be easily relicensed.

It seems much easier to remove the reference to the COPYING file in the top level directory OR create a COPYING_REACTOS / COPYING-GPLV2 (or better named) or something to reference something stable.

*Also note that the GPLv2 text allows 2 variations of licensing: ~"GPLv2 or any later version" or ~"GPLv2 only". So usually the copyright header has to be explicit about this; I'm not sure how ReactOS handles this.*

src/core/kernel/memory-manager/VMManager.cpp
```diff
@@ -27,8 +27,12 @@
...
 // Acknowledgment:
 // The core logic of the VMManager class is based on the virtual page management code of the citra emulator (GPLv2 license),
-// with heavy changes and the addition of real page tables to better suit Cxbx-Reloaded and Xbox emulation.
-// Citra website: https://citra-emu.org/
+// with heavy changes and the addition of real page tables to better suit Cxbx-Reloaded and Xbox emulation.
+// https://github.com/citra-emu/citra
+
+// Copyright 2015 Citra Emulator Project
+// Licensed under GPLv2 or any later version
+// Refer to the license.txt file included.
```
JayFoxRox: This file does not exist.

src/devices/usb/Hub.cpp
```diff
@@ -25,8 +25,32 @@
 // *
 // ******************************************************************
 
-// Acknowledgment: XQEMU (GPLv2)
+// Acknowledgment: QEMU hub device emulation as used in XQEMU (GPLv2)
```
JayFoxRox: As much as I appreciate all those mentions of XQEMU, the upstream you should care about for this is QEMU.

RadWolfie:
> create an issue for the follow-up work

I agree with @PatrickvL for simply make an issue ticket of affecting copyright files issue. Plus determine each file license with the right license. After all of us make the agreement with absolutely no confusion at all from the issue ticket's discussion. Finally, make one pull request to resolve them without the need to create new pull request**s**.

PR #1565, opened 2019-03-10
Loader reloaded fixup

PatrickvL:
This introduces the basics for #769 by introducing 3 new projects (a loader executable, an emulation DLL and an empty GUI executable) without actually making use of any of it.

This setup will allow Cxbx-Reloaded to have near-absolute control over memory ranges under Windows (and probably Wine nearly as well).

This work is not being used just yet, but the code is now at a point that it'd be helpfull to merge it into develop, so others can start contributing (like the required changes to the memory manager classes, for example).
(EDIT : Per @LukeUsher request, this PR is meanwhile updated to target a separate 'loader' branch)

Note, that existing functionality should NOT be changed, so if the existing Cxbx.exe still works, it'll be alright to merge.

Reviews & remarks welcome!

Note, this PR replaces #1552 (some comments are worth reading). Special thanks to @RadWolfie for fixing newline issues.

Also, I've taken the liberty to target the develop branch with this, as this work doesn't hinder the current Cxbx project, and having yet another branch will surely lead to this getting forgotten yet again. As I've explained in the previous PR, the loader-based approach is completely separate, and should be completed asap so we can switch over sooner rather than later.


AzurikRiseOfPerathia:
Azurik Start in game AGAIN YES !!!!! ;)

RadWolfie:
Your project files are still using LF instead of CRLF. And this PR appears did not come from my branch's work which contains so many fixes and merge fixed commits to respective commit cause the breakage.

PatrickvL:
This is branched of off your fixup branch, then rebased on develop, then one additional commit

PatrickvL:
Perhaps it's better if you make those steps?

RadWolfie:
Ah, the fixup branch didn't have the updated fixes. https://github.com/RadWolfie/Cxbx-Reloaded/tree/loader_reloaded_pr_rebase has the updated fixes which is being on hold until cmake is implemented first. Then, the rebased branch will another "edited" commits to use cmake instead of generated project files.

If you found any bugs from the work you had done. Then, create a new branch of cxbxr's develop branch and use `git pull -r https://github.com/RadWolfie/Cxbx-Reloaded loader_reloaded_pr_rebase` in the command prompt. Let me know you had perform some additional fixes in dm.

RadWolfie:
This PR can be close as I had pushed the replacement to cmake branch here: https://github.com/RadWolfie/Cxbx-Reloaded/tree/cmake_loader_reloaded

Any upstream work in development branch will need to maintenance rebase from `add-cmake-support` branch then another rebase to `cmake_loader_reloaded` branch. Plus any files added or removal will need to re-update the cmake files.

PatrickvL:
I agree, and thanks for your work!

PR #1567, opened 2019-03-11
Alternative RSA implementation

ergo720:
Solves #1559. It uses the libtomcrypt/libtommath libraries to perform RSA and the modular exponentiation. Confirmed by the signature check which passes if the the signature is valid. XcModExp was verified with the respective kernel test in the xbox_kernel_test_suite, which passes as well.


PatrickvL:
LGTM

PatrickvL:
Maybe it's better to move `swap_endianess` to EmuRsa.cpp

RadWolfie:
Issues:
* .gitmodules file is converted to `crlf` instead of `lf`.
* Both libraries does not include cmake support. Their organization seems to not accept cmake in their source code.

@PatrickvL, `swap_endianess`  doesn't apply to RSA code only by the way. `endianess` is generally base on CPU architecture and some software usage which should remain in utility group.

PatrickvL:
@RadWolfie if you look at how `swap_endianess` is implemented, you'll see it's not a generic conversion function

RadWolfie:
Okay, in my opinion `swap_endianess` should be `mbedtls_swap_endianness` to avoid confusion since the function is part of mbed TLS library.

P.S. Found a typo for `endianness`.

LukeUsher:
Regarding cmake: there’s no reason why we can’t provide our own cmakelists / build rules for libraries that don’t support it by default. We can’t expect that every third party library in exisitance will always use cmake.

PR #1568, opened 2019-03-20
Fix a bug in KiClockIsr affecting slower host CPUs

ergo720:


PatrickvL:
Thanks!

src/core/kernel/exports/EmuKrnlKi.cpp
```diff
@@ -166,7 +168,7 @@ xboxkrnl::VOID xboxkrnl::KiClockIsr(
 		Hand = KeTickCount & (TIMER_TABLE_SIZE - 1);
 		if (KiTimerTableListHead[Hand].Entry.Flink != &KiTimerTableListHead[Hand].Entry &&
 			(ULONGLONG)InterruptTime.QuadPart >= KiTimerTableListHead[Hand].Time.QuadPart) {
-			KeInsertQueueDpc(&KiTimerExpireDpc, (PVOID)&KeTickCount, 0);
+			KeInsertQueueDpc(&KiTimerExpireDpc, (PVOID)OldKeTickCount, 0);
```
RadWolfie: @ergo720, you are sure this is correct? Since the parameter asked for a pointer, not a value from DWORD. I think it should be `(PVOID)&OldKeTickCount`.

ergo720: Yes, I'm sure. It should pass the value of ```OldKeTickCount```, not its address like it was doing before (and was actually another bug). ```KiTimerExpiration``` will then use ```PtrToLong``` to retrieve the value from the pointer (```PtrToLong``` does not dereference the pointer, it just casts it to a ```LONG```).

HyperScorpio:
Just a tiny fix made JSRF a little more quick. But not 100% yet.

src/core/kernel/exports/EmuKrnlKi.cpp
RadWolfie: Huh, thanks for the update. I recently took a look at ReactOS's `dpc.c` file which appear to do the same way as well.

👍

PR #1570, opened 2019-03-23
Fix warning passing ushort to set1

NZJenkins:


src/core/hle/D3D8/Direct3D9/WalkIndexBuffer.cpp
```diff
@@ -40,7 +40,7 @@ void WalkIndexBuffer_SSE41(XTL::INDEX16 & LowIndex, XTL::INDEX16 & HighIndex, XT
 
 	// Initialize mins and maxes
 	__m128i *unalignedIndices = (__m128i*) pIndexData;
-	__m128i min = _mm_set1_epi16(static_cast<short>(USHRT_MAX)); // cast as set1 only takes signed shorts
+	__m128i min = _mm_set1_epi16((short)(USHRT_MAX)); // cast as set1 only takes signed shorts
```
RadWolfie: My suggestion is to do this method:
```cpp
__m128i min = _mm_set1_epi16(-1); // cast as USHRT_MAX; argument is signed short
```

Then we don't need to worry about cast issue here. The same for below.

PR #1572, opened 2019-03-25
Introduce Network Support (System Link/Lan)

LukeUsher:
This PR builds upon the existing NVNet network hardware emulation to allow System Link/Lan play.

Notes:
* LAN Play/System Link works across Real hardware, Cxbx-R, and XQEMU: Players from all three can join in matches together, provided they are on the same local network.
 * Playing with Real Hardware & XQEMU requires keys dumped from a real xbox. A python script for dumping keys with xboxpy is provided by JayFoxRox [Here](https://github.com/JayFoxRox/xbox-tools/blob/master/python-scripts/dump-keys.py)
* MAC address based networking is functional (System Link will work), but TCP/IP networking is still broken: so no internet access or FTP servers in Cxbx-R just yet.
* Due to limitations with PCAP drivers, LAN play is only guaranteed to work over Ethernet, Wifi is currently unsupported, but may work on certain configurations.

Special thanks to mborgerson (who reverse engineered the NVNet hardware and implemented it within XQEMU), JayFoxRox (for the key dumping script, general tips/support and testing), and the rest of the XQEMU team for helping to make this possible.

image

image

image

image


RadWolfie:
Wireshark is using git, add a gitmodule is possible. (I did tested myself with git software) Source: https://code.wireshark.org/review/#/admin/projects/winpcap

It's my understanding you had included a pre-compiled library in the repo. If the header and library files are all we need, then it is possible to minimize unused files like @PatrickvL did to distorm's import. It would be great to include a link to the source code in the winpcap's readme file.

Other than that, I had tested a system without winpcap which appear does not able to get an IP address. At least, Cxbx-Reloaded is able to run normally without winpcap.

resource/Cxbx.rc
```diff
@@ -55,6 +55,11 @@ BEGIN
     BEGIN
     END
 
+	IDD_NETWORK_CFG, DIALOG
```
RadWolfie: Why tabs in Cxbx.rc file? Seems like this was done manually than VS's tool.

src/common/EmuEEPROM.cpp
```diff
@@ -223,6 +224,34 @@ void EmuEEPROMReset(xboxkrnl::XBOX_EEPROM* eeprom)
 	eeprom->UserSettings.ParentalControlMovies = XC_PC_MAX; // = XC_PRTL_CRTL_MAX
 	eeprom->UserSettings.MiscFlags = 0;  // No automatic power down
 
-	// TODO: Online Settings
+	// Online Settings
```
RadWolfie: All three, Serial Number, MAC Address, and Online Key fields appear to remain zero'd after press "reset" button. I think this is an old bug.

resource/Cxbx.rc
LukeUsher: This was edited with Visual Studio, so no idea

src/common/EmuEEPROM.cpp
LukeUsher: Yeah, it does actually reset the eeprom, just doesn't refresh the dialog.

LukeUsher:
> Wireshark is using git, add a gitmodule is possible. (I did tested myself with git software) Source: https://code.wireshark.org/review/#/admin/projects/winpcap
> 
> It's my understanding you had included a pre-compiled library in the repo. If the header and library files are all we need, then it is possible to minimize unused files like @PatrickvL did to distorm's import. It would be great to include a link to the source code in the winpcap's readme file.
> 
> Other than that, I had tested a system without winpcap which appear does not able to get an IP address. At least, Cxbx-Reloaded is able to run normally without winpcap.

winpcap is not necessarily from wireshark: It's a separate library, wireshark just happen to also use it.
Note that we use the winpcap SDK and precompiled library, mostly because it failed to compile correctly with MSVC 2017.

src/common/EmuEEPROM.cpp
RadWolfie: Here's the steps I took.

1. Open Config Eeprom
2. Click Reset button
3. Click Accept button
4. Open Config Eeprom

Result is still the same, all zero'd. Even the EEPROM.bin file shows the same thing as well.

RadWolfie: Issue is resolved, I can verified all three fields are functional.

src/core/kernel/exports/EmuKrnlPhy.cpp
```diff
@@ -0,0 +1,337 @@
+// This is an open source non-commercial project. Dear PVS-Studio, please check it.
+// PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com
+// ******************************************************************
+// *
+// *    .,-:::::    .,::      .::::::::.    .,::      .:
```
RadWolfie: This ASCII art had been removed from all source files.

The same for...
* `src/gui/DlgNetworkConfig.cpp`
* `src/gui/DlgNetworkConfig.h`

LukeUsher: Sorry, these files were created before the ascii art removal; they date back a full year, they just never made it into the main builds until now

LukeUsher: I’ll fix it tomorrow

src/devices/EmuNVNet.cpp
```diff
@@ -385,7 +256,7 @@ void EmuNVNet_UpdateIRQ()
 {
 	if (EmuNVNet_GetRegister(NvRegIrqMask, 4) &&
 		EmuNVNet_GetRegister(NvRegIrqStatus, 4)) {
-		DBG_PRINTF("Asserting IRQ\n");
+		DBG_PRINTF("EmuNVNet: Asserting IRQ\n");
```
RadWolfie: Pretty sure this should be `NET: Asserting IRQ\n`.

fancymelonub:
It's possible to play on Wi-Fi using the Microsft KNET Loopback Adapter. Just saying.

src/devices/EmuNVNet.cpp
```diff
@@ -435,7 +306,7 @@ int EmuNVNet_MiiReadWrite(uint64_t val)
 
 uint32_t EmuNVNet_Read(xbaddr addr, int size)
 {
-	DBG_PRINTF("Read%d: %s (0x%.8X)\n", size, EmuNVNet_GetRegisterName(addr), addr);
+	DBG_PRINTF("NET : Read%d: %s (0x%.8X)\n", size * 8, EmuNVNet_GetRegisterName(addr), addr);
```
RadWolfie: NET prefix is not necessary here since `DBG_PRINTF` is forward macro to `DBG_PRINTF_EX` which already include the prefix macro.

src/devices/EmuNVNet.cpp
```diff
@@ -499,7 +474,20 @@ void EmuNVNet_Write(xbaddr addr, uint32_t value, int size)
 		break;
 	}
 
-	DBG_PRINTF("Write%d: %s (0x%.8X) = 0x%.8X\n", size, EmuNVNet_GetRegisterName(addr), addr, value);
+	DBG_PRINTF("NET : Write%d: %s (0x%.8X) = 0x%.8X\n", size * 8, EmuNVNet_GetRegisterName(addr), addr, value);
```
RadWolfie: Same here, NET prefix is already included.

RadWolfie:
Since the settings' structure has increased, the setting's version need a increment.
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/cfabb7f09ba1b4098b0d4ab865b4091846278705/src/common/Settings.cpp#L48

PatrickvL:
@RadWolfie I marked 'resolved' all remarks you made that have now been addressed. If you could do one final review, I'll be happy give you the honors of merging this gem.

Nice work @LukeUsher !

HyperScorpio:
XBOX Live is alive. The fans are waiting for this. Congrats to all is involved here!

LukeUsher:
@HyperScorpio this is NOT Xbox Live, this is System Link/LAN play.

Tom-Neverwinter:
Hopefully this helps pave the way for xlink kai support. Looking forward to possible xbox live emulation in the future. This is a massive milestone, congratulations!!!!

LukeUsher:
@Tom-Neverwinter there's no reason XLink Kai shouldn't work already: All XLink Kai does, is forward System Link packets over the internet.

PR #1579, opened 2019-04-05
Fix a deadlock in KiTimerExpiration caused by KeWaitForSingleObject

ergo720:
This was meant for LLE USB, but the deadlock can also happen on current develop so it's beneficial to upstream.


PatrickvL:
LGTM!

PR #1581, opened 2019-04-08
Respect Xbox presentation interval 

LukeUsher:
This allows us to improve our frame-limiter, so that we can disable it
if a game does not require a fixed frame-rate, and enable it for titles
that do: The uncap frame-rate hack is no longer required for those titles.

Result:
XDK Samples can now run at high frame-rates (100+ fps) without effecting actual
speed (just like on real hardware)

JSRF and other titles that tie game speed to frame-rate are limited to
60FPS.

Fixed broken VSYNC handling too: Disable VSYNC has been broken since the
D3D9 port, perhaps before then.

QueryPerformanceCounter patch has been removed: This is better handled by our RDTSC patching code, and is no longer required.


LukeUsher:
Please make sure to properly test this before merging to make sure that
1. Frame Limited games (such as JSRF) run at the correct speeds
2. RDTSC sensitive games still work as expected (Zapper is known to be problematic)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4555,19 +4546,42 @@ DWORD WINAPI XTL::EMUPATCH(D3DDevice_Swap)
...
+                multiplier = 3.0f;
+                break;
+            case D3DPRESENT_INTERVAL_FOUR:
+                multiplier = 4.0f;
+                break;
+        }
 
-		frameStartTime = std::chrono::high_resolution_clock::now();
-	}
+        auto targetDuration = std::chrono::duration<double, std::milli>(((1.0f / targetRefreshRate) * multiplier) * 1000.0f);
```
PatrickvL: The final multiplication by `1000.0f` could be removed if the initial `1.0f` numerator is changed into `1000.0f` instead.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4555,19 +4546,42 @@ DWORD WINAPI XTL::EMUPATCH(D3DDevice_Swap)
...
+        while (std::chrono::high_resolution_clock::now() - frameStartTime < targetDuration) {
+            // We use an empty while loop because actually sleeping is too unstable
```
PatrickvL: I know this code isn't new, but I'm still convinced that a more efficient wait can be employed here.

Currently, this loop is burning cycles at a 100% load of the core it's running on. 
There exist a number of ways to pass time without this burden, like SwitchToThread(), Sleep(0), Sleep(1), WaitForSingleObject, etc.
Each of those API's has specific characteristics; We're primarily interested in the highest precision.
Given these numbers, we could select the best-fitting wait function, considering the amount of time to be slept.
This way, the biggest elapses of sleep-time can be covered with a low impact on system resources.
Any remaining sleep-time that no API covers, can still be dealt with via  hot-loop like we currently have.
If not now, then at least consider creating an issue for this functionality.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -1484,21 +1484,6 @@ LPVOID WINAPI XTL::EMUPATCH(ConvertThreadToFiber)
...
-// ******************************************************************
-ULONGLONG CxbxGetPerformanceCounter(bool acpi); // implemented in EmuKrnlKe.cpp
-BOOL WINAPI XTL::EMUPATCH(QueryPerformanceCounter)
-(
-	LARGE_INTEGER * lpPerformanceCount
-)
-{
-	// NOTE: QueryPerformanceCounter runs from the tsc via RdTsc (733mhz)
-	// However, KeQueryPerformanceCounter runs at 3.375Mhz, so we can't use that here
-	lpPerformanceCount->QuadPart = CxbxGetPerformanceCounter(false);
```
PatrickvL: Please check if the boolean argument to `CxbxGetPerformanceCounter` is still required?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1916,21 +1917,11 @@ static DWORD WINAPI EmuCreateDeviceProxy(LPVOID)
...
-                    if(!g_XBVideo.bVSync && (g_D3DCaps.PresentationIntervals & D3DPRESENT_INTERVAL_IMMEDIATE) && g_XBVideo.bFullScreen)
-                        g_EmuCDPD.HostPresentationParameters.FullScreen_PresentationInterval = D3DPRESENT_INTERVAL_IMMEDIATE;
-                    else
-                    {
-                        if(g_D3DCaps.PresentationIntervals & D3DPRESENT_INTERVAL_ONE && g_XBVideo.bFullScreen)
-                            g_EmuCDPD.HostPresentationParameters.FullScreen_PresentationInterval = D3DPRESENT_INTERVAL_ONE;
-                        else
-                            g_EmuCDPD.HostPresentationParameters.FullScreen_PresentationInterval = D3DPRESENT_INTERVAL_DEFAULT;
-                    }
+					g_EmuCDPD.HostPresentationParameters.PresentationInterval = g_XBVideo.bVSync ? D3DPRESENT_INTERVAL_ONE : D3DPRESENT_INTERVAL_IMMEDIATE;
```
PatrickvL: Why is `FullScreen_PresentationInterval` no longer set? Won't this change impact `bFullScreen`-mode?

src/core/hle/XAPI/Xapi.cpp
LukeUsher: The boolean argument is required: Since false indicates it should return timing based on Xbox Clock (733mhz, used for our rdtsc intercept), but true indicates that it should return timing based on the System Timer Frequency (3.375mhz, used for the ACPI timer, as used by KeQueryPerformanceCounter).

QueryPerformanceCounter/Rdtsc use 733mhz has a clock source, 
KeQueryPerformanceCounter uses the APCI timer as a clock source.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: FullScreen_PresentationInterval is an alias for PresentationInterval in D3D9, the same field is used for both Windowed and Fullscreen

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1916,21 +1917,11 @@ static DWORD WINAPI EmuCreateDeviceProxy(LPVOID)
                 {
                     g_EmuCDPD.HostPresentationParameters.Windowed = !g_XBVideo.bFullScreen;
 
-                    if(g_XBVideo.bVSync)
-                        g_EmuCDPD.HostPresentationParameters.SwapEffect = XTL::D3DSWAPEFFECT_COPY; // Was D3DSWAPEFFECT_COPY_VSYNC;
```
PatrickvL: If `SwapEffect` is no longer set here, what value will it default to?

LukeUsher: As current windowed behavior, it defaults to whatever was in the Xbox field

PatrickvL:
Please bear with me about this I'm wondering about:

Some Xbox software might not do any timing, and totally rely on the speed of the system (CPU, GPU, memory-bandwidth, etc).
On a real machine this would result in a certain frame-rate that users are used to - a frame rate which they'd expect Cxbx-Reloaded to show too.
But if we're not limiting frame-rate at all, Cxbx-Reloaded will render as fast as the host allows, which could be many times as fast as the real machine.
Therefor, for those cases, frame-rate limiting would still be desirable.
Either via a hack, or via some sort of load-detection, so that the speed of the real machine could still be guestimated.

So, it might be wise not remove the hack entirely just yet...

src/core/hle/XAPI/Xapi.cpp
PatrickvL: Yeah, I'm aware. It just needs checking if by any chance all callers pass in the same value now?

LukeUsher:
For most games, the frame rate is entirely disconnected from the speed: weather the game runs at 100fps or 1000fps, those games run at the same speed, the only difference is a smoother appearance. 

This implementation supports that scenario, while capping the frame rate for titles that ask for it. 

Any title that ties speed to frame rate will set these values in real hardware. Otherwise they’d end up with certain low-load scenes (for example less enemies on screen) running too fast to be playable.

src/core/hle/XAPI/Xapi.cpp
LukeUsher: they don’t. Our rdtsc implementation passes false (for 733mhz) and KeQueryPerformanceCounter passes true (for 3mhz)

PatrickvL:
To further reason on that;

What if some titles, during heavy-load scenes, **temporarily** skip any timing or frame-rate limiting, to try reach the highest possible frame-rate the real machine can produce.

During these phases, Cxbx-Reloaded might run way too fast.

Granted, we currently run way too slow under many other scenario's, but running too fast is still a real issue to consider, wouldn't you agree?

LukeUsher:
>What if some titles, during heavy-load scenes, temporarily skip any timing or frame-rate limiting, to try reach the highest possible frame-rate the real machine can produce.`

This sounds like a very unlikely edge case that should be addressed if it comes up in testing: It seems especially unlikely when using the Presentation Interval as a limiter has zero performance impact: Enabling it has no drawbacks other can capping the maximum frame-rate. Even if the game disabled it's own frame limiting, if it would set a presentation interval, it would never exceed that maximum frame-rate.

So what you're proposing requires a very specific scenario (all of the below would have to be true)
1. The game is programmed in such a way that frame-rate effects game speed (This is rare/only a problem for a tiny subset of games, since Microsoft specifically state *NOT* to do this, most games are frame-rate independent, it's also undesirable as it means that if the frame-rate drops, the game would run too slow instead of just appearing less smooth)
2. For most of the game, there is a presentation interval set that limits the frame rate, to prevent the game running too fast because of point 1. 
3. A specific scene disables ALL frame-limiting because the developer *assumes* it's impossible to reach the target frame-rate (But the calls to disable the presentation interval *add* overhead and cause some GPU state to be flushed, which would defeat the point of trying to get maximum performance)
4. The specific scene must be heavy enough that real hardware cannot ever reach full speed, otherwise the game has a chance of running too fast on real hardware

Saying that, it's not impossible that there are some games effected by this, so I think we should wait until we see them, and add an exception for those specific games: as to not negatively effect titles that don't require it.

PatrickvL:
Agreed - glad we worked that out ;)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4574,13 +4578,21 @@ DWORD WINAPI XTL::EMUPATCH(D3DDevice_Swap)
...
-        auto targetDuration = std::chrono::duration<double, std::milli>(((1.0f / targetRefreshRate) * multiplier) * 1000.0f);
-        while (std::chrono::high_resolution_clock::now() - frameStartTime < targetDuration) {
-            // We use an empty while loop because actually sleeping is too unstable
-            // Sleeping causes the frame duration to jitter...
-            ;
+        auto targetDuration = std::chrono::duration<double, std::milli>(((1000.0f / targetRefreshRate) * multiplier));
+        auto targetTimestamp = frameStartTime + targetDuration;
+
+        // If we need to wait for a larger amount of time (>= 1 frame at 60FPS), we can just sleep
+        if (std::chrono::duration_cast<std::chrono::milliseconds>(targetTimestamp - std::chrono::high_resolution_clock::now()).count() > 16) {
```
LukeUsher: This approach could possibly be improved in the future, but for now it's better than what we have

slx7R4GDZM: Won't sleeping be slightly off for frametimes that use decimal milliseconds like 30 or 60 FPS?
It might be better to just have an option to choose between sleeping and busy waiting.

LukeUsher: Honestly I'm okay with just busy-waiting: sleep-related jitter is undesirable but then we have users on laptops who cry at high/wasted CPU usage :P

If it helps, in testing I'm able to get a stable 60FPS
image

PatrickvL: To avoid jitter, only a remainder of the entire wait could be busy-waited.

But lets keep that as an exercise for another day.

PatrickvL:
Status : All review remarks are addressed.

Awaiting results from testers...

LukeUsher:
So far, it has been reported that
1. JSRF still runs at the correct speed of 60FPS (except when the user's computer is too slow to keep-up)
2. Phantasy Star Online now runs at the correct speed of 30FPS, instead of too-fast at 60FPS
3. XDK samples that disable frame limiting can still reach high FPS, XDK samples that limit frame-rate are able to get stable 60FPS
4. Otogi no longer runs too-fast, but still remains unplayable due to other, existing emulation issues.

src/common/Settings.hpp
```diff
@@ -207,7 +207,6 @@ class Settings
 	// Hack settings
 	struct s_hack {
 		bool DisablePixelShaders;
-		bool UncapFramerate;
```
RadWolfie: Please restore this, then mark as unused or pad. Thanks!

P.S. You do not need to keep the name `UncapFramerate`.

LukeUsher: I could do this, but first I have to question why this matters: The Settings code saves/loads these values like this
```
	m_hacks.DisablePixelShaders = m_si.GetBoolValue(section_hack, sect_hack_keys.DisablePixelShaders, /*Default=*/false);
	m_hacks.UseAllCores = m_si.GetBoolValue(section_hack, sect_hack_keys.UseAllCores, /*Default=*/false);
	m_hacks.SkipRdtscPatching = m_si.GetBoolValue(section_hack, sect_hack_keys.SkipRdtscPatching, /*Default=*/false);
	m_hacks.ScaleViewport = m_si.GetBoolValue(section_hack, sect_hack_keys.ScaleViewPort, /*Default=*/false);
	m_hacks.DirectHostBackBufferAccess = m_si.GetBoolValue(section_hack, sect_hack_keys.DirectHostBackBufferAccess, /*Default=*/false);
```

So as far as I can tell: nothing relies on the order of values, or this being present

RadWolfie: To maintenance global memory share, the size must be exact. Otherwise running a newer or older cxbx-r build will read outside the global memory share allocation size.

There are reasons of extra paddings for most settings to allow add new setting without need to bump a version.

LukeUsher: Ah, so this is a problem if the user tries to run multiple versions of Cxbx-R simultaneously? Otherwise I don't see the issue

RadWolfie: That is correct. You may not see the issue, I do. 😛 I tend to look in (developer) and *out* (user) of the box to ensure all the safety are implemented.

Once we have individual memory shared functional, then versions and removal/additional would not matter at all.

PatrickvL:
There've been plenty of reports on this commit in our Discord channel Showcase, to justify merge

PR #1583, opened 2019-04-10
Clear released texture from EmuD3DActiveTexture

NZJenkins:


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5414,6 +5414,11 @@ ULONG WINAPI XTL::EMUPATCH(D3DResource_Release)
 			g_XboxBackBufferSurface = nullptr;
 		}
 
+		for (int i = 0; i < TEXTURE_STAGES; i++) {
+			if (pThis == EmuD3DActiveTexture[i])
```
LukeUsher: Just a nitpick, please use braces for all if statements, it doesn't matter so much in this case, but in general it's easier to work with (expand/collapse blocks), and less error-prone when adding more code.

```
for (int i = 0; i < TEXTURE_STAGES; i++) {
    if (pThis == EmuD3DActiveTexture[i]) {
        EmuD3DActiveTexture[i] = nullptr;
	}
}

LukeUsher: Such a minor thing that I'll merge for now, but it's something to keep in mind

PR #1584, opened 2019-04-11
Allow XBE Section preload to fail

LukeUsher:
Since all sections in all Xbes are usually marked as preload, the Xbox
kernel attempts to load them all.

In some titles (for example, Lego Star Wars DVD demo) the whole XBE is
too large to fit in RAM, and it is *expected* for the preload to fail.

Previously, when hit with this scenario, Cxbx-R would hard crash.

Now it mirrors hardware behavior and gracefully continues, skipping the
failed Xbe section.

Q: Why is this the case?
A: It's possible (and proven) that with titles such as the Lego Star Wars
DVD Demo, *all* game data is contained within the XBE file, rather than
on-disc as raw files. Naturally, these are too big to fit into the Xbox
RAM, so certain large sections fail to allocate, and are skipped
entirely by the Xbox Kernel.

The game then reads data from these sections by *not* loading them to
memory, but parsing the section *headers* to do raw file-io and read the
Xbe file as a binary blob, parsing the virtual filesystem.

This change allows Lego Star Wars DVD Demo (and other large Xbe titles)
to boot.

Lego Star Wars DVD Demo doesn't work yet, however, as it attempts a
dashboard update and fails, but it no longer crashes the whole emulator.

### DIFF
--- a/src/core/kernel/memory-manager/VMManager.cpp
+++ b/src/core/kernel/memory-manager/VMManager.cpp
@@ -1746,6 +1746,7 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
 
 	// Attempt to commit the requested range with VirtualAlloc *before* setting up and reserving the PT
 	// This allows an early-out in a failure scenario (Test Case: Star Wars Battlefront DVD Demo: LA-018 v1.02)
+    // We don't commit the requested range if it's within our placeholder, since that was already allocated earlier
 	if (AlignedCapturedBase >= XBE_MAX_VA)
 	{
 		if (!VirtualAlloc((void*)AlignedCapturedBase, AlignedCapturedSize, MEM_COMMIT,
--- a/src/core/kernel/memory-manager/VMManager.cpp
+++ b/src/core/kernel/memory-manager/VMManager.cpp
@@ -1744,6 +1744,19 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
 		goto Exit;
 	}
 
+	// Attempt to commit the requested range with VirtualAlloc *before* setting up and reserving the PT
+	// This allows an early-out in a failure scenario (Test Case: Star Wars Battlefront DVD Demo: LA-018 v1.02)
+	if (AlignedCapturedBase >= XBE_MAX_VA)
+	{
+		if (!VirtualAlloc((void*)AlignedCapturedBase, AlignedCapturedSize, MEM_COMMIT,
+			(ConvertXboxToWinProtection(PatchXboxPermissions(Protect))) & ~(PAGE_WRITECOMBINE | PAGE_NOCACHE)))
+		{
+			DBG_PRINTF("%s: VirtualAlloc failed to commit the memory! The error was %d\n", __func__, GetLastError());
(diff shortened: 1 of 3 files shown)


src/core/kernel/memory-manager/VMManager.cpp
```diff
@@ -1744,6 +1744,19 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
 		goto Exit;
 	}
 
+	// Attempt to commit the requested range with VirtualAlloc *before* setting up and reserving the PT
+	// This allows an early-out in a failure scenario (Test Case: Star Wars Battlefront DVD Demo: LA-018 v1.02)
+	if (AlignedCapturedBase >= XBE_MAX_VA)
```
PatrickvL: Dangling {

src/core/kernel/memory-manager/VMManager.cpp
```diff
@@ -1769,18 +1782,6 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
 		PointerPte++;
 	}
 
-	// Actually commit the requested range but don't if it's inside the placeholder or we are committing an xbe section so that
```
PatrickvL: The gist of this comment got lost, please restore the still-valid parts (about the placeholder, XeLoadSection, etc) in the replacement comment above.

src/core/kernel/memory-manager/VMManager.cpp
LukeUsher: This was just a move/copy from from lower in the code, not new code: Figured style changes was outside of the scope since it's not new code, or a replacement for existing.

I could fix it, but then the PR would be larger ;)

src/core/kernel/memory-manager/VMManager.cpp
```diff
@@ -1744,6 +1744,19 @@ xboxkrnl::NTSTATUS VMManager::XbAllocateVirtualMemory(VAddr* addr, ULONG ZeroBit
 		goto Exit;
 	}
 
+	// Attempt to commit the requested range with VirtualAlloc *before* setting up and reserving the PT
+	// This allows an early-out in a failure scenario (Test Case: Star Wars Battlefront DVD Demo: LA-018 v1.02)
+	if (AlignedCapturedBase >= XBE_MAX_VA)
+	{
+		if (!VirtualAlloc((void*)AlignedCapturedBase, AlignedCapturedSize, MEM_COMMIT,
+			(ConvertXboxToWinProtection(PatchXboxPermissions(Protect))) & ~(PAGE_WRITECOMBINE | PAGE_NOCACHE)))
+		{
```
PatrickvL: Nit : dangling { here too
(knowing this was already the case)

src/core/kernel/memory-manager/VMManager.cpp
LukeUsher: The comment wasn't entirely correct as it's not just to allow XeLoadSection to work: It's for *any* allocation that is requested within the Xbe Base address, and XeLoadSection sometimes allocated sections outside of this reserved area (which is part of what this PR addresses)

That said, I'll add a similar remark about the placeholder

LukeUsher: Command has been updated

PR #1586, opened 2019-04-12
DirectSound unfocus audible option + minor fixes

RadWolfie:
Following changes in this pull request:
* Add disable mute on unfocus. (tick it to able hear gameplay while Cxbx-Reloaded is not active window)
* Rename "Enable Hardware Video Mode" to "Enable Exclusive Fullscreen Mode".
* Repaired generator resource files.

image

**EDIT: By default, audio will still remain muted while unfocus. Unless changed in the audio's settings dialog.**


RadWolfie:
"Disable mute on unfocus" option is normally for front-end third-party and possibility some broadcast tools didn't receive audio from the emulation.

AzurikRiseOfPerathia:
Why not make a volume bar 0% to 100% to adjust the audio?

RadWolfie:
Not relative to this pull request. If a flag `DSBCAPS_GLOBALFOCUS` is not included, then you cannot hear the audio outside of the emulation.

ObiKKa:
Just a note. You might. Want to change the phrase to "Disable mute when window is unfocused" or "Disable mute on/during absence" (or "Disable mute while absent from window"). It's because that shortened ending "... on unfocused" can be confusing to many users. I wouldn't have been sure of what it was for if I hadn't read your post above.

PatrickvL:
Unfocus is a well-known term, absent/absence/away are less clear IMHO.

But to piggy-pack on this remark: This menu option could be inverted in meaning and be called "Mute when unfocused" (default on), which is much shorter, and (because like this it no longer is a double negative, and less words are used), probably much better to understand.

PR #1589, opened 2019-04-18
Unpatch Fiber Functions

LukeUsher:
Fix MmCreateKernelStack: It should return the StackBase *not* the
allocation base

Remove no-longer-requires Fiber patches

Make sure Host and Xbox Stacks are in-sync for Xbox Threads

These three points combined allow Futurama to reach in-game, *without*
breaking other fiber-sensitive titles such as DOA3 and DOAX

image


RadWolfie:
So far, I only found DoA Ultimate 2 made a silent crash once get to the menu.

The other title I found with CreateFiber, Hulk remain unaffected.

LukeUsher:
It’s going to take some time to solve the doa2x regression: feel free to carry on testing but please don’t merge until I update to say otherwise

LukeUsher:
This does wonders for Futurama  but is too unstable to merge as a whole, it breaks quite a few other titles. Closing as I work on it further.

PR #1590, opened 2019-04-21
Move EmuLog(Ex) together with the other logging functions

ergo720:


RadWolfie:
Low risk + no changes occur during the move.

PR #1592, opened 2019-04-28
Add bias modifier so bumpmap is signed

NZJenkins:
This corrects the boost dash effect in JSRF by treating the bump texture as -0.5 to 0.5 instead of 0 to 1
I'm not sure if this correction belongs in the TEXBEM code?

image

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -2961,12 +2961,14 @@ bool PSH_XBOX_SHADER::InsertTextureModeInstruction(XTL::X_D3DPIXELSHADERDEF *pPS
             Ins.Output[0].SetRegister(PARAM_R, 1, MASK_R);
             Ins.Parameters[0].SetScaleBemLumRegister(XTL::D3DTSS_BUMPENVMAT00, Stage, Recompiled);
             Ins.Parameters[1].SetRegister(PARAM_R, PSH_XBOX_MAX_R_REGISTER_COUNT + inputStage, MASK_R);
+            Ins.Parameters[1].Modifiers = (1 << ARGMOD_BIAS);
             Ins.Parameters[2].SetRegister(PARAM_R, PSH_XBOX_MAX_R_REGISTER_COUNT + Stage, MASK_R);
             InsertIns.emplace_back(Ins);
             Ins.Initialize(PO_MAD);
             Ins.Output[0].SetRegister(PARAM_R, 1, MASK_R);
             Ins.Parameters[0].SetScaleBemLumRegister(XTL::D3DTSS_BUMPENVMAT10, Stage, Recompiled);
             Ins.Parameters[1].SetRegister(PARAM_R, PSH_XBOX_MAX_R_REGISTER_COUNT + inputStage, MASK_G);
+            Ins.Parameters[1].Modifiers = (1 << ARGMOD_BIAS);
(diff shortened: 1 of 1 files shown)


LukeUsher:
Good work, thanks, it still doesn't render right (it's still too dark/covers too much of the screen/the lines shouldn't be this obvious) but is an improvement over what we have currently.

ESPPIRAL:
maybe it was taken in a shadowed area? to me doesn't look that dark, atcually I'd say it looks as it should.

bandicam 2019-04-28 11-59-34-077
bandicam 2019-04-28 11-56-55-933

PatrickvL:
Let's await a few more tests before merging this, but it looks very promising indeed!

This also illustrates that a thorough review of the current shader conversion code is a necessity instead of a luxury

PatrickvL:
Xdk samples might be good test cases

LukeUsher:
Question we need to answer now is is this dependant on the input texture format, or should it always to take place? What texture format is JSRF using here? Perhaps the bug is our conversion of a signed texture to an unsigned type or similar issue.

ESPPIRAL:
It appears that on the original console there is an adition blur effect applied to the image.

This is CXBX-R
image

This is captured directly from the XBox framebuffer

192 168 0 4-image10

Personally I do preffer how does it looks in CXBX :)

RadWolfie:
Two XDK samples do show a difference between develop and this pull request.
<details><summary>BumpEarth</summary>

Pull request commit 9694153
BumpEarth-pr1592
Hardware
BumpEarth-hardware
Latest develop branch
BumpEarth-develop
</details>

<details><summary>BumpLens</summary>

Pull request commit 9694153
BumpLens-pr1592
Hardware
BumpLens-hardware
Latest develop branch
BumpLens-develop
</details>

LukeUsher:
Okay, thanks, so it's definitely format independent, implying that rather than the shader instructions being incorrect, we converted/handled the input texture incorrectly, this needs further investigation

medievil1:
> 
> 
> Two XDK samples do show difference between develop and this pull request.
> BumpEarth
> 
> Pull request commit [9694153](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/9694153db885d78306fa5847d0332fb5c2d6b5fb)
> BumpEarth-pr1592
> Hardware
> BumpEarth-hardware
> Latest develop branch
> BumpEarth-develop
> BumpLens

do you have to still have that version of the bumpearth demo??? the one we are usign for reference is slightly different (black background, all yellow test, etc..)
@RadWolfie

RadWolfie:
> do you happen to still have that version of the bumpearth demo??? the one we are usign for reference is slightly different (black background, all yellow test, etc..)
> @RadWolfie
> 
> Or wait was that the same one on both??.. cause I looked it over and no where does it set a purple background

The hardware's bumpearth demo is from Voxel's all-in-one video showing what each demo looks like on hardware. Afaik, it's newer than what I had shown in cxbxr screenshot.

medievil1:
> 
> 
> > do you happen to still have that version of the bumpearth demo??? the one we are usign for reference is slightly different (black background, all yellow test, etc..)
> > @RadWolfie
> > Or wait was that the same one on both??.. cause I looked it over and no where does it set a purple background
> 
> The hardware's bumpearth demo is from Voxel's all-in-one video showing what each demo looks like on hardware. Afaik, it's newer than what I had shown in cxbxr screenshot.

yup after you contacted Patrick he had that version....
it uses colorsign, the other doesn't so currently fix one, break the other..lol

PR #1593, opened 2019-04-29
Selectively add bias modifier so bump-map is signed

LukeUsher:
This builds upon #1592 to fix JSRF without breaking the XDK Samples, it's a little bit of a hack, but it has been tested and improves the regressions from #1592 without re-breaking JSRF.

The issue stems from the fact that the bump-map formats X_D3DFMT_L6V5U5 and X_D3DFMT_X8L8V8U8 share an identical format specifier as the texture formats X_D3DFMT_X8L8V8U8 and X_D3DFMT_L6V5U5.

The fix applies a bias to those two input types when used in texbem or texbeml operations, to allow them to be (correctly) interpreted as signed bump-map surfaces.

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -2958,8 +2958,9 @@ bool PSH_XBOX_SHADER::InsertTextureModeInstruction(XTL::X_D3DPIXELSHADERDEF *pPS
         if (m_PSVersion >= D3DPS_VERSION(1, 4))
         {
 
-            // If the bump-map input texture is not a standard Xbox bump-map texture format, apply a bias
-            // Fixes an issue where JSRF that uses unsigned texture as an input to texbem
+            // If the bump-map texture format is X_D3DFMT_X8L8V8U8 or  X_D3DFMT_L6V5U5 we need to apply a bias
+            // This happens because these formats are an alias of unsigned texture formats.
+            // Fixes an issue with the JSRF boost-dash effect
             // NOTE: This assumes that this shader will only ever be used for the input bumpmap texture
             // If this causes regressions in other titles, we'll need to be smarter about this
             // and include the texture formats in the shader hash, somehow.
@@ -2967,7 +2968,7 @@ bool PSH_XBOX_SHADER::InsertTextureModeInstruction(XTL::X_D3DPIXELSHADERDEF *pPS
             extern XTL::X_D3DFORMAT GetXboxPixelContainerFormat(const XTL::X_D3DPixelContainer *pXboxPixelContainer);
             auto format = GetXboxPixelContainerFormat(XTL::EmuD3DActiveTexture[0]);
             bool bias = false;
-            if (format != XTL::X_D3DFMT_V8U8 && format != XTL::X_D3DFMT_V16U16) {
+            if (format == XTL::X_D3DFMT_X8L8V8U8 || format == XTL::X_D3DFMT_L6V5U5) {
                 bias = true;
             }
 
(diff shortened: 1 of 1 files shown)


RadWolfie:
Continued from https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1592#issuecomment-487411825

* BumpEarth (no change)
  * Fix the unswizzle may will show the correct position + remove the extra spotlights. So, testing with BumpEarth sample in this state is inconclusive.
* BumpLens (resolved)

Jet Set Radio Future had not revert back original boost bug.

revel8n:
Actually it looks like it uses:X_D3DFMT_X8L8V8U8
Which is a signed bump map format on the xbox.  Cxbx is just aliasing it as X8R8G8B8 which i guess is the actual issue.

revel8n:
Since they are aliased as the same enumeration value, this type of check may be the best we can do since we cannot gather intent until it is actually used, and then in this case it can actually be treated as either format.

LukeUsher:
Any more detailed tests results on this? I think we're good to merge.

PatrickvL:
LGTM (IF all other aliases have been verified?)

PatrickvL:
For completeness, here all Xbox D3DFORMAT aliasses :
```
	X_D3DFMT_X8R8G8B8 = 0x07,
	X_D3DFMT_X8L8V8U8 = 0x07, // Alias

	X_D3DFMT_R6G5B5 = 0x27,
	X_D3DFMT_L6V5U5 = 0x27, // Alias

	X_D3DFMT_G8B8 = 0x28,
	X_D3DFMT_V8U8 = 0x28, // Alias

	X_D3DFMT_LIN_G8B8 = 0x17,
	X_D3DFMT_LIN_V8U8 = 0x17, // Alias

	X_D3DFMT_LIN_X8R8G8B8 = 0x1e,
	X_D3DFMT_LIN_X8L8V8U8 = 0x1e, // Alias
```

LukeUsher:
@PatrickvL this is intentional, according to all the sources I looked at: only X_D3DFMT_X8L8V8U8 and X_D3DFMT_L6V5U5 are sharing their internal format with other types, despite sharing IDs

Of course we can always investigate this further if it proves to be incorrect.

Also note that only the following are compatible with texbem/texbeml instructions, so all others are outside the scope of this work/should not be checked here.
```
D3DFMT_V8U8
D3DFMT_V16U16
D3DFMT_L6V5U5
D3DFMT_X8L8V8U8  
```

PatrickvL:
Are you referring to https://docs.microsoft.com/en-us/windows/desktop/direct3d9/d3dformat#mixed-formats? Or the signed formats below that?

PatrickvL:
Also, https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/texbem---ps mentions:

> The red and green color data in the src register is interpreted as the perturbation data (du,dv).
> 
> This instruction transforms red and green components in the source register using the 2D bump environment-mapping matrix. The result is added to the texture coordinate set corresponding to the destination register number, and is used to sample the current texture stage.
> 
> This operation always interprets du and dv as signed quantities. For versions 1_0 and 1_1, the Source Register Signed Scaling input modifier (_bx2) is not permitted on the input argument.
> 
> This instruction produces defined results when input textures contain signed format data. Mixed format data works only if the first two channels contain signed data.

revel8n:
By sources he means Xbox related information, not Direct3d related
information.

The related information on texbem/texbml for the Xbox only list the formats
currently being checked for.

If we were still able to use the texbem/texbeml instructions directly this
would not be an issue, it only is because we have had to emulate the
instructions manually.

revel8n:
Yeah, I have a workaround for this locally. The problem is that for luminance formats L would end up in R during conversion instead of U. If you look at the conversion of the non-luminance formats V and U are assumed to be in G and B, and are handled as such. Since the blind conversion treats it as R6G5B5 the components are converted incorrectly when the format is supposed to be L6V5U5.

PR #1594, opened 2019-05-01
Prevent NVNet hogging cpu

LukeUsher:
Just a minor change, to prevent the NVNet emulation from hogging 100% cpu, low impact.


LukeUsher:
Turns out that this PR introduced networking issues/desync in many games, reverting will restore these to working status.

PR #1595, opened 2019-05-01
Use StretchRect to blit the Xbox backbuffer

LukeUsher:
This is a huge performance increase when using higher resolutions while
the Render to Host Backbuffer hack is disabled.

This was not easily possible in D3D8, but now we have D3D9, we can use
StretchRect to prevent unnecessary GPU->CPU transfers.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -4446,11 +4446,11 @@ DWORD WINAPI XTL::EMUPATCH(D3DDevice_Swap)
 				// TODO: This could be much faster if we used the XboxBackBufferSurface as a texture and blitted with a fullscreen quad
 				// This way, the scaling/format conversion would be handled by the GPU instead
                 hRet = g_pD3DDevice->StretchRect(
-                    pXboxBackBufferHostSurface,
-                    nullptr,
-                    pCurrentHostBackBuffer,
-                    nullptr,
-                    LoadSurfaceFilter
+                    /* pSourceSurface = */ pXboxBackBufferHostSurface,
+                    /* pSourceRect = */ nullptr,
+                    /* pDestSurface = */ pCurrentHostBackBuffer,
+                    /* pDestRect = */ nullptr,
+                    /* Filter = */ LoadSurfaceFilter
                 );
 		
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4443,17 +4445,14 @@ DWORD WINAPI XTL::EMUPATCH(D3DDevice_Swap)
...
-					/* pDestPalette = */ nullptr,
-					/* pDestRect = */ nullptr,
-					/* pSrcSurface = */ pXboxBackBufferHostSurface,
-					/* pSrcPalette = */ nullptr,
-					/* pSrcRect = */ nullptr,
-					/* Filter = */ LoadSurfaceFilter,
-					/* ColorKey = */ 0);
-
+                hRet = g_pD3DDevice->StretchRect(
+                    pXboxBackBufferHostSurface,
```
PatrickvL: Maybe you want to mention argument names using /*inline comment*/ blocks here, like it was before?

PatrickvL:
Good catch! If only we added more of those comments, directing us to other quick wins....

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: I really don't like that coding style, I find it unnecessary when it's documented in the DirectX docs, and in the D3D9 header (which is automatically displayed by many IDE's when typing anyway ;))

image

But if you really think it's necessary, I can amend

PatrickvL: In no way this is obligatory, I mention it purely because these comments were removed with this change, and they do help make things more clear; For example, having these comments would lead to hits when code-searches match on them, resulting in a broader insight on the usage of searched-for-terms. Also, it could prevent (or help spotting) supplying an incorrect value to an argument, as the discrepancy between argument-name and supplied argument is more clearly visible. Additionally, beginning coders (and users of a less advanced editor) will have less trouble understanding the code. So those could be reasons to apply this style. But probably not everywhere, as that would create an enormous burned on maintenance and code-bloat...

LukeUsher: Good point, amended

PR #1596, opened 2019-05-01
Fix multisampling

LukeUsher:
Thanks to PR #1595, we can now fix multisampling in Xbox titles. This has a *zero* performance impact too!

Please note that due to a limitation of host Direct3D, this only works as while the 'Render to Host Backbuffer' is disabled.

Before:
before_1
before_2

After:

after_1
after_2


PatrickvL:
Low-risk change, merging as-is, thanks!

PR #1597, opened 2019-05-01
Fix an oversight where the scaled viewport values were not used

LukeUsher:
This has no known impact, but it was an oversight I spotted while testing something else.

SetViewPort should clip the viewport dimensions to the backbuffer size, this operation was done, but the result was simply discarded in normal operation.

This fixed that minor oversight.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -168,6 +168,10 @@ static XTL::X_D3DSurface           *g_pXboxDepthStencil = NULL;
 static DWORD                        g_dwVertexShaderUsage = 0;
 static DWORD                        g_VertexShaderSlots[136];
 
+// Original Backbuffer display width/height, set during CreateDevice
+static UINT g_XboxBackBufferDisplayWidth = 0;
+static UINT g_XboxBackBufferDisplayHeight = 0;
+
 DWORD g_XboxBaseVertexIndex = 0;
 DWORD g_DefaultPresentationInterval = D3DPRESENT_INTERVAL_IMMEDIATE;
 DWORD g_PresentationIntervalOverride = 0;
@@ -2414,6 +2418,9 @@ void Direct3D_CreateDevice_Start
 	XTL::X_D3DPRESENT_PARAMETERS     *pPresentationParameters
 )
 {
+    g_XboxBackBufferDisplayWidth = pPresentationParameters->BackBufferWidth;
+    g_XboxBackBufferDisplayHeight = pPresentationParameters->BackBufferHeight;
+
 	// create default device *before* calling Xbox Direct3D_CreateDevice trampline
 	// to avoid hitting EMUPATCH'es that need a valid g_pD3DDevice
 	{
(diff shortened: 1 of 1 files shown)


RadWolfie:
If you don't mind my asking, what exactly can produce the issue? Since I'm not able to produce the issue with Jet Set Radio Future.

LukeUsher:
Reread the comment: it doesn’t fix any currently known issues in normal operation, but it’s still a mistake/bug that will matter for new features in the future

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3200,6 +3200,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
 		XboxViewPort.MinZ = pViewport->MinZ;
```
PatrickvL: Since `XboxViewPort = *pViewport` above, these two assignments of MinZ/MaxZ are unnecessary (only the clipped X/Y/Width/Height assignments are relevant).

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3200,6 +3200,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
 		XboxViewPort.MinZ = pViewport->MinZ;
 		XboxViewPort.MaxZ = pViewport->MaxZ;
 
+		// Write the scaled viewport data back to the host viewport structure
+		HostViewPort = XboxViewPort;
```
PatrickvL: Since `HostViewPort` is overwritten when `GetHostRenderTargetDimensions()` below succeeds, this assignment would be useless, and could be moved towards an `else` block below the `if (g_ScaleViewport)` block.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3200,6 +3200,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
 		XboxViewPort.MinZ = pViewport->MinZ;
 		XboxViewPort.MaxZ = pViewport->MaxZ;
 
+		// Write the scaled viewport data back to the host viewport structure
```
PatrickvL: Actually, the `XboxViewPort` isn't scaled here, but clipped. Only below it gets scaled, in the `if (g_ScaleViewport)` block.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: I wanted to avoid an if/else for cleaner-reading code

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: Thanks, the comment could be more clear

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: This was unchanged from the original code, but I could amend

LukeUsher:
Note that this is part of the preparation for actual upscaling to higher resolution, instead of the buggy backbuffer-expanding approach we currently have now: I've been investigating the work for a couple of days now

LukeUsher:
Updated to fix SetViewPort scale when a multi-sampling is used on the backbuffer, mimicking the behavior of the Xbox implementation of this function. Just a minor thing, but required now after #1596

LukeUsher:
Meh, messed up the line-endings, I'll close for now, submit a new PR with additional related fixes later.

Luffykun007:
We are looking forward to it!

PR #1599, opened 2019-05-03
Disable multisampling yet again...

LukeUsher:
It's a shame but the previous 'fix' only worked for the dashboard, and caused
issues in many games themselves.

For example, GTA3 only rendered to half-the-screen due to multisampling issues.

A better fix needs to be investigated and implemented, sadly, we can't rely on
D3D9's native multisampling as it prevents the buffers from being locked... which
Cxbx-R *requires*

So back to the drawing board on that for now...


RadWolfie:
Since this change only do the revert with more details of reason why multi-sampling is disable. Merging as-is.

PR #1601, opened 2019-05-04
Fix D3DDevice_Begin/D3DDevice_EndPush length calculation

LukeUsher:
With this fix, Rallisport Challenge is in-game even in Release builds.
Any title that uses BeginPush/EndPush is potentially effected, so it'l be good to retry any game that previously complained about PushBuffer issues.


RadWolfie:
If `g_dwPrimaryPBCount` variable no longer become necessary, it should be remove within this pull request. I only found 2 functions assigned the `g_dwPrimaryPBCount` value from BeginPush and BeginPush2.

PatrickvL:
This fixed burnout : https://github.com/Cxbx-Reloaded/game-compatibility/issues/197#issuecomment-489379010

PatrickvL:
And fixed GTA 4 : https://github.com/Cxbx-Reloaded/game-compatibility/issues/782#issuecomment-489373471

PR #1603, opened 2019-05-05
Bumpmap bias tweaks

NZJenkins:
Use inputStage for bias check:
We were always checking stage 0 but the stage can vary

SCALE_BX2 instead of BIAS:
Makes the bump effect more extreme than it is now.
The comparison of bias and bx2 is done with hacked together support for L6V5U5 conversion.
The reference shot is a bit blurry, but bx2 seems definitely closer to it

<details>
<summary>Bias / Bx2 comparison</summary>

Reference Image (despite the blurriness note the highlights in the bottom right)
image

Hardware bias (-0.5, 0.5)
hardware_bumpearth_bias

Hardware bx2 (-1, 1)
hardware_bumpearth_bx2

Software with no bias hack
image

JSRF (Makes the blur more extreme)
jsrf_texbemstage

</details>

Disable L6V5U5 biasing (for now):
If L6V5U5 is supported by hardware it is already ranged (-1, 1), so the bias breaks the sample.
Not sure this is the correct host format to ultimately use though

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -2958,7 +2958,7 @@ bool PSH_XBOX_SHADER::InsertTextureModeInstruction(XTL::X_D3DPIXELSHADERDEF *pPS
         if (m_PSVersion >= D3DPS_VERSION(1, 4))
         {
 
-            // If the bump-map texture format is X_D3DFMT_X8L8V8U8 or  X_D3DFMT_L6V5U5 we need to apply a bias
+            // If the bump-map texture format is X_D3DFMT_X8L8V8U8 or X_D3DFMT_L6V5U5 we need to apply a bias
             // This happens because these formats are an alias of unsigned texture formats.
             // Fixes an issue with the JSRF boost-dash effect
             // NOTE: This assumes that this shader will only ever be used for the input bumpmap texture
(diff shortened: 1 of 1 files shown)


revel8n:
Moved comment from #1593, since it is already merged.

The problem is that for luminance formats L would end up in R during conversion instead of U.

If you look at the conversion of the non-luminance formats V and U are assumed to be in G and B, and are handled as such. Since the blind conversion treats it as R6G5B5 the components are converted incorrectly when the format is supposed to be L6V5U5, and similarly for X8L8V8U8.

Also the modifier should be SCALE_BX2 instead of BIAS I believe.

```
            bool bias = false;
            bool swap_components = false;
            if ((format == XTL::X_D3DFMT_X8L8V8U8 && pc_format != XTL::D3DFMT_X8L8V8U8)
                || (format == XTL::X_D3DFMT_L6V5U5 && pc_format != XTL::D3DFMT_L6V5U5)) {
                bias = true;
                swap_components = true;
            }

            auto red_mask = swap_components ? MASK_B : MASK_R;
            auto blue_mask = swap_components ? MASK_R : MASK_B;
```

PatrickvL:
Any progress / recent discoveries on this?

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -2958,17 +2958,19 @@ bool PSH_XBOX_SHADER::InsertTextureModeInstruction(XTL::X_D3DPIXELSHADERDEF *pPS
...
+            // If the bump-map texture format is X_D3DFMT_X8L8V8U8 or X_D3DFMT_L6V5U5 we need to apply a bias
             // This happens because these formats are an alias of unsigned texture formats.
             // Fixes an issue with the JSRF boost-dash effect
             // NOTE: This assumes that this shader will only ever be used for the input bumpmap texture
             // If this causes regressions in other titles, we'll need to be smarter about this
             // and include the texture formats in the shader hash, somehow.
             extern XTL::X_D3DBaseTexture* XTL::EmuD3DActiveTexture[TEXTURE_STAGES];
             extern XTL::X_D3DFORMAT GetXboxPixelContainerFormat(const XTL::X_D3DPixelContainer *pXboxPixelContainer);
-            auto format = GetXboxPixelContainerFormat(XTL::EmuD3DActiveTexture[0]);
+            auto format = GetXboxPixelContainerFormat(XTL::EmuD3DActiveTexture[inputStage]);
```
RadWolfie: This part of change, I had not seen any regression at least few days ago. Somehow, it feels the right method. Although, I had not seen improvement any titles I had tested.

PatrickvL:
Okay I'll merge this. Channel swap, like @revel8n proposed can be investigated later, as can detecting host hardware support for L6G5B5 format.

Thanks!

PR #1605, opened 2019-05-08
Fix an issue with overlay blitting after StretchRects change

LukeUsher:
Thanks to @NZJenkins for pointing out that this was causing an issue with Pirates of the Caribbean, which should now be fixed with this PR


PatrickvL:
LGTM. @NZJenkins, could you add before and after screenshots, please?

PR #1606, opened 2019-05-08
Fix when games try to use a surface as a texture

LukeUsher:
This is quite a bit of a hack, but it works.

Comments in the code detail what happens, but tldr: when xbox uses a surface as a texture, we create a host texture with the same format, width/height as the surface, then copy the surface to a texture, and use that.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -942,9 +942,62 @@ XTL::IDirect3DTexture *GetHostTexture(XTL::X_D3DResource *pXboxResource, int iTe
 
 	if (GetXboxCommonResourceType(pXboxResource) != X_D3DCOMMON_TYPE_TEXTURE) {
 		// Burnout and Outrun 2006 hit this case (retrieving a surface instead of a texture)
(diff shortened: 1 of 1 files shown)


gandalfthewhite19890404:
WWE Raw 2 stops crashing at FMVs. FMVs are still missing, 1605 affected only for PotC.

LukeUsher:
Please don't merge yet, I want to do some additional testing/tweaks later today, but won't have time until after work.

PatrickvL:
Still busy with this?

LukeUsher:
It doesn't work at all in burnout, still investigating and debugging

LukeUsher:
So, turns out this doesn't work at all for the games it might effect:

StrectRects does not work if one of the surfaces is an OffScreenPlainSurface and the other is a Texture.
UpdateSurface does not work since the source surface must be in D3DPOOL_SYSMEM (which it isn't)

The input surface is not lockable, so the LockRect and manual copy approach doesn't work.

D3DXLoadSurfaceFromSurface drops the frame-rate from 60fps all the way down to 3-5, so that's not viable.

This is just going to have to wait until we either have a better way of handling textures overall, or move to a different rendering API.

Closing for now. 

If anyone wants to pick this up feel free.

PR #1609, opened 2019-05-13
Upgrade to Windows 10 SDK and fix inconsistent PlatformToolset

Margen67:


LukeUsher:
Thanks!

For more context, this allowed Cxbx-R to be compiled with both MSVC2017 and MSVC2019 with no changes, since now a Microsoft account is required for MSVC2017.

LukeUsher:
This currently causes a build failure in AppVeyor: 
`error MSB8036: The Windows SDK version 10.0 was not found`

LukeUsher:
Same issue, perhaps 2017 doesn't support the wildcard 10,0/10, maybe it was a new feature of 2019?

LukeUsher:
It turns out that the v10 SDK wildcard is a new feature in VS 2019, so cannot be used in VS 2017.

Based on that, we can't merge this PR until after *all* current contributors agree to make the switch, since it would break the build for anybody who still uses 2017, something we'll have to set a date for and co-ordinate between us

You could, however, submit an alternate PR that just fixes the PlatformToolset

PR #1612, opened 2019-05-14
Selectively unpatch Fiber functions, fixes Futurama

LukeUsher:
Unlike the previous approach, this should not break the DOA games.

It is currently not possible to completely unpatch the Fiber functions in all games, due to some limitations of our current execution model (DOA series specifically break if they are unpatched), however, Futurama breaks if they are patched.

For now, a PATCH_IS_FIBER flag has been added, so they can be selectively unpatched if the game requires it. This will do as an interim solution until we improve threading/scheduling proper (Likely requires CPU emulation)

With this patch, Futurama appears to be fully playable!

### DIFF
--- a/src/core/hle/Patches.cpp
+++ b/src/core/hle/Patches.cpp
@@ -45,6 +45,7 @@ const uint32_t PATCH_ALWAYS = 1 << 0;
 const uint32_t PATCH_HLE_D3D = 1 << 1;
 const uint32_t PATCH_HLE_DSOUND = 1 << 2;
 const uint32_t PATCH_HLE_OHCI = 1 << 3;
+const uint32_t PATCH_IS_FIBER = 1 << 4;
 
 #define PATCH_ENTRY(Name, Func, Flags) \
     { Name, { &Func, Flags} }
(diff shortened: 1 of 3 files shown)


PatrickvL:
LGTM!

Let's wait merging this until testers have had the opportunity to report improvements and/or regressions.

LukeUsher:
All we need to make sure is that this didn't unintentionally break the DOA games, once we know this, we're good

LukeUsher:
Showing proof that Futurama now works, without breaking DOA series that also use Fibers:

image

image

RadWolfie:
List of titles not affected:
* DoAU 1 & 2
* DoA 3 Demo
* The Hulk (even works without the patch)
* Futurama

Could not be tested:
* NFL Blitz 2002 (different issue, only show intro then an error about audio's buffer lock failed)

These are the titles I only have has "Fiber" relative. Plus finding titles with Fiber functions are harder to find and test against.

LukeUsher:
@RadWolfie futurama not effected? It doesn't work before this PR ;)

src/core/hle/Patches.cpp
```diff
@@ -414,6 +447,11 @@ inline void EmuInstallPatch(std::string FunctionName, xbaddr FunctionAddr)
 		return;
 	}
 
+    if ((patch.flags & PATCH_IS_FIBER) && TitleRequiresUnpatchedFibers()) {
```
RadWolfie: It's best to put a note like below to know it's a hack.
`// HACK: Enable fibers functions patch except for certain titles.`

Then it will be easier in the future to find "// HACK:" search or in IDE's built-in tasklist window.

RadWolfie: resolved

PR #1613, opened 2019-05-16
Allow Cxbx-R to compile (without debugger) if a C# compiler isn't found

LukeUsher:
Also fixed an incorrect comment while I was at it.


LukeUsher:
Going to be tweaking this until AppVeyor no longer fails, after that, it should be good to merge

CMakeLists.txt
```diff
@@ -1,10 +1,5 @@
 cmake_minimum_required (VERSION 3.8)
 
-# Enforce to use Windows 8.1 SDK
-if(${CMAKE_HOST_SYSTEM_NAME} STREQUAL "Windows")
```
LukeUsher: This has been removed because we don't actually require 8.1, was just used that as a default to support MSVC2015 and MSVC2017, but now cmake can take care of this

PR #1615, opened 2019-05-18
Add VS cache to .gitignore

NZJenkins:
Visual Studio creates a .vs cache folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 [Bb]uild/
 [Bb]uild-*/
 
+# Visual Studio Cache
+.vs/
+
 # Generated files
 *.aps
 


JayFoxRox:
This should not have been merged.

It's not the task of .gitignore to handle personal preferences (suchas IDE or compiler).

The proper solution is to configure your local copy of git. You can do this system wide, or in your repo in `./git/info/exclude` (which works like .gitignore but it's not part of the shared repository).

Also see https://git-scm.com/docs/gitignore#_description

You can typically also configure your IDE to avoid these files in the first place (ideally, by moving them into locations which are preloaded into RAM).

PatrickvL:
Having common filename patterns in the project itself (rather than relying on correctly setup contributor environments) does prevent accidental commits of files that shouldn't be in the repository.

JayFoxRox:
That's why review exists: To find poor changes (accidental changes, legal issues, design issues, ...) and fix / educate.

I could *still* add a ton of garbage manually (and many other IDEs will still do that).
It's not .gitignores purpose to prevent bad code.

If you want a permanent solution in this style (handling more editors, cache files, ...), your only option is to [`cat` all files in this repository](https://github.com/github/gitignore) and throw it into the project .gitignore... except it will cause more problems than doing anything good.

For VS cache files it's probably somewhat fine, as most people are probably currently using MSVS for Cxbx-R - so it's not worth discussing this further; but strictly speaking the change is still wrong. I'm surprised such changes are being merged (and defended), as it sets a precedent for the future.

Issue #1604, opened 2019-05-05
X_D3DFMT_L6V5U5 breaks if not supported by hardware


NZJenkins:
BumpEarth uses a bump texture stored as L6V5U5
If L6V5U5 is not supported natively, CXBX attempts to convert to ARGB using R6G5B5 to A8R8G8B8 conversion.

<details>
<summary>BumpEarth comparison</summary>

Hardware without L6V5U5
![image](https://user-images.githubusercontent.com/9897874/57190311-ddcf1d00-6f6c-11e9-8348-a5a8f2563595.png)

Software D3D Reference Device
![image](https://user-images.githubusercontent.com/9897874/57190491-d14bc400-6f6e-11e9-8712-b3e6c231e943.png)
</details>

It's tricky because this is one of a few formats that shares a value with another:
X_D3DFMT_L6V5U5 == X_D3DFMT_R6G5B5

Related formats and test cases:
```
// Test case BumpEarth -> is stored as L6V5U5
D3DFMT_L6V5U5 : TEXTURE_FORMAT_COLOR_SZ_R6G5B5 // Alias D3DFMT_R6G5B5
// Test case JSRF boost dash -> seems to be stored as X8R8G8B8 works correctly
D3DFMT_X8L8V8U8 : TEXTURE_FORMAT_COLOR_SZ_X8R8G8B8 // Alias D3DFMT_X8R8G8B8
// Test case ???
D3DFMT_LIN_L6V5U5 : TEXTURE_FORMAT_COLOR_LU_IMAGE_R6G5B5 // Alias D3DFMT_LIN_R6G5B5
```
Some observations:
- Seems all Xbox LVU formats are aliases
- Seems all internal formats are RGB

I don't know enough about textures to confidently fix this.
I suspect:
- there can only be one actual format per X_D3DFMT
- we should always convert to RGB textures

It could be worth testing a build that fixes BumpEarth, as if there are textures sharing the same format, but are actually stored as RGB instead of LVU, then it should be fairly obvious.
However conversion to ARGB doesn't maintain the same bits-per-component of the texture, so apparently could cause issues.

PatrickvL:
It'd also be worthwhile to find out (which bits of) which channels are read by the bump mapping texture fetch operation. Perhaps we can suffice by using R5G6B5 on host instead, and using the Xbox texture data as-is?

NZJenkins:
In pixel shader, LVU textures get their channels reversed so
[...]

PR #1618, opened 2019-05-22
Extracted IsSupportedFormat() method, now also called in bias-determination

PatrickvL:
This fixes a TODO in the X_D3DFMT_L6V5U5 pixel shader bias determination (a leftover from PR #1603)

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -2964,15 +2964,30 @@ bool PSH_XBOX_SHADER::InsertTextureModeInstruction(XTL::X_D3DPIXELSHADERDEF *pPS
             // NOTE: This assumes that this shader will only ever be used for the input bumpmap texture
             // If this causes regressions in other titles, we'll need to be smarter about this
             // and include the texture formats in the shader hash, somehow.
-            extern XTL::X_D3DBaseTexture* XTL::EmuD3DActiveTexture[TEXTURE_STAGES];
-            extern XTL::X_D3DFORMAT GetXboxPixelContainerFormat(const XTL::X_D3DPixelContainer *pXboxPixelContainer);
-            auto format = GetXboxPixelContainerFormat(XTL::EmuD3DActiveTexture[inputStage]);
             bool bias = false;
 			auto biasModifier = (1 << ARGMOD_SCALE_BX2);
-			// TODO L6V5U5 format is converted incorrectly if not supported by the device
-            if (format == XTL::X_D3DFMT_X8L8V8U8 /*|| format == XTL::X_D3DFMT_L6V5U5*/) {
-                bias = true;
-            }
+			auto pXboxTexture = XTL::EmuD3DActiveTexture[inputStage];
+			if (pXboxTexture != nullptr) {
+				extern XTL::X_D3DFORMAT GetXboxPixelContainerFormat(const XTL::X_D3DPixelContainer *pXboxPixelContainer); // TODO : Move to XTL-independent header file
+
+				switch (GetXboxPixelContainerFormat(pXboxTexture)) {
+					case XTL::X_D3DFMT_L6V5U5: {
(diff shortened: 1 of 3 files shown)


RadWolfie:
Before:
image

After:
image

I had checked the return value of IsSupportedFormat is `false`.

Plus I did check the difference between with and without bias for software mode. It only show a little difference and did not shift to different location like above image did.

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -2964,15 +2964,30 @@ bool PSH_XBOX_SHADER::InsertTextureModeInstruction(XTL::X_D3DPIXELSHADERDEF *pPS
...
+				extern XTL::X_D3DFORMAT GetXboxPixelContainerFormat(const XTL::X_D3DPixelContainer *pXboxPixelContainer); // TODO : Move to XTL-independent header file
+
+				switch (GetXboxPixelContainerFormat(pXboxTexture)) {
+					case XTL::X_D3DFMT_L6V5U5: {
+						extern XTL::X_D3DRESOURCETYPE GetXboxD3DResourceType(const XTL::X_D3DResource *pXboxResource); // TODO : Move to XTL-independent header file
+						extern bool IsSupportedFormat(XTL::X_D3DFORMAT X_Format, XTL::X_D3DRESOURCETYPE XboxResourceType, DWORD D3DUsage); // TODO : Move to XTL-independent header file
+
+						// L6V5U5 format is converted incorrectly if not supported by the device
+						XTL::X_D3DRESOURCETYPE XboxResourceType = GetXboxD3DResourceType(pXboxTexture);
+						DWORD D3DUsage = 0; // TODO : How to determine this?
```
RadWolfie: Also, this is the main reason I had not merge yet.

PatrickvL: This is less an issue than the screenshots showing a regression. So yeah, better hold this one off.

RadWolfie:
> Before, L6V5U5 was excluded from bias, now bias is applied (when the host doesn't support the format).
> So BumpEarth should definitely change on hardware that doesn't support it?
> We still need to correct the data itself though

I'm thinking the format was not converted correctly. Though, I'm not entirely sure yet.

> No change on software mode, it still renders correctly as expected

I'm not exactly sure what reason the software mode did not made any change. Which remind me, I do not remember if I did test `IsSupportedFormat` value in software mode.

LukeUsher:
@RadWolfie what you are saying is exactly why the bias is needed :P

If the host GPU supports L6V5U5 (or software mode is used, which also supports L6V5U5), the bumpearth sample works as expected, and looks okay.

If the host GPU does *not* support L6V5U5, it gets interpreted as the wrong format, because R6G5B5 and L6V5U5 use the same format ID, it's not possible to tell the difference until usage time. This means that the data is interpreted incorrectly, and the 'bias' is an attempt to fix this scenario.

The bias fix works for JSRF but is not enough to properly fix bumpearth

For more info, please see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1604

LukeUsher:
So is this ready for merge or are we awaiting further testing? I'm a little reluctant to merge this without seeing many test results because it doesn't fix the depth buffer format problem, just hides it, and the resulting behavior is very hardware-specific (it differs based on what formats the host GPU supports)

RadWolfie:
D3DUsage being hard code to 0 is what prevent me from merge this pull request. Link: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1618#discussion_r287601582

It would be nicer to have D3DUsage detection implemented before the merge other wise it will be consider as a hack.

src/core/hle/D3D8/XbPixelShader.cpp
PatrickvL: Perhaps rendertarget and depthstencil usage can be detected by comparing the texture at hand to the corresponding global variables. But I'm not entirely sure this will work under all circumstances. Nor do I know of any method to detect other uses...

PR #1620, opened 2019-05-26
Always apply scalar swizzle fix

NZJenkins:
We fix up swizzling when writing into scalar output registers (`oPts`, `oFog`)
This PR removes the selective application of the fix.
This fixes visual issues in Mechassault
And something in Amped (no visible difference to me)
Both of these games `MUL` into scalar registers, so perhaps we keep things selective and just add MUL?

image

image


RadWolfie:
Tested against commit 55e2643
<details>
<summary>GTA 3</summary>

So far seems the roads are now visible. Other graphic issues appear to remain. Also, the car's shadow from streetlights remain the same working condition.

Before:
GTA3-before

After:
GTA3-after

</details>

Titles with washout texture, like GTA3 and MechAssault, are hard to find by the way. I had thought of Oddworld: Munch's Oddysee (demo version) title, except the visual remain the same.

LukeUsher:
I think it's safe to merge this now

PR #1621, opened 2019-05-26
Use stringstream instead of fixed-size buffer to prevent overrun

NZJenkins:
Previously we used generous estimate for the size of the vertex shader.
In SSX Tricky it's not enough to fit one of the vertex shaders, and the heap gets corrupted.
This commit changes things over to use `std::stringstream`

I'm sure there are issues with this commit, I'm not hugely familiar with c++
And I think the syntax is awful, but as far as I could tell better syntax isn't available, without using a library like fmt?

### DIFF
--- a/src/core/hle/D3D8/XbVertexShader.cpp
+++ b/src/core/hle/D3D8/XbVertexShader.cpp
@@ -712,44 +712,39 @@ static char *VshGetRegisterName(VSH_PARAMETER_TYPE ParameterType)
     }
 }
 
-static void VshWriteOutputMask(boolean *OutputMask,
-                               char    *pDisassembly,
-                               uint32_t *pDisassemblyPos)
+static void VshWriteOutputMask(boolean     *OutputMask,
+                               std::stringstream *pDisassembly)
 {
     if(OutputMask[0] && OutputMask[1] && OutputMask[2] && OutputMask[3])
     {
         // All components are there, no need to print the mask
         return;
     }
-    *pDisassemblyPos += sprintf(pDisassembly + *pDisassemblyPos, ".%s%s%s%s",
-                                OutputMask[0] ? "x" : "",
-                                OutputMask[1] ? "y" : "",
-                                OutputMask[2] ? "z" : "",
-                                OutputMask[3] ? "w" : "");
+	*pDisassembly << "." << (OutputMask[0] ? "x" : "")
+						 << (OutputMask[1] ? "y" : "")
+						 << (OutputMask[2] ? "z" : "")
+						 << (OutputMask[3] ? "w" : "");
 }
 
 static void VshWriteParameter(VSH_IMD_PARAMETER *pParameter,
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -712,44 +712,39 @@ static char *VshGetRegisterName(VSH_PARAMETER_TYPE ParameterType)
     }
 }
 
-static void VshWriteOutputMask(boolean *OutputMask,
-                               char    *pDisassembly,
-                               uint32_t *pDisassemblyPos)
+static void VshWriteOutputMask(boolean     *OutputMask,
```
LukeUsher: You'll be able to get cleaner code and avoid manually dereferencing pointers if you pass as a reference instead of a pointer, eg:

`std::stringstream& disassembly`

would allow the below code to become
```
disassembly << "." << (OutputMask[0] ? "x" : "")
                                OutputMask[0] ? "x" : "",							 << (OutputMask[1] ? "y" : "")
                                OutputMask[1] ? "y" : "",							 << (OutputMask[2] ? "z" : "")
                                OutputMask[2] ? "z" : "",							 << (OutputMask[3] ? "w" : "");
```

This applies to all functions you pass a stringstream pointer along to.

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -820,35 +815,34 @@ static const BYTE DeclAddressUsages[][2] =
...
 						   XTL::D3DVERTEXELEMENT *pRecompiled,
                            boolean Truncate)
 {
-    uint32_t DisassemblyPos = 0;
     switch(pShader->ShaderHeader.Version)
     {
         case VERSION_VS:
 #ifdef CXBX_USE_VS30
-            DisassemblyPos += sprintf(pDisassembly + DisassemblyPos, "vs.3.0\n");
+			*pDisassembly << "vs.3.0\n";
```
LukeUsher: When using C++ stringstreams, you can/should use `<< std::endl` rather than `\n`

NZJenkins: For aesthetic reasons? I think it's weird to call flush for a stringstream
Here is a discussion on the topic, it was recommended to be avoided generally https://github.com/isocpp/CppCoreGuidelines/issues/357

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: Done

src/core/hle/D3D8/XbVertexShader.cpp
LukeUsher: Ah, don't worry about it then, I wasn't actually aware it had a behavioural change, I was always of the assumption it was the C vs C++ way :P

LukeUsher:
nice job

PR #1623, opened 2019-05-30
Fix an issue where the wrong depth buffer was selected

LukeUsher:
This fixes a (stupid) bug where the Default Depth Stencil was always set, even if there was another active depth buffer.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -159,8 +159,9 @@ static UINT                         QuadToTriangleD3DIndexBuffer_Size = 0; // = 
 static XTL::INDEX16                *pQuadToTriangleIndexBuffer = nullptr;
 static UINT                         QuadToTriangleIndexBuffer_Size = 0; // = NrOfQuadVertices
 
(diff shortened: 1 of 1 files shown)


RadWolfie:
After careful code review, looks good to me.

I had tested known titles (list below) for any regress and have not find any regression.
* Red Faction 2
* Crimson Skies
* Jet Set Radio Future (improved, expand "JSRF improvement" for screenshots)

Other tittles did not appear to regress.

<details><summary>JSRF improvement</summary>

Before:
JSRF-before

After:
JSRF-after
</details>

However, Jet Set Radio Future is still using same position for 2nd viewpoint when using internal upscale.

RadWolfie:
I am leaving this pull request open if Luke wants to implement any further fix toward internal upscale or not.

PatrickvL:
Also, when handling device reset, this global variable should be nilled

LukeUsher:
> Also, when handling device reset, this global variable should be nilled

We don't HLE the D3D Reset function, DirectX takes care of it on the Xbox side.
Internally, it frees the framebuffers which will trigger D3DResource_Release, causing the variables to be cleared.

LukeUsher:
@RadWolfie thanks for testing

PR #1627, opened 2019-06-04
Align TLSData to 16 byte boundary

ergo720:
Currently, the alignment of the TLSData in Cxbx-R is incorrect. This breaks compatibility with xbes built with nxdk, since the `_PDCLIB_xbox_thread_startup` in pdclib asserts an alignment of 16 byte for TLSData (see here for more details https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/issues/60). This PR fixes this by enforcing the expected 16 byte alignment. I tested it with most of my games and I didn't observe any ill effects. Someone should also test Futurama with this since it uses a thread hack but I don't have the game to test.

### DIFF
--- a/src/core/kernel/support/EmuFS.cpp
+++ b/src/core/kernel/support/EmuFS.cpp
@@ -561,11 +561,12 @@ void EmuGenerateFS(Xbe::TLS *pTLS, void *pTLSData)
 			}
 
 			/* + HACK: extra safety padding 0x100 */
-			pNewTLS = (void*)g_VMManager.AllocateZeroed(dwCopySize + dwZeroSize + 0x100);
-
+			pNewTLS = (void*)g_VMManager.AllocateZeroed(dwCopySize + dwZeroSize + 0x100 + 0xC);
+			/* Skip the first 12 bytes so that TLSData will be 16 byte aligned (addr returned by AllocateZeroed is 4K aligned) */
+			pNewTLS = (uint8_t*)pNewTLS + 12;
 
 			if (dwCopySize > 0) {
-				memcpy(pNewTLS, pTLSData, dwCopySize);
+				memcpy((uint8_t*)pNewTLS + 4, pTLSData, dwCopySize);
 			}
 
 #ifdef _DEBUG_TRACE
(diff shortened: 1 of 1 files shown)


PatrickvL:
It appears as if this is actually two fixes into one:
* Align TLS data
* Discern the difference between the TLS header and TLS data
This last part makes me wonder if they could/should be allocated separately, and if so, then maybe TLS data is meant to be allocated in entire page increments? (That'd explain the 0x100 safeguard padding)

LukeUsher:
Most of this part of Cxbx-R is *very* wrong in a lot of ways: especially regarding TLS data and its relationship with the stack base. (The memory should be allocated as a kernel stack, with TLS data at the end of the allocated region, but with stack base being set to the address of TLS data, such that the stack grows down from there. I've done some work on fixing this in the past, but many things fall apart due to assumptions elsewhere) 

The entire KPCR and Thread init code do need replacing at some point: it's *very* wrong, and is part of the reason we cannot get consistently working Fibers.

We also do unnecessary copying of data that the running executable will then repeat on it's own.

This fix will do for now, as it solves a problem, but this code does need revisiting in the future.

LukeUsher:
Pending on test results from Fiber-sensitive games: DOA3, DOA2 Ultimate, and Futurama

RadWolfie:
Test results list:

Before pull request
- no modification
   - DoA 3 Demo - **works**
   - DoA 2 Ultimate - **works**
   - Futurama - **works** (default is force unpatch)
     - Enforce patch - **doesn't work**
- force disable fiber patch
   - DoA 3 Demo - **works**
   - DoA 2 Ultimate - **doesn't work**
   - Futurama - **works**

With pull request
- no modification
   - DoA 3 Demo - **works**
   - DoA 2 Ultimate - **works**
   - Futurama - **works**
     - Enforce patch - **doesn't work**
- force disable fiber patch
   - DoA 3 Demo - **works**
   - DoA 2 Ultimate - **doesn't works**
   - Futurama - **works**


Seems like DoA 3 Demo can't be use as test case.

@ergo720, the fiber patches are forward to Windows API btw.

LukeUsher:
thanks, that's good enough for me.

PR #1629, opened 2019-06-07
Remove DBG_PRINTF

NZJenkins:
And some logging fixes


PatrickvL:
A few questions about this PR :
* Why replace `DBG_PRINTF(".*\n"` with `EmuLog(LOG_LEVEL::DEBUG, ".*"` at all?
* The replacement is more verbose, so perhaps a more compact replacement is preferable?
* Also, which uses of `DBG_PRINTF` remain? 
* And why are they remaining?
* Can't `DBG_PRINTF` be removed entirely (including it's `#define`)?
* The remark "And some logging fixes" - which ones are that, specifically?

src/core/kernel/exports/EmuKrnlKe.cpp
```diff
@@ -2236,7 +2236,7 @@ XBSYSAPI EXPORTNUM(159) xboxkrnl::NTSTATUS NTAPI xboxkrnl::KeWaitForSingleObject
 					Timer->Header.WaitListHead.Blink = &WaitTimer->WaitListEntry;
 					WaitTimer->NextWaitBlock = WaitBlock;
 					if (KiInsertTreeTimer(Timer, *Timeout) == FALSE) {
-						DBG_PRINTF("%s: KiInsertTreeTimer(Timer, *Timeout) == FALSE\n", __func__);
+						EmuLog(LOG_LEVEL::DEBUG, "%s: KiInsertTreeTimer(Timer, *Timeout) == FALSE", __func__);
 						WaitStatus = (NTSTATUS)STATUS_TIMEOUT;
```
ergo720: You can remove this logging here. I initially added it to diagnose a timer bug but I forgot to remove it after I fixed it in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1579

RadWolfie:
> * Why replace `DBG_PRINTF(".*\n"` with `EmuLog(LOG_LEVEL::DEBUG, ".*"` at all?

I can answer to this question, `DBG_PRINTF` is an ancient macro which didn't have a newline include. Instead, it is place in the `printf` function directly. `EmuLogEx` already include the newline, so extra newline aren't necessary.

> * The replacement is more verbose, so perhaps a more compact replacement is preferable?

As for this question, what do you mean? `EmuLog` / `EmuLogEx` is what we should be using directly.

PatrickvL:
> > * Why replace `DBG_PRINTF(".*\n"` with `EmuLog(LOG_LEVEL::DEBUG, ".*"` at all?
> 
> I can answer to this question, `DBG_PRINTF` is an ancient macro which didn't have a newline include. Instead, it is place in the `printf` function directly. `EmuLogEx` already include the newline, so extra newline aren't necessary.
> 
> > * The replacement is more verbose, so perhaps a more compact replacement is preferable?
> 
> As for this question, what do you mean? `EmuLog` / `EmuLogEx` is what we should be using directly.

This PR changes one symbol (`DBG_PRINTF`) by two symbols (`EmuLog` and `LOG_LEVEL::DEBUG`), that combined take up more characters than what we had before.
This means there's more code to maintain, more room for error, more risk of slight deviations creeping in, etc.

Hence my initial question : Why do this at all? (Apart from no longer newline requiring a terminating newline.)

Calling the old macro "ancient" is at best an incomplete motivation.

I could live with "this change paves the way for {something}", or whatever.
I'm not trying to be harsh, but for now its just more code for no apparent benefit.

RadWolfie:
While I understand your concern, I don't see a reason to continue use misleading name `DBG_PRINTF` when it's being handle on release build. `DBG_PRINTF` is more of a common sense understanding it will only output in debug build, not in release build for other programs' environment.

We can still find `LOG_LEVEL::DEBUG` easier the same way with `DBG_PRINTF` with unix's command or IDE's search tool. Plus our `EmuLogEx` function has been in finalize stage for very long time.

Also I am fine with the macro rename if that helps.

LukeUsher:
I'm all for removing DBG_PRINTF: We have the central EmuLog system, so why not use it?

If we really want to keep DBG_PRINTF, we could update the macro so that it forwards to EmuLog internally, but I still think there's little point in keeping it.

LukeUsher:
You could use std::stringstream to construct a string that you pass to a single emulog call. That might be the best approach.

PatrickvL:
Or make it one call by concatenating all strings and arguments

RadWolfie:
@LukeUsher fyi 😛
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/e37de96ed2aec1450df6c6b5eaf533628364ae4e/src/common/Logging.h#L403-L404

---
@NZJenkins For new lines, you can still do use `\n`. So, in your case it will be:

```
EmuLog(LOG_LEVEL::DEBUG, "Looking at ring descriptor %d (0x%llx): "
    "\n   Buffer: 0x%x, "
    "\n   Length: 0x%x, "
    "\n   Flags: 0x%x",
    s->rx_ring_index, rx_ring_addr, desc.packet_buffer, desc.length, desc.flags);
```

PatrickvL:
Ah, indeed, IF a replacement is done, then at least replace it with something that will still use the `LOG_PREFIX` which is defined differently per source code file!

LukeUsher:
Looks good to me, thanks for the work!

PR #1637, opened 2019-06-17
'Fix' Smashing drive non-nvidia hardware

LukeUsher:
The issue seems to be because non-nvidia hardware clamps the fog values when below 0, making the formula for fog evaluate to (0 - z) / 0, invoking undefined behavior and breaking the scene.

This fix attempts to recover from this situation by taking the absolute value of the fog start/end, in the case when the value is < 0. 

A LOG_TEST_CASE has been added to detect this situation in future, so we can collect test-cases.

Before:
image

After:
image

### DIFF
--- a/src/core/hle/D3D8/XbState.cpp
+++ b/src/core/hle/D3D8/XbState.cpp
@@ -27,6 +27,7 @@
 #define _XBOXKRNL_DEFEXTRN_
 #define LOG_PREFIX CXBXR_MODULE::D3DST
 
+#include "core\kernel\init\CxbxKrnl.h"
 #include "core\kernel\support\Emu.h"
 #include "core\kernel\support\EmuXTL.h"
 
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/XbState.cpp
```diff
@@ -26,7 +26,8 @@
 // ******************************************************************
 #define _XBOXKRNL_DEFEXTRN_
 #define LOG_PREFIX CXBXR_MODULE::D3DST
-
+
+#include "core\kernel\init\CxbxKrnl.h"
```
LukeUsher: This is required because LOG_TEST_CASE doesn't work without this,

Margen67:
Confirmed working on NVIDIA
image

literalmente-game:
Checked with the games listed below, no regressions found with them:
- Alien Hominid
- Bloodrayne 2
- Crash 'N' Burn
- Dinotopia - The Sunstone Odyssey
- FIFA Street
- Jet Set Radio Future
- Outlaw Golf
- Outrun 2
- Panzer Dragoon Orta
- Psyvariar 2 - Extend Edition
- Star Wars - Jedi Starfighter
- Tennis Master Series 2003

LukeUsher:
Okay, cool, happy to merge based on this

PR #1639, opened 2019-06-18
Implement Vertex Buffer Caching

### LINKED ISSUE #1277: Implement vertex buffer caching

LukeUsher:
Implements vertex buffer caching by means of a LRU (least recently used) cache.

The cache uses a std::list (linked list) of entries for fast insertion and removal, as well as a std::unordered_map of iterators into the linked list, for fast lookup.

There are two variables for controlling the caching algorithm, m_MaxCacheSize, which determines the ideal maximum size of the cache, and m_CacheElasticity, which defines how much larger than the maximum size the cache is allowed to grow before being purged back to the maximum size. This is done to prevent constant thrashing once the cache is full.

Cache entries are maintained in order of most to least used in the linked list, so purging the cache is as simple as popping the last n entries out of the cache.

This PR also updates to a newer version of xxHash for significantly increased hashing performance (it's now vectored with SSE instructions). A side-effect is that this will invalidate the HLE/Symbol cache, but since the hashes are different, no user interaction is required: Cxbx-R will automatically re-scan.

In terms of real world performance, most users should see a significant decrease in lag/stutter across the board, and many games that previously ran 30-40fps are now able to reach 60fps.

Even the Dashboard is noticeably smoother/has less stutter/lag.

A small number of titles may be negatively effected by the cache, but these should be rare, they can be investigated on a case-by-case basis if they are effected to an extreme. 

A hotkey (F1) has been added to dump current cache status to console.

Fix #1277

(as a side-note: This PR doubles the performance of JSRF on my laptop, so the results are exciting)


LukeUsher:
Note that people with *very* slow CPUs might experience reduced performance in this case: The result of this PR is more-time is spent hashing resources, but much less time is spent converting resources, and less time is spent uploading data to the GPU, so there is less CPU->GPU resource thrashing.

But the vast majority of people should see noticeable improvements, and performance tuning will be done elsewhere to reduce the impact of the additional hashing.

literalmente-game:
These numbers are related to the commit before this PR, please check against them for comparison against the version PR'ed so we can see what improved/regressed
preview

RadWolfie:
If there's one thing I preferred to change is use `xxHash` as a submodule. XbSymbolGenCacheTest is using it as submodule. Here's the commit which @JayFoxRox did. https://github.com/Cxbx-Reloaded/XbSymbolCacheGenTest/commit/24d72145f643e170defd11f63ac9113d0543e856

LukeUsher:
@RadWolfie nice suggestion, feel free to make a PR to do that, but I'm not going to do that as part of this PR ;)

RadWolfie:
Yeah, you're right. We should do it in a separate PR than in here.

PatrickvL:
@literalmente-game gunmetal seems to have worsened, is that correct?

literalmente-game:
@PatrickvL Yes, but this is using _that_ commit before some of the improvements made in this PR, I (or someone else) need to check if it's better now

LukeUsher:
Some games having slightly worse performance isn’t completely unexpected, but the majority of games have significant improvements and those which have not improved still have noticeably less stutter than before.  Like everything in emulation it’s a trade off. 

This work prevents unnecessary cpu-gpu transfers, unnecessary conversions,and unnecessary copies.

On the flip side, more time is spent on hashing. For most games this is a huge improvement since CPU-GPU bandwidth is usually the biggest performance impact. 

The amount of titles negatively effected is minimal, and those that are, will be investigated and addressed at a later time.

Right now, improving the vast majority of games is worth the slight performance hit in the remaining few.

RadWolfie:
I found two titles has big impact so far. I had confirmed both titles has vertex functions in the cache file.

<details><summary>Burnout</summary>

Before:
Burnout-before

After:
Burnout-after

After (during gameplay):
Burnout-after2

</details>

<details><summary>Turok Evolution</summary>

Before:
Turok Evolution-before

After:
Turok Evolution-after

After (when not close to the enemy):
Turok Evolution-after2

</details>

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -207,7 +122,7 @@ size_t GetVerticesInBuffer(DWORD dwOffset, DWORD dwVertexCount, PWORD pIndexData
 	DWORD highestVertexIndex = 0;
 	for (DWORD i = 0; i < dwVertexCount; i++) {
 		if (highestVertexIndex < pIndexData[i]) {
-			highestVertexIndex = pIndexData[i];
+ 			highestVertexIndex = pIndexData[i];
```
RadWolfie: I see an unintended space here.

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -274,6 +189,45 @@ inline FLOAT NormShortToFloat(const SHORT value)
...
+        } while (count > 0);
+    }
+
+    return stream;
+}
+
+void XTL::CxbxVertexBufferConverter::PrintStats()
+{
+    printf("Vertex Buffer Cache Status: \n");
+    printf("- Cache Size %d\n", m_PatchedStreams.size());
```
RadWolfie: Missing semi-dot character here.

literalmente-game:
1 regression (Gunmetal crashes in the menus with the vertex creation error message + worse performance than the commit before this PR in general:
preview 2

LukeUsher:
Okay cool, so we have found that the newer xxhash is faster than the older xxhash, but the sse4 crc32c hash extension is even faster, so I’ll restore our modifications to use hardware crc32c when available and that should restore performance for you. As for gunmetal, no idea why it crashes now but I’ll look into it

LukeUsher:
@RadWolfie feedback addresses
@literalmente-game we now use a different hash algorithm based on if the users hardware supports fast crc32c or not, so this should fix your performance hit between your initial testing and the final PR

src/common/util/hasher.h
```diff
@@ -0,0 +1,34 @@
...
+// *
+// *  All rights reserved
+// *
+// ******************************************************************
+#ifndef _HASHER_H
+#define _HASHER_H
+
+#include <stdint.h>
+
+extern __forceinline uint64_t ComputeHash(void* data, size_t len);
```
LukeUsher: __forceinline is used as the compiler may choose to ignore standard inline, but we know that inlining improves performance vs the function call overhead because of how often hashing is done, so we override the compiler in this one case.

src/common/util/hasher.cpp
```diff
@@ -0,0 +1,43 @@
...
+    HASH_CRC32C
+};
+
+static int g_HashAlgorithm = HASH_NONE;
+
+void InitHasher()
+{
+    // Detect the best hashing algorithm to use for the host machine
+    // TODO/Future Improvement: This could be expanded to support even more hash algorithims
+    // And we could hash a random buffer to calculate the fastest hash to use on a given host
```
LukeUsher: Right now, selecting crc32c when hardware acceleration is available is a good bet, but there may be faster hash algorithms available in future: Once we extend this in a later PR to support more algorithms, we'll update this function to time each hash, so the best hash algorithm can be automatically chosen based on the current host machine.

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -715,38 +708,32 @@ void XTL::CxbxVertexBufferConverter::ConvertStream
...
 		}
 	}
+
+    patchedStream.isValid = true;
+    patchedStream.pCachedXboxVertexData = pXboxVertexData;
+    patchedStream.uiCachedXboxVertexDataSize = xboxVertexDataSize;
+    patchedStream.uiVertexDataHash = vertexDataHash;
+    patchedStream.uiCachedXboxVertexStride = uiXboxVertexStride;
+    patchedStream.uiCachedHostVertexStride = uiHostVertexStride;
+    patchedStream.bCacheIsStreamZeroDrawUP = (pDrawContext->pXboxVertexStreamZeroData != NULL);
```
RadWolfie: `NULL` should be `xbnullptr`.

LukeUsher: nullptr, because that's a C++ construct but null is standard C:P

I'm fully of the opinion xbnullptr is pointless, it serves no purpose. There's no different, functionally between xbnullptr and nullptr.

We shouldn't have such useless constructs around in the code: xbnullptr has zero benefits to us.

PatrickvL: If used correctly all throughout our codebase, xbnullptr clearly distinguishes Xbox from host data. The same applies to all Xbox type declarations, which have an X_ prefix. Not adhering to such conventions would remove information

LukeUsher: I understand that, but it's better suited for a variable name (which in this case clearly shows that it is XboxVertexData)

It's different than the X_ D3D stuff, because the values of such differ between Xbox and Host.
A null pointer is *always* a null pointer.

Additionally, if it was any other assignment than a null pointer, context would be lost too. making it useless at it's 'intended' purpose.

nullptr == xbnullptr, so it's not like type checks can be used to rely on it either, since either would evaluate to true.

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -392,14 +336,63 @@ void XTL::CxbxVertexBufferConverter::ConvertStream
...
+    // No need to hash or patch at all in this case!
+    if (pDrawContext->pXboxVertexStreamZeroData != xbnullptr && !bNeedStreamCopy) {
+        pHostVertexData = pXboxVertexData;
+
+        CxbxPatchedStream stream;
+        stream.isValid = true;
+        stream.uiCachedHostVertexStride = uiHostVertexStride;
+        stream.bCacheIsStreamZeroDrawUP = true;
+        stream.pCachedHostVertexStreamZeroData = pHostVertexData;
+        stream.bCachedHostVertexStreamZeroDataIsAllocated = bNeedStreamCopy;
```
RadWolfie: I suggest to set this to `false` to avoid confusion.

LukeUsher: It's not even needed: that field is default false always anyway.

src/core/hle/D3D8/XbVertexBuffer.cpp
PatrickvL: But by that reasoning we could also use 0 instead of nullptr, xbnullptr, false, heck, even for 0.0f and NT_STATUS_OK.

Still, those symbols are used instead of a zero digit, because they convey meaning.

LukeUsher: NULL, 0 and nullptr have differences: nullptr does not behave as an integer but NULL and 0 do, the difference is not purely semantic. 

For NT_STATUS_OK, that's because the data type of the variable is NTSTATUS, not an integer.

src/core/hle/D3D8/XbVertexBuffer.cpp
LukeUsher: Latest commit addresses this

src/core/hle/D3D8/XbVertexBuffer.cpp
RadWolfie: I'm pretty sure we all had agreed to use xbnullptr for xbox vs (nullptr) host data distinguish.

LukeUsher: I never did,but I was overruled :p

LukeUsher: updated to clarify the differnce between NULL, 0 and nullptr

LukeUsher:
Regression effecting Burnout has been fixed, this probably fixes Turok too, but could do with re-testing.

The cause was that I did not invalidate cache entires if the vertex shader information had changes, which made the existing conversion invalid. Now a hash of the vertex stream info is included in the cache.

image

RadWolfie:
I can confirm Burnout works, except for Turok which didn't resolve the issue.

I wonder if it's because of `bNeedVertexPatching` if statement, line 422 as of commit 3eb75d5, check to convert the requirement every time it get activate. Since the value can change at any time. I could be wrong on this.

LukeUsher:
We're at a good place now, let's get this merged!

PR #1646, opened 2019-06-26
Fix a deadlock when some LTCG titles attempt to wait for vertical blank

LukeUsher:
With this, conker no longer hangs. It still doesn't work, however.

The cause was that some LTCG titles replace WaitForVerticalBlank with nothing more than a pure wait for interrupt stub, which means that interrupts must be working

### DIFF
--- a/src/devices/video/nv2a.cpp
+++ b/src/devices/video/nv2a.cpp
@@ -1209,12 +1209,12 @@ void NV2ADevice::Init()
 	// Setup the conditions/mutexes
 	pgraph_init(d);
 
-	// Only spawn VBlank thread when LLE is enabled
+	// Only init PVIDEO when LLE is enabled
 	if (d->pgraph.opengl_enabled) {
 		pvideo_init(d);
-
-		vblank_thread = std::thread(nv2a_vblank_thread, d);
 	}
+
+    vblank_thread = std::thread(nv2a_vblank_thread, d);
 
     qemu_mutex_init(&d->pfifo.pfifo_lock);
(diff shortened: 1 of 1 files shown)


PatrickvL:
As small this change may be, it does result in significantly different behaviour, so rigorous testing is desirable for this...

LukeUsher:
@PatrickvL this is actually much less impact than it appears: All it does is allow the D3D VBlank ISR to be called, which updates some internal D3D variables, triggers the Vblank event, and (optionally) passes through a swap command to PGRAPH, nothing more: The impact is minimal.

It does, however, mean that games that wait for VBlank by waiting on the internal D3D event instead of calling WaitForVerticalBlank(), no longer hang.

LukeUsher:
No comments or test results? This is quite an important piece to kicking more games from 'hanging/nothing' to 'booting' state or higher.

LukeUsher:
After further investigation/testing, this has been found to have no unwanted/unexpected side-effects: The VBlank routine has minimal functionality. 

This might eventually lead to the unpatching of D3DDevice_WaitForVerticalBlank and perhaps even removal of some of our SwapCount related hacks!

numchuk12:
This implementation prevents Panzer Dragoon Orta to stop booting and either hangs on a black screen or goes right to dashboard. Any build before this, it works fine.

RadWolfie:
@numchuk12 Thanks for your investigation. However, the actual source of the regression is from #1662 pull request. We already had verified by 2+ contributors.

PR #1648, opened 2019-07-02
Fix black screen in some titles when switching resolutions during gameplay

LukeUsher:
This contains a bug fix that can be seen on the MS dashboard (Clipped viewport values were never copied back to HostViewPort)

And one that was a previously unknown behavior (changing backbuffer width/height without a call to SetRenderTarget)
a LOG_TEST_CASE has been added for the new behavior, and a fix has been applied by destroying the existing host render target, and creating a new one.

Known PR Test Cases:
 - MS Dashboard 'Video Mode' settings menu (When switching between 'widescreen', 'letterbox', 'normal')
 - Chihiro Factory Test Program (requires chihiro-work branch)

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -3207,13 +3207,31 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
 		DWORD top = std::max((int)pViewport->Y, 0);
 		DWORD right = std::min((int)pViewport->X + (int)pViewport->Width, (int)XboxRenderTarget_Width);
 		DWORD bottom = std::min((int)pViewport->Y + (int)pViewport->Height, (int)XboxRenderTarget_Height);
+		DWORD width = right - left;
+		DWORD height = bottom - top;
 
 		XboxViewPort.X = left;
 		XboxViewPort.Y = top;
-		XboxViewPort.Width = right - left;
-		XboxViewPort.Height = bottom - top;
+		XboxViewPort.Width = width;
+		XboxViewPort.Height = height;
 		XboxViewPort.MinZ = pViewport->MinZ;
 		XboxViewPort.MaxZ = pViewport->MaxZ;
+
+		// This operation is often used to change the display resolution without calling SetRenderTarget!
+		// This works by updating the underlying Width & Height of the Xbox surface, without reallocating the data
+		// Because of this, we need to validate that the associated host resource still matches the dimensions of the Xbox Render Target
+		// If not, we must force them to be re-created
+		// TEST CASE: Chihiro Factory Test Program
+		DWORD HostRenderTarget_Width = 0, HostRenderTarget_Height = 0;
+		GetHostRenderTargetDimensions(&HostRenderTarget_Width, &HostRenderTarget_Height);
+		if (HostRenderTarget_Width != XboxRenderTarget_Width || HostRenderTarget_Height != XboxRenderTarget_Height) {
+			LOG_TEST_CASE("RenderTarget width/height changed without calling SetRenderTarget");
+			FreeHostResource(GetHostResourceKey(g_pXboxRenderTarget)); g_pD3DDevice->SetRenderTarget(0, GetHostSurface(g_pXboxRenderTarget, D3DUSAGE_RENDERTARGET));
+			FreeHostResource(GetHostResourceKey(g_pXboxDepthStencil)); g_pD3DDevice->SetDepthStencilSurface(GetHostSurface(g_pXboxRenderTarget, D3DUSAGE_DEPTHSTENCIL));
+		}
+						
+		// Store the updated viewport data ready to pass to host SetViewPort
+		HostViewPort = XboxViewPort;
 
 		if (g_ScaleViewport) {
 			// Get current host render target dimensions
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3207,14 +3207,32 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
...
+		XboxViewPort.Height = height;
 		XboxViewPort.MinZ = pViewport->MinZ;
 		XboxViewPort.MaxZ = pViewport->MaxZ;
 
+		// This operation is often used to change the display resolution without calling SetRenderTarget!
+		// This works by updating the underlying Width & Height of the Xbox surface, without reallocating the data
+		// Because of this, we need to validate that the associated host resource still matches the dimensions of the Xbox Render Target
+		// If not, we must force them to be re-created
+		// TEST CASE: Chihiro Factory Test Program
+		DWORD HostRenderTarget_Width = 0, HostRenderTarget_Height = 0;
```
LukeUsher: While this behavior may look like a hack, it's the best solution to this particular problem.

If we defer this change until the render target is next used (During _swap), it's too late as the 'broken' render target will have already been drawn too.

This code block only is entered when SetViewPort is called on the active render target, in which case, the render target is checked to make sure the Xbox and Host version of the resource still match.

This works because SetViewPort is *always* called as part of the resolution switch process, and *always* after the desired render target is already active.

So far the only known test-case for this is the Chihiro Factory Test program, hence why I added a LOG_TEST_CASE to catch any others.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3207,14 +3207,32 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
...
+		GetHostRenderTargetDimensions(&HostRenderTarget_Width, &HostRenderTarget_Height);
+		if (HostRenderTarget_Width != XboxRenderTarget_Width || HostRenderTarget_Height != XboxRenderTarget_Height) {
+			LOG_TEST_CASE("RenderTarget width/height changed without calling SetRenderTarget");
+			FreeHostResource(GetHostResourceKey(g_pXboxRenderTarget)); g_pD3DDevice->SetRenderTarget(0, GetHostSurface(g_pXboxRenderTarget, D3DUSAGE_RENDERTARGET));
+			FreeHostResource(GetHostResourceKey(g_pXboxDepthStencil)); g_pD3DDevice->SetDepthStencilSurface(GetHostSurface(g_pXboxRenderTarget, D3DUSAGE_DEPTHSTENCIL));
+		}
+						
+		// Store the updated viewport data ready to pass to host SetViewPort
+		HostViewPort = XboxViewPort;
```
LukeUsher: This line is what fixes the dashboard, and is a genuine bug fix. 
This has no impact on the Chihiro Factory Test program issue as described above.

RadWolfie:
List of titles has trigger the test case:
* Crimson Skies - in-game
* Constantine - menu
* Splinter Cell (Exhibition Demo Disc) - in-game
* Dead or Alive 3 (Exhibition Demo Disc) - menu (character selection)
* Dead or Alive Ultimate 2 - menu (character selection)

RadWolfie:
**UPDATE:** It appear titles with backbuffer hack with higher resolution is causing regression.

LukeUsher:
It's becoming increasingly important that we find the cause of the issues in the games that require the host back buffer hack, since that hack is holding back progress in many areas.

Until that hack is removed, we cannot implement proper higher resolution rendering, and there has to be two code paths in many places, for with/without the hack.

The host backbuffer hack is the most nasty/invasive hack we have right now. 

For now, we'll preserve existing backbuffer hack behavior by disabling this new code when the hack is in use and the backbuffer is the current render target.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3236,12 +3236,10 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
...
 		// Store the updated viewport data ready to pass to host SetViewPort
-		// Again, we can only do this when the host backbuffer hack is disabled, otherwise, we break
-		if (!g_DirectHostBackBufferAccess && g_pXboxRenderTarget != g_XboxBackBufferSurface) {
-			HostViewPort = XboxViewPort;
-		}
+		HostViewPort = XboxViewPort;
 
-		if (g_ScaleViewport) {
+		// We *must* scale the viewport when using the DirectHostBackBuffert hack for the backbuffer render target
+		if (g_ScaleViewport || (g_DirectHostBackBufferAccess && g_pXboxRenderTarget == g_XboxBackBufferSurface)) {
```
LukeUsher: Without this change, the dashboard only renders to part of the screen instead of the entire screen when changing resolutiosn while using the host backbuffer hack.

LukeUsher:
This PR is ready now provided the regressions have been solved: Further work will follow in the coming weeks towards fixing the problematic games (that require the host backbuffer hack) and by replacing the hack with a proper upscaling solution.

Until then, this PR is enough to prevent viewport-reloated black screen issues in most cases they occurred.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3207,15 +3207,40 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetViewport)
 		DWORD top = std::max((int)pViewport->Y, 0);
 		DWORD right = std::min((int)pViewport->X + (int)pViewport->Width, (int)XboxRenderTarget_Width);
 		DWORD bottom = std::min((int)pViewport->Y + (int)pViewport->Height, (int)XboxRenderTarget_Height);
+		DWORD width = right - left;
```
RadWolfie: @PatrickvL mention in the discord about
> like in how width is calculated: width = right-left is wrong, it needs a +1.

The same for below too.

PR #1649, opened 2019-07-03
DSound Mixbin's RW Support

RadWolfie:
Since Mixbin/voice properties isn't directly used by Cxbx-Reloaded, this pull request provide a read/write functionality to avoid any further breakage in xbox applications attempting to set a value when it doesn't preserve the value at all.

It is more of a placeholder to allow xbox applications progress further if any applications get stuck in a loop attempting to apply the volume or using custom mixbin.


src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
```diff
@@ -185,17 +185,148 @@ inline void GenerateXboxBufferCache(
...
+inline void GenerateMixBinDefault(
+    XTL::X_DSVOICEPROPS    &Xb_VoiceProperties,
+    LPCWAVEFORMATEX         lpwfxFormat,
+    XTL::X_DSMIXBINS*       mixbins_output,
+    bool                    is3D)
+{
+    // HACK: Below XDK 4039 does not use a pointer,
+    //       plus mixbins_output is a mask (DWORD).
+    extern uint32_t g_BuildVersion;
+    if (g_BuildVersion > 0 && g_BuildVersion < 4039) {
```
RadWolfie: Somehow, `g_BuildVersion` is 0 if xbe does not have D3D8 section. We need to fix this first before I can remove this part.

RadWolfie:
Test the titles from https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1647#issuecomment-506216510 would be more than enough to know they have not regress.

literalmente-game:
Tested all the games from that list, only Soldier Of Fortune II - Double Helix regressed (crash right after booting and rendering the first screen)

RadWolfie:
Turns out Soldier Of Fortune II can progress further if emulation is slower. The issue came from `IDirectSoundBuffer_SetVolume` with no pointer to the host's audio buffer class, and was attempted from a different thread. The fix will require a separate pull request to relocate all of enter/leave critical section into patched functions instead of hybrid functions.

src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
LukeUsher: That is to be expected: Historically this variable was always taken from d3d because that’s where it mattered most, and there are multiple versions of 39XX that use the same xapi and dsound version, but a different D3D version. 

Going forward I think either this variable should be renamed to g_D3DBuildVersion OR we could make a global struct containing the library version of each lib we care about.

RadWolfie: It will be better to have both. Use g_BuildVersion (in the cache) as highest known build version and global struct of each lib versions.

LukeUsher: Sounds good to me

RadWolfie: Resolved in a separate pull request.

RadWolfie:
Ready for merge, I have another branch depend on this pull request to be merge first.

PR #1650, opened 2019-07-04
Separate D3D8 Build Version Away from General Build Version

RadWolfie:
Doing this method will allow D3D8 library version not to select general build version.

Plus prepare DSOUND build version to be use for #1649 pull request. I had confirmed DSOUND's 4039 detection is functional.

### DIFF
--- a/src/core/hle/Intercept.cpp
+++ b/src/core/hle/Intercept.cpp
@@ -76,9 +76,6 @@ static const char* section_symbols = "Symbols";
 std::map<std::string, xbaddr> g_SymbolAddresses;
 bool g_SymbolCacheUsed = false;
 
-// D3D build version
-uint32_t g_BuildVersion = 0;
-
 bool bLLE_APU = false; // Set this to true for experimental APU (sound) LLE
 bool bLLE_GPU = false; // Set this to true for experimental GPU (graphics) LLE
 bool bLLE_USB = false; // Set this to true for experimental USB (input) LLE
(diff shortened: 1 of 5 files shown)


src/core/hle/Intercept.cpp
```diff
@@ -348,23 +345,46 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
...
 
 			if (xdkVersion < BuildVersion) {
 				xdkVersion = BuildVersion;
 			}
-			XbLibScan |= XbSymbolLibrayToFlag(std::string(pLibraryVersion[v].szName, pLibraryVersion[v].szName + 8).c_str());
+			XbLibFlag = XbSymbolLibrayToFlag(std::string(pLibraryVersion[v].szName, pLibraryVersion[v].szName + 8).c_str());
+			XbLibScan |= XbLibFlag;
+
+			// Keep certain library versions for plugin usage.
+			switch (XbLibFlag) {
```
RadWolfie: If any case future plugin need different library version, it can be place in here and use the settings' file for variable accessible.

src/core/hle/Intercept.cpp
```diff
@@ -474,17 +494,6 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
 
 		std::printf("Symbol: Detected Microsoft XDK application...\n");
 
-		// TODO: Is this enough for alias? We need to verify it.
-		if ((XbLibScan & XbSymbolLib_D3D8) > 0 || (XbLibScan & XbSymbolLib_D3D8LTCG) > 0) {
```
RadWolfie: No longer necessary

src/core/hle/Intercept.cpp
```diff
@@ -474,17 +494,6 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
...
-			g_BuildVersion = xdkVersion;
-		}
-#if 0 // NOTE: This is a note for what we should do for above.
```
RadWolfie: I had searched for any reference of 5558, 5659, and 5788 in Cxbx-Reloaded. The result is either none or not relative to D3D8.

I'm 99% sure this old code was actually used for XbSymbolDatabase's scanning process.

src/core/hle/Intercept.cpp
```diff
@@ -537,7 +546,7 @@ void EmuHLEIntercept(Xbe::Header *pXbeHeader)
 		symbolCacheData.SetLongValue(section_libs, LibraryName.c_str(), pLibraryVersion[i].wBuildVersion, nullptr, /*UseHex =*/false);
 	}
 
-	symbolCacheData.SetLongValue(section_libs, sect_libs_keys.BuildVersion, g_BuildVersion, nullptr, /*UseHex =*/false);
+	symbolCacheData.SetLongValue(section_libs, sect_libs_keys.BuildVersion, xdkVersion, nullptr, /*UseHex =*/false);
```
RadWolfie: Since we already have xdkVersion to use as highest build version. We'll simply use this value to store in the cache file.

src/core/hle/Intercept.cpp
RadWolfie: Correction, use the if statement. Not the case statement since the return result can have a combination. Such as D3D8 may will be sharing with D3D8LTCG library.

src/core/hle/Intercept.cpp
RadWolfie: DxbxRenderStateInfo return highest version is 4627 and lowest version is 3424.

PatrickvL:
LGTM

PR #1654, opened 2019-07-09
Fix DSound's Multi-thread Issue

RadWolfie:
The changes contain three things, relocate the lock to inside the patch function as the first thing it get used. And replaced Windows' critical section function to `std::mutex` method. Plus a minor fix of the warning message in compiler.

With these changes, Soldier of Fortune II no longer crash on boot often or the need to slow down the emulation process.

I have tested at least 6 titles to ensure regression had not occurred.


RadWolfie:
Also, I had performed a manual regex replacement than manual insertion work to be sure everything is placed where it's needed or not.

PR #1657, opened 2019-07-10
Show Disabled Selection in Network Configuration

### LINKED ISSUE #1656: CXBX Crashes if "Accept" is clicked in Network Configuration with default blank selection
The default selection for a network adapter is blank. If you click accept with the blank field the program will crash.

blank_network_select

Proposing PR shortly.

gellis713:
Close #1656 

Before:
blank_network_select

After:
default_network_select

### DIFF
--- a/src/gui/DlgNetworkConfig.cpp
+++ b/src/gui/DlgNetworkConfig.cpp
@@ -60,8 +60,12 @@ VOID FetchNetworkInterfaces(HWND hwnd)
 
 	pAdapterInfo = AdapterInfo;
 
+	LRESULT counter = SendMessage(g_hNetworkAdapters, CB_ADDSTRING, 0, (LPARAM)"Disabled");
+	SendMessage(g_hNetworkAdapters, CB_SETITEMDATA, counter, (LPARAM)"");
+	SendMessage(g_hNetworkAdapters, CB_SETCURSEL, counter, 0);
+
 	do {
-		LRESULT counter = SendMessage(g_hNetworkAdapters, CB_ADDSTRING, 0, (LPARAM)pAdapterInfo->Description);
+		counter = SendMessage(g_hNetworkAdapters, CB_ADDSTRING, 0, (LPARAM)pAdapterInfo->Description);
 		SendMessage(g_hNetworkAdapters, CB_SETITEMDATA, counter, (LPARAM)pAdapterInfo->AdapterName);
 
 		if (strcmp(g_XBNetwork.adapter_name, pAdapterInfo->AdapterName) == 0) {


LukeUsher:
A better fix would be to allow blank selection: The internal code already disables network emulation if a valid selection isn't made (and this is an intentional design choice)

So the front-end should allow this to be possible

src/gui/DlgNetworkConfig.cpp
```diff
@@ -109,6 +108,11 @@ INT_PTR CALLBACK DlgNetworkConfigProc(HWND hWndDlg, UINT uMsg, WPARAM wParam, LP
                 case IDC_AC_ACCEPT:
 					HWND hNetworkAdapter = GetDlgItem(hWndDlg, IDC_NETWORK_ADAPTER);
 					LRESULT lRet = SendMessage(hNetworkAdapter, CB_GETCURSEL, 0, 0);
+					if (lRet == -1)
+					{
+						EndDialog(hWndDlg, wParam);
+						break;
```
LukeUsher: Don't forget to set g_XBNetwork.adapter_name to an empty string too, otherwise the empty selection will not be saved. This just exits the dialog without saving anything..

RadWolfie:
@LukeUsher Once you made a selection, there is no option to select as "blank" after you accept the change.
image

LukeUsher:
> @LukeUsher Once you made a selection, there is no option to select as "blank" after accept the change.
> image

Oh really? Well then that should be addressed too

src/gui/DlgNetworkConfig.cpp
gellis713: Ok, I will take some more time to flesh out this as well as the inability to select blank after accepting an adapter.

BiatuAutMiahn:
Why not use 'Disabled' instead of blank?

gellis713:
> Why not use 'Disabled' instead of blank?

Good idea, added and tested.

gellis713:
> > @LukeUsher Once you made a selection, there is no option to select as "blank" after accept the change.
> > image
> 
> Oh really? Well then that should be addressed too

I still do not have the hang of git and it seems I have messed up the commit history while trying to keep my branch up to date. Apologies, and I will try not to have it happen again.

RadWolfie:
Hmm, I think maybe send a message to select disable first then allow while loop to find the adapter has been selected from the settings. At least will reduce the confusion from the users.

A preview from this pull request as of commit ed27697.
image

P.S. The regression has return if nothing is selected when click accept button.

src/gui/DlgNetworkConfig.cpp
```diff
@@ -60,6 +60,9 @@ VOID FetchNetworkInterfaces(HWND hwnd)
 
 	pAdapterInfo = AdapterInfo;
 
+	LRESULT counter = SendMessage(g_hNetworkAdapters, CB_ADDSTRING, 0, (LPARAM)"Disabled");
+	SendMessage(g_hNetworkAdapters, CB_SETITEMDATA, counter, (LPARAM)"");
```
RadWolfie: Will need to add `SendMessage(g_hNetworkAdapters, CB_SETCURSEL, counter, 0);` after this line in order to avoid a crash when nothing had been selected when click accept button.

src/gui/DlgNetworkConfig.cpp
```diff
@@ -60,6 +60,9 @@ VOID FetchNetworkInterfaces(HWND hwnd)
...
+
 	do {
 		LRESULT counter = SendMessage(g_hNetworkAdapters, CB_ADDSTRING, 0, (LPARAM)pAdapterInfo->Description);
```
RadWolfie: Also, while you're at it, remove `LRESULT ` here to avoid compiler's warning.

PR #1666, opened 2019-07-16
Fix a bug with D3DDevice_Reset that could trigger a black screen

LukeUsher:
Hopefully the code comments do a good job of documenting this:

Basically, some titles do not call D3DDevice_SetRenderTarget after/during Reset, meaning that these titles completely lost backbuffer access!

This PR addresses that by freeing the associated host resources, rather than setting the global pointers to nullptr.

This works because these pointers are static: even if the Xbox allocates a new backbuffer, the address of the default backbuffer header will *not* change, thanks to it being a class-member function of the pD3DDevice object, so actually, g_XboxBackBufferSurface and g_XboxDefaultDepthStencilSurface  remain valid after the free/re-allocation!

Test Case:
Chihiro Factory Test Program (You'll have to trust me on this, as it doesn't work in develop yet, only in Chihiro branches)
Likely many others

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2587,14 +2587,21 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_Reset)
 	// We should leave the rest of the params as is, and let the Xbox code/surface cache take care of it for us
 	pPresentationParameters->MultiSampleType = XTL::X_D3DMULTISAMPLE_NONE;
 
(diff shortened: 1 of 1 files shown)


LukeUsher:
> Oh, but then the comment must reflect this, instead of saying "we don't have to" (which makes it sound like an option, which it isn't)

I thought it too obvious to mention: of course overwriting a valuable loses the value stored it.  And when that value is the address of the Xbox backbuffer surface, Of course swap will start to fail. You can’t blit a nullptr.

PatrickvL:
I don't want to whine, and it might indeed be obvious for the experienced reader, but still the comment isn't accurate (and yeah, I know, most people, you and me included,  have a tendency to say/write statements less strict than what's required to make them unambiguous).

PR #1667, opened 2019-07-16
Remove 'Render to Host Backbuffer' hack

LukeUsher:
Removing this hack as it is negatively effecting Cxbx-R at this point: 

1. It's confusing for users, it was previously abused to get high-resolution rendering in a hacky method before we supported that functionality, but it is incompatible with actual high-resolution rendering support
2. It is harming progress on fixing bugs: Most games previously requiring the use of this hack have been fixed. The small number remaining are not being investigated, or even *reported* by users because they're happy with this hack. By removing it, hopefully they'll come forward so we can fix the remaining bugs
3. It involves supporting multiple code paths, which complicates things and makes debugging issues harder. Currently many code changes need testing twice, with/without this hack and it wastes time.

At the end of the day, it's a hack we added to fix a specific condition, which was a limitation of DirectX8 and no longer applies, it's outlived it's usefulness.


RadWolfie:
After checking for global usage:
* DirectHostBackBufferAccess (Settings' source and header files)
* g_DirectHostBackBufferAccess (Emu's source and header files)
* ID_HACKS_RENDERDIRECTLYTOHOSTBACKBUFFER (ResCxbx.h)

I am confident all references are removed and debug build did passed.

Margen67:
This hack made Reshade able to access the depth buffer for things like Ambient Occlusion, Depth of field, etc

PR #1668, opened 2019-07-16
Update memory functions to work with the loader

ergo720:
This updates the memory functions to work with both the loader and non-loader approaches. The non-loader code cannot be deleted yet because the loader project still needs to split the gui from the emu code, which can be done in a later PR. The changes I made to memory manager with this PR are the following:
1.	`Memory.bin` and `PageTable.bin` are no longer created, and to persist memory across quick reboots a memory mapping backed by the paging file is used instead.
2.	 Slightly simplified initialization by splitting `AllocateContiguousMemory` into two functions.
3.	Fixed a serious bug where some deallocation functions would start to free the ptes from EndingPte + 1 instead of StartingPte.
4.	Since memory reservation is now done by `VirtualAlloc`, `MapViewOfFileEx` cannot be used any longer to make allocations, so the corresponding code path has been deleted.


LukeUsher:
I haven’t had chance to look at the code yet: is this still opt-in or will this PR switch over? 

If it switches over, we cannot merge without extensive testing, as this is the largest single change we have done since the beginning.

ergo720:
The loader is disabled, this will still use the one exe approach like before (it can be enabled with the macro `CXBX_LOADER`). However, the changes applied to the memory manager will also apply to the non-loader approach.

.travis.yml
```diff
@@ -31,7 +31,7 @@ script:
 #  - mkdir export
 #  - 7z u export/${configuration} COPYING README.md
 #  - cd build/win32/${configuration}
-#  - 7z u ../../../export/${configuration} cxbxr.exe cxbxr-ldr.exe cxbxr-emu.dll glew32.dll subhook.dll CxbxVSBC.dll
+#  - 7z u ../../../export/${configuration} cxbx.exe glew32.dll subhook.dll CxbxVSBC.dll
```
RadWolfie: This file should be restored, not sure how it got reverted in this commit.

ergo720: I changed it because originally this was meant for the develop branch, and since the loader wasn't ready to use yet, I removed its artifacts from the script so that they would not be exported, which would otherwise confuse users. Since this PR has been re-targeted to the loader branch, this can be reverted now.

RadWolfie:
I can confirmed the emulation is working from GUI and loader, `cxbxr-ldr.exe`.

RadWolfie:
I notice there's `ipc_send_gui_update` call whenever there's reboot occur. Doing this method will not work with standalone emulation's reboots. May I ask why do this method?

Issue #1665, opened 2019-07-15
D3D Resources are erronously freed under certain conditions.


LukeUsher:
If D3DResource_Release is called such that it sets the reference count to zero, we free the associated host resource and reset any variables containing the referenced resource.

However, it has been discovered that it is incorrect to free the host resource at this point: If the internal reference count (not returned by D3DResource_Release) is > 0, then the freeing of the resource is delayed until the internal reference count reaches zero.

The correct point to do this free operation would be when the XDK calls DestroyResource(); At this point the Xbox resource is freed and the host resource should be too.

Fixing this issue depends on https://github.com/Cxbx-Reloaded/XbSymbolDatabase/issues/82, as we currently have no method to hook the DestroyResource function.

This is the cause for `LOG_TEST_CASE("SetRenderTarget fallback to backbuffer");`

And also potentially the cause for some of the remaining black screen issues, so I consider this quite important.

(Documenting the issue so it doesn't get forgotten, but I'll address this as soon as XbSymbolDatabase allows)

ObiKKa:
> This is the cause for `LOG_TEST_CASE("SetRenderTarget fallback to backbuffer");`

He's referring to the issue put up recently before this issue here: #1663

PatrickvL:
I believe Luke fixed this in merged pull request #1701

LukeUsher:
Yes this issue is resolved

PR #1669, opened 2019-07-16
Workaround & LOG_TEST_CASE for missing DestroyResource patch

LukeUsher:
This is a temporary workaround until https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1665 can be solved.

Added a LOG_TEST_CASE so we can track occurances so we know what to retest once we do finally patch DestroyResource (we're waiting on XbSymbolDatabase for that)

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -5576,45 +5576,37 @@ ULONG WINAPI XTL::EMUPATCH(D3DResource_Release)
 
 	// Was the Xbox resource freed?
 	if (uRet == 0) {
-		// We should free any variables storing the resource when freed
-        // HACK: For now, we skip the release when the resource has a non-zero internal ref count
-		// We also skip the release if we still have a cached variable containing this value
-        // This is a (small) memory leak, but prevents incorrectly freeing an in-use resource
-        // This is a *temporary* work around until solving https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1665
+	
+        // Generate some test cases so we know what to investigate/re-test after
+		// solving https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1665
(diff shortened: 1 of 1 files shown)


gandalfthewhite19890404:
Should we report titles here?

LukeUsher:
> Should we report titles here?
> Titles:
> WWE Raw 2 - LOG_TEST_CASE: Skipping release of active Xbox Texture
> In EmuPatch_D3DResource_Release (C:\projects\cxbx-reloaded\src\core\hle\D3D8\Direct3D9\Direct3D9.cpp line 5610)

Like the other test cases, this should probably get it’s own entry in the issue tracker, so they can be collected in one central place.

LukeUsher:
> Would it be possible to check the internal reference count too before skipping a release? That way we get less logging and we'll know for sure this issue needs a patch on DestroyResource

It is possible for the internal reference count to be zero and still have a left-over assignment in our code, so we'll need to check both cases regardless: But I can add an early-out for non-zero internal reference count.

This is because DestroyResource can be called *other* places than just the Release function: There's an internal release function that we don't patch, that also calls DestroyResource.

Hooking DestroyResource is a reliable way to track resource deletion.

Parovozik:
Hi. I see regression for Futurama (a406fb095529639919413a4cfb38fe934cf95281 build)
Critical error immediately after start 'New Game'
bandicam 2019-07-17 13-59-27-421

LukeUsher:
Ah, yeah, sorry this isn't going to work: It avoids black-screen in some titles, but in others it leaves dangling references because DestroyResource was called on the Xbox Side, but the host doesn't know.

I'll update this PR to purely be a LOG_TEST_CASE without skipping the free operation: This will allow us to collect test cases, with no side-effects. (But it also won't avoid the problem)

The following issues are blockers for a real fix:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1665
https://github.com/Cxbx-Reloaded/XbSymbolDatabase/issues/82

PatrickvL:
Non-functional change, merging as-is. Thanks, and looking forward to the real fix!

PR #1670, opened 2019-07-17
Improve some LTCG specific patches

LukeUsher:
Slightly improves behavior of Conker and Halo 2, they still don't do anything worth mentioning, however.
But they run for a little bit longer before crashing now, so it's progress,


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2456,30 +2456,43 @@ void Direct3D_CreateDevice_End()
 	// Set g_XboxD3DDevice to point to the Xbox D3D Device
 	auto it = g_SymbolAddresses.find("D3DDEVICE");
 	if (it != g_SymbolAddresses.end()) {
-		xbaddr dwD3DDevice = it->second;
-		if (dwD3DDevice != xbnull) {
-			g_XboxD3DDevice = *((DWORD**)dwD3DDevice);
-		}
+        g_XboxD3DDevice = (DWORD*)it->second;
```
LukeUsher: Turns out we were deriving this incorrectly before: This approach matches what I see on the Xbox side.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2456,30 +2456,43 @@ void Direct3D_CreateDevice_End()
...
-		}
+        g_XboxD3DDevice = (DWORD*)it->second;
     }
 }
 
 // LTCG specific Direct3D_CreateDevice function...
-// This uses a custom calling convention passed unknown parameters
-// Test-case: Ninja Gaiden
+// This uses a custom calling with parameters passed in eax, ecx and the stack
+// Test-case: Ninja Gaiden, Halo 2
```
LukeUsher: Fixing this prevents an issue where CreateDevice would corrupt the D3DDevice pointer on the Xbox side

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2515,6 +2528,35 @@ HRESULT WINAPI XTL::EMUPATCH(Direct3D_CreateDevice_16)
 	return hRet;
 }
 
+// ******************************************************************
+// * patch: D3DDevice_SetIndices_4
+// LTCG specific D3DDevice_SetIndices function...
+// This uses a custom calling convention where parameter is passed in EBX and Stack
+// Test Case: Conker
+// ******************************************************************
+VOID WINAPI XTL::EMUPATCH(D3DDevice_SetIndices_4)
```
LukeUsher: Fixes a crash in Conker caused by corrupted internal D3D index buffer state

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7436,6 +7478,27 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetPixelShader)
 	D3DDevice_SetPixelShaderCommon(Handle);
 }
 
+// ******************************************************************
+// * patch: D3DDevice_DrawVertices_4
+// LTCG specific D3DDevice_DrawVertices function...
+// This uses a custom calling convention where parameter is passed in ECX, EAX and Stack
+// Test Case: Conker
+// ******************************************************************
+VOID WINAPI XTL::EMUPATCH(D3DDevice_DrawVertices_4)
```
LukeUsher: Draw function that was detected, but not patched in Conker

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie: Hmm... I don't think the fix from targeted XbSymbolDatabase did occur to find `D3DDEVICE` symbol in LTCG applications.

LukeUsher: It’s not a problem. Nowhere uses this variable at the moment anyway. It was used in the past before we had trampolines working. 

I just noticed it was derived wrong by cxbxr and fixed it

RadWolfie: Hopefully soon we'll update the target XbSymbolDatabase to take advantage of existing symbols detecting properly.

PatrickvL:
LGTM, merging

PR #1671, opened 2019-07-17
HOTFIX: GenerateMixBinDefault Check for lpwfxFormat is null pointer

RadWolfie:
After discussion with other dev, I had discover a design flaw for not checking lpwfxFormat if has a null pointer or not.


LukeUsher:
Thanks,

PR #1672, opened 2019-07-18
Fix shader constants: It's per-function not per-XDK

LukeUsher:
Previously, Cxbx-R interpreted shaders from earlier XDKs as requiring +96 and later XDKS not, but this is incorrect.

After disassembling various games, it can be seen that instead, it goes per-function.

The standard D3DDevice_SetVertexShaderConstant function *always* takes constants ranging -96 to 96 on the Xbox, so we need to apply a +96 bias to convert to the host representation.

This also matches how the shaders are mapped to nv2a registers 0-196.

The functions that we implemented as forwards to D3DDevice_SetVertexShaderConstant take registers in the range 0-196. (already in NV2A register range). This is a perforamce optimization on the Xbox side: Later XDKs use a macro to adjust the range at compile time (using these functions) instead of run-time.

It's safe to implement these as forwards, but rather than implementing a per-XDK version check, we can simply subtract 96 to convert the range back into the expected range for D3DDevice_SetVertexShaderConstant.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -3705,10 +3705,10 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexShaderConstant)
 		LOG_FUNC_END;
 
 /*#ifdef _DEBUG_TRACK_VS_CONST
-    for (uint32 i = 0; i < ConstantCount; i++)
+    for (uint32_t i = 0; i < ConstantCount; i++)
     {
-        printf("SetVertexShaderConstant, c%d (c%d) = { %f, %f, %f, %f }\n",
-               Register - 96 + i, Register + i,
+        printf("SetVertexShaderConstant, c%d  = { %f, %f, %f, %f }\n",
+               Register + i,
                *((float*)pConstantData + 4 * i),
                *((float*)pConstantData + 4 * i + 1),
                *((float*)pConstantData + 4 * i + 2),
@@ -3716,12 +3716,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexShaderConstant)
     }
 #endif*/ // _DEBUG_TRACK_VS_CONST
 
-	// TODO: HACK: Since Xbox vertex shader constants range from -96 to 96, during conversion
-	// some shaders need to add 96 to use ranges 0 to 192.  This fixes 3911 - 4361 games and XDK
-	// samples, but breaks Turok.
-	// See D3DDevice_GetVertexShaderConstant
-	if(g_LibVersion_D3D8 <= 4361)
-		Register += 96;
+    // Xbox vertex shader constants range from -96 to 96
+    // The host does not support negative, so we adjust to 0-192
+	Register += 96;
 
     HRESULT hRet;
 	hRet = g_pD3DDevice->SetVertexShaderConstantF(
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -9218,11 +9223,6 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetPixelShaderConstant_4)
     //    LOG_FUNC_END;
     EmuLog(LOG_LEVEL::DEBUG, "D3DDevice_SetPixelShaderConstant_4(Register : %d pConstantData : %08X ConstantCount : %d);", Register, pConstantData, ConstantCount);
 
-    // TODO: This hack is necessary for Vertex Shaders on XDKs prior to 4361, but if this
-    // causes problems with pixel shaders, feel free to comment out the hack below.
-    if(g_LibVersion_D3D8 <= 4361)
```
LukeUsher: Pixel shader registers only range from 0-15, so this hack doesn't make sense here, it's possible it was breaking some functions in games with early XDKs

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3705,23 +3705,20 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetVertexShaderConstant)
 		LOG_FUNC_END;
 
 /*#ifdef _DEBUG_TRACK_VS_CONST
-    for (uint32 i = 0; i < ConstantCount; i++)
+    for (uint32_t i = 0; i < ConstantCount; i++)
     {
-        printf("SetVertexShaderConstant, c%d (c%d) = { %f, %f, %f, %f }\n",
-               Register - 96 + i, Register + i,
+        printf("SetVertexShaderConstant, c%d  = { %f, %f, %f, %f }\n",
```
LukeUsher: Even though this macro has been disabled, I updated it purely to log the input format: This is how I was able to diagnose this problem and fix it without relying on d3d build version

LukeUsher:
When testing this I have not noticed any improvements so far, but I haven't noticed any regressions either..
If user-testing goes well, it's worth merging just to remove the version-check hack, since the code here better matches actual XDK behavior.

LukeUsher:
> https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/fdaa102dc83470318f42f876a4f60f595eeb21ac/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp#L3683-L3684
> 
> `D3DDevice_SetVertexShaderConstant_8`'s edx, register parameter, did not subtract from -96. So, @LukeUsher You missed one.

Intentional, this patch looks nasty and I’m not sure it’s right at all. I just don’t have the game mentioned in the test case right now to take a look and fix it. 

But I really don’t trust the behaviour of that particular patch, so didn’t want to touch it.

PatrickvL:
It'd be good to have this PR tested some more, and await a few reports back from the field, before merging...

LukeUsher:
> It'd be good to have this PR tested some more, and await a few reports back from the field, before merging...

I expect zero impact: I tested more than 20 games across the entire range of xdks, including XDK samples from every version we have, and there was no change in behaviour. 

This just results in one less XDK version related hack.

RadWolfie:
Since I already tested known titles may cause regression base on the comments, including 12 other titles to tests as well. There is no regression found so far.

PR #1673, opened 2019-07-19
Add sleeps/waits in various places

NZJenkins:
Add some sleeps/waits here and there

* Fixes performance issue in Panzer Dragoon Orta (currently broken due to input hack though)
* Fixes performance issue in Prince of Persia 2 loading screen
* Lowers DirectSound worker thread CPU usage


LukeUsher:
Thanks! Since I came up with two of these fixes and tested them myself, I can confirm they work and haven’t caused any issues on any game I’ve tried. 

This PR should be entirely safe to merge as is now

PatrickvL:
Taking your word for it Luke, and thanks NZJenkins! Merged.

LukeUsher:
Very unexpected, might need to revert that single fix then, the rest can stay. I’ll look into why later

PR #1674, opened 2019-07-19
Fix a crash in JSRF

ergo720:
Fixes a crash in JSRF introduced by #1669. Below is the original error:

JSRF


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5621,7 +5624,7 @@ ULONG WINAPI XTL::EMUPATCH(D3DResource_Release)
 	
         // Generate some test cases so we know what to investigate/re-test after
 		// solving https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1665
-		if ((pThis->Common & X_D3DCOMMON_INTREFCOUNT_MASK) != 0) {
+		if ((Common & X_D3DCOMMON_INTREFCOUNT_MASK) != 0) {
```
PatrickvL: Note, that when internal reference count is store before calling Release, it's value is probably 1 higher than after Release was called

PR #1676, opened 2019-07-20
Fix Universal Error Message Output

RadWolfie:
While this pull request may not change anything, except it will fix the loader branch's UEM window issue when `cxbxr-emu.dll` try to output missing UEM message.


RadWolfie:
@PatrickvL, the reason it's not target to loader branch is to prepare for loader branch and better cross-platform. Even though the fix in the develop will fix the AppVeyor's bug.

<details><summary>Long standing bug (screenshot)</summary>

image
</details>

<details><summary>The PR fix (screenshot)</summary>

image
</details>

PR #1677, opened 2019-07-20
Preserve caller esp value

NZJenkins:
We patch instructions like `mov fs esp`
But `esp` is modified by calling the patch and pushing registers

This PR reverses `esp` changes before storing it in the emulated `fs`

Regressing titles:
PoP2 (black screen on boot)
Mechassault (black screen on boot)
Mercenaries (hang on boot)

### DIFF
--- a/src/core/kernel/support/EmuFS.cpp
+++ b/src/core/kernel/support/EmuFS.cpp
@@ -426,6 +426,7 @@ __declspec(naked) void EmuFS_MovFs00Esp()
 		push eax
 		mov eax, fs : [TIB_ArbitraryDataSlot]
 		mov [eax], esp
+		add [eax], 8 // account esp changes from pushed registers and return address
 		pop eax
 		call UnlockFS
 		ret


src/core/kernel/support/EmuFS.cpp
```diff
@@ -423,9 +423,13 @@ __declspec(naked) void EmuFS_MovFs00Esp()
 	{
 		call LockFS
 		call EmuFS_RefreshKPCR
-		push eax
+		push eax
+		push ebx
 		mov eax, fs : [TIB_ArbitraryDataSlot]
-		mov [eax], esp
```
RadWolfie: IMHO, the below working example is better and shorter than need to borrow another register.
```
mov [eax], esp
add [eax], 8
```

LukeUsher:
Any input as to why it breaks PoP2? Have you looked into the reason?

src/core/kernel/support/EmuFS.cpp
NZJenkins: done

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -676,7 +676,6 @@ static unsigned int WINAPI CxbxKrnlInterruptThread(PVOID param)
 		if (g_bEnableAllInterrupts) {
 			TriggerPendingConnectedInterrupts();
 		}
-		Sleep(1);
```
RadWolfie: Was this your intentional change in this pull request?

NZJenkins: No will remove it

PatrickvL:
So close this without merging?

PR #1678, opened 2019-07-20
Allow LoadVertexShaderProgram to work in games without their own CreateVertexShaderFunction

LukeUsher:
This is the best we can do without a huge overhaul of our vertex shader system: Some games don't use CreateVertexShader at all and load pre-compiled vertex shader programs. On the Xbox, this functionality doesn't even require a shader handle, but with Cxbx-Reloaded, we do.

This fixes a crash in all games that use LoadVertexShaderProgram but don't use CreateVertexShader.

One such example is Spy Vs Spy, that no longer crashes during early boot.


LukeUsher:
With this, Halo 2 gets further: Black screen but music can be heard.

RadWolfie:
Additional 2 titles can be improved:
* Ghost Recon 2 (before: crash on boot; after: goes in-game with broken graphic)
* Avatar - The Last Airbender (no change, however the current issue is not relative to this pull request)

RadWolfie:
Base on discussing with @LukeUsher, I am confidence enough to merge it.

PR #1680, opened 2019-07-21
Fix when CreateDevice doesn't call SetRenderTarget

LukeUsher:
This allows for the Xbox backbuffer location to be properly determined, even in titles that do not call
SetRenderTarget at all!

This also fixes the situation where games didn't call SetRenderTarget internally within CreateDevice.

This works because GetRenderTarget always returns the assigned render target, which immediately after
CreateDevice, is *always* the backbuffer.

This allows us to reliably determine the default backbuffer and depth stencil, even without a CreateDevice call.

This fixed a regresion in Shin Megami Tensei: Nine , and has potential to fix some more blackscreen issues.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2453,10 +2453,66 @@ void Direct3D_CreateDevice_Start
 
 void Direct3D_CreateDevice_End()
 {
-	// Set g_XboxD3DDevice to point to the Xbox D3D Device
-	auto it = g_SymbolAddresses.find("D3DDEVICE");
-	if (it != g_SymbolAddresses.end()) {
+    // Set g_XboxD3DDevice to point to the Xbox D3D Device
+    auto it = g_SymbolAddresses.find("D3DDEVICE");
+    if (it != g_SymbolAddresses.end()) {
         g_XboxD3DDevice = (DWORD*)it->second;
+    }
+
+    // If the Xbox version of CreateDevice didn't call SetRenderTarget, we must derive the default backbuffer ourselves
+    // This works because CreateDevice always sets the current render target to the Xbox Backbuffer
+    // In later XDKs, it does this inline rather than by calling D3DDevice_SetRenderTarget
+    // meaning our patch doesn't always get called in these cases.
+    // We fix the situation by calling the Xbox GetRenderTarget function, which immediately after CreateDevice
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2453,11 +2453,67 @@ void Direct3D_CreateDevice_Start
 
 void Direct3D_CreateDevice_End()
 {
-	// Set g_XboxD3DDevice to point to the Xbox D3D Device
-	auto it = g_SymbolAddresses.find("D3DDEVICE");
-	if (it != g_SymbolAddresses.end()) {
+    // Set g_XboxD3DDevice to point to the Xbox D3D Device
```
LukeUsher: tabs to spaces, for consistency with the rest of the function

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2453,11 +2453,67 @@ void Direct3D_CreateDevice_Start
...
+    // If the Xbox version of CreateDevice didn't call SetRenderTarget, we must derive the default backbuffer ourselves
+    // This works because CreateDevice always sets the current render target to the Xbox Backbuffer
+    // In later XDKs, it does this inline rather than by calling D3DDevice_SetRenderTarget
+    // meaning our patch doesn't always get called in these cases.
+    // We fix the situation by calling the Xbox GetRenderTarget function, which immediately after CreateDevice
+    // WILL always return the Backbuffer!
+    // Test Case: Shin Megami Tensei: Nine
+    if (g_XboxBackBufferSurface == xbnullptr || g_XboxDefaultDepthStencilSurface == xbnullptr) {
+        // First, log the test case
+        LOG_TEST_CASE("Xbox CreateDevice did not call SetRenderTarget");
```
LukeUsher: This is done in it's own if statement because we only want to log once, even if both code paths are hit.

RadWolfie:
Titles report an error about unable to determine default Xbox backbuffer:
* The Suffering: Ties That Bind
* Superman: The Man of Steel
* Tao Feng

Reported from discord tester.

LukeUsher:
> Titles report an error about unable to determine default Xbox backbuffer:
> 
> * The Suffering: Ties That Bind
> * Superman: The Man of Steel
> * Tao Feng
> 
> Reported from discord tester.

Thanks. This would have been an existing problem that was being silently ignored in the past. I’ll look into it

PR #1682, opened 2019-07-26
Improve LockFS performance

NZJenkins:
Follow up to #1673 
Only call `SwitchToThread` after we fail to obtain `fs_lock`

* Transworld Surf goes in game again
* Fixes DoAX and Outrun 2006 stutter and low fps during startup and in menus (sub 30fps to 60fps)

PoP2 still fine

### DIFF
--- a/src/core/kernel/support/EmuFS.cpp
+++ b/src/core/kernel/support/EmuFS.cpp
@@ -128,12 +128,12 @@ __declspec(naked) void LockFS()
 		// Backup Registers
 		pushfd
 		pushad
+		jmp entry
 
 		// Spin until we can aquire the lock
 		spinlock :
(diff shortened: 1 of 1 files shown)


RadWolfie:
Improved:
* Call of Duty: Finest Hour
  * Goes back in-game
  * Faster loading time (used to hang base on latest dev build)
* Ghost Recon: Island Thunder
  * Shorter boot wait time
  * Faster loading time

RadWolfie:
Since we do not need to call `SwitchToThread` every single time we make the first attempt to spinlock and could not find any regression so far. Looks good to me.

Plus calling `SwitchToThread` function first does prevent emulation to process correctly if running on xbox's single, or two, thread.

LukeUsher:
Great work!

PR #1687, opened 2019-08-02
Preserve caller esp value

NZJenkins:
Rehash of #1677 (Which did not preserve the flags registers correctly, causing hanging in some games)

We patch instructions like `mov fs esp`
But `esp` is modified by calling the patch and pushing registers

This PR reverses `esp` changes before storing it in the emulated `fs`

It also removes an extra `UnlockFS` call

Haven't observed any improvements or regressions


LukeUsher:
Thanks

PR #1688, opened 2019-08-02
Frame limiter toggle

NZJenkins:
F9 toggles the frame limiter off and on

This makes it much faster to pass intros and click through menus on some titles such as Outrun 2006

Also stop the GUI consuming keys (it has a few commands that only do anything when emulation is not running)


gandalfthewhite19890404:
What is about GUI indication of limit?

LukeUsher:
It’s fine as is, thanks

PR #1692, opened 2019-08-16
Remove LockFS from the functions that don't require locking

LukeUsher:
A slight stability improvement, fixing one known cause of a deadlock in some titles.
Thanks @NZJenkins for identifying the problem


PatrickvL:
Could you tell us a bit more about 'the known cause for deadlock', and how this commit avoids that precisely?

LukeUsher:
See discussion on Discord https://discordapp.com/channels/293096198808666122/426419647454248960/611438659131932694

Basically, some games were hitting a deadlock situation where two threads were both waiting on the same lock, so it never resumed.

The situaton occured in FS functions that use no global data, and as such, don't require locking to begin with.

PR #1694, opened 2019-08-21
.appveyor.yml fixups

Margen67:
* Don't trigger an AppVeyor build when .travis.yml is changed.
* Removed - after | since it's unneeded.
* Made %CONFIGURATION% lowercase to be consistent with actual env var.
* Made after_build into on_success so artifacts aren't tried on build fail, and made Release check apply to all commands.
* Made 7z archiving into one step since speed > cosmetics.
* Removed empty space(s), and line at end of file.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,9 @@
 # Documentation is here: https://www.appveyor.com/docs/appveyor-yml/
 
+skip_commits:
+  files:
+    - .travis.yml
+
 init:
   - ps: Update-AppveyorBuild -Version "$env:appveyor_repo_commit"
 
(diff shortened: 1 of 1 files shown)


LukeUsher:
Nice work

RadWolfie:
To be clear, I do not approved this pull request for the following reasons:
- We have the files zipped per project relative (such as Cxbx-Reloaded and Cxbx-Reloaded Debugger).
- literal style, `|`, does require a dash to remove unnecessary empty command. Two options are use `|-` or `>`.

Margen67:
@RadWolfie Should I make a PR to fix that?

RadWolfie:
You can, currently busy with other tasks. 👨‍🏭 It will be great to have the fix push in develop before rebase loader branch.

PR #1695, opened 2019-08-22
Set up CI with Azure Pipelines

Margen67:
This will only create artifacts for Release just like AppVeyor.

Reasons to add Azure:
* Faster than AppVeyor.
* Can build multiple configurations at the same time.

However, merging this PR alone isn't enough to add it. You must register for [Azure DevOps](https://azure.microsoft.com/en-in/services/devops) then set up a build pipeline using the .azure-pipelines.yml from this PR.
The build status badge may not work as-is so it may need to be fixed. (it's broken currently since no such Azure repo exists.)

Here's what my Azure build looks like:
https://margen67.visualstudio.com/Cxbx-Reloaded/_build/results?buildId=215&view=results

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,6 +2,7 @@
 
 skip_commits:
   files:
+    - .azure-pipelines.yml
     - .travis.yml
 
 init:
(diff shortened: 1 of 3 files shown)


RadWolfie:
> The build status badge may not work as-is so it may need to be fixed.

That is correct, if any "target" repos aren't using specific CI such as TravisCI, the badge will not work either. 😉

PR #1700, opened 2019-08-25
Update XbSymbolDatabase to latest version as of today

LukeUsher:
This brings in some additional symbols that are required for some upcoming work


PatrickvL:
Ignoring the build error, since that occasionally happens in visual studio 2019 as well, and requires avoiding any dependency on D3DX

PR #1707, opened 2019-08-25
Fix Azure artifact double-zipping

Margen67:


RadWolfie:
Since the solution for remove "Release" folder inside the zip folder is not found or is not possible base on Azure's artifact procedure design. Looks good to me!

Issue #666, opened 2017-08-14
Galleon - Shows loading screen, slows down a lot, then crashes


BSoD38:
### Description

Using Aug 13 2017 build. (bf5719e1)

The game now shows the initial loading screen of the game, and the 3D model of the main character is displayed correctly, but the waves in the backdrop have glitched colours. The initial title screen seems to have missing/glitchy graphics. The progress bar wil more often than not have graphics glitches.
The loading progress bar will fill up at about a little less than a quarter, then the game will slow down to a crawl (about 3 FPS), then end up crashing, with an error message. Terminating the emulation at this moment will sometimes make Cxbx crash.

Also, I think the game has no sound.

---

### Game info
[XBE Dump, KrnlDebug, HLE Cache, Ect]
XBE Dump : [https://hostr.co/pfxg1K4ipzsY](url)
HLE Cache : [https://hostr.co/ovjgStEYQMMR](url)
KrnlDebug : [https://hostr.co/q16PA4JJLk23](url)

---

### Status
**Previous behavior:**
In builds older than one or two months ago, the game would crash instantly, without displaying anything. Then, builds would start showing the initial game logos, then crash before showing anything else.

**Current behavior:**
Shows the initial title and loading screens, but crashes while loading.

---

### Screenshots [Errors, Graphics, Glitches, ECT]
![Title Screen 2](http://i.imgur.com/rlM9vcC.png)
![Loading screen](http://i.imgur.com/AnwHI0q.png)
![Loading screen glitch](http://i.imgur.com/6qHR4sk.png)
![Error message](http://i.imgur.com/6A3K6ZH.png)
---

### Additional Information

Default graphics settings, LLE disabled, using automated Release build.

PatrickvL:
This issue was moved to Cxbx-Reloaded/game-compatibility#1

PR #1708, opened 2019-08-28
Fix Xapi controller capabilities structure

aav7fl:
There's a bug with the Xapi controller capabilities structure which places it in the wrong memory address. This fixes it. 

Resolves controller input issues in: https://github.com/Cxbx-Reloaded/game-compatibility/issues/666


LukeUsher:
Excellent work, great find

x1nixmzeng:
The same fix was needed to fix controller vibration, It may be worth sanity checking other xinput structs for packing

ObiKKa:
Ooh, #666, the Mark for the Beast; and also rather similar to the Order 66 sub plot in Star Wars Episode 3 movie. "Ironic".

gamecare:
Has anyone found out the issue with the missing/invisible TEXT ?
So far its running perfect (apart from the missing text that is)

PR #1709, opened 2019-08-29
Lots of cleanup regarding vertex declarations and shaders

PatrickvL:
Mostly, this helps to make this code a little bit easier to maintain.
Along the way a few small bugs have been fixed as well.

### DIFF
--- a/src/CxbxVersion.h
+++ b/src/CxbxVersion.h
@@ -2,10 +2,6 @@
 
 #include "version.h"
 
-#ifndef _GIT_VERSION
-	#define _GIT_VERSION
-#endif
-
 /*! version string dependent on trace flag */
 #ifndef _DEBUG_TRACE
 #define _CXBX_VERSION _GIT_VERSION " (" __DATE__  ")"
--- a/src/CxbxVersion.h
+++ b/src/CxbxVersion.h
@@ -2,6 +2,10 @@
 
 #include "version.h"
 
+#ifndef _GIT_VERSION
+	#define _GIT_VERSION
+#endif
+
 /*! version string dependent on trace flag */
 #ifndef _DEBUG_TRACE
 #define _CXBX_VERSION _GIT_VERSION " (" __DATE__  ")"
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -164,7 +164,7 @@ static XTL::X_D3DSurface           *g_XboxDefaultDepthStencilSurface = NULL;
 XTL::X_D3DSurface                  *g_pXboxRenderTarget = NULL;
 static XTL::X_D3DSurface           *g_pXboxDepthStencil = NULL;
 static DWORD                        g_dwVertexShaderUsage = 0;
-static DWORD                        g_VertexShaderSlots[136];
+static DWORD                        g_VertexShaderSlots[VSH_XBOX_MAX_INSTRUCTION_COUNT];
 
 DWORD g_XboxBaseVertexIndex = 0;
 DWORD g_DefaultPresentationInterval = D3DPRESENT_INTERVAL_IMMEDIATE;
@@ -3065,25 +3065,8 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_LoadVertexShader_4)
 	DWORD           Handle;
 	__asm mov Handle, eax;
 
-	//LOG_FUNC_BEGIN
-	//	LOG_FUNC_ARG(Handle)
-	//	LOG_FUNC_ARG(Address)
-	//	LOG_FUNC_END;
-	EmuLog(LOG_LEVEL::DEBUG, "D3DDevice_LoadVertexShader_4(Handle : %d Address : %08x);", Handle, Address);
-
-	// Handle is always address of an Xbox VertexShader struct, or-ed with 1 (X_D3DFVF_RESERVED0)
-	// An Xbox VertexShader contains : a 'Vshd' signature, flags, a size, a program (and constants)
-	// Address is the slot (offset) from which the program must be written onwards (as whole DWORDS)
-	// D3DDevice_LoadVertexShader pushes the program contained in the Xbox VertexShader struct to the NV2A
-    if(Address < 136 && VshHandleIsVertexShader(Handle))
-    {
-		CxbxVertexShader *pCxbxVertexShader = GetCxbxVertexShader(Handle);
-        for (DWORD i = Address; i < pCxbxVertexShader->XboxNrAddressSlots; i++)
-        {
-            // TODO: This seems very fishy
-            g_VertexShaderSlots[i] = Handle;
-        }
-	}
+	LOG_FORWARD("D3DDevice_LoadVertexShader");
+	return EMUPATCH(D3DDevice_LoadVertexShader)(Handle, Address);
 }
 
 // ******************************************************************
@@ -3098,10 +3081,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_LoadVertexShader)
 	LOG_FUNC_BEGIN
 		LOG_FUNC_ARG(Handle)
 		LOG_FUNC_ARG(Address)
-		LOG_FUNC_END;
+	LOG_FUNC_END;
 
-	// Handle is always address of an Xbox VertexShader struct, or-ed with 1 (X_D3DFVF_RESERVED0)
-	// An Xbox VertexShader contains : a 'Vshd' signature, flags, a size, a program (and constants)
+	// Handle is always address of an X_D3DVertexShader struct, thus always or-ed with 1 (X_D3DFVF_RESERVED0)
 	// Address is the slot (offset) from which the program must be written onwards (as whole DWORDS)
 	// D3DDevice_LoadVertexShader pushes the program contained in the Xbox VertexShader struct to the NV2A
     if(Address < 136) {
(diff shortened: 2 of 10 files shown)


RadWolfie:
I had not look in the vertex relative changes yet. One thing do caught my attention, `
Avoid compiles failing on a missing _GIT_VERSION symbol` commit doesn't make sense to me. If CMake didn't generate _GIT_VERSION inside the `version.h` header file, then the whole project wouldn't be able to compile since linkage is already there. See relative sources below: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/6a29876b5eb4348e31e49515d4b5d07092070ef1/src/CxbxVersion.h#L3
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/a1731b6bfc42b4bc7a035414545d6d621158a60f/CMakeLists.txt#L52-L55
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/a1731b6bfc42b4bc7a035414545d6d621158a60f/CMakeLists.txt#L100

So if version.h header file was manually created, that would mean CMake steps wasn't taken. What is the current problem with CMake? Other than XbSymbolDatabase's [latest fix, PR 93,](https://github.com/Cxbx-Reloaded/XbSymbolDatabase/pull/93) for CMake had been upstream with develop branch few days ago.

RadWolfie:
@PatrickvL If you don't mind, list the bugs you had discover?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3171,38 +3171,51 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SelectVertexShader)
 	// Either way, the given address slot is selected as the start of the current vertex shader program
 	g_CurrentXboxVertexShaderHandle = Handle;
 
+    CxbxVertexShader *pCxbxVertexShader = nullptr;
+	DWORD HostFVF = 0;
```
LukeUsher: Indenting seems off here/in other places

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4750,17 +4754,17 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_Clear)
...
-// * patch: D3DDevice_CopyRects
-// ******************************************************************
-VOID WINAPI XTL::EMUPATCH(D3DDevice_CopyRects)
-(
-    X_D3DSurface* pSourceSurface,
-    CONST RECT* pSourceRectsArray,
-    UINT                cRects,
-    X_D3DSurface* pDestinationSurface,
-    CONST POINT* pDestPointsArray
+
```
LukeUsher: Why did this change?

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1771,630 +1753,601 @@ static boolean VshConvertShader(VSH_XBOX_SHADER *pShader,
...
+XTL::D3DVERTEXELEMENT *XTL::EmuRecompileVshDeclaration
+(
+    DWORD                *pXboxDeclaration,
+    bool                  bIsFixedFunction,
+    DWORD                *pXboxDeclarationCount,
+	DWORD                *pHostDeclarationSize,
+    CxbxVertexShaderInfo *pCxbxVertexShaderInfo
+)
+{
+	class XboxVertexDeclarationConverter Converter;
```
LukeUsher: Class keyword shouldn't be required here?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -8693,70 +8702,68 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_LoadVertexShaderProgram)
...
+		pXboxVertexDeclaration = pCxbxVertexShader->pXboxDeclarationCopy;
     } else {
-        // This is an unusual scenario in which an FVF shader is being replaced with an actual shader
+        // This is an unusual scenario in which an FVF-based shader is being replaced with an actual shader
         // But without calling CreateVertexShader: This means we need to parse the current FVF and generate
-        // a declaration to use when converting/setting this new shader
+        // our own Xbox-like declaration to use when converting/setting this new shader
 
-        // Allocate a large enough definition to contain all possible FVF types
+        // Define a large enough definition to contain all possible FVF types
```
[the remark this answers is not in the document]
PatrickvL: Using stack instead of heap allocation is faster, reduces fragmentation and avoids potential leaks.

Since the below code will never access the array above index 15, the reserved 20 indices are already more than required. (There's at most 4 texture coordinates, hence the assert for that, since the bitfield could contain a highest value of 15, which would be wrong)

src/CxbxVersion.h
```diff
@@ -1,9 +1,13 @@
 #pragma once
 
 #include "version.h"
+
+#ifndef _GIT_VERSION
```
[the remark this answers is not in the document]
PatrickvL: If this goes wrong on my end, then others could hit this as well.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Yeah, we definitely should do something about that! (I did take care to keep the EOL style intact, but indenting is another matter)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: SmartGit told me the prevalent encoding of this file was Unix-style.
It also told me this file used a mixed EOL style, which is probably not what we want either.
So, I decided to make the EOL style consistent again.

I guess this part used the Windows style, which then got changed when I used Notepad++ to set the EOL style of the entire file to Unix.

src/core/hle/D3D8/XbVertexShader.cpp
PatrickvL: Good to know!

src/CxbxVersion.h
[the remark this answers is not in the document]
RadWolfie: See more details at https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1709#issuecomment-526435166.

@LukeUsher's comment is only half truth. If you had seen my comment and read the relative lines I had linked. Then you should know if git isn't installed, then CMake will generate "unknown" string for generated `_GIT_VERSION` define within `version.h` file.

So, your change is only apply if you never ran CMake at all. If you had never ran CMake, then the compiler can't build unless `version.h` file is created.

As for git requirement is for the submodules. Yet it's possible someone "could" use a local git executable to download the submodule repos. Then the chances is CMake doesn't know about the local git executable's location. CMake will still output "unknown" string in version.h file anyway.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -8568,9 +8550,9 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_GetVertexShaderConstant)
 		LOG_FUNC_ARG(ConstantCount)
 		LOG_FUNC_END;
 
-	// Xbox vertex shader constants range from -96 to 96
-    // The host does not support negative, so we adjust to 0-192
-	Register += 96;
+	// Xbox vertex shader constants range from -96 to 95
+    // The host does not support negative, so we adjust to 0..1912
```
RadWolfie: Found a typo mistake, `1912` should be `191`. 😉

src/CxbxVersion.h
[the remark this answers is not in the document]
RadWolfie: Also, instead of doing this hack. We should create an issue/discussion about the issues there are then implement the proper fix than making the hacks. Even if there's only one person having the issue.

LukeUsher:
Finally got around to testing this, I didn't notice any issues in the games I tested so I am happy to merge once the points mentioned by myself and @RadWolfie are addressed.

PatrickvL:
Yeah, I should just revert the parts that had some remarks

src/CxbxVersion.h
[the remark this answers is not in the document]
PatrickvL: I've reverted this commit for now.

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1108,7 +1119,8 @@ X_VERTEXSHADERINPUT;
 // ******************************************************************
 typedef struct _X_VERTEXATTRIBUTEFORMAT
 {
-    X_VERTEXSHADERINPUT pVertexShaderInput[16];
+	// TODO : Does this need #pragma pack(1) / #include "AlignPrefix1.h" (and it's closure)?
```
RadWolfie: X_VERTEXSHADERINPUT structure has size of 14 (0x0E) which will have extra 2 bytes at the end to align 4 bytes properly. So, yes it is a requirement since the structure is xbox data. Without align by 1 byte fix, the arrays will be not be the same as xbox's data input/output.

PatrickvL:
Ready for merge. Unless you want me to cherry pick some of improvements I'm doing ok my other Vsh branch?

src/core/hle/D3D8/XbD3D8Types.h
RadWolfie: What about this todo note? From what @PatrickvL told me the structure is align correctly. Since it's align correctly, this todo note should be removed.

PatrickvL: Done.

PR #1558, opened 2019-03-04
Enable LLE USB + new input gui

ergo720:
This does the following:
1.	Unpatch all xinput and MU functions
2.	Permanently enables LLE USB, which becomes the only input mode available
3.	Disables all HLE xinput code
4.	Replace the old input gui with a new one (supports DInput, XInput and SDL input + vibration)

However, there a couple of things I couldn’t test myself:
1.	My SDL controller doesn’t support vibration, so that’s untested. Based on the existing code, I think it will only vibrate once, and none afterwards
2.	I tested JSRF and input works fine, but I don’t have the other two problematic games (lego star wars and gunvalkyre), so someone should test those for me and see if input works

Testing is welcome.

PR #1713, opened 2019-09-01
New input gui

ergo720:
This implements a new input GUI which completely replaces the old one. It uses Sdl input to handle all non XInput controllers (which are instead handled by XInput) and the keyboard is handled by DInput. The appropriate API is automatically selected so you don’t have to select it now like you had to with the old GUI. Code to support the mouse is present but disabled as it has an unresolved issue which leads to weird movements in games and needs more investigation. Rumble functionality has been implemented so your controllers can also vibrate now (tested with XInput, Sdl needs testing). I also made some changes to the Xbox XInput functions so you will want to retest many games to see if I haven’t accidentally broken something. Please note that, according to here https://github.com/JayFoxRox/xqemu/pull/4#issue-199335814 , not all games accept input from ports other than 1. So, if a game doesn’t accept input from a port, try to connect it to port 1 and if it works, than it’s not a regression/bug.

**Edit**: Ready for merge


src/common/input/DInputKeyboardMouse.cpp
```diff
@@ -0,0 +1,262 @@
...
+// *
+// *  (c) 2019 ergo720
+// *
+// *  All rights reserved
+// *
+// ******************************************************************
+
+// Copyright 2010 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
```
RadWolfie: If you don't mind my asking... what `license.txt` file? The license file we have at the moment is `COPYING`. Since we're still using the same license perhaps change the sentence to
```suggestion
// Refer to top of this file's copyright license.
```

or to the `COPYING` file.

Though, I think we should rename `COPYING` to `LICENSE`.

ergo720: That file is of Dolphin, not Cxbx-R. I just copied their copyright header.

src/common/input/Button.cpp
```diff
@@ -0,0 +1,74 @@
...
+// *
+// *  This program is distributed in the hope that it will be useful,
+// *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+// *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// *  GNU General Public License for more details.
+// *
+// *  You should have recieved a copy of the GNU General Public License
+// *  along with this program; see the file COPYING.
+// *  If not, write to the Free Software Foundation, Inc.,
+// *  59 Temple Place - Suite 330, Bostom, MA 02111-1307, USA.
```
RadWolfie: Boston* 😉

We can do this fix later on.

src/common/input/DInputKeyboardMouse.cpp
RadWolfie: If you want to refer to their own file, then you should use a direct link. Since `license.txt` file isn't included in Cxbx-R nor is in the pull request.

Does it make sense?

RadWolfie:
Did we discussed about missing save/cancel buttons before? Ah found the mention about missing buttons. Well, it's also missing in this pull request.
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1558#issuecomment-472262510

If we're going to merge this without the missing buttons, then we'll need to open a new issue for it. Since I had not heard of general UI dialog always save the change.

.gitmodules
```diff
@@ -18,3 +18,7 @@
 	path = import/libtomcrypt
 	url = https://github.com/libtom/libtomcrypt
 	branch = master
+[submodule "import/SDL2"]
+	path = import/SDL2
+	url = https://github.com/spurious/SDL-mirror
```
RadWolfie: I wouldn't advice to use this link. Use https://github.com/SDL-mirror/SDL instead.
Sources:
- https://github.com/XboxDev/nxdk-sdl
- https://github.com/XboxDev/nxdk/blob/686d5140617262f225f13b82e27da8b4ff576c92/.gitmodules#L17

Using https://github.com/SDL-mirror is better since it also mirror the other repos as well.

.gitmodules
```diff
@@ -18,3 +18,7 @@
...
+	branch = release-2.0.10
```
RadWolfie: Why specific branch? We should be able to use the latest update from master branch.

ergo720: I believe it's generally better to use the latest stable release of a repository instead of the latest update, since the release has a less chance of bugs and/or regressions.

ergo720:
>Did we discussed about missing save/cancel buttons before? Ah found the mention about missing buttons. Well, it's also missing in this pull request.
#1558 (comment)
If we're going to merge this without the missing, apply/cancel, buttons, then we'll need to open a new issue for it. Since I had not heard of general UI dialog always save the change.

1. I could, but I don't see the utility of adding a popup to the Clear button? I feel it would instead be an hindrance if every time the button is pressed, the popup shows up.
2. Ok, but with one change: the user must have provided a name for the profile. This is because empty profile names are not allowed and are used during profile searches.
3. `Save/Apply/Cancel` buttons weren't asked in that comment. Anyway, there's already a `Save` button in the GUI which does what you want, namely, saving your configuration, so adding another one is just confusing. `Apply` also feels unnecessary since the button is supposed to save your configuration without closing the window, which is what `Save` already does. Not sure about the `Cancel` button at the moment.

.gitmodules
LukeUsher: Agreed, for some libraries (such as XbSymbolDatabase or even Subhook) it's beneficial to have bleeding-edge code, sometimes we *need* it.

But for something like SDL, using the stable branch is fine.

src/common/Settings.cpp
```diff
@@ -48,8 +50,15 @@ std::string g_exec_filepath;
 uint16_t g_LibVersion_D3D8 = 0;
 uint16_t g_LibVersion_DSOUND = 0;
 
-// NOTE: Update settings_version when add/edit/delete setting's structure.
-const unsigned int settings_version = 4;
+// NOTE: Update settings_version when add/edit/delete setting's structure.
+///////////////////////////
+// * History:
```
LukeUsher: Interesting addition, but is it necessary? Version control history should provide enough information if you do a 'blame' on the settings_version line?

src/common/input/Button.cpp
LukeUsher: Didn't we already fix the Bostom=>Bostom globally? If so, then it should be fixed here rather than in a later PR.

src/common/util/CxbxUtil.cpp
```diff
@@ -255,3 +255,21 @@ void unix2dos(std::string& string)
 		position += 2;
 	}
 }
+
+// Copyright 2008 Dolphin Emulator Project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
```
LukeUsher: We don't have license.txt, see earlier comment by RadWolfie

jeltaqq:
Tested with Sdl
Working
Gunvalkyrie
Quantum Redshift
Not working (regression)
Tenerezza

I'm fond of the stick and D-Pad settings "Up down left right" like old GUI. In addition, the same order as the trigger can be set intuitively.

.gitmodules
RadWolfie: Then we can choose specific tag inside the submodule's repo to specify the stable version's commit which is `c7b8b49`.

src/common/input/Button.cpp
RadWolfie: Nope, we didn't. Source:
- https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/develop/src/gui/DlgAbout.h#L18
- https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/search?q=bostom&unscoped_q=bostom

.gitmodules
RadWolfie: So, the change here is not necessary except for the url field.

RadWolfie:
> 1. I could, but I don't see the utility of adding a popup to the Clear button? I feel it would instead be an hindrance if every time the button is pressed, the popup shows up.

Supposedly you have keyboard input in progress to setup then you had accidentally clear it. You'll have to manually re-input 25 fields. You can check if any inputs had occur then show the popup.

> 2. Ok, but with one change: the user must have provided a name for the profile. This is because empty profile names are not allowed and are used during profile searches.

👍 Yeah, users will be confuse why there's a profile name field. Then assume it's not necessary after setting up their inputs.

> 3. `Save/Apply/Cancel` buttons weren't asked in that comment. Anyway, there's already a `Save` button in the GUI which does what you want, namely, saving your configuration, so adding another one is just confusing. `Apply` also feels unnecessary since the button is supposed to save your configuration without closing the window, which is what `Save` already does. Not sure about the `Cancel` button at the moment.

After 2nd review, looks like the native close button just close it. Perhaps add a close button next to clear button?

RadWolfie:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1713#pullrequestreview-282515069

To follow up @LukeUsher's comment. Once saved then close/re-open, the profile name is blank which should be showing the profile name.

src/common/Settings.cpp
ergo720: No, I just thought it would be nice to have, but there's no compelling reason to keep it if you don't want.

RadWolfie: I think it's fine to have a text version of history revisions.

gandalfthewhite19890404:
1. What is "XInput Default" labeled button?
2. Settings are not saving until you create profile or choose (if you need reconfigure it later)
3. Modifier - what is it, thumb (analogue stick button)?

LukeUsher:
I think it's still unclear that the user must type a profile name, and cannot simply configure-and-go.
Other than that, this is good and I think it can be merged after this is solved, perhaps by setting a default name if no profiles or present, or even just by adding some instruction text to the dialog.

RadWolfie:
@LukeUsher There is current review change I had requested. Source: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1713#discussion_r319782952

Beside that, I had not re-test the recent change at the moment. I will test it out very soon.

RadWolfie:
> 1. What is "XInput Default" labeled button?

Hmm, supposedly in the future. We may will have default setup for keyboard/mouse. So, it's better to use "Default" button and disable the button if using keyboard/mouse is selected.

> 2. Settings are not saving until you create profile or choose (if you need reconfigure it later)

Cause by inexperience users or forgot to save new change(s). We'll have to expect multiple false positive report for this. Plus complaints too.

ergo720:
I know that there are still issues and the work is still incomplete. I'm just pushing commits to allow early testing and discover eventual bugs sooner rather than later.

RadWolfie:
As of commit a001e46, the GUI look complete to me even the recent GUI code changes too.

src/common/input/InputWindow.cpp
```diff
@@ -0,0 +1,460 @@
...
+		std::thread([this, dev, ms]() {
+			EnableWindow(m_hwnd_rumble, FALSE);
+			SendMessage(m_hwnd_rumble_list, WM_SETTEXT, 0, reinterpret_cast<LPARAM>("..."));
+			auto outputs = dev->GetOutputs();
+			for (const auto out : outputs) {
+				if (out->GetName() == m_rumble) {
+					out->SetState(1.0, 1.0);
+				}
+			}
+			std::this_thread::sleep_for(std::chrono::milliseconds(ms));
```
RadWolfie: Seems to be missing set the rumble state back to 0 after this sleep function. Otherwise the rumble will continue to vibrate until the whole input UI dialogs are close.

ergo720:
I suppose that happens with SDL controllers? Because with my XInput controller it rumbles for a fixed amount of time and stops normally, independently of the input GUI. I don't have SDL controllers which support vibration so I can't see it myself.

src/common/input/InputWindow.cpp
RadWolfie: > I suppose that happens with SDL controllers? Because with my XInput controller it rumbles for a fixed amount of time and stops normally, independently of the input GUI. I don't have SDL controllers which support vibration so I can't see it myself.

Depending on hardware/driver/SDK design is my guess. Although, the games are responsible to stop the rumble at specific time. It is possible for long time term (10+ seconds) continue vibrate under certain environment. Like the ground is shifting which sometimes doesn't need to adjust the value of vibration until the end of shift event.

With commit edc503e, the rumble after 3 seconds has stop vibrate with my wireless 360 controller. 👍

gandalfthewhite19890404:
"XInput Default" labeled button is present even with dinput/keyboard

ergo720:
I don't see a problem with that since it will do nothing for non-XInput devices and it's labeled XInput Default to explicitly communicate it's for XInput deivces.

PR #1715, opened 2019-09-04
RenderTarget and CopyRect fixes

LukeUsher:
This PR fixes the following issues:

1. Fix a stupid bug that preventing the CopyRects patch from ever working to begin with
2. Fix an issue where attempting to copy a backbuffer to a texture would fail, due to mismatching pool types (StrechRects only works with D3DPOOL_DEFAULT)
3. Fix an issue where converting a RenderTarget to a Texture would not assign the child surface

Two Notable Improvements:

**Crash Bandicoot: The Wrath of Cortex (Fixed Pause screenshot)**
_Before:_
Pause screen has corrupted graphics
image

_After:_
Pause screen renders as expected
image

**Jet Set Radio Future (Fixed Graffiti Editor)**

_Before:_ 
Graffiti Editor partially works, some sizes don't update the view
image

_After:_
Graffiti Editor works as expected
image

PR #1716, opened 2019-09-05
Fix CopyRects when upscaling is used

LukeUsher:
This is a hotfix to #1715 in which CopyRects would still fail if upscaling was being used.


RadWolfie:
Which title made a regression?

LukeUsher:
> Which title made a regression?

Everything that the PR was supposed to fix. 
It worked without upscaling, but the PR was useless when upscaling was used, it would've been caught if someone tested it after my rebase, but nobody did :P

PatrickvL:
Ah well, good that you did ;)

Judging by the change, the bug was already present before your prior fix, but it wasn't hit yet

PR #1717, opened 2019-09-06
Create GitHub Actions for some CI tasks

GXTX:
These commits will add 2 different actions to CXBX-R. One will run only on pull requests and will verify that the build passes with both `Release` & `Debug` configurations, the second will run on a push event to `develop` and publish an artifact for it.

Currently GitHub has little to no information published about artifacts but I'm assuming they'll be similar to GitHub Releases and will be indefinite, there's also seems to be no way to directly link to 'latest build' so users will have to manually find them (Actions > Release Artifacts > develop > Artifacts).

The end goal here is to remove all 3rd party requirements such as AppVeyor, Travis-CI, and Azure.

Because this feature is still in "invite only" beta I'm leaving this as a draft PR, however those with access can fork and verify functionality.

Suggestions welcomed.

### DIFF
--- a/.github/workflows/master_artifact_publish.yml
+++ b/.github/workflows/master_artifact_publish.yml
@@ -1,0 +1,34 @@
+name: Release Artifacts
+
+on:
+  push:
+    branches: develop
+
+jobs:
+  windows:
+    name: Windows
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        configuration: [Release]
+    steps:
+      - uses: actions/checkout@master
+      - name: Generate cmake files
+        run: |
+          git submodule update --init --recursive --force
+          mkdir build && cd build
+          cmake .. -G "Visual Studio 16 2019" -A Win32
+      - name: Build
+        run: cmake --build build/ --config ${{ matrix.configuration }}
+      - name: Organize artifact files
+        shell: bash
+        run: |
+          mkdir dist
+          cp {COPYING,README.md} dist/
+          cd build/bin/${{ matrix.configuration }}
+          [ "${{ matrix.configuration }}" = "Debug" ] && sdl="SDL2d.dll" || sdl=SDL2.dll; cp {Cxbx.exe,glew32.dll,subhook.dll,$sdl} ../../../dist/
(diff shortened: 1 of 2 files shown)


RadWolfie:
We can deactivate the other CIs at any time, it's best to leave the CI files in case of contributors' preference to choose alternative CI other than GitHub Action.

Hmm... There's a GitHub Action task, "GitHub Releases", can make things easier for us for tag a release to include the build automatically. Except it may not work correctly base on `master_artifact_publish.yml` file. I could be wrong.

GXTX:
I think we should keep releases the way we have it right now where it's only manually pushed and tagged. Of course anything would work.

GXTX:
I've looked into this some more, this is going to require me to submit this to `master` and not `develop`. Apparently Actions will *only* run workflows located on `master` but it will still track other branches.

Margen67:
@GXTX Actually, it's because of this:
```
on:
  push:
    branches: develop
```

RadWolfie:
So, if we merge this as is. Then make any attempt to fix Action's CI in later pull request?

I'm not sure how to verify if this setup is correct or not.

GXTX:
I'm actually going to abandon this PR. It's mostly done for someone else to pick up, a better solution may be to use releases like proposed in xqemu/xqemu#266 instead of artifacts. Maybe @Margen67 will pickup?

Margen67:
If you mean make a Release for every build, that has pros and cons.
I have it set up here: https://github.com/xenia-canary/xenia-canary/releases

Pros:
 * You can link the latest build (like AppVeyor) which you can't do with artifacts: https://github.com/username/repo/releases/latest/download/file
 *The file must always have the same name, and releases can't be marked as prerelease.*

Cons:
 * It requires making a tag for every commit, which can get messy.
* Branch/tag notifications can't be disabled individually, so that can clutter up Webhooks. (like on a Discord #github channel)
  * There's probably a hack/workaround for this, but that's beyond me.

---

Anyway, feedback on this:

You can probably make this into one yml by just conditionally running steps for develop branch and push event. (not sure why you wouldn't want artifacts for all builds, though..)
You can also conditionally only have artifacts for Release so you can still verify debug builds from develop branch.

PR #1719, opened 2019-09-09
Add default bindings for the keyboard

ergo720:
Addition suggested by @PatrickvL Please test to see if I haven't accidentally broken something.


PatrickvL:
Splendid!

If I may, I've been using the following keyboard layout, maybe this can serve as default?
* identity-map `A`, `B`, `X` and `Y` inputs to their alphabetically corresponding keys
* map the Left and Right Triggers to the corresponding left and right `SHIFT` keys
* map the Back input to `Backspace`
* map the Start input to `Space` (although `Enter` *is* closer to the backspace key, ymmv.)
* map the two Thumb-press keys adjacent to their corresponding set of Axis keys

This would require shifting the Left Axis key bindings a bit aside, like in this layout :
```
const char* button_xbox_ctrl_names[XBOX_CTRL_NUM_BUTTONS][3] = {
	"D Pad Up",      "Pad N",      "UP",
	"D Pad Down",    "Pad S",      "DOWN",
	"D Pad Left",    "Pad W",      "LEFT",
	"D Pad Right",   "Pad E",      "RIGHT",
	"Start",         "Start",      "SPACE", // or "RETURN" ?
	"Back",          "Back",       "BACK",
	"L Thumb",       "Thumb L",    "W", // or "R" ?
	"R Thumb",       "Thumb R",    "O", // or "U" ?
	"A",             "Button A",   "A",
	"B",             "Button B",   "B",
	"X",             "Button X",   "X",
	"Y",             "Button Y",   "Y",
	"Black",         "Shoulder R", "COMMA",
	"White",         "Shoulder L", "PERIOD",
	"L Trigger",     "Trigger L",  "LSHIFT",
	"R Trigger",     "Trigger R",  "RSHIFT",
	"Left Axis X+",  "Left X+",    "F",
	"Left Axis X-",  "Left X-",    "S",
	"Left Axis Y+",  "Left Y+",    "E",
	"Left Axis Y-",  "Left Y-",    "D",
	"Right Axis X+", "Right X+",   "L",
	"Right Axis X-", "Right X-",   "J",
	"Right Axis Y+", "Right Y+",   "I",
	"Right Axis Y-", "Right Y-",   "K",
	"Motor",         "LeftRight",  "",
};
```

PatrickvL:
Tested, works fine!

LukeUsher:
@PatrickvL 

Identity mapping A/B/X/Y is not a good idea: It's terrible for gameplay. The keys need to be close together, ideally, so you can hit them all with one hand, with minimal movement.

A/B/X/Y mapped to A/B/X/Y will make many games extremely difficult to play, and it doesn't lend itself as a default mapping for games.

Personally, I map my inputs such that A/B/X/Y/Black/White are ZXCVB

Left Analog Stick is the arrow keys, right analog stick is numpad 2468, d-pad is IJKL, L/R are A/B, Start is Enter, Select is backspace and 1/2 are analog click left/right.

Not the best mapping and not once most people would like to use, but the point applies: Gameplay buttons (ABXY,black,white) should be close together

PatrickvL:
As discussed on discord, the proposed keyboard default bindings are modeled after https://github.com/espes/xqemu/wiki/Input-Options which avoids non-alphanumeric keys is that might not be present in all the keyboards.

Let's go with that for now, it's already great that we HAVE defaults!

Thanks!

PR #1721, opened 2019-09-10
Fix Bind Input Detection

RadWolfie:
DetectInput function has additional checking if input has been released before provide input than after it was pressed. This fix false positive of space (release) and enter (holding) keys trigger on Windows asking for input again.

Since DetectInput is only checking first pressed and released, the button's text will display "[_____]" while input is still pressed. If the input isn't released after certain time, then original bind will be restore.

Tested with no issues:
- Keyboard & mouse
- Wireless Xbox 360 Controller

Plus only check for changed state instead of repeatedly set unchanged states. This prevent spam in the console while debugging and prevent unnecessary checks.

### DIFF
--- a/src/common/input/InputWindow.cpp
+++ b/src/common/input/InputWindow.cpp
@@ -59,6 +59,7 @@ void InputWindow::Initialize(HWND hwnd, int port_num, int dev_type)
 	m_max_num_buttons = dev_num_buttons[dev_type];
 	m_port_num = port_num;
 	m_bHasChanges = false;
+	m_bIsBinding = false;
 
 	// Set window icon
 	SetClassLong(m_hwnd_window, GCL_HICON, (LONG)LoadIcon(GetModuleHandle(NULL), MAKEINTRESOURCE(IDI_CXBX)));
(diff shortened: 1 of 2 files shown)


src/common/input/InputWindow.cpp
```diff
@@ -214,12 +223,18 @@ void InputWindow::BindButton(int ControlID)
 	if (dev != nullptr) {
 		// Don't block the message processing loop
 		std::thread([this, dev, ControlID]() {
+			// testcase: spacebar and enter keys; without this fix will cause unstable binding result.
+			// Plus EnableButton set to false for prevent enter key create useless new threads from holding enter key down.
+			if (!m_mtx_binding.try_lock()) {
+				return;
+			}
```
ergo720: I believe you can probably replace this mutex with a simple boolean `m_bIsDetecting`, which should be `true` during an "input attempt" and thus return immediately. Also, I think you should place the check before creating the `std::thread` to avoid creating the thread object entirely when the condition holds.

RadWolfie: A while ago, I thought of `std::thread` will return a handle to for us check if thread is running or not. Turns out I was wrong, `std::thread` apparently only will provide an id which is useless in this situation.

So, I went ahead use atomic boolean since it's the only one I know of has multi-thread safe in C++ standard.

RadWolfie:
What's been changed from this pull request after rebased:
- Replaced mutex to atomic boolean
- Removed optional disable bind button. Since atomic boolean is able to prevent unnecessary new thread creations.

PR #1722, opened 2019-09-10
Moving, reordering, renaming and other cleanup of D3D symbols

PatrickvL:
(all taken from my Vsh_unpatching branch)

This moves symbol declarations to more appropriate locations, completes some unknown types, and applies symbols where there, well, apply ;)

This doesn't change any functionality (compiles should result in a nearly binary identical executable).

Having this merged, makes related code better maintainable.

(Apologies for a few CRLF<>LF changes, but at least it's now consistently used within the files that have changes because of this, instead of a mixed EOL style.)


src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -193,43 +193,43 @@ inline FLOAT NormShortToFloat(const SHORT value)
...
+    
+    return stream;
+}
+
+void XTL::CxbxVertexBufferConverter::PrintStats()
+{
+    printf("Vertex Buffer Cache Status: \n");
+    printf("- Cache Size: %d\n", m_PatchedStreams.size());
+    printf("- Hits: %d\n", m_TotalCacheHits);
+    printf("- Misses: %d\n", m_TotalCacheMisses);
```
ergo720: Nit: are these required to always be printed or can they use `EmuLog` instead?

PatrickvL: It hasn't changed other than EOL style, but yeah, that could be done. I'd prefer in another PR though.

LukeUsher: These are printed when F1 is hit, it's an intentional debug feature, unrelated to general logging:

LukeUsher:
Looks good, as long as it is indeed code-style only, I'm happy for it to be merged.

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1061,45 +972,51 @@ constexpr DWORD X_D3DCOLORWRITEENABLE_ALL = 0x01010101; // Xbox ext.
...
-#define VSH_MAX_TEMPORARY_REGISTERS 32
-#define VSH_VS11_MAX_INSTRUCTION_COUNT 128
-#define VSH_VS2X_MAX_INSTRUCTION_COUNT 256
-#define VSH_VS30_MAX_INSTRUCTION_COUNT 512
+/// nv2a microcode header
+typedef struct
+{
+	uint16_t Version; // See VERSION_XVS*
+	uint16_t NumInst;
+}
```
RadWolfie: This structure is changed. Originally was:
```
typedef struct
{
    uint8_t Type;
    uint8_t Version;
    uint8_t NumInst;
    uint8_t Unknown0;
}
X_VSH_SHADER_HEADER;
```
I believe testing is a requirement to ensure it doesn't break the existing code.

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -38,10 +38,16 @@
...
 
-#ifdef CXBX_USE_VS30
 //#define CXBX_USE_VS30 // Separate the port to Vertex Shader model 3.0 from the port to Direct3D9
+#ifdef CXBX_USE_VS30
+	#define VSH_MAX_INSTRUCTION_COUNT VSH_VS30_MAX_INSTRUCTION_COUNT // == 512
+#else 
+	#define VSH_MAX_INSTRUCTION_COUNT VSH_VS2X_MAX_INSTRUCTION_COUNT // == 256
 #endif
 
+// Host Vertex Shader version (mustn't conflict with any VERSION_XVS*)
```
RadWolfie: In my opinion, the comment need little bit better description. Since host's graphic can't support unofficial value. Include "internal" in the description will help understanding better.

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1285,7 +1203,146 @@ typedef enum _X_D3DVSD_TOKENTYPE
 #define X_D3DFVF_TEXCOORDSIZE3(Index) (X_D3DFVF_TEXTUREFORMAT3 << (Index * 2 + 16))
 #define X_D3DFVF_TEXCOORDSIZE4(Index) (X_D3DFVF_TEXTUREFORMAT4 << (Index * 2 + 16))
 
-
+// NV2A Push buffer command masks
 typedef DWORD NV2AMETHOD;
 
+#define PUSH_TYPE_MASK         0x00000002 // 2 bits
```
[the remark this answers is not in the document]
LukeUsher: PushBuffers are fully exposed by DirectX on the Xbox: there’s a set of APIs for generating and playing back raw pushbuffer command streams and many, many games use this functionality. Defines like this are helpful for decoding them and generating host D3D draw calls.

src/core/hle/D3D8/XbD3D8Types.h
PatrickvL: True, but the existing code was wrong; The `Version` field must be two bytes, and the two places in our code where the `Version` field is accessed, now also check against `VERSION_X*` defines that are similarly updated to two bytes. Given the input data will always correspond to the newly defined values anyway, this will result in identical functionality.

PatrickvL: The `NumInst` field is also a 16 bit field, which can be seen by disassembling Xbox code that reads it. In practice it makes no difference though, since it isn't allowed to contain values above 136, so one byte would've sufficed. Still, it's better to define the types like they're supposed to.

src/core/hle/D3D8/XbVertexShader.cpp
PatrickvL: I'll address this, and more in a follow-up PR.

RadWolfie: I can confirm it is resolved in #1725

src/core/hle/D3D8/XbD3D8Types.h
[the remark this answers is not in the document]
PatrickvL: After discussing this, I've moved them back to `nv2a.h`.

PR #1723, opened 2019-09-11
Initialize pCurrentvertexShaderStreamInfo to nullptr

NZJenkins:
There's a branch that checks if this is `nullptr`, but it might not be unless we initialize it
Fixes Lego Star Wars II


LukeUsher:
Nice spot, all pointers that are unset should be initialized to nullptr/xbnullptr/similar to allow them to be detected/checked for validity.

PR #1725, opened 2019-09-11
Moving d3d types

PatrickvL:
This is a follow-up on #1722 and moves a few types outside of the XTL namespace.

The past few years, way too many types have ended up being declared in the XTL namespace, which is wrong - only Xbox-specific types should be in this namespace.
This PR moves the pushbuffer, vertexbuffer and vertexshader types out of XTL, and fixes the implications for the code referencing those symbols.

Otherwise, there's no functional change in this PR.


src/core/hle/D3D8/XbPushBuffer.cpp
```diff
@@ -31,8 +31,11 @@
 
 #include "core\kernel\support\Emu.h"
 #include "core\kernel\support\EmuXTL.h"
-#include "XbD3D8Types.h" // For X_D3DFORMAT
+namespace XTL {
+	#include "core\hle\D3D8\XbD3D8Types.h" // For X_D3DFORMAT
```
RadWolfie: nit: Instead of risk re-introduce namespace hell in the future, nest entire `XbD3D8Types.h` and `XbD3D8Types.cpp` as XTL namespace which should had been done in the first place. Doing so will remove `EmuXTL.h`'s XTL hell conflict.

PatrickvL: There's no XbD3DTypes.cpp, but yes, this should really be addressed at some point. Right now, this wrapper was needed because XBD3DTypes.h is already included (in EmuXTL.h) in the XTL namespace too.
However, this PR is not about fixing XTL hell, but about moving types to a more appropriate place, where they belong.

RadWolfie: 👌

src/core/kernel/support/EmuXTL.h
```diff
@@ -32,10 +32,6 @@ namespace XTL
 {
     #include "core\hle\XAPI\Xapi.h"
     #include "core\hle\D3D8\Direct3D9/Direct3D9.h"
-    #include "core\hle\D3D8\XbConvert.h"
-    #include "core\hle\D3D8\XbVertexBuffer.h"
-    #include "core\hle\D3D8\XbPushBuffer.h"
-    #include "core\hle\D3D8\XbVertexShader.h"
```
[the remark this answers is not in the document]
PatrickvL: Yeah baby!

src/core/hle/D3D8/XbPushBuffer.cpp
```diff
@@ -492,7 +500,7 @@ extern void XTL::EmuExecutePushBufferRaw
 	uint32_t *dma_put; // pushbuffer current end address
 	uint32_t *dma_get; //pushbuffer current read address
 	struct {
-		NV2AMETHOD mthd; // Current method
+		XTL::NV2AMETHOD mthd; // Current method
```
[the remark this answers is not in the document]
PatrickvL: I'll correct this, thanks.

PatrickvL:
Ready for merge - anyone?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7564,7 +7570,7 @@ void XTL::CxbxDrawIndexed(CxbxDrawContext &DrawContext)
 
 			// Emulate a quad by drawing each as a fan of 2 triangles
 			HRESULT hRet = g_pD3DDevice->DrawIndexedPrimitive(
-				D3DPT_TRIANGLEFAN,
+				XTL::D3DPT_TRIANGLEFAN,
```
RadWolfie: Was this intended change? Since it's calling directly to the host's graphic driver.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7642,12 +7648,12 @@ void XTL::CxbxDrawPrimitiveUP(CxbxDrawContext &DrawContext)
 		INDEX16 HighIndex = (INDEX16)(DrawContext.dwVertexCount - 1);
 
 		HRESULT hRet = g_pD3DDevice->DrawIndexedPrimitiveUP(
-			D3DPT_TRIANGLELIST, // Draw indexed triangles instead of quads
+			XTL::D3DPT_TRIANGLELIST, // Draw indexed triangles instead of quads
```
RadWolfie: See comment above.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7642,12 +7648,12 @@ void XTL::CxbxDrawPrimitiveUP(CxbxDrawContext &DrawContext)
...
 
 		HRESULT hRet = g_pD3DDevice->DrawIndexedPrimitiveUP(
-			D3DPT_TRIANGLELIST, // Draw indexed triangles instead of quads
+			XTL::D3DPT_TRIANGLELIST, // Draw indexed triangles instead of quads
 			LowIndex, // MinVertexIndex
 			(HighIndex - LowIndex) + 1, // NumVertexIndices
 			PrimitiveCount,
 			pIndexData,
-			D3DFMT_INDEX16, // IndexDataFormat
+			XTL::D3DFMT_INDEX16, // IndexDataFormat
```
RadWolfie: See comment above.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Each time a function is moved outside of the XTL namespace, it's arguments and implementation could still refer to symbols that reside in the XTL namespace.
Those symbols won't compile unless they get access to the XTL namespace, either through a prefix, or a mention of `using namespace XTL` in their scope.

Now, of course, ultimately non-Xbox symbol should all be moved outside the XTL namespace, but during this transition these intermediate bandaids are required to keep everything compiling successfully.

RadWolfie: Currently, CxbxDrawIndexed function is in XTL namespace. So are the other comments below.

So, using Windows' D3D8/9 header files aren't ideal to do so in XTL namespace. We should make a copy of our own since some may not match to xbox version. I had done similar method for Xbox's DirectSound in the past except it does not match Windows' DirectSound macros. I had to revert back to copy from Windows' DirectSound marcos with proper values. Including add values not listed in Windows' DirectSound.

It's the main reason I suggested to use XTL namespace inside `XbD3D8Types.h` file. 😛 Then Windows' D3D9 includes will not be in XTL namespace at all.

PatrickvL: Again, we all agree XTL namespace should be cleaned up thoroughly so that it solely contains Xbox symbols and nothing else.

However, reaching that end goal is a huge task.

This PR is a step in that direction, but I can't and don't want to do it all in one go.
Instead, I'm working towards it in manageable increments.

This approach makes it easier to review, reduces chance of failure and incrementally restores sanity in this mess sooner, rather than later.

RadWolfie: @PatrickvL replied:
> Since they're going to be removed anyway, is probably more efficient to merge it as-is and pick up the remainder in a enfollowing PR.

Also said XTL-hell fixup progress will be easier to remove XTL namespace from the function itself. Then again remove the XTL namespace from the nits I made.

PR #1728, opened 2019-09-14
Exclude PDO from input hack, fixes regression from hack being introduced

LukeUsher:
The input hack is required for many games to function (JSRF, Gunvalkyre, Lego, others), but it breaks Panzer Draggon Orta

This adds an exception to prevent the hack being applied for this game.

With this change, Panzer Dragoon Orta works again.


RadWolfie:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1729 is most likely better replacement than include specific title to bypass the hack.

PR #1729, opened 2019-09-15
Fix Delay Input Hack

RadWolfie:
This resolve Panzer Dragoon ORTA title regression.

Titles tested with no regression:
- Lego Star Wars
- Jet Set Radio Future
- GunValkyrie
- Otogi

Also tested with at least 8 various titles with no regression.


PatrickvL:
Be aware that this is still (and also) a hack, regardless that it's less 'hacky' than the other PR

RadWolfie:
I got a report this pull request break the Transworld Surf intro.

LukeUsher:
> I got a report this pull request break the Transworld Surf intro.

This is why I decided to do an exclusiong for PDO. But we really need to test on real hardware to find out what is going wrong here

RadWolfie:
Closed for better solution by #1730 pull request.

PR #1730, opened 2019-09-16
Implement Input Initialize Latency Timer & Device Emulation Status

RadWolfie:
More or less a proper fix than #1729 and #1728 pull requests.

Instead of increment counter for every call occurs, using a delay timer to simulate initialize processing after call XInitDevices is accurate solution.

Works well with:
- Lego Star Wars
  - Without delay timer will cause not to call the input polling even after devices are open.
- Panzer Dragoon ORTA
  - After called and within 500 milliseconds will cause unstable crash for the emulator.

Did not get a regression from:
- Jet Set Radio Future
- GunValkyrie
- Otogi

### DIFF
--- a/src/core/hle/XAPI/Xapi.cpp
+++ b/src/core/hle/XAPI/Xapi.cpp
@@ -67,8 +67,9 @@ PFARPROC1 fnCxbxVSBCOpen;
 //typedef DWORD(*fnCxbxVSBCGetState)(UCHAR *);
 XTL::PXPP_DEVICE_TYPE g_DeviceType_Gamepad = nullptr;
 
-// Flag is set after gamepad state has been queried by the title. Supports defer gamepad connection hack
-bool g_gamepadStateQueriedHackFlag = false;
+// Flag is unset after initialize devices is done by simulate LLE USB thread.
+std::atomic<bool> g_bIsDevicesInitializing = true;
+std::atomic<bool> g_bIsDevicesEmulating = false;
 
 //global bridge for xbox controller to host, 4 elements for 4 ports.
 X_CONTROLLER_HOST_BRIDGE g_XboxControllerHostBridge[4] = {
@@ -99,6 +100,7 @@ bool operator==(XTL::PXPP_DEVICE_TYPE XppType, XBOX_INPUT_DEVICE XidType)
 	default:
 		break;
 	}
+
 	return false;
 }
 
@@ -109,6 +111,9 @@ bool operator!=(XTL::PXPP_DEVICE_TYPE XppType, XBOX_INPUT_DEVICE XidType)
 
 bool ConstructHleInputDevice(int Type, int Port)
 {
+	g_bIsDevicesEmulating = true;
+	bool ret = true;
+
 	switch (Type)
 	{
 	case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE): {
@@ -137,19 +142,24 @@ bool ConstructHleInputDevice(int Type, int Port)
 	case to_underlying(XBOX_INPUT_DEVICE::MEMORY_UNIT):
 	case to_underlying(XBOX_INPUT_DEVICE::IR_DONGLE): {
 		EmuLog(LOG_LEVEL::INFO, "%s: device %s is not yet supported", __func__, GetInputDeviceName(Type).c_str());
-		return false;
+		ret = false;
 	}
+	break;
 
 	default:
 		EmuLog(LOG_LEVEL::WARNING, "Attempted to attach an unknown device type (type was %d)", Type);
-		return false;
+		ret = false;
+		break;
 	}
 
-	return true;
+	g_bIsDevicesEmulating = false;
+	return ret;
 }
 
 void DestructHleInputDevice(int Port)
 {
+	g_bIsDevicesEmulating = true;
+
 	g_XboxControllerHostBridge[Port].XboxType = XBOX_INPUT_DEVICE::DEVICE_INVALID;
 	g_XboxControllerHostBridge[Port].XboxPort = PORT_INVALID;
 	while (g_XboxControllerHostBridge[Port].bIoInProgress) {}
@@ -162,6 +172,8 @@ void DestructHleInputDevice(int Port)
 	g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucInputStateSize = 0;
 	g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucFeedbackSize = 0;
 	g_XboxControllerHostBridge[Port].XboxDeviceInfo.dwPacketNumber = 0;
+
+	g_bIsDevicesEmulating = false;
 }
 
 void SetupXboxDeviceTypes()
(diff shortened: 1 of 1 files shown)


src/core/hle/XAPI/Xapi.cpp
```diff
@@ -1375,9 +1381,9 @@ DWORD WINAPI XTL::EMUPATCH(XGetDeviceEnumerationStatus)()
 
 	LOG_FUNC();
 
-	LOG_UNIMPLEMENTED();
+	DWORD ret = g_bIsDevicesInitialized ? XDEVICE_ENUMERATION_IDLE : XDEVICE_ENUMERATION_BUSY;
```
ergo720: This is not completely correct. `XGetDeviceEnumerationStatus` should return `XDEVICE_ENUMERATION_BUSY` also during device insertions/removals, not just during the call to `XInitDevices`. You can easily simulate this behaviour by setting a boolean `g_bIsEnumerating ` to true during the calls to `ConstructHleInputDevice` and `DestructHleInputDevice` (called by the input manager during device insertions/removals respectively),  and then put `g_bIsEnumerating ` back to false just before the functions return. So the code here would become `DWORD ret = g_bIsDevicesInitialized ? (g_bIsEnumerating  ? XDEVICE_ENUMERATION_BUSY : XDEVICE_ENUMERATION_IDLE) : XDEVICE_ENUMERATION_BUSY;`

RadWolfie: I'm not sure why, this should reveal as outdated in github browser.

Anyway, yeah it wasn't completely correct. Thanks for the guidance to also interface with HLE's insertion/removal functions. 😄 While your example code is a bit messy. I had pushed a cleaner version of your example. 😉

RadWolfie:
What's been changed after force pushed:
- Replace `g_bIsDevicesInitialized` to `g_bIsDevicesInitializing` for cleaner code review in `XGetDeviceEnumerationStatus` patch function.
- Added `g_bIsDevicesEmulating` check as suggested by @ergo720.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -81,6 +82,8 @@ X_CONTROLLER_HOST_BRIDGE g_XboxControllerHostBridge[4] = {
 
 bool operator==(XTL::PXPP_DEVICE_TYPE XppType, XBOX_INPUT_DEVICE XidType)
 {
+	g_bIsDevicesEmulating = true;
```
ergo720: Why put this here? This operator is used to check if the DeviceType supplied by the title matches the xid type of the device. For example, it will be called by `XInputOpen` and thus set the flag, which is incorrect as calling that function doesn't imply that enumeration is in progress. The correct fix is to simply move this to `DestructHleInputDevice` instead.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -145,6 +152,7 @@ bool ConstructHleInputDevice(int Type, int Port)
 		return false;
 	}
 
+	g_bIsDevicesEmulating = false;
```
ergo720: Please notice that the cases for the other xid types use `return` and not `break`, which means that the flag will stay set if those are hit. However, the `Type` argument comes from the device type selection that the user has made in the input gui, and since at the moment they can only set `DEVICE_INVALID` and `MS_CONTROLLER_DUKE`, we will never reach the other cases (note that, for `DEVICE_INVALID` the input manager doesn't call this function). It's still a bug though, so it might be worth fixing it anyway.

src/core/hle/XAPI/Xapi.cpp
RadWolfie: Whoops, the operator looks identical to `ConstructHleInputDevice` function. So when I start to look for `DestructHleInputDevice` except it was `ConstructHleInputDevice` function from my mind trick. 😅

src/core/hle/XAPI/Xapi.cpp
RadWolfie: Didn't notice the returns at the time. I went ahead remove the returns then only use return once with given `ret` value.

As for the input manager doesn't call with `DEVICE_INVALID` type, that will need to be done in a separate pull request.

RadWolfie:
What's been changed after force pushed:
- Put `g_bIsDevicesEmulating` in `DestructHleInputDevice` function, not `operator==` function.
- Replace returns to a single return for `ConstructHleInputDevice` function and insert breaks to other types and default cases.

src/core/hle/XAPI/Xapi.cpp
ergo720: >As for the input manager doesn't call with DEVICE_INVALID type, that will need to be done in a separate pull request.
>
That's by design, it's not an error. If `Type` is `DEVICE_INVALID`, it means it's a removal event, and so the input manager will call `DestructHleInputDevice`, so there's nothing to fix here.

ergo720:
Looks good now and ready for merging, unless someone has reported input regressions from this PR.

src/core/hle/XAPI/Xapi.cpp
RadWolfie: > It's still a bug though, so it might be worth fixing it anyway.

Yet you said it's a bug? I'm confused.

RadWolfie:
Actually, I just remembered. Even device emulating also need to block the UpdateConnectedDeviceState  change since it's in "busy" state. That's one fix I had pushed.

src/core/hle/XAPI/Xapi.cpp
ergo720: The bug I was referring to was hitting the other cases with a `return` and thus leave the flag set. This function not being called with `DEVICE_INVALID` is fine instead.

RadWolfie: Ah okay! 😄 Then it is resolved.

ergo720:
>If the delay is commented out the following games work:
>
This contradicts the findings from @RadWolfie, which said that with this PR (and thus with the delay active), PDO and JSRF work.

RadWolfie:
> The games I mention work with or without the delay for me

This is because of titles doesn't have the bugs. Except for Panzer Dragoon ORTA and Lego Star Wars. Also if you do disable the delay thread, then the status will remain as busy. It's a fail-safe feature since if XInitDevices were not called, then device inputs will not do anything.

**Lego Star Wars** is a perfect example of game's bug did check for device change on first call yet *never did start the polling calls* for any input from the devices. On real xbox hardware, the bug never occur.

*To be clear, I'm not talking about the 2nd sequel of Lego Star Wars.*
**P.S.** Lego Star Wars 2 never had an issue with the input before, it was the first sequel of Lego Star Wars which did had issues.

Then further testing with Panzer Dragoon ORTA after certain counter did produce reboot to dashboard bug in the emulator (with experimental code). However with latency delay confirmation, it never occur on real xbox hardware, hence the game's bug. It is just being busy doing other tasks and then got checked that devices are ready too early.

In brief summary, `XInitDevices` internal calls create a thread to perform device changes check. This thread take a while, under half a second, to finish the tasks. So `XGetDevices`, `XGetDeviceChanges`, `XGetDeviceEnumerationStatus` functions are just a reader with partial write internal structure to know API had been called via reset the changed state and preserved the old changed state until the next call occur.

PR #1731, opened 2019-09-17
Fix a large part of XTL-hell

PatrickvL:
This PR fixes a large part of what we've been calling "XTL-hell", which is (was) the problem that non-Xbox symbols were included in the Xbox-only "namespace XTL".

The remaining XTL mentions are now really needed to indicate Xbox-specific symbols.

(Note, that the sound-code is left largely untouched, so there's still XTL-issues aroud that part of our code.)

This PR should not have any functional impact.

There's a slight risk that some log-renderings are impacted, since symbols now no-longer-can or now-cannot be resolved to logging intrinsics. If this turns out to be the case, it should be resolved in a followup-PR.


src/vsbc/CxbxVSBC.h
```diff
@@ -2,9 +2,12 @@
 #define CXBXVSBC_API __declspec(dllexport) 
 #else
 #define CXBXVSBC_API __declspec(dllimport) 
-#endif#pragma once
-//#include <windef.h>
-//#include "../CxbxKrnl/EmuXapi.h"
+#endif
+
+#ifdef _MSC_VER
```
RadWolfie: We don't need this, `#pragma once` is supported for all compilers.

src/common/xdvdfs-tools/xdvdfs.h
```diff
@@ -4,8 +4,6 @@
 #define __XDVDFS_H__
 
 #ifdef __cplusplus
-#pragma once
```
RadWolfie: Unintentional change, please revert this.

import/OpenXDK/include/xboxkrnl/xboxkrnl.h
```diff
@@ -175,6 +175,7 @@ typedef long                            NTSTATUS;
 // * Registry value types
 // ******************************************************************
 // Used in ExQueryNonVolatileSetting and ExSaveNonVolatileSetting
+#ifndef REG_NONE // Avoid "warning C4005:  'REG_NONE': macro redefinition" (conflicting with winnt.h)
```
RadWolfie: Is it exactly the same with `winnt.h` file?

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -622,7 +623,7 @@ typedef struct _D3DVBLANKDATA
     DWORD           Swap;
     DWORD           Flags;
 }
-D3DVBLANKDATA;
+X_D3DVBLANKDATA;
```
RadWolfie: For what reason do we need to add `X_`? The same for below changes. The plan is to remove `X_` once we completely once there's no conflict with host's stuff out of XTL namespace. **(excluding `#define`)**

src/core/hle/XONLINE/XOnline.h
```diff
@@ -22,8 +22,10 @@
 // *  All rights reserved
 // *
 // ******************************************************************
-#ifndef XONLINE_H
-#define XONLINE_H
+#if 0 // XOnline.h isn't used, but below is still usefull documentation.
```
RadWolfie: found a typo, should be **useful**. 😉

src/core/hle/D3D8/XbD3D8Types.h
PatrickvL: Up until now, `X_` prefix has been applied to nearly all Xbox symbols (and *especially* `#defines`), to easily discern them from non-Xbox types.

Now that this PR has mostly resolved 'XTL-hell', this `X_`-prefix can indeed be dropped, since everything should already mention the namespace `XTL::`-prefix.

However, since `#define`'s are not scoped by namespace, they still need to be discerned to avoid conflicts with host `#define`'s.

Therefor, for now, it's probably best to still apply the `X_`-prefix to all Xbox symbols, hence the above change.

src/common/xdvdfs-tools/xdvdfs.h
PatrickvL: Actually, this is intentional, since this file is already using include guards.

For compatibility with C, include guards should stay, so this `#pragma once` is needless and can thus be removed.

import/OpenXDK/include/xboxkrnl/xboxkrnl.h
PatrickvL: Good question, I haven't checked. This change has no real value yet, since the clash between xboxkrnl and windows types still exists anyway. So I could revert this?

src/common/xdvdfs-tools/xdvdfs.h
RadWolfie: Well, I don't approve this change unless it match upstream change. And it doesn't block compiler to build anyway. So, it's not a critical issue.

RadWolfie: Actually upstream doesn't have `#pragma once`. So this change is approve.

import/OpenXDK/include/xboxkrnl/xboxkrnl.h
RadWolfie: Let's review first before decide to revert. 😉

src/common/xdvdfs-tools/xdvdfs.h
PatrickvL: Which upstream? This code is not submodule from https://github.com/multimediamike/xbfuse
And even in that repo, there's no mention of `#pragma once` - I think I may have added it a long time ago, without realizing it was unnecessary.

From what I remember, this code was copied from somewhere on the internet years back, has been translated from C to Delphi (Dxbx), back to C(++), and has been updated and fixed in various ways.

This would need verification, and it would be nice to contribute back if there's enough motivation for that. But that's a task for another time.

import/OpenXDK/include/xboxkrnl/xboxkrnl.h
RadWolfie: Confirmed there's no change from `winnt.h` file. Except they included `ul` after the numbers. I don't think there will be any issue with value type difference?

src/core/hle/D3D8/XbD3D8Types.h
ergo720: Now that you mentioned it, I remembered that I added the  `X_` prefix to some host Xapi variables in `XapiCxbxr.h` to distinguish them, so you might want to remove the prefix from those if possible.

PatrickvL: https://github.com/PatrickvL/Cxbx-Reloaded/blob/more_cleanup/src/core/hle/XAPI/XapiCxbxr.h#L26 does mention that, indeed.
How about changing that into a `CXBX_` prefix instead? That's what was done in other places too.

x1nixmzeng:
I'm glad this is going ahead. I had started this on branch of my own but there were far too many files!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -549,17 +549,14 @@ void DrawUEM(HWND hWnd)
 
 	SetTextColor(hMemDC, RGB(0, 204, 0));
 
-	std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> UTF8toUTF16;
```
RadWolfie: This change will produce a regression see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1676 pull request.

src/core/kernel/common/strings.hpp
```diff
@@ -25,12 +25,10 @@
...
 #include <string>
 
-// The file is saved as UTF-8 format, some strings may not output correctly at runtime.
-// To work around this, you will need to use std::wstring_convert class
-// or any UTF-8 to UTF-16 conversion function.
+// The file is saved as UTF-8 format.
 
 // Universal Error Message (UEM)
-static std::string uem_str = "Your Xbox requires service.\n\nPlease call Xbox Customer Support.\n"
+static std::wstring uem_str = L"Your Xbox requires service.\n\nPlease call Xbox Customer Support.\n"
```
RadWolfie: UEM string isn't wide character (two bytes). UTF-8 is *actually* single character.

PatrickvL: Duly noted!

src/common/xdvdfs-tools/xdvdfs.h
RadWolfie: Wrong repo and doesn't match, you made a copy from Rocky5's repo. Pull request source: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/210 with a mention of Rocky5 repo link. Although, the link is invalid since the file has been moved.

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -538,6 +539,7 @@ void PrintCurrentConfigurationLog()
 		// The correct solution is to use GetProductInfo but that function
 		// requires more logic to parse the response, and I didn't feel
 		// like building it just yet :P
+#pragma warning(suppress: 4996) // GetVersion is deprecated, but there's no simple alternative
```
RadWolfie: GetVersion was used to determine which version of WINE. I preferred this warning to be shown at compile time for awareness we still have legacy code.

Also, the check is part of the old Cxbx codebase.

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -1358,15 +1360,15 @@ __declspec(noreturn) void CxbxKrnlInit
...
-			"   pLibraryVersion     : 0x%.08X\n"
+			"   hwndParent          : 0x%.08p\n"
+			"   pTLSData            : 0x%.08p\n"
+			"   pTLS                : 0x%.08p\n"
+			"   pLibraryVersion     : 0x%.08p\n"
 			"   DebugConsole        : 0x%.08X\n"
 			"   DebugFilename       : \"%s\"\n"
-			"   pXBEHeader          : 0x%.08X\n"
+			"   pXBEHeader          : 0x%.08p\n"
 			"   pXBEHeaderSize      : 0x%.08X\n"
```
RadWolfie: **REMARK:** `pXBEHeaderSize` should be `dwXBEHeaderSize`.

src/devices/usb/Hub.cpp
```diff
@@ -391,9 +391,9 @@ void Hub::UsbHub_HandleControl(XboxDeviceState* dev, USBPacket* p,
 			port = &m_HubState->ports[n];
 			EmuLog(LOG_LEVEL::DEBUG, "GetPortStatus -> Address 0x%X, wIndex %d, wPortStatus %d, wPortChange %d",
 				m_HubState->dev.Addr, index, port->wPortStatus, port->wPortChange);
-			data[0] = port->wPortStatus;
+			data[0] = port->wPortStatus & 0xFF;
 			data[1] = port->wPortStatus >> 8;
-			data[2] = port->wPortChange;
+			data[2] = port->wPortChange & 0xFF;
```
RadWolfie: Interesting... I think this change should also go upstream in QEMU repo.

RadWolfie:
There's two remarks I am requesting:
- Allow warning about `GetVersion` being depreciated.
- Fix the inaccurate wording of `pXBEHeaderSize` to `dwXBEHeaderSize`. Since the string format had been changed.

src/devices/usb/Hub.cpp
ergo720: I noticed this too. This code writes `wPortStatus` and `wPortChange` to the output buffer of the current USB request (`GetPortStatus`). Since both variables are `uint16_t` but data[i] is `uint8_t`, this will result in the truncation of the upper byte, which is intentional but the compiler will probably generate a warning about it and it should be why @PatrickvL has done this change. In practice, this won't change the end result of the operation so it's fine.

RadWolfie: > In practice, this won't change the end result of the operation so it's fine.

Right, except every software's source code shouldn't output any warnings at all. They should be fix anyway. Since QEMU is a large code base, and if developers do this way. It's likely the warnings are up by over ten of thousands or maybe hundred of thousands. When this occur, it will hide critical bugs for ages and developers will just ignore the warnings. So, upstream should also have this change anyway.

PatrickvL:
DrawContext is merely a carrier of data that doesn't survive longer than the draw call itself. So storing a handle in there is equal to but storing but using the source value directly, because the source value doesn't change in between.

src/devices/usb/Hub.cpp
PatrickvL: Explicit truncation indeed avoids a warning on implicit truncation. A comment for just that felt like overkill.

src/core/kernel/init/CxbxKrnl.cpp
PatrickvL: GetVersion is not used to determine wine version. Its merely printed under environment variable in the debug log.

src/devices/usb/Hub.cpp
PatrickvL: Where does Hub.cpp originate from?

RadWolfie: Originate from: https://github.com/qemu/qemu/blob/master/hw/usb/dev-hub.c
😉

RadWolfie:
Thanks for the remark about D3D stuff. 😄

src/devices/usb/Hub.cpp
LukeUsher: As previously discussed, any useful changes or bug fixes to xqemu code should be upstreamed to them

PatrickvL: Again, this change merely avoids a compiler warning, but is otherwise is functionally equivalent.

I'm not going to solve compiler warnings for another project, it's already bad enough we have to go through so much hassle to get our own code in a better shape.

Also, since the commit was made that actually introduced this specific piece of code, it has been left untouched for FOURTEEN years. And we're not talking about Xqemu here, but *qemu* history, which predates Xqemu by a long shot. So I really don't think anyone else would care enough to be notified about little things like this.

I also feel discussing this any further would drag things on for no sane and valid reason and would become a major waste of time and energy for everyone involved, which I'd much rather see spend on more important things, like actually improving our emulator.

PatrickvL: https://bugs.launchpad.net/qemu/+bug/1844644

PatrickvL: https://github.com/qemu/qemu/blob/bb36d4708bc93874af95db54da2a9cee0d30a84d/hw/usb.c#L482

RadWolfie: cba discussing QEMU's bad quality codes. Mark a resolved, any upstream to QEMU will not be in further discussion.

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -527,34 +528,8 @@ void CxbxPopupMessageEx(CXBXR_MODULE cxbxr_module, LOG_LEVEL level, CxbxMsgDlgIc
 void PrintCurrentConfigurationLog()
 {
 	// Print environment information
```
RadWolfie: As discussed in discord, @LukeUsher forgot to remove this comment.

PatrickvL: I'd say this comment can stay, since it also applies to the logging of wine version number.

RadWolfie: My original reply to the commented line.
> Perhaps replace it to a proper comment or re-add "Windows Detected". :wink:

Then @LukeUsher said:
> I removed that comment :wink:
> at least i thought i did :stuck_out_tongue:

So it's two options either re-word the comment to such as `// Print wine information if running under wine environment.` or remove the comment. Though, I agree with @LukeUsher to just remove the comment.

PatrickvL: Stop w(h)ineing ;p

PR #1735, opened 2019-09-25
Fix New GUI Input Breakage

RadWolfie:
The changes are to fix #1713 pull request to able allow loader's emulator project to compile. Since some includes were not meant for emulation usage. And moved emulation codes into proper common files.

Plus move SDL2 linkage directly into target_link_libraries, CMake, call since SDL2 is not Windows only platform. 😉

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,13 +73,11 @@ file (GLOB CXBXR_HEADER_COMMON
  "${CXBXR_ROOT_DIR}/src/common/CxbxDebugger.h"
  "${CXBXR_ROOT_DIR}/src/common/EmuEEPROM.h"
  "${CXBXR_ROOT_DIR}/src/common/Error.h"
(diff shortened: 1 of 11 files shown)


RadWolfie:
AppVeyor failed because of...
`C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.23.28105\include\experimental/filesystem(26,1): fatal error C1189: #error:  The <experimental/filesystem> header providing std::experimental::filesystem is deprecated by Microsoft and will be REMOVED. It is superseded by the C++17 <filesystem> header providing std::filesystem. You can define _SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING to acknowledge that you have received this warning.`
Which is not relative to this pull request change. Since other two CIs are passing, even though I did verified on my end, decision to merge is still safe to use.

PR #1736, opened 2019-09-25
Remove Any Experimental filesystem Linkage

RadWolfie:
Since AppVeyor CI's Visual Studio start making errors about experimental filesystem. Let's move on to use std filesystem than experimental version. After all, C++17 does have std filesystem implemented.

**EDIT:** And to replace `std::wstring_convert` to SDL's iconv library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,7 @@ file (GLOB CXBXR_HEADER_COMMON
  "${CXBXR_ROOT_DIR}/src/common/util/CPUID.h"
  "${CXBXR_ROOT_DIR}/src/common/util/crc32c.h"
  "${CXBXR_ROOT_DIR}/src/common/util/CxbxUtil.h"
+ "${CXBXR_ROOT_DIR}/src/common/util/strConverter.hpp"
  "${CXBXR_ROOT_DIR}/src/common/win32/AlignPosfix1.h"
  "${CXBXR_ROOT_DIR}/src/common/win32/AlignPrefix1.h"
  "${CXBXR_ROOT_DIR}/src/common/win32/EmuShared.h"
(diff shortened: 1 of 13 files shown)


RadWolfie:
Currently under investigation...

RadWolfie:
http://www.unicode.org/faq//utf_bom.html#utf16-11
Hmm... I'm not sure if xbox is using UCS-2LE or UTF-8LE (not UTF-8) for wide character string in xbe's structure. Although, both does return the string correctly.

LukeUsher:
> http://www.unicode.org/faq//utf_bom.html#utf16-11
> Hmm... I'm not sure if xbox is using UCS-2LE or UTF-8LE (not UTF-8) for wide character string in xbe's structure. Although, both does return the string correctly.

Xbox does not use UTF8-LE, it is using the exact same variant of UTF-16 as the standard Unicode Windows API on PC

RadWolfie:
Whoops, it's a typo. I meant UTF16LE (without BOM). We can't use UTF-16 by itself due to decision maker inside SDL_iconv. If Cxbx-Reloaded is compiled with different architecture which use big endian then reading xbe's UTF-16LE (without BOM) string will be corrupted.

gandalfthewhite19890404:
What is this - it caused by this PR?
image

RadWolfie:
The message is from `src/core/kernel/support/EmuFile.cpp`, since MSVC's experimental filesystem was forwarded to official C++17 filesystem.

@gandalfthewhite19890404 I'm not able to reproduce your issue. Can you create a new issue with steps to reproduce a bug you found? Use the latest develop commit to reproduce the issue instead of this pull request's commit.

LukeUsher:
I see that issue ocasionally: but it doesn't always have the same message.
It happens when std::filesystem is asked to delete a directory that is still in-use/locked.

RadWolfie:
Base on discussion with @LukeUsher, the issue did not come from this pull request. Most likely cause is from third-party utility system application cause the directory to lock, prevent deletion, to output the error message.

PR #1737, opened 2019-09-28
Add support for Xbox Controller S

LukeUsher:
This PR adds support for 'Xbox Controller S' to the input system.

The Xbox Controller S is functionally identical to the standard controller, it has all the same inputs, but the physical layout of the buttons is different: You can see the layouts represented in the XDK Controller Test sample screenshots below.

Device Selection Dialog:
image

XDK Controller Test (Controller S Connected):
image

XDK Controller Test (Controller 'Duke' Connected):
image

### DIFF
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -478,11 +478,19 @@ bool Settings::LoadConfig()
 
 	// ==== Input Profile Begin ====
 	std::array<std::vector<std::string>, to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX)> control_names;
-	for (int i = 0; i < dev_num_buttons[0]; i++) {
-		char control_name[30];
-		std::sprintf(control_name, sect_input_profiles.control, button_xbox_ctrl_names[i][0]);
-		control_names[0].push_back(control_name);
+
+	for (int device = 0; device < to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX); device++) {
+		if (dev_num_buttons[device] == 0) {
+			continue;
+		}
+
+		for (int i = 0; i < dev_num_buttons[device]; i++) {
+			char control_name[30];
+			std::sprintf(control_name, sect_input_profiles.control, button_xbox_ctrl_names[i][0]);
+			control_names[device].push_back(control_name);
+		}
 	}
+
 	// TODO: add the control names of the other devices
 
 	index = 0;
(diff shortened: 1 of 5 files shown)


src/common/input/InputWindow.cpp
```diff
@@ -39,7 +39,7 @@ constexpr ControlState INPUT_DETECT_THRESHOLD = 0.55; // arbitrary number, using
 InputWindow* g_InputWindow = nullptr;
 int dev_num_buttons[to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX)] = {
 	XBOX_CTRL_NUM_BUTTONS,
-	0,
+    XBOX_CTRL_NUM_BUTTONS,
```
LukeUsher: This change adds the button count to the 'S' device type

LukeUsher:
Note that the purpose of this task was mostly to familiarize myself with the input system, since I'll need to add support for Sega Chihiro parts soon.

src/common/Settings.cpp
```diff
@@ -477,13 +477,22 @@ bool Settings::LoadConfig()
...
+        std::sprintf(control_name, sect_input_profiles.control, button_xbox_ctrl_names[i][0]);
+        control_names[to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE)].push_back(control_name);
+    }
+
+    for (int i = 0; i < dev_num_buttons[to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S)]; i++) {
+        char control_name[30];
+        std::sprintf(control_name, sect_input_profiles.control, button_xbox_ctrl_names[i][0]);
+        control_names[to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S)].push_back(control_name);
+    }
+
```
ergo720: You can avoid adding a new loop for every device type by doing something like this:
```
	for (int j = to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE); j < to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX); j++) {
		if (dev_num_buttons[j] != 0) {
			for (int i = 0; i < dev_num_buttons[j]; i++) {
				char control_name[30];
				std::sprintf(control_name, sect_input_profiles.control, button_xbox_ctrl_names[i][0]);
				control_names[j].push_back(control_name);
			}
		}
	}
```

src/common/Settings.cpp
```diff
@@ -477,13 +477,22 @@ bool Settings::LoadConfig()
...
 	// TODO: add the control names of the other devices
 
+
```
ergo720: Was this line added intentionally?

src/common/Settings.cpp
LukeUsher: This is only an option when the device types we support are in sequential order: If we implement a later device type with a gap in the middle, it will fail, which is why I decided against a loop.

src/common/Settings.cpp
LukeUsher: probably not :P

ergo720:
Code indentation is inconsistent with the surrounding code (tabs/space problem?).

LukeUsher:
> Code indentation is inconsistent with the surrounding code (tabs/space problem?).

All new code uses spaces, but there is existing code in many places that still uses tabs

RadWolfie:
I do use spaces unless other function is using tabs, then I just convert my spaces to tabs. It's how I roll for now. 😉

LukeUsher:
Latest commit was a rebase + feedback addressing

src/common/Settings.cpp
```diff
@@ -442,72 +442,80 @@ bool Settings::LoadConfig()
...
+	std::array<std::vector<std::string>, to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX)> control_names;
 
-	index = 0;
-	while (true) {
-		std::string current_section = std::string(section_input_profiles) + std::to_string(index);
-		if (m_si.GetSectionSize(current_section.c_str()) == -1) {
-			break;
+	for (int device = 0; device < to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX); device++) {
+		if (dev_num_buttons[device] == 0) {
+			continue;
```
ergo720: It seems you have accepted my previous proposal, so the gap problem wasn't an issue?

LukeUsher: The check for dev_num_buttons == 0 prevents any problem occuring

Issue #193, opened 2017-02-24
HLE : Centrally convert D3D resources


PatrickvL:
The current method to associate D3D resources to emulated resources (or the other way around) sucks.
We'd be much better off with an unordered map, that associates Xbox D3D resources to the host counterpart.

Doesn't matter much which implementation we choose, let's first get the rendering correct. Perhaps std::map is already good enough.

The trick there would be to get a reference to all active resources.
D3D stores them in a structure somewhere, but it could have changed between xdk versions.
This needs to be investigated.

Each type requires a type a specific conversion.

We can call all required conversions from one central function, which just needs to be called from the actual drawing patches. Drawing patches are about the only D3D patches that we should keep. Most of the other D3D patches can be removed, especially those that don't access hardware and just call into our kernel. These can run just fine as they are, without a patch.

Before conversion, a fast hash can be calculated over the Xbox resource and associate data, to determine if a conversion is needed compared to the hash of the previous conversion.

To speed it up, we could skip the hash check when no changes are detected in a configurable number of prior checks. After a configurable number of skips, the hash must forcibly be checked, in order to detect sudden changes in a resource.

PatrickvL:
As a fast hash, xxxhash could be a well-suited candidate, I've changed to this with this commit : https://github.com/PatrickvL/Cxbx-Reloaded/commit/1585b75445f503682aa62760fd968c31ac4bfc7c

Please test before cherry-picking this!

PatrickvL:
Eventually most functions handling D3D resources must be left unpatched, so they will work just as the original. This requires a bit more work in the kernel. Especially MmAllocateContiguousMemory should allocate from the same address range as the Xbox does. Once that's realized, both HLE and LLE graphics emulation must take into account that each Data-pointer inside a D3D resource, is stored as a GPU-address, not a physical address. (Converting back and forth is just a matter of xor-ring with 0xF800000)

PatrickvL:
Another step into this direction is done with this commit (which also mentions the steps Dxbx did) : https://github.com/PatrickvL/Cxbx-Reloaded/commit/d326db68d5ca4cf41b8fcec5397a56f702c0dbfa.

Again: test well before cherry-picking this in!

PatrickvL:
We've reached a new milestone with this, now that we copy index buffers correctly when drawing from them.
Also, I'm working on the same approach for textures, which is also coming along nicely - stay tuned for more news on this subject!

[...]

PR #1742, opened 2019-10-03
Remove move RenderState/TextureState Patches

LukeUsher:
With this work, all of these patches are no longer required, and
instead, are replaced with a RenderStateConverter and
TextureStateConverter object

One single render state patch (D3DDevice_SetRenderState_Simple) is still
required, as that bypasses the D3D__RenderStates array in some XDK
versions.

The bulk of this is removing patches, and replacing the patching mechanism and ugly/awkward to maintain deferred state handling with nice RenderState/TextureState converter objects.

The result is the removal of 40 HLE patches.

NOTE: Draft status will be removed after regression testing, as this is a huge change.


PatrickvL:
Please put the removed patches in the file we keep around for that

src/core/hle/Patches.cpp
```diff
@@ -455,9 +415,9 @@ inline void EmuInstallPatch(std::string FunctionName, xbaddr FunctionAddr)
 		printf("HLE: %s: Skipped (LLE OHCI Enabled)\n", FunctionName.c_str());
 		return;
 	}
-
-    // HACK: Some titles require unpatched Fibers, otherwise they enter an infinte loop
-    // while others require patched Fibers, otherwise they outright crash
+
+    // HACK: Some titles require unpatched Fibers, otherwise they enter an infinte loop
```
Margen67: *infinite

gandalfthewhite19890404:
WWE RAW 2:
image

Margen67:
Instructions and characters now (sort of) render in Sneakers.
Master:
Master
PR:
PR

medievil1:
buffy the vampire slayer doesn't render at all (black screen) with video and game 3d audio in the background

PatrickvL:
Please also list the xdk version of apparent regressions, so we're able to compare against titles using the same xdk version

LukeUsher:
@gandalfthewhite19890404 please don't post screenshots without context. If the game improved, that's good, but it's not a regression.

If the game got worse, please explain how/what changed, with a comparison screenshot if possible.

Posting a screenshot with no context is unhelpful and it will probably be ignored.

medievil1:
previous comment updated with xdk version

gandalfthewhite19890404:
FIFA World Cup 2006 - XDK 5849:
image

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6791,7 +6791,15 @@ void CxbxUpdateNativeD3DResources()
 VOID __declspec(noinline) D3DDevice_SetPixelShaderCommon(DWORD Handle)
 {
     // Cache the active shader handle
-	g_D3DActivePixelShader = (XTL::X_PixelShader*)Handle;
+    g_D3DActivePixelShader = (XTL::X_PixelShader*)Handle;
+
+    // Copy the Pixel Shader data to our RenderState handler
+    // This mirrors the fact that unpathed SetPixelShader does the same thing!
```
RadWolfie: unpatched* 😉

jackchentwkh:
In Otogi, when you pressed A in menu, there will be a water wave effect. The effect is broken now. I guess it's a regression.

LukeUsher:
@jackchentwkh if it happens in the development build also, then it doesn't count as part of this PR
But it should be looked into as part of another task.

PatrickvL:
@gandalfthewhite19890404 Could you re-test with the latest build? (PS Your screenshots are somehow not showing, but their links do work.)

LukeUsher:
I went ahead and updated @gandalfthewhite19890404's post to fix the broken screenshot links.

gandalfthewhite19890404:
Tested with de2b891
- **FIFA World Cup 2006** <details> <summary>Fixed rainbow graphics </summary> 
image
</details>

- **Lord of the rings Third age** - now crashing just after start:
Received Exception Code 0xC0000005 @ EIP := 0x040CCFA9

- **The Elder Scrolls Morrowind** <details> <summary>Fixed rainbow water </summary>
image
</details>

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6819,7 +6820,7 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_SetPixelShader_0)
 	// Call the Xbox function to make sure D3D structures get set
 	XB_trampoline(VOID, WINAPI, D3DDevice_SetPixelShader_0, ());
     __asm {
-        push eax
+        mov eax, Handle
```
LukeUsher: This function (as documented in the patch comments above) accepts it's parameter in eax
For some reason, we were putting eax on the stack and passing it that way?

That was obviously never going to work:
1. The above call to XB_trampoline would have trashed eax
2. The function doesn't check the stack, the value contained within Handle should be placed in EAX before the call.

LukeUsher:
Draft status removed: I think this is pretty much ready.

Notable improvments:
Super Monkeyball now works in-game
Vex has improved
The Incredible Hulk has improved

Voxel9:
DolphinClassic is missing distance fog.

Master:
image

This PR:
image

LukeUsher:
@Voxel9 which version of DolphinClassic? Many XDKs have a DolphinClassic sample, so the XDK version you're using is important.

LukeUsher:
So it would appear that some LTCG games are experiencing new crash issues with this PR, however the behavior is weird.

Both Star Wars Battlefront II and Lord of the Rings: The Third Age crash on builds compiled with appveyor, but don't crash on builds I compile locally.

After investigating further, it appears that the issue is compiler version: Builds compiled with recent versions of Visual Studio 2019 will crash in this games, but earlier versions of 2019 and all versions of 2017 will not crash.

Compiler optimisation is probably fucking us over, needs further investigation.

LukeUsher:
Added a fix for the crash caused within the LTCG pixel shader patch.
A similar fix should be applied to other patches later (as a seperate PR), but this is enough to avoid the crash in LOTR: The Third Age.

A seperate PR should address the crashing because it's unrelated to this PR: The actual cause is a change in AppVeyor's compiler version.

gandalfthewhite19890404:
TES 3 Morrowind now shows dirty disc error after you start new game.

PatrickvL:
This PR addresses part of #193 (linking for future reference)

Issue #1744, opened 2019-10-06
Loader: Emulation Crashed in 2nd Reboot Process


RadWolfie:
I noticed there's a crash occur some time last week while working on loader branch with my demo disc. First reboot works fine as usual, however when perform a second reboot. It produced a crashed from `VMManager::SavePersistentMemory`. Then recently, I was testing dashboard launch to Xbox Live then back. Which also produce the same issue. I believe previous persistent memory were not cleared after first reboot.

Steps to reproduce:
1) Launch dashboard
2) Enter Xbox Live
3) Exit to dashboard (by press B)
4) Emulation crashed

![VMManager_SavePersistentMemory ](https://user-images.githubusercontent.com/3938312/66266650-72c09580-e7ed-11e9-9d17-6051186b6880.png)

PR #1745, opened 2019-10-10
Fix memory manager crashing upon second reboot

ergo720:
Fixes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1744

### DIFF
--- a/src/core/kernel/memory-manager/VMManager.h
+++ b/src/core/kernel/memory-manager/VMManager.h
@@ -84,16 +84,14 @@ typedef enum _MemoryRegionType
 	COUNTRegion,
 }MemoryRegionType;
 
-#pragma warning(push)
-#pragma warning(disable: 4200)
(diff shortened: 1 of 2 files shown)


src/core/kernel/memory-manager/VMManager.h
```diff
@@ -84,14 +84,16 @@ typedef enum _MemoryRegionType
 	COUNTRegion,
 }MemoryRegionType;
 
-
+#pragma warning(push)
+#pragma warning(disable: 4200)
```
RadWolfie: Let's use `#pragma warning(suppress: 4200)` instead. See @PatrickvL's changes for simplify method from the commit below.
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1731/commits/58f9033c6ad32489bfe5fd32420da55d4269062f

PR #1748, opened 2019-10-13
LOADER: Add Toggle Ability Experimental Loader Project

RadWolfie:
Currently only accessible via edit settings file. Since the value toggle is not intended for users to try out yet. Plus show read-only value from GUI end.

Tested and confirm working from my end for launch old vs new launcher.

### DIFF
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -89,6 +89,7 @@ static struct {
 	const char* AllowAdminPrivilege = "AllowAdminPrivilege";
 	const char* LoggedModules = "LoggedModules";
 	const char* LogLevel = "LogLevel";
+	const char* LoaderExperiment = "LoaderExperiment";
 } sect_core_keys;
 
 static const char* section_video = "video";
@@ -350,6 +351,8 @@ bool Settings::LoadConfig()
 		index++;
 	}
 
+	m_core.loaderExperiment = m_si.GetBoolValue(section_core, sect_core_keys.LoaderExperiment, /*Default=*/false);
+
 	// ==== Core End ============
 
 	// ==== Hack Begin ==========
@@ -524,6 +527,8 @@ bool Settings::Save(std::string file_path)
 		stream << "0x" << std::hex << m_core.LoggedModules[i];
 		m_si.SetValue(section_core, sect_core_keys.LoggedModules, stream.str().c_str(), nullptr, false);
 	}
+
+	m_si.SetBoolValue(section_core, sect_core_keys.LoaderExperiment, m_core.loaderExperiment, nullptr, true);
 
 	// ==== Core End ============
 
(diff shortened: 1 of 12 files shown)


src/common/Settings.cpp
```diff
@@ -89,6 +89,7 @@ static struct {
 	const char* AllowAdminPrivilege = "AllowAdminPrivilege";
 	const char* LoggedModules = "LoggedModules";
 	const char* LogLevel = "LogLevel";
+	const char* LoaderExperiment = "LoaderExperiment";
```
ergo720: You have changed the setting's structure with this addition. As such, `settings_version` should be updated then.

PatrickvL: Well, yes and no; yes, there's a new value, but no, it's not an extension - it's merely reusing a reserved field. So increasing the version number is not a hard requirement

RadWolfie: It's unnecessary for the following reasons:
1. Backward compatible since the structure size had not changed. See `src/common/Settings.hpp` change in this pull request, like @PatrickvL said.
2. Previous PR is no longer using global memory shared structure. See `sid`.
   - This make `settings_version` no longer necessary since global memory sharing used to include the setting's version.
3. Settings file doesn't have a version stored.

ergo720: >Settings file doesn't have a version stored.

It has it, the `Revision` key which is also used by one of the Setting class functions (`RemoveLegacyConfigs`), so please don't delete it.

RadWolfie: > It has it, the `Revision` key which is also used by one of the Setting class functions (`RemoveLegacyConfigs`), so please don't delete it.

Ah okay, then perhaps we should rename `settings_version` to `settings_revision` in the later pull request. 😉 I can understand the reason to have revision in the settings file for general purpose.

src/common/win32/EmuShared.h
```diff
@@ -191,7 +191,7 @@ class EmuShared : public Mutex
 		// ******************************************************************
 		void GetDebuggingFlag(bool *value) { Lock(); *value = m_bDebugging; Unlock(); }
 		void SetDebuggingFlag(const bool *value) { Lock(); m_bDebugging = *value; Unlock(); }
-#ifndef CXBX_LOADER
+#ifndef CXBX_LOADER // Temporary usage for cxbx.exe's emu
```
PatrickvL: CXBX_LOADER is replaced with CXBXR_EMU in other files in this PR - Was this occurrence left here intentionally?

RadWolfie: In order to keep structure align between cxbx vs loader emulation, it can't be rename to CXBXR_EMU. It is left intentionally, although it seems unnecessary unless it has any relative to memory pages of what @ergo720 rely on to restore memory.

PR #1750, opened 2019-10-14
Restore fog hack: Fixes Silent Hill 2

LukeUsher:
This restores a hack that was accidentally removed during the work on #1742.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/RenderStates.cpp
+++ b/src/core/hle/D3D8/Direct3D9/RenderStates.cpp
@@ -311,6 +311,15 @@ void XboxRenderStateConverter::ApplyDeferredRenderState(uint32_t State, uint32_t
             }
         } break;
         case XTL::X_D3DRS_FOGENABLE:
+            if (g_LibVersion_D3D8 == 3925) {
+                // HACK: Many 3925 games only show a black screen if fog is enabled
+                // Initially, this was thought to be bad offsets, but it has been verified to be correct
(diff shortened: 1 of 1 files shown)


PatrickvL:
This could also fix the reported Halo regression

src/core/hle/D3D8/Direct3D9/RenderStates.cpp
```diff
@@ -311,6 +311,15 @@ void XboxRenderStateConverter::ApplyDeferredRenderState(uint32_t State, uint32_t
             }
         } break;
         case XTL::X_D3DRS_FOGENABLE:
+            if (g_LibVersion_D3D8 == 3925) {
+                // HACK: Many 3925 games only show a black screen if fog is enabled
+                // Initially, this was thought to be bad offsets, but it has been verified to be correct
+                // Unitl we find out the true underlying cause, disable fog and carry on
```
PatrickvL: s/Unitl/Until

jackchentwkh:
why not make it a option/hack switch?

LukeUsher:
> how about making it a option/hack switch?

This is not a new hack, it's just restoring previous behavior: It was accidentally removed during other work, and wasn't caught during testing

LukeUsher:
> One typo...

Even the typo was an existing problem, this code is not new, it's just a restoration of what was accidentally removed during renderstate cleanup.

That being said, i did fix it

LukeUsher:
> This could also fix the reported Halo regression

it does, it's the same issue.

PR #1752, opened 2019-10-16
LOADER: Scrap Win32 GUI Draft

RadWolfie:
As of #1748 pull request, new draft GUI (Win32) project is no longer necessary.

Plus modify Cxbx's CMakeList project file to exclude emulator can be done to verify any emulation codebase no longer reside in existing GUI's codebase.

---

After this merge, we should fix any remaining bugs in the progress before merge into develop branch.


PatrickvL:
Not sure this is wise. I prepared an empty project, so that all GUI-related code could be appended to it, such seemed easier than to sort out which non-GUI code could be removed from the existing Cxbx project.

PatrickvL:
Also, I had planned to keep both the single-executable-based version intact, alongside the setup that's composed of separate GUI+loader+emulation DLL

PatrickvL:
That way, would problems arise due to the separated design, we'd still have the all-in-one project that can be used

PatrickvL:
Once we're convinced the new setup is reliable (and hopefully even better than the single exe), switching over would merely require deletion of the old project - the bare-bones GUI would already have been prepared

PatrickvL:
This bare-bones GUI is nothing new, it's just the GUI-related sub-selection of the source code files we already have, put in a dedicated project for just the GUI purpose.
It's not intended as a rewrite of the GUI. That's another task, for another day.

PatrickvL:
Having a dedicated GUI project, cleaned from all emulation related code, would also make it easier to see what a rewritten GUI could be needing, since emulation-related code wouldn't be a part of the dedicated GUI anymore.

RadWolfie:
> That way, would problems arise due to the separated design, we'd still have the all-in-one project that can be used

Solution: (not accessible from GUI end to prevent end users from using experimental project)
image

> This bare-bones GUI is nothing new, it's just the GUI-related sub-selection of the source code files we already have, put in a dedicated project for just the GUI purpose.
> It's not intended as a rewrite of the GUI. That's another task, for another day.

Well, it is a re-write anyway since the resource file is created from scratch plus the window messages are not handled. Plus `cxbxr.exe` is reserved for `Qt`'s GUI.

> Having a dedicated GUI project, cleaned from all emulation related code, would also make it easier to see what a rewritten GUI could be needing, since emulation-related code wouldn't be a part of the dedicated GUI anymore.

As I was saying... remove `${CXBXR_xxxxx_EMU}` from CMakeLists.txt from Cxbx's project file will tell us what left to do after attempt to compile. See below:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/9b87f6c88dadbaa69c4f01770afcaaa6f97b950c/projects/cxbx/CMakeLists.txt#L83-L105

PatrickvL:
Although it's nice to be able to enable a hidden setting and let that run the loader, I'm talking about a phase beyond that: Once we switch over by default to the 3 separate projects (including a cut-down version of the current GUI), it's still wise to have the all-in-one project available as a last resort, for when unforseen issues suddenly arise.

PatrickvL:
The cut-down GUI is not a rewrite. It's a subset of the current GUI source code (which currently still includes emulation) files. The current .RC resource file can be reused or copied over to the cut-down GUI project as well.

PatrickvL:
Reserved executable file names can always be reconsidered.

PatrickvL:
I fail to see how handling window messages would be any different in the stripped-down GUI?

RadWolfie:
> Although it's nice to be able to enable a hidden setting and let that run the loader, I'm talking about a phase beyond that: Once we switch over by default to the 3 separate projects (including a cut-down version of the current GUI), it's still wise to have the all-in-one project available as a last resort, for when unforseen issues suddenly arise.

Here's a counter offer, create a separate project which only handle emulation executable (not dll) itself without any GUI codebase. Of course, exactly like virtual memory placeholder you had done in the current codebase. Since there's not much need for all-in-one codebase in one project and it's holding us back to create a brand new GUI plus ability to work on 64-bit executable.

P.S. Do you think we can use Qt's GUI codebase in the all-in-one project? I think not since it will increase chance of regression. 😛

PatrickvL:
We probably all agree that maintenance on the current GUI should be kept to a minimum.
And we share the same vision of having a new cross-platform GUI, that won't contain emulation code.

Still, as much as I hope that the new (tiny loader + emulation DLL) approach will be at least as good as the current approach (of an all-in-one GUI+emulation executable), there's still the risk that it won't work well initially.
Therefore, we probably all agree that it's best to keep the all-in-one approach around for a while.

BUT, having this as a fallback doesn't mean we have to keep using it!

Until we have a new GUI, I'd rather use a light-weight subset of the current GUI as front-end.

Such a temporary front-end would compile much faster (since it won't contain emulation code), it's executable would be much smaller (without the 128 MB placeholder) and take less system resources during use.

Of course, a new cross-platform GUI would replace this light-weight sub-set of the current GUI.
But in the mean time, having such a GUI would be preferable over using the all-in-one GUI we have right now.

RadWolfie:
> Still, as much as I hope that the new (tiny loader + emulation DLL) approach will be at least as good as the current approach (of an all-in-one GUI+emulation executable), there's still the risk that it won't work well initially.
> Therefore, we probably all agree that it's best to keep the all-in-one approach around for a while.

If tiny loader + emulation DLL doesn't work out, it will still be in separate executable (which is cxbx.exe without GUI codebase) than all-in-one. The best time to have emulation code in GUI is virtual emulation support. Having both GUI and emulation in host memory is not a great idea anyway.

> BUT, having this as a fallback doesn't mean we have to keep using it!

I don't plan on maintenance the current, "Win32", GUI codebase any time in the future once new GUI is fully done. New GUI should already have IPC both way and no longer use EmuShare data between GUI and emulation.

> Until we have a new GUI, I'd rather use a light-weight subset of the current GUI as front-end.
>
> Such a temporary front-end would compile much faster (since it won't contain emulation code), it's executable would be much smaller (without the 128 MB placeholder) and take less system resources during use.

If you're using tiny loader method, then just compile Cxbx project once and unload the project. All you have to do is just compile the emulation project then do your thing. Until a new setting had been added which of course shouldn't be affected long as the EmuShared data doesn't have a different size. If EmuShared do have a different size, then emulation will reject and exit.

There is still third-party front-end you can also use or even use command prompt (can use batch script) to run standalone emulation which doesn't rely on our own GUI. YOU can still edit the settings file any time you want. Run standalone emulation once will insert new setting parameter in the settings file if there is one.

> Of course, a new cross-platform GUI would replace this light-weight sub-set of the current GUI.
> But in the mean time, having such a GUI would be preferable over using the all-in-one GUI we have right now.

I rather strip the whole current GUI out. Working on current and "new" sub-set, Win32, GUI is not going to work out for me at least. Since I rather not to work on 2 separate GUIs to ensure they're up to date. Even if I did add one to the current GUI by the way.

---

TLDR:
I don't have a problem compiling "all-in-one", tiny loader, and emulation projects at the same time. The only time one of the project get a re-compile task if the source files has been changed. And that is much shorter time to compile. The header files are different topic which should be only used in certain source files than in the header files to reduce most of the re-compile process.

PR #1755, opened 2019-10-19
Dump vertex declaration separately from conversion

PatrickvL:
This PR extracts the code that dumps a vertex declaration from the code that converts it to host, so that the dumping and the conversion are two independent parts.


medievil1:
found no issues with this yet...seems to be working well

PatrickvL:
If there's no objection to this, it'd be good to have this merged soon

PR #1756, opened 2019-10-20
Use a file to store persistent memory across quick reboots

ergo720:
Currently, the loader uses shared memory to persist the pages across a quick reboot. @RadWolfie  has brought to my attention that this approach will not work when using a third-party GUI (so only when the cxbxr process is running). This fixes this problem by saving the pages on a disk `memory.bin` file.
@RadWolfie I also included your commit `20b301954378c9aeb1c020fd1172b7eaa5237253` because, without it, it triggers the error "Data path directory is currently in used. Use different data path directory or stop emulation from another process.", which is a bug introduced in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1741 and your commit fixes it.

### DIFF
--- a/src/common/IPCHybrid.hpp
+++ b/src/common/IPCHybrid.hpp
@@ -38,7 +38,6 @@ typedef enum class _IPC_UPDATE_GUI {
 	, XBOX_LED_COLOUR
 	, LOG_ENABLED
 	, KRNL_IS_READY
-	, VM_PERSIST_MEM
 } IPC_UPDATE_GUI;
 
 void ipc_send_gui_update(IPC_UPDATE_GUI command, const unsigned int value);
--- a/src/common/win32/IPCWindows.cpp
+++ b/src/common/win32/IPCWindows.cpp
@@ -65,10 +65,6 @@ void ipc_send_gui_update(IPC_UPDATE_GUI command, const unsigned int value)
 			cmdParam = ID_GUI_STATUS_KRNL_IS_READY;
 			break;
 
-		case IPC_UPDATE_GUI::VM_PERSIST_MEM:
-			cmdParam = ID_GUI_VM_PERSIST_MEM;
-			break;
-
 		default:
 			cmdParam = 0;
 			break;
--- a/src/core/kernel/exports/EmuKrnlHal.cpp
+++ b/src/core/kernel/exports/EmuKrnlHal.cpp
@@ -577,6 +577,26 @@ XBSYSAPI EXPORTNUM(49) xboxkrnl::VOID DECLSPEC_NORETURN NTAPI xboxkrnl::HalRetur
 				if (!CxbxExec(false, nullptr, false)) {
 					CxbxKrnlCleanup("Could not launch %s", XbePath.c_str());
 				}
+
+				// This is a requirement to have shared memory buffers remain alive and transfer to new emulation process.
+				unsigned int retryAttempt = 0;
+				unsigned int curProcID = 0;
+				unsigned int oldProcID = GetCurrentProcessId();
+				while(true) {
+					std::this_thread::sleep_for(std::chrono::milliseconds(100));
+					g_EmuShared->GetKrnlProcID(&curProcID);
+					// Break when new emulation process has take over.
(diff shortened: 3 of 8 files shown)


PatrickvL:
Looking good guys!

RadWolfie:
Here's the pro/con for using file method:

**PRO**
- good for cross-platform by using std library's file stream
- file always exist after reboot (even after a crash)

**CON**
- Not good for active multi-emulation
  - **NOTE:** Loader branch at the moment has support for multi-emulation even from same executable.
- Take time to save the file to disk.

PatrickvL:
As a bit of background for the closing of this PR, here an excerpt of the discussion we had on this on discord:

> ergo720 10/20/2019
> From your original comment here https://discordapp.com/channels/@me/434091123183583235/633020115348357140, it seems to me that my shared memory approach wouldn’t work with a 3rd party GUI (environment ?), which makes sense, hence I changed it to a file based approach. Now you are saying that it works and it’s better to use the shared memory, hence why I’m confused. This confusion probably arises because I don’t understand how a 3rd party GUI is supposed to access the emu process. Is it supposed to interact with EmuShared? Windows only? Which language? Also in your review comment on the PR “Not good for active multi-emulation NOTE: Loader branch at the moment has support for multi-emulation even from same executable.” To be honest, I have no idea what this means.
> 
> Patrickv L10/20/2019
> I think radwolfie used the term 'multi-emulation' when indicating concurrently running multiple emulations side by side on the same machine, perhaps with network play in mind
> As such that could be done with file persistation as well, just as long as the file name used the recently introduced SessionID in its filename
>  
> PatrickvL 10/20/2019
> From the code I've seen, it appears that shared memory (as allocated in EmuShared) and memory mapped files (marked as temporary, by means of a delete-on-close creation flag) would be lost if they're released by an emulation process before the follow-up process got a handle to those resources again.
> Until thus far, this "stay-alive-until-the-resources-are-reopened" purpose, was handled by sending a signal to the main GUI process
> As far as I understand, Radwolfie offers a solution for this without the need for a signal back to the GUI. So having less dependency is favourable.
> Now that this purpose can be reached without the signal back to the GUI, it might be possible to keep using a memory mapped file for transfer of persistent memory to the follow-up process. Am I interpreting this correctly thus far, @RadWolfie ?
> So that'd remove the immediate necessity to change the way persistent memory is transferred to the follow-up process - it could keep using the memory mapped file approach
> It's not portable to other operating systems than Windows'es, but that's a rather moot point right now, given that we've got much more code that depends on the Windows API's anyway.
> I personally do like the fact that memory mapped files wouldn't need actual I/O when accessed from the follow-up proces (since the data is most likely still present in the Windows file cache anyway). But truth be told, reading back data from a normal file will be fetched from the same Windows file cache as well, so that's not much of a difference
> One benefit of using a delete-on-close file (memory mapped or not), is that it won't accidentally stay behind (as that would pollute the file system).
> But as I said, that aspect isn't dependent on memory mapped file versus regular file, both can use the same creation flag for that.
> I can see one advantage of using memory mapped files though: once mapped onto a view, the data is there, nothing else needs to be done to access the data.
> While reading from a regular file requires allocation of a buffer and a copy from file cache to that buffer, so it's a bit more expensive in that regard
>  
> RadWolfie 10/20/2019
> @PatrickvL That is correct, without the commit 20b3019. Then first emulation process got terminated too early for both GUI and standalone side. EmuShared remained open got luckly tbh. I didn't catch that bug.
>  
> PatrickvL 10/20/2019
> Suspending proces termination until the kernel ProcessID gets changed in EmuShared, is a great idea indeed
> Not sure if 5 seconds will suffice, but that can be tuned
>  
> RadWolfie 10/20/2019
> That is the fix. :wink: If 5 seconds isn't enough we can increase it any time in the future. :smiley:
>  
> PatrickvL 10/20/2019
> But now the point of discussion: Shall we keep using a memory mapped file to transfer persistent memory through to the follow-up process, or shall we switch to a regular file?
> I feel somewhat more at ease with the memory mapped file approach, but I do appreciate that using regular file makes at least that (small) part of our code portable
>

PR #1759, opened 2019-10-25
LOADER: Fix Shared Memory Reboot Loss

RadWolfie:
Previously, there's a chance of memory shared data didn't get transferred over to new emulation (standalone). With this pull request, previous emulation process will wait, at least 5 seconds, until new emulation has launch, open shared memory, then take over by update kernel process id. Finally, previous emulation process will terminate itself or after 5 seconds timeout.


PatrickvL:
LGTM!

PatrickvL:
Perhaps next time we work in `CxbxKrnlEmulate`, we could split the various steps into separate functions. Since movements like this cause entire blocks of code being deleted and re-added elsewhere, while moving function calls around would alter only a few lines.

PR #1761, opened 2019-10-27
Disable LLE GPU option in the gui

ergo720:
Disables the LLE GPU option in the GUI, so that end users cannot enable it any longer (it's too slow to play games at acceptable speeds any way). It can still be enabled in the settings file, by writing 0x2 to the variable "FlagsLLE", in the case developers and/or power users need it for testing purposes.


PatrickvL:
Now that we have multiple settings unavailable in the GUI, but still configurable through the settings file, it's probably wise to document those in our wiki (for as far as they aren't already documented?)

LGTM otherwise.

RinMaru:
shouldnt it be removed from the GUI as an entry aswell?

PR #1765, opened 2019-11-01
Fix sdl controller input bindings

ergo720:
Some users reported that the sdl controller axes are bound erroneously in the input gui. I managed to trace back the original broken commit to PR #1721. This fixes the problem by reverting most of those changes and using an alternative implementation to fix the SPACE/ENTER problem mentioned in that PR.


RadWolfie:
Well, space key is working fine. Except enter key isn't accepting as input. Even though enter key did trigger to listen for inputs. I'll create a new issue for this.

PR #1767, opened 2019-11-04
Fixed CMake designer settings for debugger gui

### LINKED ISSUE #1758: Corrupt cmake settings for cxbxr-debugger project
I've synced to c0cdb01 and opened up the **cxbxr-debugger** project.

Saving any changes made to the Form1 winforms designer will trash the entire `Form1.designer.cs` file and remove it from the solution. The designer tab will just show a blank winform and break the project.

I'm clueless about cmake, but it looks like the wrong type is assigned to the designer.cs file. It's being treated as a "strongly-typed resource class"?

x1nixmzeng:
Fixes #1758

### DIFF
--- a/src/CxbxDebugger/CMakeLists.txt
+++ b/src/CxbxDebugger/CMakeLists.txt
@@ -95,12 +95,8 @@ file (GLOB SOURCES
 csharp_set_windows_forms_properties(
  "${CXBXR_DEBUGGER_SRC_DIR}/Form1.Designer.cs"
  "${CXBXR_DEBUGGER_SRC_DIR}/Form1.cs"
+ "${CXBXR_DEBUGGER_SRC_DIR}/Form1.resx"
  "${CXBXR_DEBUGGER_SRC_DIR}/Properties/AssemblyInfo.cs"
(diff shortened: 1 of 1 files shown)


x1nixmzeng:
@RadWolfie the old project was for VS2015, VS2019 may handle these files differently. Doing a resave doesn't resave the project with any major differences and there are no warnings about broken project files.

I've also gone over the documentation a few times and this is correct https://cmake.org/cmake/help/v3.8/module/CSharpUtilities.html

RadWolfie:
How is it correct? You have to use one of two options for working on a UI layout per above's link documentation. Which is `csharp_set_windows_forms_properties` or `csharp_set_designer_cs_properties`.

See both examples in repo: https://github.com/bemehiser/cmake_csharp_example
And answer: https://stackoverflow.com/a/46247927

Apparently, I was wrong. It should be using `csharp_set_windows_forms_properties`.

P.S. Visual Studio 2017 is still active support.

RadWolfie:
Hmm... I obviously didn't look higher up, my mind was looking at different function which is "include files".

PR #1769, opened 2019-11-05
Support Xbox Direct3D "read" callback events

PatrickvL:
<details>
 <summary>Previously, Xbox Direct3D "read" callback events where not processed</summary>

As can be seen in the Notifier XDK sample by the PBCount counter staying at zero :
image
</details>

<details>
 <summary>This pull request improves this situation.</summary>

As can be seen in the Notifier XDK sample by the increasing PBCount counter :
image
</details>

This PR also introduces synchronization of callback events with host GPU.

Since read-callbacks where previously ignored, this PR now executes callback code that wasn't hit before. This could lead to improvements, but also regressions (because this 'new' code might hit cases that we don't support well enough yet).

If testing shows more regressions than improvements, it may be wise to alter this PR to disable read-callbacks for the time being.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -6877,6 +6877,50 @@ void CxbxUpdateNativeD3DResources()
 */
 }
 
+// This function mimicks NV2A software callback events.
+// Normally, these would be handled by actual push-buffer
+// command handling at the point where they where inserted.
+// Since our HLE mostly circumvents the NV2A pushbuffer,
+// this function has to be called after 'pushing' functions.
+void CxbxHandleXboxCallbacks()
+{
+	// Execute callback procedure
+	if (g_CallbackType == XTL::X_D3DCALLBACK_WRITE) {
+		if (g_pCallback) {
+			g_pCallback(g_CallbackParam);
+			// TODO: Reset pointer?
+		}
+	}
+}
+
+// On Xbox, this function inserts push-buffer commands that
+// will trigger the software handler to perform the callback
+// when the GPU processes these commands.
+// The type X_D3DCALLBACK_WRITE callbacks are prefixed with an
+// wait-for-idle command, but otherwise they're identical.
+// (Software handlers are triggered on NV2A via NV097_NO_OPERATION) 
+void CxbxImpl_InsertCallback
+(
+	XTL::X_D3DCALLBACKTYPE	Type,
+	XTL::X_D3DCALLBACK		pCallback,
+	XTL::DWORD				Context
+)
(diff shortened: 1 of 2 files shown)


PatrickvL:
I've update PR title and description to mention "Xbox Direct3D" instead of "NV2A"

medievil1:
buffy and indiana jones both crash (no error, just goes back to cxbx app screen) after playing opening videos

dreimer1986:
Halo 1 and 2 seem to do the same. PAL and NTSC

PatrickvL:
I was afraid this was going to happen. These crashes are likely the reason we skipped Read callbacks. Like I said it's easy enough to keep skipping Read callbacks, but it'd be better to know WHY these crashes occur...

medievil1:
maybe add a log with all the info so we can compare it?, right now it just returns to cxbx.. this is the last part of the log for indiana jones when logging d3d8
`[0x4588] DEBUG: D3D8    RenderState_Simple: D3DRS_ALPHABLENDENABLE = 0x00000001
[0x4588] D3D8    D3DDevice_SetTexture(
   Stage                : 0
   pTexture             : 0x00000000
);
[0x450C] D3D8    D3D_BlockOnTime(
   Unknown1             : 5
   Unknown2             : 2
);
[0x450C] D3D8    Direct3D_CreateDevice(
   Adapter              : 0x00000000
   DeviceType           : (D3DDEVTYPE)0x00000001 = D3DDEVTYPE_HAL
   hFocusWindow         : 0x00000000
   BehaviorFlags        : 50
   pPresentationParameters : 09EBFA90
   ppReturnedDeviceInterface : 0x00717660
);
[0x4914] DEBUG: D3D8    CreateDevice proxy thread received request.
[0x4914] DEBUG: D3D8    CreateDevice proxy thread releasing old Device.
`

PatrickvL:
I've disabled Read callbacks again for now. However, this PR still holds value since it syncs better with host GPU, and handles Write callbacks better (don't trigger them repeatedly, but do handle multiple requests).

LukeUsher:
I'm going to spend some time attempting to debug the crash issues above, so holding off merging for now, thanks for understanding

LukeUsher:
@PatrickvL I fixed the crash with Halo: It was unrelated to Read callbacks so they can be enabled again.
The problem was a double-free that only occured in Release builds due to a compile-time if statement.

medievil1:
confirmed it fixed indiana jones and buffy as well, although, at least in the case of buffy, some textures on the character (the exploded stuff) are now black

medievil1:
the black texture thing, after reviewing some old pic's might not be such a new thing, she has black pants on so I think it just might be that particular texture looking darker cause it is stretched out

PR #1772, opened 2019-11-09
Fix IP network traffic

LukeUsher:
image

DNS queries and other IP network traffic is now working within Cxbx-R.

This was achieved by means of a hack to allow handles created with the Nt Event APIs to function properly when passed to Ke APIs.

The Xbox network drivers depend on this functionality.

This also has potential to solve crash/hang issues in other games too.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlKe.cpp
+++ b/src/core/kernel/exports/EmuKrnlKe.cpp
@@ -746,6 +746,14 @@ XBSYSAPI EXPORTNUM(108) xboxkrnl::VOID NTAPI xboxkrnl::KeInitializeEvent
 		LOG_FUNC_ARG(SignalState)
 		LOG_FUNC_END;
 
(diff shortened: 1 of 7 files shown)


PatrickvL:
Could the call to `GetHandleInformation` be wrapped in a find called `IsWindowsHandle`, and then called before instead of after all
`KiLockDispatcherDatabase(&OldIrql)` calls?

LukeUsher:
@PatrickvL: Doing the action before the LockDIspatcher calls breaks the IP stack: the locks (and unlocks) are required. 

Splitting into a IsWindowsHandle function is still a viable option though.

PatrickvL:
If the problem with an `IsWindowsHandle` (or perhaps better named `IsHostHandle`) helper function can't be fixed, please create an issue for that - this PR is ready for merge as far as I'm concerned. Some testing may be needed?

PatrickvL:
MechAssault is reported (by Strelok) to be broken by this PR

PR #1774, opened 2019-11-12
Truncate input to mova

NZJenkins:
Make sure we truncate the contents of the 'a' register, so it isn't rounded to the nearest integer.

This fixes Buffy the Vampire Slayer, which was scaling it's indexes back from the 0-1 range using the formula:
`index = x * 256 + 0.5`
which would round up, when it should always round down

<details>

<summary>Before/After</summary>
Before

image

After

image
</details>

Thanks medievil for tracking down the issue and @PatrickvL for the suggested regex approach


LukeUsher:
nicely done!

PR #1776, opened 2019-11-14
Update expp instruction handling

NZJenkins:
Update handling of the Xbox `expp` instruction, by doing calculations for various components.
Before, `frc` was used, but it's valid for the `y` component only

Seems we want the `vs_1_1` version of this instruction:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/expp---vs

This improves Mechassault, which uses
`expp r1.z, r7.w`

<details>

<summary>Before / After</summary>
Before:

frc

After:
exp

</details>

TODO
* ~This implementation (and the `mova` post process) currently blows away whatever is in `r31`!~
* Test the `y` case (Tony Hawk Pro Skater 2x)
* ~`w` is unimplemented (wasn't sure how to proceed without setting a constant somewhere?)~
* Find test cases for the `x`, `w` and no swizzle cases


PatrickvL:
Here an attempt that you could cherry-pick : https://github.com/PatrickvL/Cxbx-Reloaded/commit/0598ca93a9fbd43917559a034d07cb62a3a55a0c

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -2563,22 +2532,85 @@ D3DVERTEXELEMENT *EmuRecompileVshDeclaration
...
+std::string UsingScratch(std::string input) {
+	return std::regex_replace(input, std::regex("tmp"), "r" + std::to_string(X_VSH_TEMP_SCRATCH));
+}
+
+// Xbox expp seems to behave the as vs_1_1
+std::string VshPostProcess_Expp(std::string shader) {
+	// Find usages of exp with each swizzle
+	// If there's no swizzle, we should still match so we do the calculation
+	// for all components
+	const auto exp_x = std::regex("expp (\\w\\d\\d?)(\\.x)?, (.+)$");
```
PatrickvL: Perhaps rename these to Xbox_`EXP_[XYZW]` and the host versions to `Host_EXP_[XYZW]`?

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -2563,22 +2532,85 @@ D3DVERTEXELEMENT *EmuRecompileVshDeclaration
     return pHostVertexElements;
 }
 
+std::string UsingScratch(std::string input) {
```
PatrickvL: This might as well be dropped, when all `tmp` occurrences would be replaced by `r13` (or a const static string variable that's defined as `"r" + std::to_string(X_VSH_TEMP_SCRATCH)`).

LukeUsher:
Did anyone confirm that THPS2X still works?
It should, but it's good to be safe.

PatrickvL:
Player model in-game in Tony Hawk Pro Skater 2X looks fine too : 
image

Merging

PR #1781, opened 2019-11-16
Turn Cxbx version define into a variable

PatrickvL:
This is intended to reduce the number of files that need to be
recompiled when updating a repository.


src/CxbxVersion.cpp
```diff
@@ -0,0 +1,44 @@
...
+// *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+// *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// *  GNU General Public License for more details.
+// *
+// *  You should have recieved a copy of the GNU General Public License
+// *  along with this program; see the file COPYING.
+// *  If not, write to the Free Software Foundation, Inc.,
+// *  59 Temple Place - Suite 330, Bostom, MA 02111-1307, USA.
+// *
+// *  (c) 2017 Patrick van Logchem <pvanlogchem@gmail.com>
```
ergo720: nit: the date is incorrect, should be 2019

src/CxbxVersion.cpp
```diff
@@ -0,0 +1,35 @@
+// This is an open source non-commercial project. Dear PVS-Studio, please check it.
+// PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com
+// ******************************************************************
+// *
+// *  This file is part of the Cxbx project.
```
RadWolfie: Why license this? There's no functions or unique usage for specifically Cxbx-Reloaded projects.
HINT: `CxbxVersion.h` file doesn't have a license for the same reason.

src/CxbxVersion.h
```diff
@@ -2,9 +2,4 @@
 
 #include "version.h"
 
-/*! version string dependent on trace flag */
-#ifndef _DEBUG_TRACE
-#define _CXBX_VERSION _GIT_VERSION " (" __DATE__  ")"
-#else
-#define _CXBX_VERSION _GIT_VERSION "-Trace (" __DATE__  ")"
-#endif
+extern char * CxbxVersionStr;
```
RadWolfie: Missing new line

src/common/xbe/XbePrinter.cpp
```diff
@@ -170,7 +170,9 @@ std::string XbePrinter::GameRatingToString()
 std::string XbePrinter::GenDumpHeader()
 {
     std::string text;
-    text.append("XBE information generated by Cxbx-Reloaded (Version " _CXBX_VERSION ")\n\n");
+	text.append("XBE information generated by Cxbx-Reloaded (Version ");
```
RadWolfie: Indent mistakes?

src/CxbxVersion.cpp
```diff
@@ -0,0 +1,35 @@
...
+// *
+// *  All rights reserved
+// *
+// ******************************************************************
+#include "Version.h"
+#include "CxbxVersion.h"
+
+/*! version string dependent on trace flag */
+#ifndef _DEBUG_TRACE
+char * CxbxVersionStr = _GIT_VERSION " (" __DATE__  ")";
```
RadWolfie: Unnecessary an extra space before asterisk. Same for below and header file.

RadWolfie: Also it should be `const char*` since variable is not suppose to be writable.

RadWolfie:
After the remarks are done, thanks for fix this outstanding issue of take longer time to re-compile once git hash is updated. 😅

CMakeLists.txt
```diff
@@ -226,6 +226,7 @@ file (GLOB CXBXR_SOURCE_COMMON
  "${CXBXR_ROOT_DIR}/src/common/xdvdfs-tools/buffered_io.cpp"
  "${CXBXR_ROOT_DIR}/src/common/xdvdfs-tools/xdvdfs.cpp"
  "${CXBXR_ROOT_DIR}/src/gui/DbgConsole.cpp"
+ "${CXBXR_ROOT_DIR}/src/CxbxVersion.cpp"
```
RadWolfie: I recently notice this wrong placement (in the order), we're using case insensitive order. It should be above `"${CXBXR_ROOT_DIR}/src/gui/DbgConsole.cpp"`, below `"${CXBXR_ROOT_DIR}/src/common/xdvdfs-tools/xdvdfs.cpp"`, line.

src/CxbxVersion.cpp
```diff
@@ -0,0 +1,11 @@
...
+// PVS-Studio Static Code Analyzer for C, C++ and C#: http://www.viva64.com
+#include "Version.h"
+#include "CxbxVersion.h"
+
+/*! version string dependent on trace flag */
+#ifndef _DEBUG_TRACE
+const char* CxbxVersionStr = _GIT_VERSION " (" __DATE__  ")";
+#else
+const char* CxbxVersionStr = _GIT_VERSION "-Trace (" __DATE__  ")";
+#endif
```
RadWolfie: missing newline

RadWolfie:
LGTM

There's one side effect in this pull request. When you open an executable, it does not show the build hash/date until load an xbe. Is this intentional?

src/gui/WndMain.cpp
```diff
@@ -173,7 +173,7 @@ WndMain::WndMain(HINSTANCE x_hInstance) :
 	// initialize members
 	{
 		m_classname = "WndMain";
-		m_wndname   = "Cxbx-Reloaded " _CXBX_VERSION;
+		m_wndname   = "Cxbx-Reloaded";
```
PatrickvL: This change results in initially not-showing the version, until an xbe is loaded.

If we want to see the version information before that as well, we need to put this caption in a temporary string variable, and use that.

I didn't do that, because I was being lazy.

RadWolfie: I have a solution for this issue. Replace m_wndname's `const char*` to `std::string`. I only see three references of `m_wndname` so far.

Though, we should drop char array and use `std::string` or the like instead.

PatrickvL: Done

PatrickvL:
@RadWolfie The side-effect has been addressed, and Luke approved the PR. Are you okay to merge this?

PR #1782, opened 2019-11-16
Some more cleanup

PatrickvL:
Various non-functional changes:

Make resource file language-neutral, so that Visual Studio won't complain when regional settings aren't compatible. (See https://stackoverflow.com/a/58011025/12170)

Also applied the advisory steps from our GPL license to our About dialog, and trimmed this advise from COPYING itself.


resource/Cxbx.rc
```diff
@@ -13,11 +13,11 @@
...
 
 /////////////////////////////////////////////////////////////////////////////
-// English (United States) resources
+// Language neutral resources
 
 #if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_ENU)
-LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
-#pragma code_page(1252)
+LANGUAGE LANG_NEUTRAL, SUBLANG_NEUTRAL
+#pragma code_page(65001)
```
RadWolfie: Did Visual Studio made this change? If not, try remove the `#pragma code_page...` then edit something in Visual Studio, and save, to see if it will re-add again. Reason I'm asking since resource file is saved in UTF-8 which make this code page unnecessary.

CONTRIBUTORS
```diff
@@ -1,17 +1,41 @@
-Developers:
-Luke Usher [SoullessSentinel]
-PatrickvL
-StrikerX3
-blueshogun96
-donwayo
-hrydgard
-jagotu
+Contributors:
+PatrickvL (Patrick van Logchem)
```
RadWolfie: I'm not happy with this change. Luke Usher is the founder of Cxbx-Reloaded project.

src/gui/DlgAbout.cpp
```diff
@@ -81,7 +81,10 @@ INT_PTR CALLBACK DlgAboutProc(HWND hWndDlg, UINT uMsg, WPARAM wParam, LPARAM lPa
 			SendMessage(GetDlgItem(hWndDlg, IDC_TAB1), TCM_ADJUSTRECT, FALSE, (LPARAM)&tabRect);
 			// Tab Pane 1
 			HWND tab = CreateWindowEx
-        	(NULL, "STATIC", "\nCxbx-Reloaded\nVersion " _CXBX_VERSION "\n� The Cxbx-Reloaded Team",
+        	(NULL, "STATIC", "\nCxbx-Reloaded\nVersion " _CXBX_VERSION "\n� The Cxbx-Reloaded Team"
```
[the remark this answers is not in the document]
PatrickvL: Yeah, I'll fix the conflict once either one is merged.

CONTRIBUTORS
PatrickvL: Okay, how about the new version?

RadWolfie: resolved

resource/Cxbx.rc
RadWolfie: Okay, I see more details in the *commit* since OP post doesn't have complete details of what's been change and the support of reason(s).

Hmm... I can see likely reason to have a code page as a safety precaution if Visual Studio become confuse for what code page the resource file is saved in. In my opinion, I'm leaning on a bug in Visual Studio not reading alternative configuration earlier or skipped.

PR #1783, opened 2019-11-16
Introduce two CxbxImpl_*'s for as-yet unimplemented functions.

PatrickvL:
This introduces a foundation for later work (which will also allow unpatching a few vertex shader related functions).

Ultimately, all patches could be forwarded to similar `CxbxImpl_`* functions, that can be grouped into separate source flies.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -141,6 +141,13 @@ static bool                         g_bHack_DisableHostGPUQueries = false; // TO
 static IDirect3DQuery              *g_pHostQueryWaitForIdle = nullptr;
 static IDirect3DQuery              *g_pHostQueryCallbackEvent = nullptr;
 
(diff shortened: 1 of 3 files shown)


LukeUsher:
I do like the 'CxbxImpl' idea: Seperating implementations from patches is great, especially when it comes to allowing implementations to be swapped out via user configuraton (eg: OpenGL vs DirectX renderer backends)

But for this to work, there needs to be some kind of interface in which implementations define CxbxImpl functions, otherwise there's not much benefit to splitting them out.

I always envisioned this by having a base class interface, for example, for D3D we could have an interface class `IXboxD3DInterface`, which has 'virtual' functions for CreateDrive, SetRenderTarget, etc.

Then the actual implementations could derive from that class, so our existing implementation could be `XboxD3D9Implementation`, which implements all the required functions.

This way, our HLE file could contain a global pointer to an IXboxD3DInterface class, and we get to choose at boot-time which implementation to instanciate into that pointer, completely decoupling our patches and rendering code.

PR #1785, opened 2019-11-21
Implement extra log component calculations

NZJenkins:
Implement extra `log` cases, similar to `exp`.
The `z` component remains the same as before, a straight `log` instruction
This should fix usages of the `x`, `y`, and `w` components, but I can't find any test cases

### DIFF
--- a/src/core/hle/D3D8/XbVertexShader.cpp
+++ b/src/core/hle/D3D8/XbVertexShader.cpp
@@ -2588,6 +2588,64 @@ std::string VshPostProcess_Expp(std::string shader) {
 	return shader;
 }
 
+std::string VshPostProcess_Log(std::string shader) {
+	const auto xbox_log_x = std::regex("log (\\w\\d\\d?)(\\.x)?, (.+)$");
+	const auto xbox_log_y = std::regex("log (\\w\\d\\d?)(\\.y)?, (.+)$");
+	const auto xbox_log_z = std::regex("log (\\w\\d\\d?)(\\.z)?, (.+)$");
+	const auto xbox_log_w = std::regex("log (\\w\\d\\d?)(\\.w)?, (.+)$");
+
+	if (std::regex_search(shader, xbox_log_x))
+		LOG_TEST_CASE("Title uses the x component result of log");
+	if (std::regex_search(shader, xbox_log_y))
+		LOG_TEST_CASE("Title uses the y component result of log");
+	if (std::regex_search(shader, xbox_log_w))
+		LOG_TEST_CASE("Title uses the w component result of log");
+
+	// exponent and mantissa functions such that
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -2588,6 +2588,64 @@ std::string VshPostProcess_Expp(std::string shader) {
...
+		"log tmp.x, $3\n"
+		"frc $tmp.y, $tmp.x\n"
+		"sub tmp.x, tmp.x, tmp.y\n" // tmp.x = exponent(x) = floor(log(x))
+		"exp tmp.x, tmp.x\n"
+		"rcp tmp.x, tmp.x\n" // tmp.x = 1 / (2 ^ exponent(x))
+		"mul $1.y, $3, tmp.x");
+	shader = std::regex_replace(shader, xbox_log_y, host_log_y);
+
+	// dest.z = log(x)
+	// Test Case: Mechassault
```
PatrickvL: It helps to also describe shortly when this part of the converted shader code can be seen in effect.

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -2588,6 +2588,64 @@ std::string VshPostProcess_Expp(std::string shader) {
...
+		"; patch log: dest.z = log(x)\n"
+		"log $1.z, $3";
+	shader = std::regex_replace(shader, xbox_log_z, host_log_z);
+
+	// dest.w = 1
+	// Test Case: ???
+	// TODO do a constant read here
+	const auto host_log_w = UsingScratch(
+		"; patch log: dest.w = 1\n"
+		"sub tmp.x, tmp.x, tmp.x\n" // Get 0
```
PatrickvL: When we get more and more cases where we need constants, it's probably better to just set a boolean per constant, use pre-defined constant indices here, and set those constants somewhere else (either use actual constants, and set them once after CreateDevice or whatnot, or use temporary registers, and insert initialization code to the shader header once all "constant-X-required" booleans have their final values).

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: true, will add it

NZJenkins: fair call, will add it

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: Yes definitely possible but IMO not worth doing it unless we end up setting some CXBX constants anyway (even then debatable - we may never find a test case for these as it's a bit of an obscure result)

PR #1789, opened 2019-11-21
Corrected old ordering of D3DTOP enum elements.

PatrickvL:
This fixes alpha texture blending on titles that use D3D XDK 3948 and older.

Note : Together with D3DRS (Render States) and D3DTSS (Texture Stage States), this D3DTOP (Texture blending OPerations) are the only three enumerate sets that had meaningful different declarations between XDK versions.

To be complete, there are two other enumerate sets that had differences, but those have no impact on emulation : D3DAPI (used for measuring API call heuristics) and D3DSBD (State Block Dirty flags).

EDIT : It'd be good to test titles between D3D XDK 3944 and 4039;

Right now, we assume the re-ordering took place at or before 3948, but it could have happened anywhere past 3944 (which is verified to use the old ordering) and before 4039 (which is verified to use the new ordering).


LukeUsher:
I might have some good test cases for this, I'll get back to you later

PR #1791, opened 2019-11-21
Speed up paletized texture

PatrickvL:
Include a palette hash in the texture lookup key, instead of patching  Palette_Lock functions and freeing textures when setting palette's.

To avoid overflowing the resource cache, PruneResourceCache now applies cache eviction as well.
Thanks to that change, two `Palette_Lock` patches could be removed as well.

This speeds up titles that use paletized textures a lot, here `Yu-Gi-Oh! The Dawn Of Destiny` (YuGiOh DOD) running in-game at 30fps :
image

### DIFF
--- a/src/core/hle/D3D8/XbD3D8Types.h
+++ b/src/core/hle/D3D8/XbD3D8Types.h
@@ -444,7 +444,7 @@ struct X_D3DResource
 #define X_D3DCOMMON_INTREFCOUNT_MASK   0x00780000
 #define X_D3DCOMMON_INTREFCOUNT_SHIFT  19
 #define X_D3DCOMMON_INTREFCOUNT_1      (1 << X_D3DCOMMON_INTREFCOUNT_SHIFT) // Dxbx addition
-#define X_D3DCOMMON_VIDEOMEMORY        0x00800000 // Not used.
+#define X_D3DCOMMON_VIDEOMEMORY        0x00000000 // Was 0x00800000, but Xbox doesn't have this flag!
 #define X_D3DCOMMON_D3DCREATED         0x01000000
 #define X_D3DCOMMON_ISLOCKED           0x02000010 // Surface is currently locked (potential unswizzle candidate)
 #define X_D3DCOMMON_UNUSED_MASK        0xFE000000
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -173,7 +173,7 @@ static XTL::PVOID                   g_pXbox_Palette_Data[XTL::X_D3DTS_STAGECOUNT
 static unsigned                     g_Xbox_Palette_Size[XTL::X_D3DTS_STAGECOUNT] = { 0 }; // cached palette size
 
 
-       XTL::X_D3DBaseTexture       *EmuD3DActiveTexture[XTL::X_D3DTS_STAGECOUNT] = {0,0,0,0}; // Set by our D3DDevice_SetTexture and D3DDevice_SwitchTexture patches
+       XTL::X_D3DBaseTexture       *g_pXbox_SetTexture[XTL::X_D3DTS_STAGECOUNT] = {0,0,0,0}; // Set by our D3DDevice_SetTexture and D3DDevice_SwitchTexture patches
 static XTL::X_D3DBaseTexture        CxbxActiveTextureCopies[XTL::X_D3DTS_STAGECOUNT] = {}; // Set by D3DDevice_SwitchTexture. Cached active texture
 
 /* Unused :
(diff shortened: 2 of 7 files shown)


JusicP:
Half-Life 2 Crashes on startup.
```
[0x32F8] MAIN: Received Exception (Code := 0xC0000005)

 EIP := 0x04108AA0
 EFL := 0x00010246
 EAX := 0x00000100 EBX := 0x00000400 ECX := 0x00000000 EDX := 0xFFFFFFFF
 ESI := 0x00000000 EDI := 0x00000000 ESP := 0x099FF9D0 EBP := 0x099FF9DC
 CR2 := 0x00000000
```

PatrickvL:
Try the latest build (I force-pushed a safe-guard against that crash)

JusicP:
Now it works

literalmente-game:
These don't go in-game with the PR:

- MechAssault
- Metal Slug 5
- Crash N Burn

literalmente-game:
> > These don't go in-game with the PR:
> > 
> > * MechAssault
> > * Metal Slug 5
> > * Crash N Burn
> > 
> > These have new rendering issues:
> > 
> > * Kill Switch (wrong textures being loaded for some models)
> > * RalliSport Challenge (black floor)
> > * Bloodrayne (black textures for many models, also black floor)
> > * Bloodrayne 2 (black textures for most of the models, black floor)
> 
> Mechassault appears broken since [6b08ee8](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/6b08ee81cc15b5d3c2209857deddcf5ffc07287d) so may need to double check with the previous build

Updated the comment, you're right, this list of games was affected by the mentioned change, not this PR:
- MechAssault
- Metal Slug 5
- Crash N Burn

AzurikRiseOfPerathia:
Azurik : Rise Of Perathia : Black texture, and some texture poorly modeled (Image has the support)

hyperspeedgx:
Does we need to set some kinda of settings for Yu-Gi-Oh works? My only shows a black screen, I saw that pal ver. works but I don't have it.

PatrickvL:
@AzurikRiseOfPerathia the exploding vertices are not related to this pull request, since it doesn't touch vertices at all.

It's probably caused by the recent MOVA fix, @NZJenkins is already looking into it.

As for the black textures, I'm investigating the root cause, but it's eluding me for now...

PatrickvL:
@literalmente-game commit db45305 may fix black textures, could you do a quick test?

PatrickvL:
Although the latest commit might improve/fix a few regressions, the bigger problem this PR still has, is that when the resource cache is pruned (which right now means it's cleared entirely), the association between Xbox render targets and the host counterparts is lost.

This results in the creation of new host-side render-targets, but those won't contain the same data as before anymore.

Also, some resources (like the backbuffer and other globals) should not be removed at all, but are also cleared right now. For this, we have to come up with two solutions :

1. a real cache eviction algorithm that's fast, efficient and easy to implement and maintain
2. a way to avoid removing entries that must be kept alive at all times

RinMaru:
Vexx No longer gets ingame

PatrickvL:
With the most recent commit, I've implemented an easier solution for now;

Instead of implementing an actual cache eviction policy, I've split up the cache in two parts.

This way, only the paletized textures get pruned (still with the poor-mans cache eviction policy : clean all), and all other resource types stay in the main cache, which implicitly prevents render target mappings getting lost.

RinMaru:
Annotation 2019-11-28 183826
getting this error in vexx due to this PR

PatrickvL:
I think I'll have to redo this entire pr, in smaller increments, making sure each step doesn't regress

EternalShackles:
Well, latest build fixed the black textures in all games that were broken with this pr earlier
GW OG
GW FIXED

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -444,7 +444,7 @@ struct X_D3DResource
 #define X_D3DCOMMON_INTREFCOUNT_MASK   0x00780000
 #define X_D3DCOMMON_INTREFCOUNT_SHIFT  19
 #define X_D3DCOMMON_INTREFCOUNT_1      (1 << X_D3DCOMMON_INTREFCOUNT_SHIFT) // Dxbx addition
-#define X_D3DCOMMON_VIDEOMEMORY        0x00800000 // Not used.
+#define X_D3DCOMMON_VIDEOMEMORY        0x00000000 // Was 0x00800000, but Xbox doesn't have this flag!
```
LukeUsher: If this flag is not present on Xbox, why not just remove the definition entirely?

PatrickvL: Removing this is possible of course, but it'd also be information loss. Keeping this single line in, with the end of line comment, at least documents it clearly in code. If we ever document all of our findings somewhere that'll last, it can be removed from code, but until then I'd politely request to keep these little nuggets of information in.

PatrickvL:
@LastBreathToday reported :
> That latest PR on Appvoyer breaks Crash TWOC the main rooms glass floors don't render correctly when walking around plus the gates don't render anymore
> Might just be an AMD problem
> Not sure if the crystals rendering black is a AMD problem either
> Ugh AMD issue it is
> For me the floor will flicker black then normal but the elemental gates don't render and this all started from the recent PR that was causing regressions

RinMaru:
@PatrickvL nvr mind false alarm on AMD all is good

PR #1796, opened 2019-12-14
Eeprom Config: Correctly render Serial Number as Numeric Text, not HEX

LukeUsher:
Serial Number is a 12 digit string stored as ASCII. Unlike the keys, this field should not be displayed as hex.

This matches up with what end users would expect to see (serial number matches the one displayed in the Xbox Network Settings)

Before:
image

After:
image

### DIFF
--- a/src/gui/DlgEepromConfig.cpp
+++ b/src/gui/DlgEepromConfig.cpp
@@ -95,18 +95,9 @@ void WriteEepromInMemory(HWND hDlg)
 
 	// Serial number
 	{
-		char Buffer[25];
+		char Buffer[13];
 		SendMessage(GetDlgItem(hDlg, IDC_EE_SERIAL_NUMBER), WM_GETTEXT, 25, reinterpret_cast<LPARAM>(Buffer));
-		std::string hex(Buffer);
-		size_t len = hex.length();
-		std::string ByteSequence;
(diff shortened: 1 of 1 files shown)


src/gui/DlgEepromConfig.cpp
```diff
@@ -95,18 +95,11 @@ void WriteEepromInMemory(HWND hDlg)
...
-		std::string ByteSequence;
-		for (unsigned int i = 0; i < len; i += 2)
-		{
-			std::string byte = hex.substr(i, 2);
-			char chr = static_cast<int>(stol(byte, nullptr, 16));
-			ByteSequence.push_back(chr);
-		}
-		std::memcpy(&pEEPROM_GUI->FactorySettings.SerialNumber, ByteSequence.c_str(), 12);
+		std::string string(Buffer);
+		size_t len = string.length();
```
RadWolfie: Any reason to have this unused variable here?

RadWolfie:
All code changes look good to me. I also had verified pr build for anything may cause breakage, no breakage were found.

I found one variable is unused. Waiting for removal or a response back.

src/gui/DlgEepromConfig.cpp
LukeUsher: just didn't remove it when modifiying the existing code :P

PR #1797, opened 2019-12-14
nvnet: Accept only broadcast and targeted traffic, not multicast

LukeUsher:
Fixes warnings in titles using debug version of xnet library.


RadWolfie:
LTGM

PR #1800, opened 2019-12-17
Fix when titles attempt to find the XBEH header in memory

LukeUsher:
This allows titles that manually access the XBE header and verify the XBEH signature to function in Cxbx-Reloaded.

It works by patching the game to replace the XBEH string with the string we actually have at the Xbe base address, satisfying the header test.

Test Case: Half Life 2.


PatrickvL:
Please include the work 'HACK' in a comment in code (and not just the commit description), so we can locate all hacks more easily

PR #1805, opened 2019-12-22
HLSL fixes

PatrickvL:
This commit should have been a part of the HLSL port PR, but didn't make it because the problem we saw was eluding us.

Re-reading various sources on HLSL, the problem we had (over-saturated ground color in XDK Fog sample) turned out to be caused by us not-saturating the COLOR-semantic outputs (oD0, oD1, oB0 and oB1).

Applying saturation to those outputs, fixed the problem!

image

Look, Indy reading a map, rendered like never before : 
image

Also updated the ARL bias, applied the frexp intrinsic to our LOGP function and updated LIT according to https://github.com/google/swiftshader/blob/6e3a387c01598240a7a8722ab8cbd1015ec1cd0a/src/Shader/ShaderCore.cpp#L1168


PatrickvL:
For reference: This was how the XDK Fog sample looked while it was still over-saturated before :
image

PR #1806, opened 2019-12-22
Implement BeginVisibilityTest, EndVisibilityTest and GetVisibilityTestResult

Voxel9:
The following implementations fix the VisibilityTest XDK sample, and also slightly improves the LensFlare sample. Very likely won't affect much else in the way of games or other samples.

Before:
=====
image

After:
=====
image
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -3239,7 +3239,11 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_BeginVisibilityTest)()
 {
 	LOG_FUNC();
 
-	g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &g_pVisibilityQuery);
+	HRESULT hRes = g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &g_pVisibilityQuery);
+	if (hRes != D3D_OK) {
+		LOG_TEST_CASE("BeginVisibilityTest: CreateQuery failed");
+		RETURN(hRes);
+	}
 
 	g_pVisibilityQuery->Issue(D3DISSUE_BEGIN);
 
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -3322,7 +3322,14 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_GetVisibilityTestResult)
 		LOG_FUNC_ARG(pTimeStamp)
 		LOG_FUNC_END;
 
-	while(S_FALSE == g_pVisibilityQuery->GetData(pResult, sizeof(DWORD), D3DGETDATA_FLUSH));
+	if (g_pVisibilityQuery != nullptr)
+	{
+		while (S_FALSE == g_pVisibilityQuery->GetData(pResult, sizeof(DWORD), D3DGETDATA_FLUSH));
+	}
+	else
+	{
+		LOG_TEST_CASE("GetVisibilityTestResult: g_pVisibilityQuery = nullptr");
+	}
 
     return D3D_OK;
 }
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -113,6 +113,7 @@ static IDirect3DVertexBuffer       *g_pDummyBuffer = nullptr;  // Dummy buffer, 
 static IDirect3DIndexBuffer        *g_pClosingLineLoopHostIndexBuffer = nullptr;
 static IDirect3DIndexBuffer        *g_pQuadToTriangleHostIndexBuffer = nullptr;
 static IDirect3DSurface            *g_pDefaultHostDepthBufferSurface = nullptr;
+static IDirect3DQuery              *g_pVisibilityQuery = nullptr;
 
 // cached Direct3D state variable(s)
 static size_t                       g_QuadToTriangleHostIndexBuffer_Size = 0; // = NrOfQuadIndices
@@ -3231,8 +3232,6 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_EndPush)(DWORD *pPush)
 	}
 }
 
-IDirect3DQuery9* VisibilityQuery;
-
 // ******************************************************************
 // * patch: D3DDevice_BeginVisibilityTest
 // ******************************************************************
@@ -3240,9 +3239,9 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_BeginVisibilityTest)()
 {
 	LOG_FUNC();
 
-	g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &VisibilityQuery);
+	g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &g_pVisibilityQuery);
 
-	VisibilityQuery->Issue(D3DISSUE_BEGIN);
+	g_pVisibilityQuery->Issue(D3DISSUE_BEGIN);
 
 	return D3D_OK;
 }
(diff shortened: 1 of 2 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3231,14 +3231,20 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_EndPush)(DWORD *pPush)
 	}
 }
 
+IDirect3DQuery9* VisibilityQuery;
```
PatrickvL: Move towards where all globals are declared, assign a default nullptr, and prefix appropriately

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3231,14 +3231,20 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_EndPush)(DWORD *pPush)
...
 // ******************************************************************
 // * patch: D3DDevice_BeginVisibilityTest
 // ******************************************************************
-VOID WINAPI XTL::EMUPATCH(D3DDevice_BeginVisibilityTest)()
+HRESULT WINAPI XTL::EMUPATCH(D3DDevice_BeginVisibilityTest)()
 {
 	LOG_FUNC();
 
-	LOG_UNIMPLEMENTED();
+	g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &VisibilityQuery);
```
PatrickvL: Check if this call succeeds before accessing the output argument

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3267,7 +3273,7 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_EndVisibilityTest)
 {
 	LOG_FUNC_ONE_ARG(Index);
 
-	LOG_UNIMPLEMENTED();
+	VisibilityQuery->Issue(D3DISSUE_END);
```
PatrickvL: Check if this global is assigned before using it

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3301,13 +3307,7 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_GetVisibilityTestResult)
...
 		LOG_FUNC_END;
 
-    // TODO: actually emulate this!?
-
-    if(pResult != 0)
-        *pResult = 640*480;
-
-    if(pTimeStamp != 0)
-        *pTimeStamp = 0;
+	while(S_FALSE == VisibilityQuery->GetData(pResult, sizeof(DWORD), D3DGETDATA_FLUSH));
```
PatrickvL: Check if the global is assigned before using it
Make sure an endless loop can never occur, and add a comment explaining how this is done, please

RadWolfie:
Missing two remark reviews fix request:
- https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1806#discussion_r360732631
- https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1806#discussion_r360732741

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie: Keep in mind, there is a chance of `D3DDevice_BeginVisibilityTest` symbol missing in cache or a personal homebrew application may not be coded properly. Hence the reason of @PatrickvL's change request remark.

Voxel9:
@RadWolfie I merely didn't get around to finishing it off, give me chance...

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
Voxel9: This should be done now.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
Voxel9: This should also be done now.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
Voxel9: After further research, I would argue to leave this as-is.
Using D3DGETDATA_FLUSH means that, in the event that the D3D device gets lost, GetData will be able to detect that, and return a failure value instead, meaning the loop will be broken out of when that happens.
Also, by cxbx's design, we can't perform drawing within the while loop either, so I'm not really sure how else we'd approach it.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
Voxel9: This should also be fixed now.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie: ^ should be in the loop's comment to explain the reason. That way, in the future we might change it by accidentally which would break stuff. 😉

Voxel9: Pushed a small explanation now. 🙂
If I could just get a once-over on everything at some point, that would be great

PatrickvL: Looking good, thanks for adressering most remarks!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3234,11 +3235,15 @@ VOID WINAPI XTL::EMUPATCH(D3DDevice_EndPush)(DWORD *pPush)
 // ******************************************************************
 // * patch: D3DDevice_BeginVisibilityTest
 // ******************************************************************
-VOID WINAPI XTL::EMUPATCH(D3DDevice_BeginVisibilityTest)()
+HRESULT WINAPI XTL::EMUPATCH(D3DDevice_BeginVisibilityTest)()
 {
 	LOG_FUNC();
 
-	LOG_UNIMPLEMENTED();
+	g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &g_pVisibilityQuery);
```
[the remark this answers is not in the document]
Voxel9: I didn't think there was more than 1 query object?
I was under the assumption CreateQuery overwrites on top of the existing query stored in the global, without adding anything extra to memory with each call. Essentially just creating a fresh query.
Right now, I'm honestly not sure.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3267,7 +3272,7 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_EndVisibilityTest)
 {
 	LOG_FUNC_ONE_ARG(Index);
 
-	LOG_UNIMPLEMENTED();
+	g_pVisibilityQuery->Issue(D3DISSUE_END);
```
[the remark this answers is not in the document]
Voxel9: Depends. If I'm wrong and CreateQuery does in fact leak memory with multiple BeginVisibilityTest calls, then it might potentially be possible to free the global at the beginning of that function instead. That way, we know there will always _only_ be 1 query occurring at a time.

RadWolfie:
As of commit `fd2a050`, release build failed only from Azure CI.
`d3dx9.lib(d3dxvec3.obj) : fatal error LNK1103: debugging information corrupt; recompile module [D:\a\1\s\build\projects\cxbx\cxbx.vcxproj]`

It is indeed from Visual Studio 2019's end of the issue since AppVeyor's had that issue before.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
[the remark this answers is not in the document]
LukeUsher: From the D3D9 MSDN documentation 
```
When you are finished with the query, release it like any other interface
```

So yeah, this will leak memory, because it will allocate a new Query object on every call.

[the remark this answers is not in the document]
LukeUsher: You should check if Query != nullptr, and if so, call` g_pVisibilityQuery->Release()` and set it to nullptr.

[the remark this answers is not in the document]
PatrickvL: I'm sorry to add to this, but one extra remark : To be entirely 'clean', any created query should also be `Release`d when the device is reset, and/or when Cxbx closes.

PatrickvL:
Replaced by a reworked branch in #1809

PR #1809, opened 2019-12-25
Implement BeginVisibilityTest, EndVisibilityTest and GetVisibilityTestResult

PatrickvL:
(Note : An initial PR for this was submitted by @Voxel9, I merely finished it.)

The following implementations fix the VisibilityTest XDK sample, and also slightly improves the LensFlare sample. Very likely won't affect much else in the way of games or other samples.

Before:
=====
image

After:
=====
image
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -3239,7 +3239,11 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_BeginVisibilityTest)()
 {
 	LOG_FUNC();
 
-	g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &g_pVisibilityQuery);
+	HRESULT hRes = g_pD3DDevice->CreateQuery(D3DQUERYTYPE_OCCLUSION, &g_pVisibilityQuery);
+	if (hRes != D3D_OK) {
+		LOG_TEST_CASE("BeginVisibilityTest: CreateQuery failed");
+		RETURN(hRes);
+	}
 
 	g_pVisibilityQuery->Issue(D3DISSUE_BEGIN);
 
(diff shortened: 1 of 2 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3301,13 +3361,32 @@ HRESULT WINAPI XTL::EMUPATCH(D3DDevice_GetVisibilityTestResult)
...
+		pHostQueryVisibilityTest->Release();
+	} else {
+		// Fallback to old faked result when there's no host occlusion query :
+		if (pResult != xbnullptr) {
+			*pResult = 640 * 480; // TODO : Use actual backbuffer dimensions
+		}
+	}
+
+	if (pTimeStamp != xbnullptr) {
+		LOG_TEST_CASE("pTimeStamp = nullptr");
```
RadWolfie: Pretty sure it should be "pTimeStamp != nullptr" or "requested value from pTimeStamp". First option in my opinion will throw off maybe few testers and a lot of users. They may see it as an error message.

RadWolfie:
Hmm... I see a setback with this change due to `g_bHack_DisableHostGPUQueries` is force set to false since it's not configurable.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: I agree, since LOG_TEST_CASE's are meant to be reported back, a somewhat clear description is better than this cryptic message - I'll update it.

RadWolfie:
Only two XDK samples has a reference to `D3DDevice_BeginVisibilityTest` symbol, VisibilityTest and Lensflare. While both samples' message status do report change accurately with this pull request. Lensflare's effect is broken. I am unable to verify if this effect is effectively filtered with objects in front of the sun.

reference:
<details><summary>Xbox - courtesy of @Voxel9's recorded samples</summary>

Lensflare - xbox
</details>
<details><summary>Pull request</summary>

Lensflare - after
</details>

Before pull request, the message would alway say "Full Visibility Optimization". Unless look away from the sun, then message will change to "Frustum Clipped".

LGTM

PR #1811, opened 2019-12-26
Prevent error X5765 on pixel shader LPR opcode

PatrickvL:
This prevents "error X5765: Dest register for LRP cannot be the same as first or third source register" in WWE RAW2.

A little background information :

When compiling pixel shaders, to facilitate later optimizations, FixOverusedRegisters() inserts MOV opcodes reading from constant, vertex color or texture registers, and writing to a free R register slot.

In this case, this transformed :

````
lrp r0, c0,r2,c1
````

into

````
mov r0, c0
lrp r0, r0,r2,c1
````

Not only introduces this a completely useless additional MOV (to R0, which is overwritten by the next LRP opcode), but it also results in LRP reading from R0, which is the same register it's writing to, which seems to be forbidden.

This fix prevents that from happening, by not-inserting a MOV when this situation would arise.

It's possible there exist more such scenario's, but let's tackle them when they are reported.

As for the other change in this commit : Instead of repeating a nearly identical while-loop three times, this loop is now wrapped in a for-loop, that handles all three versions (for PARAM_C, PARAM_V and PARAM_T type registers).

PS: This is merely a fix I did in one of my older branches, which is small and contained, and mentions a test-case, so I figured I might as well put it up as a PR...

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -5613,63 +5613,35 @@ bool PSH_XBOX_SHADER::FixOverusedRegisters()
 
         int InsertPos = i;
 
-        int addressCount = 0;
-        int total = 0;
-        while (Intermediate[i].ReadsFromRegister(PARAM_C, -1, addressCount, total) && (addressCount > 1 || total > 2))
-        {
-            for (int p = 0; p < PSH_OPCODE_DEFS[Intermediate[i].Opcode]._In; ++p)
+        // Handle PARAM_C, PARAM_V and PARAM_T (in that order) :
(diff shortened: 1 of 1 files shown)


gandalfthewhite19890404:
WWE Raw 2 now get hang after pressing Start - as in previous attempt to fix that error.

gandalfthewhite19890404:
And now WWE Raw 2 hangs after pressing Start. Why it was merged in non-tested state?

LukeUsher:
@gandalfthewhite19890404 the hang is most likely an unrelated issue

All this does is fix a compilation error in pixel shaders, it has no impact on execution.
The game (for whatever reason) seems to have only worked because the pixel shader was broken.

PR #1812, opened 2019-12-31
Trampoline optimization

PatrickvL:
This moves construction of lookup strings outside call-locations towards a central LookupTrampolines() function, reducing the overhead of calling a trampoline down to a minimum.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -267,6 +267,58 @@ g_EmuCDPD = {0};
 		} \
 	} while (0)
 
+// Declare trampolines
+#define XB_TRAMPOLINES(XB_MACRO)                                                                                                                                       \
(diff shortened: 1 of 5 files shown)


PatrickvL:
Here the disassembly of a call to the `Direct3D_CreateDevice` trampoline - as you can see, there's no overhead whatsoever :
```
// call Xbox CreateDevice function
HRESULT hRet = XB_TRMP(Direct3D_CreateDevice)(Adapter, DeviceType, hFocusWindow, BehaviorFlags, pPresentationParameters, ppReturnedDeviceInterface);

04148B25  push        edi  
04148B26  push        esi  
04148B27  push        dword ptr [BehaviorFlags]  
04148B2A  push        dword ptr [hFocusWindow]  
04148B2D  push        dword ptr [DeviceType]  
04148B30  push        dword ptr [Adapter]  
04148B33  call        dword ptr [XB_TRAMPOLINE_Direct3D_CreateDevice (04514B0Ch)]  
04148B39  mov         esi,eax  
```

PR #1813, opened 2020-01-01
Fix a crash when converting certain, large vertex shaders

LukeUsher:
This allows DOA3 to play again after it was broken by the HLSL work.

The cause of the crash was a Stack Overflow when compiling the shader,
caused by allocating a large object on the stack.

Instead, this commit allocates the XboxVertexShaderDecoder on the heap
instead, which thanks to the unique_ptr smart pointer, automatically
gets freed when it goes out of scope.

### DIFF
--- a/src/core/hle/D3D8/XbVertexShader.cpp
+++ b/src/core/hle/D3D8/XbVertexShader.cpp
@@ -1615,7 +1615,7 @@ extern HRESULT EmuRecompileVshFunction
 {
 	XTL::X_VSH_SHADER_HEADER* pXboxVertexShaderHeader = (XTL::X_VSH_SHADER_HEADER*)pXboxFunction;
 	uint32_t* pToken;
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1615,7 +1615,7 @@ extern HRESULT EmuRecompileVshFunction
 {
 	XTL::X_VSH_SHADER_HEADER* pXboxVertexShaderHeader = (XTL::X_VSH_SHADER_HEADER*)pXboxFunction;
 	uint32_t* pToken;
-	XboxVertexShaderDecoder VshDecoder;
+	std::unique_ptr<XboxVertexShaderDecoder> VshDecoder = std::make_unique<XboxVertexShaderDecoder>();
```
LukeUsher: This object was too large for the stack, causing D3DCompile to trigger a stack overflow exception in some titles.

PR #1821, opened 2020-01-24
Move D3D9-specific vertex shader code out of XbVertexShader

NZJenkins:
PR is just for reorganizing and tidying code - behaviour should be unchanged


LukeUsher:
Great idea, I'll test this with some 'problem games' later but it looks good to me at first glance.

RadWolfie:
As JayFoxRox would had nit at me before. Wouldn't it better to name `Direct3D9VertexShader` as `VertexShader`? Since it is in Direct3D9 plugin folder, unless the file is outside Direct3D9 folder then it would make sense to name it `Direct3D9VertexShader`.

PatrickvL:
Please expand the description stating if there are functional changes as well, or not.

CMakeLists.txt
```diff
@@ -261,6 +262,7 @@ file (GLOB CXBXR_SOURCE_EMU
  "${CXBXR_ROOT_DIR}/src/common/util/gloffscreen/gloffscreen_common.cpp"
  "${CXBXR_ROOT_DIR}/src/common/util/gloffscreen/gloffscreen_wgl.cpp"
  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp"
+ "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/VertexShader.cpp"
```
RadWolfie: This line had been overlooked. It should be between `TextureStates.cpp` and `WalkIndexBuffer.cpp`.

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -0,0 +1,317 @@
...
+#include "VertexShader.h"
+#include "core\kernel\init\CxbxKrnl.h"
+#include "core\kernel\support\Emu.h"
+
+#include <sstream>
+
+// HLSL generation
+void OutputHlsl(std::stringstream& hlsl, VSH_IMD_OUTPUT& dest)
+{
+	static const char* OReg_Name[/*VSH_OREG_NAME*/] = {
```
RadWolfie: From github's web viewing, it shows as if there is a space and a tab (indent). However when I download the file as raw, it reveal tabs only for indent. Reviewers, if you notice this then please ignore inaccurate indent from github's web view in this pull request.

P.S. **Do not mark this as resolved. It is not a review, only a comment.**

PR #1823, opened 2020-01-24
Fix Audio Stream Flush(Ex) Functions

RadWolfie:
After re-review disassembly of CDirectSoundStream_Flush(Ex) and CDirectSoundStream_Discontinuity, along with later callers. There was some mistakes were made.
- FlushEx with rtTimeStamp given with zero were not properly checked since it is asking for immediate trigger to flush. It was originally check within `X_DSSFLUSHEX_ENVELOPE` flag conditional process.
- When Flush is called, it does halt the buffer play-state. Once audio is stopped, then flush all of packets.
- Discontinuity is only used after final active packet in case of need more input packets. In a way, it free up inactive pending packets in order to accept new packets.

Titles been tested and has improve:
- Hunter The Reckoning
- WWE RAW 2

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
@@ -298,3 +298,26 @@ bool DSStream_Packet_Process(
     }
     return 1;
 }
+
+bool DSStream_Packet_Flush(
+    XTL::X_CDirectSoundStream* pThis
+    )
+{
+    // If host's audio is still playing then return busy-state until buffer has stop playing.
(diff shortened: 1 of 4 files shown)


medievil1:
breaks indiana jones 
image

ESPPIRAL:
It no longer softlocks after the shooting practice intro.

Cxbx-Reloaded c1a882bf (Jan 24 2020) _ Emulating AV-014 v1 06 (True Crime - Streets of LA) 25_01_2020 12_08_30

RadWolfie:
Tester report MechAssault had regressed from this pull request. I have found the cause, however the fix didn't work 100 percentage. It is currently under investigation.

PR #1825, opened 2020-02-02
Minor Fixup in DirectSound Plugin

RadWolfie:
The following changes in this pull request are:
- Disable XAudioCreateAdpcmFormat patch.
- DirectSoundCreateBuffer, move hack (not really a hack, but a hack for the host) fix after `DSBufferDesc.dwFlags` set.
   - Shouldn't be tampering with xbox's original input data.
- Fix copy and paste code of channelMask flag statement check.
  - Originally had duplicate of SPEAKER_FRONT_LEFT flag checks, yet doesn't have SPEAKER_FRONT_RIGHT flag check.
  - This change should improve audio's loudness.
- Move `LOGRENDER_HEADER(GUID)` into header file for upcoming pull request requirement.


RadWolfie:
In my opinion, the changes are low risk. However, I'm not entirely sure why there's a patch for `XAudioCreateAdpcmFormat` function (no documentation). If there's a patch requirement for `XAudioCreateAdpcmFormat`, then it's likely need a patch for `XAudioCreatePcmFormat` function as well.

LukeUsher:
Can confirm XAudioCreateAdpcmFormat doesn't need to be patched: It's simply setting up a struct.

PR #1827, opened 2020-02-07
Fix argument order in HLSL lerp call

NZJenkins:
- Fix argument order in HLSL lerp call
- Renaming for clarity

Before, the order of `vOverride` and `vOverrideValue` were reversed


PatrickvL:
In our Discord, @Strelok reported back a comparison screenshot :

> image
> Baldur's Gate: Dark Alliance ^

Issue #1129, opened 2018-05-04
Correctly implement D3DDevice_GetViewportOffsetAndScale


PatrickvL:
Currently, D3DDevice_GetViewportOffsetAndScale is implemented rather simple (it just sets all scale coordinates to 1.0 and all offset coordinates to 0.0).

To implement this correctly, we need to know a few test cases. Currently reported callers are:
* 007: From Russia with Love
* Army Men®: Sarge's War
* BeatDown - Fists of Vengeance
* Big Mutha Truckers
* Big Mutha Truckers 2
* Classified: The Sentinel Crisis
* Cold Fear
* Commandos SF
* Darkwatch
* Dr. Muto
* [GTA: Vice City](https://github.com/Cxbx-Reloaded/game-compatibility/issues/738)
* Jacked
* [Madagascar](https://github.com/Cxbx-Reloaded/game-compatibility/issues/111#issuecomment-386831434)
* [Max Payne 2](https://github.com/Cxbx-Reloaded/game-compatibility/issues/711#issuecomment-386829426)
* NBA LIVE 2005
* [NFS: Underground](https://github.com/Cxbx-Reloaded/game-compatibility/issues/82#issuecomment-386826795)
* [NFS: Underground 2](https://github.com/Cxbx-Reloaded/game-compatibility/issues/83#issuecomment-386827339)
* [PoP: The Sands of Time](https://github.com/Cxbx-Reloaded/game-compatibility/issues/396#issuecomment-386825771)
* Prince of Persia: T2T- see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1109#issuecomment-386488587
* RPM Tuning
* Street Racing Syndicate
* Spongebob - Battle for Bikini Bottom
* The Incredibles: Rise of the Underminer
* The SpongeBob SquarePants Movie
* TMNT(R)2
* TMNT(R)3

TODO : Add more

Note : This has become more important now that we have a viewport scaling option

Margen67:
Is this for games that show this error with or without the viewport scaling hack enabled?

PatrickvL:
It'll be needed both with, and without the viewport-scaling hack enabled (that hack won't alter triggering the popup anyway).

JohnGodgames:
**Jacked**
Please report that Jacked shows this test-case: D3DDevice_GetViewportOffsetAndScale
[...]

PR #1833, opened 2020-02-15
Remove some patches

NZJenkins:
Because we trampoline `CreateVertexShader`
* D3DDevice_GetVertexShaderSize
* D3DDevice_GetVertexShaderType

Because we trampoline `SetViewport`
* D3DDevice_GetViewportOffsetAndScale

Not sure if this issue is still needed if it's OK to unpatch https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1129


src/core/hle/D3D8/Direct3D9/Direct3D9.h
```diff
@@ -329,17 +329,6 @@ VOID WINAPI EMUPATCH(D3DDevice_GetViewport)
...
-// ******************************************************************
-// * patch: D3DDevice_GetViewportOffsetAndScale
-// ******************************************************************
-VOID WINAPI EMUPATCH(D3DDevice_GetViewportOffsetAndScale)
-(
-	X_D3DXVECTOR4 *pOffset,
-	X_D3DXVECTOR4 *pScale
-);
-
-VOID __stdcall EMUPATCH(D3DDevice_GetViewportOffsetAndScale_0)();
```
RadWolfie: These header patches cannot be remove until we have an generic header interface file. Which is likely will be in HLE verbose plugin for better LLE research. Plus basically allow past and future developers to able reuse the APIs information provided.

src/core/hle/Patches.cpp
```diff
@@ -95,9 +95,6 @@ std::map<const std::string, const xbox_patch_t> g_PatchTable = {
 	PATCH_ENTRY("D3DDevice_GetVertexShaderDeclaration", XTL::EMUPATCH(D3DDevice_GetVertexShaderDeclaration), PATCH_HLE_D3D),
 	PATCH_ENTRY("D3DDevice_GetVertexShaderFunction", XTL::EMUPATCH(D3DDevice_GetVertexShaderFunction), PATCH_HLE_D3D),
 	PATCH_ENTRY("D3DDevice_GetVertexShaderInput", XTL::EMUPATCH(D3DDevice_GetVertexShaderInput), PATCH_HLE_D3D),
-	PATCH_ENTRY("D3DDevice_GetVertexShaderSize", XTL::EMUPATCH(D3DDevice_GetVertexShaderSize), PATCH_HLE_D3D),
-	PATCH_ENTRY("D3DDevice_GetVertexShaderType", XTL::EMUPATCH(D3DDevice_GetVertexShaderType), PATCH_HLE_D3D),
-	PATCH_ENTRY("D3DDevice_GetViewportOffsetAndScale", XTL::EMUPATCH(D3DDevice_GetViewportOffsetAndScale), PATCH_HLE_D3D),
```
RadWolfie: While it may be okay to be remove. In the past I did not suggest to do this, instead do *commented* them out to ensure HLE verbose plugin has every patches for research purpose in the future.

NZJenkins: Will comment them instead

src/core/hle/D3D8/Direct3D9/Direct3D9.h
RadWolfie: resolved

RadWolfie:
LTGM, to me this is a low risk change. Is other developers okay with this change?

src/core/hle/Patches.cpp
LukeUsher: I'd much rather have them gone than just commented out. They're in version control history + the symbol scanning state (and xbsymboldatabase) list all symbols found, so that can be used as a list for the purposes of 'hle verbose'

src/core/hle/D3D8/Direct3D9/Direct3D9.h
LukeUsher: I also think these can be removed: Since the removed functions were copied into the .unused-patches file, we have a copy of the signatures there.

PR #1834, opened 2020-02-19
Delete Travis

Margen67:
Why?
  * Doesn't have VS2019, or artifacts.
  * Azure Pipelines and GitHub Actions both support Linux/macOS.


RadWolfie:
> Why?

- Travis may not have artifacts support from their domain, however can be upload to any sites (including github) for hosting the artifacts.
- Minimal requirement is Visual Studio 2017 since the project require C++17 or later.
- Allow forked repo users to use Travis other than Azure, GitHub, and AppVeyor remain in sync.

The removal is really unnecessary since we can always opt'd in or out at any time we want from Travis CI's end.

GXTX:
@RadWolfie I believe keeping only adds additional support requirements.

Margen67:
It would probably be better to delete all of them at once.

PR #1843, opened 2020-03-13
Fix Audio Flush Async and Discontinuity 

RadWolfie:
First commit is intend to fix MechAssault's regression. However, I see another fault in discontinuity and how GetStatus is handled. GetStatus does not apply the status right away, instead it is applied internally base on the voice's activity.

These changes are reflect how xbox behave base on modification to the xbe.

Plus extended SynchPlayback debug message to include stream class and pointers.

**EDIT:**
Specials thanks to @NZJenkins for further investigation on missing audio dialogues from MechAssault title. He had exposed two bugs which has been added to this pull request. 💪


medievil1:
178ec00 or da9e727 Fixed indiana jones background music restarting over and over!!!

PR #1846, opened 2020-03-17
AppVeyor: Minor changes

Margen67:
Add yml validater to top.
Skip files unrelated to build.
Reformat comments.
Only build Release since Azure Pipelines already builds Debug and is faster.
Remove cmd since it's the default shell, and unneeded characters.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,38 +1,43 @@
 # Documentation is here: https://www.appveyor.com/docs/appveyor-yml/
+# You can validate ymls here: https://ci.appveyor.com/tools/validate-yaml
 
 skip_commits:
   files:
+    - doc/*
+    - doc/*/*
+    - .github/*
+    - .github/*/*
     - .azure-pipelines.yml
     - .travis.yml
+    - CONTRIBUTORS
+    - COPYING
+    - README.md
+    - gen-msvc-project.bat
+    - setup.bat
 
 init:
   - ps: Update-AppveyorBuild -Version "$env:appveyor_repo_commit"
 
-image:
-  # If this is modified, please also update the build script
+image: # If this is modified, please also update the build script
   - Visual Studio 2019
 
-configuration:
-  # The builds will be run in this order
+configuration: # The builds will be run in this order
   - Release
-  - Debug
 
 before_build:
-  - cmd: |-
+  - |-
       git submodule update --init --recursive
       mkdir build
       cd build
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -1,38 +1,46 @@
...
-  # If this is modified, please also update the build script
+image: # If this is modified, please also update the build script
   - Visual Studio 2019
 
-configuration:
-  # The builds will be run in this order
+configuration: # The builds will be run in this order
   - Release
-  - Debug
+#  - Debug
```
GXTX: Even if you say it's faster on Azure, I'm not sure I see a reason to not build here? We suggest users to download artifacts *from* AppVeyor.

GXTX:
All of these tiny changes to each different CI, we should be consolidating to 1 as it just adds additional complexity and requires maintainers to have org accounts open for each one of these different services.

.appveyor.yml
Margen67: Debug doesn't have artifacts so having it on AppVeyor is redundant.

Margen67:
@GXTX I'm not sure what you mean: 
image

GXTX:
CXBX-R currently has 3 different CI providers (AppVeyor, Travis-Ci, Azure) which adds support complexity. Even just recently you've been going back to modify and change small issues with each provider. It would seem worth it to cut support for at least 2 of those and instead move to a provider which is giving up-to-date images and software.

.appveyor.yml
GXTX: Missed that, thanks.

Margen67:
> 
> 
> CXBX-R currently has 3 different CI providers (AppVeyor, Travis-Ci, Azure) which adds support complexity. Even just recently you've been going back to modify and change small issues with each provider. It would seem worth it to cut support for at least 2 of those and instead move to a provider which is giving up-to-date images and software.

I agree. Travis should be first since it doesn't have artifacts, and only supports VS2017.

GXTX:
> (or better yet, GitHub Actions) provide a way to link to the latest artifacts.

There is a way to direct link however users needs to be logged into GitHub to access them (See https://github.com/actions/upload-artifact/issues/51), this can be mitigated by using releases instead - although somewhat dirty it would be possible to "keep it clean" by using *more actions* to delete non-tagged releases.

Margen67:
Not much point to this if it might be removed soon: #1848

Margen67:
Reopening since AppVeyor keeps getting triggered by my PR: #1854

PR #1857, opened 2020-04-02
LOADER: Perform Some Clean Up

RadWolfie:
Follow up base on @PatrickvL comments in pull request #1851. Which is my plan to make the code human readability.

Loader's project and memory relative codes have been move into their respective locations.
Replace hardcoded memory address to compile time values,
And fix compiler warning for XbeType enum to `enum class` type.

There are some memory relative codes which I'm not sure where to put them so I left them alone.


src/common/AddressRanges.h
```diff
@@ -29,35 +29,151 @@
 
 #include <cstdint> // For uint32_t
 
-#ifndef CXBXR_EMU_EXPORTS // Only trim Windows symbols in cxbxr-ldr.exe, not in cxbxr-emu.dll (TODO : What about cxbxr.exe and cxbx.exe?)
-#include <SDKDDKVer.h>
```
PatrickvL, line 33: I don't remember why `SDKDDKVer.h` was used, might have something to do with it's impact on including `windows.h` - could you please check if this needs to be moved towards the new places where `WIN32_LEAN_AND_MEAN` is moved to as well?
And if it's no longer necessary, could you maybe comment here why including `SDKDDKVer.h` is no longer needed?

src/common/AddressRanges.h
```diff
@@ -29,35 +29,151 @@
...
-#define WIN32_LEAN_AND_MEAN             // Exclude rarely-used stuff from Windows headers
-#endif // CXBXR_EMU_EXPORTS
-
-#undef FIELD_OFFSET     // prevent macro redefinition warnings
-#include <windows.h> // For DWORD, CALLBACK, VirtualAlloc, LPVOID, SIZE_T, HMODULE 
-
 #define KB(x) ((x) *    1024 ) // = 0x00000400
```
PatrickvL, line 32: Being fully aware that I introduced these two macro functions (`KB` and `MB`), it might be good to correct their name together with this commit (if that won't impact too many files), to become `KiB` and `MiB` respectively (as that's a more accurate name). Just a thought. Ignore if you think it isn't worth the effort.

src/common/AddressRanges.h
```diff
@@ -29,35 +29,151 @@
...
+#define MAX_VIRTUAL_ADDRESS                 0xFFFFFFFF
+
+#define LOWEST_USER_ADDRESS                 XBE_IMAGE_BASE // = 0x00010000
+#define HIGHEST_USER_ADDRESS                0x7FFEFFFF
+#define HIGHEST_VMA_ADDRESS                 (HIGHEST_USER_ADDRESS - X64KB) // for NtAllocateVirtualMemory
+#define USER_MEMORY_SIZE                    (HIGHEST_USER_ADDRESS - LOWEST_USER_ADDRESS + 1) // 0x7FFE0000 = 2 GiB - 128 KiB
+
+// Memory size per system
+#define XBOX_MEMORY_SIZE                    (MB(64))
+#define CHIHIRO_MEMORY_SIZE                 (MB(128))
```
PatrickvL, line 164: I'm curious, why have some been changed into a `constexpr`, while others stayed `#define`'s?

src/common/AddressRanges.h
```diff
@@ -76,48 +192,12 @@ const struct {
...
-	RANGE_ENTRY(0xFF800000, 0xFFBFFFFF, MB(  4)         , PROT_NAC, SYSTEM_ALL              , "DeviceFlash_c (Flash mirror 3)"),
-	RANGE_ENTRY(0xFFC00000, 0xFFFFFFFF, MB(  4)         , PROT_NAC, SYSTEM_ALL    | MAY_FAIL, "DeviceFlash_d (Flash mirror 4) Optional (will probably fail reservation, which is acceptable - the 3 other mirrors work just fine"),
-	#undef RANGE_ENTRY
-};
-
-#define BLOCK_REGION_DEVKIT_INDEX_BEGIN 0
-#define BLOCK_REGION_DEVKIT_INDEX_END   4096
-#define BLOCK_REGION_SYSTEM_INDEX_BEGIN 4096
-#define BLOCK_REGION_SYSTEM_INDEX_END   12288
+} XboxAddressRanges[];
```
PatrickvL, line 195: Again I'm curious; Why did you move the array contents towards the .cpp file, instead of keeping it here in the header file?

src/common/ReserveAddressRanges.h
```diff
@@ -26,8 +26,22 @@
...
 #pragma once
 
-extern bool ReserveAddressRanges(const unsigned int system, uint32_t blocks_reserved[384]);
+#include <cstdint> // For uint32_t
 
-extern void FreeAddressRanges(const unsigned int system, unsigned int release_systems, uint32_t blocks_reserved[384]);
+// First block section will be devkit region (256MiB) = 262144KiB / 64 KiB = 4096 bits needed -> 4096 / 32 (sizeof(uint32_t)) = 128 entries blocks_reserved array.
+// Next block section will be system region (512MiB) = 524288KiB / 64KiB = 8192 bits needed -> 8192 / 32 (sizeof(uint32_t)) = 256 entries of blocks_reserved array.
+// 1 bit per block_size (64KiB).
+typedef uint32_t blocks_reserved_type[384]; // 384 = 128 (devkit region) + 256 (system region)
```
PatrickvL, line 34: Using a `_type` suffix isn't a common practice, while using a `_t` suffix IS common practice; Therefor, please consider renaming this type's symbol name into `blocks_reserved_t`?

src/common/xbe/Xbe.cpp
```diff
@@ -806,3 +809,20 @@ bool Xbe::CheckXbeSignature()
 	memcpy(xboxkrnl::XePublicKeyData, xboxkrnl::XePublicKeyDataRetail, 284);
 	return false;  // signature check failed
 }
+
+// ported from Dxbx's XbeExplorer
+XbeType Xbe::GetXbeType()
+{
+	// Detect if the XBE is for Chihiro (Untested!) :
+	// This is based on https://github.com/radare/radare2/blob/master/libr/bin/p/bin_xbe.c#L45
+	if ((m_Header.dwEntryAddr & 0xf0000000) == 0x40000000)
```
PatrickvL, line 818: `0xF0000000` could be replaced by `XBOX_WRITE_COMBINED_BASE` now.
`0x40000000` could be replaced by a new separate `_BASE` symbol as well (although I don't have a suggestion ready for the name).

src/common/xbe/Xbe.cpp
```diff
@@ -806,3 +809,20 @@ bool Xbe::CheckXbeSignature()
...
+XbeType Xbe::GetXbeType()
+{
+	// Detect if the XBE is for Chihiro (Untested!) :
+	// This is based on https://github.com/radare/radare2/blob/master/libr/bin/p/bin_xbe.c#L45
+	if ((m_Header.dwEntryAddr & 0xf0000000) == 0x40000000)
+		return XbeType::xtChihiro;
+
+	// Check for Debug XBE, using high bit of the kernel thunk address :
+	// (DO NOT test like https://github.com/radare/radare2/blob/master/libr/bin/p/bin_xbe.c#L49 !)
+	if ((m_Header.dwKernelImageThunkAddr & 0x80000000) > 0)
```
PatrickvL, line 823: `0x80000000` could be replaced by `KSEG0_BASE` now.

src/common/AddressRanges.h
RadWolfie: I'm not sure why `SDKDDKVer.h` file is here, I did tested without this and did not cause issues for compile and run-time emulation. Plus `#include <sdkddkver.h>` is include in `Windows.h` header file. Also, there is no `WIN32_LEAN_AND_MEAN` check inside `sdkddkver.h` file either.

src/common/AddressRanges.h
```diff
@@ -29,35 +29,151 @@
...
+#define XBOX_KERNEL_BASE                    (PHYSICAL_MAP_BASE + XBE_IMAGE_BASE)
+#define KERNEL_PHYSICAL_ADDRESS             XBE_IMAGE_BASE // = 0x10000
+#define KERNEL_SIZE                         sizeof(DUMMY_KERNEL)
+#define KERNEL_STACK_SIZE                   12288 // 0x03000, needed by PsCreateSystemThreadEx, however the current implementation doesn't use it
+
+// Miscellaneous memory variables
+// Xbox pages are (1 << 12) = 0x00001000 = 4096 bytes in size. Large pages are 4 MiB instead
+// NOTE: PAGE_SIZE is also defined in xfile.h (oxdk) and linux_wrapper.h (oxdk)
+#define PAGE_SHIFT                          12 // 2^12 = 4 KiB
+#define PAGE_SIZE                           (1 << PAGE_SHIFT)
```
PatrickvL, line 146: For consistency, consider adding an end-of-line comment here as well : ` // = 0x00001000 = KB(4)` (or `KiB` instead of `KB` if you'd apply that renaming suggestion as well)

src/common/AddressRanges.h
PatrickvL: There must have been some reason, but I just can't remember it. Ah well, it's fine.

src/common/AddressRanges.h
RadWolfie: Hmm... there are strings that can be update to `MiB`, and if there's `KB` as well, as well in `AddressRanges.cpp` file.

Also, the changes will only affect from loader's branch. 😃

src/common/AddressRanges.h
RadWolfie: For following reason: https://github.com/RadWolfie/Cxbx-Reloaded/commit/cc74551a4febc5528d07a9307fd9ff87d181faa9#diff-fde45dd8c0aa5b5aada10ce7c0a8b158R34-R40
Which are using `Windows.h` header file. We do not want `Windows.h` include in any header files to make things easier to include a "generic" header without dependency on Windows exclusive stuff.

We could ideally customize this to tell specific platform what protection each range can have. Then we can move this back to the top.

src/common/AddressRanges.h
RadWolfie: They're not used by loader design and/or VMemory class I believe. It's best to leave those intact until we have loader branch merge into develop/master branch. Then make further clean up from develop/master branch.

src/common/AddressRanges.h
PatrickvL: Those `PROT_`* shorthands could be turned into an enum and used inside this table, indeed.
But that would also require a conversion of the enum values to their related (`windows.h` based) `PAGE_`* symbols in the few places where `InitialMemoryProtection` is passed to `VirtualAlloc`, is that correct?

src/common/AddressRanges.h
PatrickvL: Okay, let's try to remember ;)

ergo720: Nit: why `constexpr`? Generally, you want it only when you require that the variable is computed at compile time, which, as far as I can tell, it's not the case here, so a plain `const` should be enough. Also, you should probably add a TODO comment to help remembering the cleanup.

RadWolfie: Isn't that what KB/MB do? Since it's doing math calculation to determine what will be actual value.

I'm not the expert with anything new in C++17 and later, and some of C++11 as well. I only have some basic understanding `constexpr` is use for. It was my understanding `constexpr` will convert to `const` at compile time.

src/common/AddressRanges.h
RadWolfie: That's correct.

PatrickvL: Okay, so if that enum was added, then that enum type, and the array declaration, could both be declared back in the header file, and not in this source file, correct?

RadWolfie: Correct. 😛 Since it doesn't rely on any references from Windows specific values and functions nor any of std library functions itself.

src/common/AddressRanges.h
RadWolfie: ❓ Can you make a suggestion code please?

src/common/xbe/Xbe.cpp
RadWolfie: @LukeUsher, since you're an expert with Chihiro at this time. I think `0x40000000` memory address is where sega boot's xbe always reside in this memory address? Maybe should be delcared as `SEGA_BOOT_BASE`.

If not, then what reside in `0x40000000` memory address? Since I don't find the info from your chihiro branch. Even both CxbxKrnl and AddressRanges header files has this info.

PatrickvL: Here a related piece of code : https://github.com/radareorg/radare2/blob/master/libr/bin/p/bin_xbe.c#L33

It seems to me, 0x40000000 isn't an actual memory address, per se, but more likely a result of the XOR key that Chihiro uses when 'decoding' the Chihiro Xbe's entry point.

RadWolfie: All Chihiro titles does not use `0x40000000` base for entry point and is signed with debug key. Chihiro's signature key is only signed with sega boot's xbe and indeed has `0x40000000` base for entry point. Luke had verified this in the past base on the signed keys.

I'm aware of radare2's information, however it is outdated and unreliable.

RadWolfie: So far, any Chihiro titles doesn't get trigger for `0x40000000` xor address for maybe part of reason they are signed with debug key.

Here's some example of what entry points looks like.
```
Segaboot (Chihiro)
0x40B682F4 (Retail: 0xE84AD55F, Debug: 0xD4331FBF)

Virtua Cop 3 (Chihiro)
0x948EF94C (Retail: 0x3C72AEE7, Debug: 0x000B6407)

Ghost Squad (Chihiro)
0x948B1DA3 (Retail: 0x3C774A08, Debug: 0x000E80E8)

GTA 3 (Retail)
0xA8FE9DF3 (Retail: 0x0002CA58, Debug: 0x3C7B00B8)
```

src/common/AddressRanges.h
PatrickvL: `#define PAGE_SIZE                           (1 << PAGE_SHIFT) // = 0x00001000 = KB(4)`
or
`#define PAGE_SIZE                           (1 << PAGE_SHIFT) // = 0x00001000 = KiB(4)`
(the latter would apply if `KB` was renamed into `KiB`)

src/common/AddressRanges.h
ergo720: Ok, I'll dismiss it.

src/common/AddressRanges.h
RadWolfie: resolved

src/common/ReserveAddressRanges.h
RadWolfie: resolved

src/common/xbe/Xbe.cpp
RadWolfie: resolved

src/common/xbe/Xbe.cpp
RadWolfie: resolved

src/common/AddressRanges.h
RadWolfie: resolved in "change KB/MB to KiB/MiB" commit.

src/common/AddressRanges.h
PatrickvL: Are you planning to mind the declaration back to the header in this pull request?
If not, we could merge this, since everything else seems to be (well) done now.

RadWolfie: Nope, it will require platform specific changes which do require a separate pull request. I don't think I will work on this.

PatrickvL: Fair enough, you've done great work already, thanks!

Issue #1819, opened 2020-01-18
Ability to build using Winelib


Serentty:
@PatrickvL suggested that I create a new issue for this.

As it stands, Cxbx-Reloaded works fairly well using Wine. However, the fact that it can only be distributed as a Windows EXE makes it harder for distributions to package, plus it means that users are responsible for manually installing Wine in order to run it.

If Cxbx-Reloaded could be built and linked with Winelib, this would allow ELF binaries that could be packaged more easily, and those packages could automatically pull in Wine as a dependency for users. But I think an even greater advantage is the ability to include Linux-specific code and fixes, while still being able to lean on all of those convenient Windows APIs.

LukeUsher:
Copy/paste from other issue:

we've looked into it, the problem is that Cxbx-Reloaded can not be compiled with any compiler that isn't Microsoft Visual C++.

There's effort being taken to allow compilation with GNU based compilers, but doing so requires a large rewrite of significant portions of the codebase, and will take a long time.

Much of the calling convention handling and assembly language parts need completely replacing, since inline assembly using GNU and Microsoft compilers has a different syntax.

Serentty:
Oh yeah, it didn't occur to me that using GCC to compile it might be an issue. Well, I'm in no hurry to see this happen, so it's good to see that this is something being looked at, even if it won't happen anytime soon.

Issue #860, opened 2018-01-14
Work towards portable codebase


PatrickvL:
Cxbx-Reloaded is currently using Windows API's.
It's is build with Visual Studio.
It uses many code constructs of MSVC.
It's written in a (bad) mix of c and c++.
It targets x86 architecture (due to the direct code execution method).
It runs as a 32 bit process under a 64 bit OS.

In order to become portable, all these aspects need to be addressed.

Some possible solutions:

Platform: c++11 or 14 standards (?)
Build system: cmake
Compiler : gcc or LLVM clang
Architecture: Linux x86, Android x86
CPU emulating: unicorn, KVM, HAXM

It's probably wise to create separate issues for all aspects that can be done independently. Let's use this issue as a central collection of options and ideas.

LukeUsher:
For C++ standards, we already target C++17 (We make use some functionality that is not available in earlier standards, so we have that covered already :P)

For a build system, I highly recommend CMake, not only because it is easy to use, but because it can export build scripts/project files for many different development environments and compilers. The same CMakefile can output build scripts for Linux Makefiles, Windows Makefiles (for mingw32) or even full MSVC Solutions/Project files, it's a great option.

For compiler, we should try not to use any compiler specific functionality, any commonly used compiler should be compatible (So gcc, clang and MSVC compilers should all work), this isn't too difficult, just make sure not to use compiler extensions, and if you must, #ifdef them out based on the compiler being used.

CPU emulation i wouldn't worry about just yet, Cross platform across Windows and Linux on X86 is a much easier target (and still requires a huge rework of the codebase to stop using the Windows API functionality directly), but is much less of a headache than porting to other architectures at this point.

We should keep it simple and take it in steps.

That said, the best approach for the future would likely to start a new, blank source code project, migrating Cxbx-Reloaded code into it and being cross platform from the get go, but such a thing is a mammoth task that I doubt everybody would be onboard with :P (Although this is what Cxbx-Reloaded was originally planned to be, it just got a little out of control!)

Fisherman166:
Very rarely does the 'throw out the code base and restart' go well. A lot of time has to be spent fixing all of the new bugs introduced with the rewrite before the project returns to the same level of functionality. A more realistic goal is to make sure that whenever someone touches a piece of code, they leave it better than it started. Over time the code base will naturally improve through many small incremental updates instead of a massive throw it all out and restart methodology.

LukeUsher:
[...]

PR #1859, opened 2020-04-07
Various fixes for standards compliance and compiler portability

darkf:
This is the first of a few PRs I intend to write to port Cxbx-R to Clang. 

The eventual goal here is to have it compile on non-MSVC platforms and be more standards-compliant (#860), facilitate compiling with MinGW (e.g. for compilation on Linux or on Windows without requiring MSVC), and investigate compiling with winelib (#1819). 

This PR is a first wave of attacking this, getting it down to only 50 compilation errors with Clang (down from around ~1000). The rest will be done in future PRs.

It should still compile perfectly fine on MSVC and have no regressions, please test that!


import/OpenXDK/include/xboxkrnl/xboxkrnl.h
```diff
@@ -13,8 +13,8 @@
 #define XBOXKRNL_H
 
 #if defined(__cplusplus)
-extern "C"
-{
+//extern "C"
```
LukeUsher, line 16: This was mostly a leftover from the original OpenXDK base, so it's perfectly fine to remove this

darkf: Went ahead and removed them entirely

src/core/hle/DSOUND/XbDSoundTypes.h
```diff
@@ -201,7 +201,7 @@ XMEDIAPACKET, *PXMEDIAPACKET, *LPXMEDIAPACKET;
 #define XMP_STATUS_PENDING          E_PENDING
 #define XMP_STATUS_FLUSHED          E_ABORT
 #define XMP_STATUS_FAILURE          E_FAIL
-#define XMP_STATUS_RELEASE_CXBXR    0xFFFFFFFF
+#define XMP_STATUS_RELEASE_CXBXR    -1
```
RadWolfie, line 204: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/262627d8-3418-4627-9218-4ffe110850b2

Unsigned integer does not accept any negative values. So this would be incorrect, it will produce a warning by the way.

src/devices/video/EmuNV2A_DEBUG.cpp
```diff
@@ -34,13 +34,13 @@
...
 #define DEBUG_START(DEV) \
-const char *DebugNV_##DEV##(xbaddr addr) \
+const char *DebugNV_##DEV(xbaddr addr) \
 { \
 	switch (addr) {
 #define DEBUG_CASE(a) \
 		case a: return #a;
 #define DEBUG_CASE_EX(a, c) \
-		case a: return #a##c;
+		case a: return #a c;
```
RadWolfie, line 43: This doesn't look right...

RadWolfie:
I'm not sure what's the main reason behind enums needing `: int` afterward. Compilers should be handle it on its own.

src/devices/video/EmuNV2A_DEBUG.cpp
darkf: It's right, `c` is a constant string literal so it gets concatenated, `##` does not append strings. It is an error.

darkf:
@RadWolfie Forward declarations without `: type` are non-standard, because the compiler has no way of knowing the size before they're defined.

See https://stackoverflow.com/a/72599

src/core/hle/DSOUND/XbDSoundTypes.h
darkf: That's true, I reverted it and just made the cast where it's used as an int now.

RadWolfie:
Then in that case, we will need to remove empty enums into an actual functional enums. Since using `int` for everything isn't healthy and reliable to use. (For a separate pull request of course.)

src/devices/video/EmuNV2A_DEBUG.cpp
LukeUsher: Looks good to me!

Issue #1803, opened 2019-12-22
HLSL: AMD GPU Regression


RinMaru:
#1801 Has cause major bugs with AMD Hardware, Mostly BlackScreens

**Reported AMD GPUs**
- Radeon RX 580
- Radeon RX 570
- Radeon RX 560 (reported [here](https://github.com/Cxbx-Reloaded/game-compatibility/issues/954))
- Radeon RX 550
- Radeon RX 530
- Radeon RX 480
- Radeon RX Vega 56
- Radeon Pro Vega 20
- Radeon HD 7970
- Radeon HD 4200 (Vulkan not supported)
- Radeon R9 380X

**Reported Games**
- Vexx AC-005 v1.03/ Menu is rendered but 3D assets are all black
- Crash Tag Team Racing VU-065 v1.02/ 2D elements loaded but crashes at first loading screen
- Crash Wrath Of Cortex VU-003 v1.01/ Some 3D models fail to render
- Xbox Dashboard FFFE0000 v1.269657855/ 2D elements only rendered
- Jet Set Radio Future / Invisible NPCs/character and menu

**_NOTE_** Use [DXVK](https://github.com/doitsujin/dxvk) as a temporary workaround. Use the 32bit d3d9.dll.
The file path will be `dxvk-1.#.tar.gz\dxvk-1.#.tar\dxvk-1.#\x32\d3d9.dll`

hinksta47:
Hi @RinMaru, I can confirm that my AMD Radeon Pro Vega 20 (Apple MBP 2018 - running BootCamp obviously) also has the exact same issues.

Earth271072:
Can confirm I'm having issues with Jet Set Radio Future on my Radeon RX Vega 56

SimonDenton:
Can confirm with RX 580

KuroRaiden:
Graphical issues for JSRF and Ninja Gaiden Black.
**JSRF:**
- Invisible Characters
- Laggy main menu

**Ninja Gaiden Black**
- Strange texture errors 


Hardware:
GPU - Radeon RX 580 Series - Primary/Discrete
CPU - AMD Ryzen 5 2600 Six-Core Processor

suchturista:
Same thing happening for JSRF on my Radeon RX 570. No 3D models.

DeathAdrift:
JSRF on my Radeon RX 550. No 3D models.

RadWolfie:
[...]

PR #1865, opened 2020-04-09
prevent missing polygons with AMD GPU

LukeUsher:
AMD's current GPU drivers seem to have problems when using Shader Model 3.0 with DirectX9

This work-around falls back to Shader Model 2.0 when AMD hardware is detected.

Partial fix/workaround for #1803 

Before:
image

After:
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
@@ -48,11 +48,16 @@ float4 c(int register_number)
 	// Map Xbox [-96, 95] to Host [0, 191]
 	// Account for Xbox's negative constant indexes
     register_number += X_D3DSCM_CORRECTION;
-    if (register_number < 0)
-        return 0;
-    
+
+    // Like Xbox, out-of-range indices are guaranteed to be zero in HLSL
+    // so no need to bounds check negative numbers
+    // if (register_number < 0)
+    //    return 0;
+
+    // If the index is too large, set it to -1 so 0 is returned
+    // Note: returning 0 directly requires many more instructions
     if (register_number >= X_D3DVS_CONSTREG_COUNT) // X_D3DVS_CONSTREG_COUNT
-        return 0;
+        register_number = -1;
 
     return C[register_number];
 }
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -1670,7 +1670,7 @@ VOID EmuD3DInit()
 		// 1002 and 1022 are vendor ids of AMD gpus
 		if (status == D3D_OK && (adapter_info.VendorId == 0x1002 || adapter_info.VendorId == 0x1022)) {
 			g_vs_model = vs_model_2_a;
-			EmuLog(LOG_LEVEL::WARNING, "AMD GPU Detected, falling back to shader model 2.X to prevent missing polygons");
+			EmuLogInit(LOG_LEVEL::WARNING, "AMD GPU Detected, falling back to shader model 2.X to prevent missing polygons");
 		}
 	}
 }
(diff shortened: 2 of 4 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -109,7 +109,7 @@ static DWORD						g_dwPrimPerFrame = 0;	// Number of primitives within one frame
 static Settings::s_video            g_XBVideo;
 
 // D3D based variables
-static IDirect3D                   *g_pDirect3D = nullptr;
+       IDirect3D                   *g_pDirect3D = nullptr;
```
LukeUsher, line 112: static was removed because this variable needed to be available in order to call GetAdapterIdentifier

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -251,14 +252,30 @@ extern HRESULT EmuCompileShader
 	// TODO Can we recompile an optimized shader in the background?
 	UINT flags1 = D3DCOMPILE_OPTIMIZATION_LEVEL0;
 
+	static bool vendor_detected = false;
```
LukeUsher, line 255: Static was used for these variables to ensure they only get initialized once

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -251,14 +252,30 @@ extern HRESULT EmuCompileShader
...
+	static bool gpu_is_amd = false;
+
+	if (!vendor_detected) {
+		vendor_detected = true;
+		D3DADAPTER_IDENTIFIER9 adapter_info;
+		HRESULT status = g_pDirect3D->GetAdapterIdentifier(D3DADAPTER_DEFAULT, 0, &adapter_info);
+		// 1002 and 1022 are vendor ids of AMD gpus
+		if (status == D3D_OK && (adapter_info.VendorId == 0x1002 || adapter_info.VendorId == 0x1022)) {
```
LukeUsher, line 263: These numbers are PCI device IDs for AMD GPUs, AMD are not known to use any other identifier.

LukeUsher:
This seemed to improve some simple cases (dashboard, etc) but JSRF is still broken
Needs further investigation

LukeUsher:
Update: This fix allows simple vertex shaders to compile and function, but some games (like JSRF) use more complex shaders, and shader model 2.0 doesn't have enough slots for them.

The only true fix would be to move our entire codebase to DirectX10/11, which requires a lot of work.

Until then, this is the best we can offer outside of suggesting a wrapper.

RadWolfie:
I added couple of fixes to the pull request per @LukeUsher request.

LukeUsher:
In hindsight, I also have a concern that reducing to sm2.0 will break some games (eg JSRF) for people who currently use dxvk as a work-around for this problem.

Unfortunately, thanks to AMDs buggy drivers, we don't have many options until we are able to port to D3D10/D3D11

PatrickvL:
Perhaps a wrapper can be detected as well?

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -48,11 +48,16 @@ float4 c(int register_number)
...
-        return 0;
-    
+
+    // Like Xbox, out-of-range indices are guaranteed to be zero in HLSL
+    // so no need to bounds check negative numbers
+    // if (register_number < 0)
+    //    return 0;
+
+    // If the index is too large, set it to -1 so 0 is returned
+    // Note: returning 0 directly requires many more instructions
```
PatrickvL, line 58: Lovely optimization, thanks for the idea @NZJenkins!

LukeUsher:
Thanks to @NZJenkins's suggestion to optimize the hlsl shader, JSRF is now functional on AMD with shader model 2.X, this might be a valid solution for AMD afterall!

I would appreciate of some AMD users could test with the games listed in #1803, and then we can merge this.

RadWolfie:
Strange, I have tested commit 4659714 with JSRF with Radeon HD 4200 (integrated graphic). Characters are still missing. Even if I go in the tutorial too. Dashboard isn't affected so far.
image

PR #1866, opened 2020-04-09
LOADER: Create two independent file mappings for a chihiro system

ergo720:
This is intended to fix the assertion I reported in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1851#issuecomment-604477695, which was caused by a corruption of the pfn database which resides on the top 128 MiB of memory on a chihiro system. The problem was that the previous setup was creating two 64 MiB memory blocks from a single file mapping. That’s incorrect, because in this way they will be identical and every change done in a block will immediately be reflected in the other, while we instead want the two blocks to operate independently. This also explains why the pfn database becomes corrupted: as soon as the title writes to the top of the first block, the changes will be reflected on the top of second block, corrupting the database.

### DIFF
--- a/src/common/ReserveAddressRanges.cpp
+++ b/src/common/ReserveAddressRanges.cpp
@@ -54,32 +54,49 @@ bool ReserveMemoryRange(int index, blocks_reserved_t blocks_reserved)
 	const DWORD Protect = XboxAddressRanges[index].InitialMemoryProtection;
 	bool NeedsReservationTracking = false;
 	unsigned int arr_index = BLOCK_REGION_DEVKIT_INDEX_BEGIN;
+	static HANDLE hFileMapping1;
+	static HANDLE hFileMapping2;
+
 #ifdef DEBUG
 	std::printf("DEBUG: ReserveMemoryRange call begin\n");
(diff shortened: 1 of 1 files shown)


RadWolfie:
I'm curious if we can also free the file handles as well?

src/common/ReserveAddressRanges.cpp
```diff
@@ -54,32 +54,49 @@ bool ReserveMemoryRange(int index, blocks_reserved_t blocks_reserved)
...
+
 		case PHYSICAL_MAP2_BASE:
 		case TILED_MEMORY_BASE: {
-			static bool NeedsInitialization = true;
-			static	HANDLE hFileMapping;
-			if (NeedsInitialization) {
-				hFileMapping = CreateFileMapping(
+			static bool NeedsInitializationMap = true;
+
+			if (NeedsInitializationMap) {
```
RadWolfie, line 83: Instead of "NeedsInitializationMap", it would be simple to check if `hFileMapping2` is nullptr. The same for `hFileMapping1` above as well. Since both `hFileMapping1` and `hFileMapping2` are set to static which can be initialize as nullptr.

Plus it would be great to have ability to free `hFileMapping2`. (Either in this or separate pull request.)

Issue #1848, opened 2020-03-19
Migrate to GitHub Actions as sole CI provider


GXTX:
This comes after noticing a pattern of having to update CI configurations for tiny details. Currently CXBX-R supports 3 different CI providers, this issue is to track and gather thoughts about moving to just GitHub Actions. From my eyes at least Actions is actively developed and actively supported, travis until late last year didn't even have a Ubuntu 18.04 runner and still defaults to using 16.04.

Benefits
- Consolidated support requirements
- No need to create 3rd party accounts (note AppVeyor with it being attached to LukeUshers personal account)
- No need to give additional org permissions to 3rd party providers

LukeUsher:
Going forward, I propose that we

1. Drop all CI except github actions
2. Configure actions to build artefacts for all branches and PRs
3. Configure actions to *publish* artefacts from master to Github Releases, no other branches should be published
4. Once all this is configured, merge develop into master, and delete the develop branch.

All work should continue based on master.

RadWolfie:
> Going forward, I propose that we
> 
> 1. Drop all CI except github actions

As in deactivate CIs which I am fine with it.

...

> 3. Configure actions to _publish_ **artifacts** from master to Github Releases, no other branches should be published

Tags should generate official release whilst merged pull request (to master) should mark as snapshot releases. Since I notice xqemu devs were talked about the artifacts aren't accessible to public unless login to the site.

RadWolfie:
See #1698 for possible actions to use for this task.

Margen67:
> 
> 
> > Going forward, I propose that we
> > 
> > 1. Drop all CI except github actions
> 
> As in deactivate CIs which I am fine with it.
> 
If we only deactivate them they will still need to be maintained, otherwise they are useless.

RadWolfie:
Little to no maintenance requirement for third-party CIs. GitHub Action's workflow is fairly new and of course will have more maintenance requirement.

What will cause "more" maintenance to third-party CIs configuration are anything other than builds system. The contributors, which is not us aka the devs, will be responsible for maintaining deactivated CIs configuration file. Mostly the work will on "artifacts" and rarely CMake's generator. That's it.

LukeUsher:
Why should the responsibility of managing CIs that people who fork the repo fall on us?

We should only manage and maintain the CI systems that *we* use.

If any third parties maintain their own forks, setting up CI should be up to them, not us.

[...]

PR #1868, opened 2020-04-13
Remove unused CIs

Margen67:
Related: #1848.

#
Other CIs have disadvantages:
  * Require a seperate account, and setup.
  * Require maintenance even if they're not used, otherwise they will eventually become broken, making them even more useless.
    **Furthermore, they're not active, so there's no way of knowing when or if they're broken from the official Cxbx-Reloaded repository.**
### Travis:
  * Doesn't have VS2019, or artifacts.
### AppVeyor:
  * Can only build one job at once, and is very slow.
    * This is exacerbated by having to build for both VS2017 and VS2019, and other build systems in the future.
### Azure Pipelines:
  * Almost the same as GitHub Actions.


RadWolfie:
> Require a seperate account, and setup.

90% setup is from configuration file. It take a lot of work to setup configuration file. However, linkage to the repo does not.

> Require maintenance even if they're not used, otherwise they will eventually become broken, making them even more useless.

That is actually half false statement. It will continue to work, **as is**, at least 4 years more or less. All we need to do is update the visual studio's generator. Then it will continue to work, **as is**, for another 4 years more or less. Or take the risk to remove the generator name. Finally, continue to work for more than who knows.

> Can only build one job at once, and is very slow.

AppVeyor may will change this in the future due to competitors are using parallel jobs. Visual Studio 2017 won't be included because of doing the job one at a time at the moment. In fact, eventually WINE will have a fix for Visual C++2019 redist by the time AppVeyor change to parallel jobs.


@Margen67 If you are not using any of third-party CIs, then you are not require to submit a pull request for them. Cheers.

Respectfully, this pull request is denied.

PR #1870, opened 2020-04-14
Move develop Branch into master Branch

RadWolfie:
Per developers' discussion, we have agree it is unlikely we will make a stable release yearly more or less. Current development builds are much better and more stable than stale stable release.

With GitHub's Actions feature and preparation from @GXTX and @Margen67, we are able to make a pre-release for every merged pull request within the same repository.


PatrickvL:
LGMT ;)

PR #1871, opened 2020-04-15
LOADER: Enable Loader's UI and Set to Default

RadWolfie:
Title says it all. 😉

We welcome any testers to try out and report back any regressions before we make another pull request into master branch.


PatrickvL:
Looking good

PR #1872, opened 2020-04-15
Near absolute memory control

PatrickvL:
This pull request has been in the works for about two years, and was a joined effort of me, ergo720 and radwolfie.
With this work, Cxbx-Reloaded gains near absolute control over all memory ranges an Xbox requires.
This has been a challenge to achieve under Windows, because applications normally load all sorts of binaries before an application even gets executed. Those binaries tend to get loaded into the address space that we have to control ourselves. We've overcome this by using a tiny loader executable, that claims all those ranges before doing *anything* else, and only then we start our emulation code through a DLL, that gets loaded only after claiming all those memory ranges.

As a side effect, we now deploy three binaries instead of one very big one;
First, the GUI executable, that looks and functions just like before, but can now be stripped of the emulation code.
Second, we deploy the tiny loader executable, described above.
And third, the emulation code, contained in a DLL.

With this setup, we believe to have reached a much better position to build upon in the future;
* Our kernel memory sub system will now hand out the exact same memory ranges to emulated titles as the real kernel would on a real Xbox
* Having a separate GUI allows us to replace it some day without hampering emulation
* We no longer have to link in a huge block of reserved address space, reducing executable size and compilation times.

For now, it's offered with a switch that can disable the new approach. If anyone observes a regression, please retry such a title with the loader option disabled.

Besides required changes to the kernel memory subsystem, no significant other emulation has changed in this pull request. This means that only memory allocation related problems might be fixed because of this work, but otherwise existing emulation behaviour should be mostly the same.

Please report any confirmed regression in a comment on this issue.


LukeUsher:
This is a massive change, and as such will require a longer testing period than most pull-requests that we make. 

Aggressive regression testing is required.

RadWolfie:
Here's the download links for testing. They are from commit 10b7d06's artifacts.
 - [CxbxReloaded-Release-VS2019-PR1871.zip](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/4480320/CxbxReloaded-Release-VS2019-PR1871.zip)
 - [CxbxReloaded-Release-VS2017-PR1871.zip](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/4480319/CxbxReloaded-Release-VS2017-PR1871.zip)
   - For WINE users

src/core/kernel/init/CxbxKrnl.h
```diff
@@ -45,110 +47,54 @@ extern "C" {
...
 // Sizes
 #define ONE_KB 1024
-#define ONE_MB (1024 * 1024)
-#define X64KB 64 * ONE_KB
-
-// NT_TIB (Thread Information Block) offsets - see https://www.microsoft.com/msj/archive/S2CE.aspx
-#define TIB_ArbitraryDataSlot  0x14 // = 20 = offsetof(NT_TIB, ArbitraryUserPointer)
-#define TIB_LinearSelfAddress  0x18 // = 24 = offsetof(NT_TIB, Self)
+#define ONE_MB (1024 * ONE_KB)
+#define X64KB (64 * ONE_KB)
```
revel8n, line 51: Should you use the new defines from AddressRange.h?

src/core/kernel/init/CxbxKrnl.h
```diff
@@ -45,110 +47,54 @@ extern "C" {
...
 // fit in 51 MB. If we ever encounter an even larger XBE, this
 // value will have to be increased likewise (maybe up to 64 MB
-// for XBOX_MEMORY_SIZE or even 128 MB for CHIHIRO_MEMORY_SIZE).
-#define XBE_MAX_VA	(64 * ONE_MB)
+// for XBOX_MEMORY_SIZE or even 128 MB for CHIHIRO_MEMORY_SIZE).
+#ifdef CXBXR_EMU
+#define XBE_MAX_VA  (128 * ONE_MB)
+#else
+#define XBE_MAX_VA  (64 * ONE_MB)
+#endif
```
revel8n, line 89: Should you use the new defines from AddressRange.h?

src/core/kernel/memory-manager/VMManager.cpp
```diff
@@ -56,45 +74,64 @@ bool VirtualMemoryArea::CanBeMergedWith(const VirtualMemoryArea& next) const
 	return false;
 }
 
-void VMManager::Initialize(HANDLE memory_view, HANDLE pagetables_view, int BootFlags)
+void VMManager::Initialize(unsigned int SystemType, int BootFlags, blocks_reserved_t blocks_reserved)
 {
+	if ((BootFlags & BOOT_QUICK_REBOOT) == 0) {
+#ifndef CXBXR_EMU
+		SystemType = g_bIsRetail ? SYSTEM_XBOX : g_bIsChihiro ? SYSTEM_CHIHIRO : SYSTEM_DEVKIT; // TODO: Temporary placeholder until loader is functional.
+#endif
```
revel8n, line 82: Is this placeholder still desired/needed?

src/core/kernel/init/CxbxKrnl.h
RadWolfie: Part of the plan is not to make any unnecessary upstream changes in loader branch. So, once loader branch is merge. All of us can breath easier to clean up within upstream's code. Without worry about continuous rebasing may will have aggressive rebase conflict from a separate branch.

src/core/kernel/init/CxbxKrnl.h
RadWolfie: We or someone will eventually clean this code up in upstream branch.

src/core/kernel/memory-manager/VMManager.cpp
RadWolfie: There's some talk about wanting to keep all-in-one executable. 🤷‍♂️ I mean, it's nice to have a backup just in case.

I'm currently checking if it is possible to replace it with `emulate_system` variable pass down from `CxbxKrnlEmulate` function.

RadWolfie: resolved

revel8n: Ok, was just curious as it looked like the code was passing in the detected system properly, but had not gotten around to testing it yet.

src/core/kernel/memory-manager/VMManager.cpp
```diff
@@ -77,20 +77,12 @@ bool VirtualMemoryArea::CanBeMergedWith(const VirtualMemoryArea& next) const
...
 	}
 	else {
 		g_EmuShared->GetMmLayout(&SystemType);
 	}
 
-#ifndef CXBXR_EMU
-	if (!ReserveAddressRanges(SystemType, blocks_reserved)) {
-		CxbxKrnlCleanup("Failed to reserve required memory ranges!", GetLastError());
-	}
-#endif
```
PatrickvL: This code would best be kept here, I think

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -1130,10 +1130,16 @@ void CxbxKrnlEmulate(unsigned int reserved_systems, blocks_reserved_t blocks_res
 #endif
 
 #ifndef CXBXR_EMU
+		// Only for GUI executable with emulation code.
+		blocks_reserved_t blocks_reserved_gui = { 0 };
+		// Reserve console system's memory ranges before start initialize.
+		if (!ReserveAddressRanges(emulate_system, blocks_reserved_gui)) {
+			CxbxKrnlCleanup("Failed to reserve required memory ranges!", GetLastError());
+		}
```
PatrickvL, line 1138: This code came from VMManager::Initialize, but since that gets called anyway, this code could just as well be kept in there (not be deleted and moved here), because this piece of code is closely related to the VMManager initialization (and it doesn't make much sense calling it in here, in CxbxKrnlEmulate).

src/core/kernel/memory-manager/VMManager.cpp
PatrickvL: Radwolfie explained this was the sole conditional code in this file, so moving it away is better.

src/core/kernel/init/CxbxKrnl.cpp
PatrickvL: Radwolfie explained this was the sole conditional code in that file, so moving it here is better.

gandalfthewhite19890404:
Just for information - there is bug with WWE RAW 2, it is crashing at Press start screen after you are pressing any button. It happens only when UseLoadedExecutable is enabled along with console logging.
That issue was already discussed on discord.

medievil1:
V1.01 morrowind (not GOTY) crashes back to cxbx UI with the loader exe.. works fine with loader unticked

LukeUsher:
@medievil1 morrowind is known to use the Xbox xbe reload functionality, so perhaps launcher has a problem with xbe reload / launching other executables properly? Needs further testing. Innocent Tears is another good test case for this behaviour.

medievil1:
> 
> 
> @medievil1 morrowind is known to use the Xbox xbe reload functionality, so perhaps launcher has a problem with xbe reload / launching other executables properly? Needs further testing. Innocent Tears is another good test case for this behaviour.

no it loads into it, it crashes as you talk to the character when you go to name yourself...

RadWolfie:
> it crashes as you talk to the character when you go to name yourself

Sound identical to WWE RAW 2's hidden bug. For some reason, invalid pointer are letting it, from `EmuX86_Opcode_xxxx`, through when it shouldn't have. Therefore is not a regression, yet we, or maybe just I, will need to look into it then make a pull request *after* this pull request is merge.

literalmente-game:
Tested with Windows 7, Intel CPU (i3 2100) and NVidia GPU (GTX 1050 TI), no regressions or changes on the behavior when compared to older builds:
* ATV Quad Power Racing 2
* Battlestar Galactica
* Blackstone
* Burnout
* Chicago Enforcer
* Egg Mania
* Fatal Frame
* Futurama
* Galleon
* Gun Valkyrie
* Halo
* Jet Set Radio Future
* MechAssault
* Miami Vice
* Smashing Drive
* Tony Hawk's Pro Skater 2X

.travis.yml
```diff
@@ -18,3 +18,25 @@ script:
...
+#deploy:
+#  provider: releases
+#  api_key: $GH_TOKEN
+#  file: "export/${configuration}.7z"
+#  skip_cleanup: true
+#  on:
+#    # Can disable this to deploy all builds appears as "untagged-buildhash" on GH
+#    tags: true
+#    branch: develop
+#    condition: $configuration = Release
```
Margen67, line 42: This shouldn't have been added.

PR #1888, opened 2020-05-07
Cleanup EmuCreateDeviceProxy & Clarify 'Automatic' option

LukeUsher:
The key part of this pull-request is simplifying/removing operations from our CreateDevice emulation that are managed by the Xbox side of things, and should *NOT* be applied to the host backbuffer.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -121,7 +121,6 @@ static IDirect3D                   *g_pDirect3D = nullptr;
 static IDirect3DVertexBuffer       *g_pDummyBuffer = nullptr;  // Dummy buffer, used to set unused stream sources with
 static IDirect3DIndexBuffer        *g_pClosingLineLoopHostIndexBuffer = nullptr;
 static IDirect3DIndexBuffer        *g_pQuadToTriangleHostIndexBuffer = nullptr;
-static IDirect3DSurface            *g_pDefaultHostDepthBufferSurface = nullptr;
 
 static bool                         g_bEnableHostQueryVisibilityTest = true;
 static std::stack<IDirect3DQuery*>  g_HostQueryVisibilityTests;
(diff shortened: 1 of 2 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2239,108 +2238,67 @@ static DWORD WINAPI EmuCreateDeviceProxy(LPVOID)
 				// Apply render scale factor for high-resolution rendering
 				g_RenderScaleFactor = g_XBVideo.renderScaleFactor;
 
-                if(g_EmuCDPD.XboxPresentationParameters.BufferSurfaces[0] != xbnullptr)
-                    EmuLog(LOG_LEVEL::WARNING, "BufferSurfaces[0] : 0x%.08X", g_EmuCDPD.XboxPresentationParameters.BufferSurfaces[0]);
```
LukeUsher, line 2243: Removed because this doesn't matter for us at all: The xbox DirectX handles this for us, and the game passes the correct targets to SetRenderTarget.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2239,108 +2238,67 @@ static DWORD WINAPI EmuCreateDeviceProxy(LPVOID)
...
 				g_RenderScaleFactor = g_XBVideo.renderScaleFactor;
 
-                if(g_EmuCDPD.XboxPresentationParameters.BufferSurfaces[0] != xbnullptr)
-                    EmuLog(LOG_LEVEL::WARNING, "BufferSurfaces[0] : 0x%.08X", g_EmuCDPD.XboxPresentationParameters.BufferSurfaces[0]);
-
-                if(g_EmuCDPD.XboxPresentationParameters.DepthStencilSurface != xbnullptr)
-                    EmuLog(LOG_LEVEL::WARNING, "DepthStencilSurface : 0x%.08X", g_EmuCDPD.XboxPresentationParameters.DepthStencilSurface);
-
-				// Make a binary copy of the Xbox D3DPRESENT_PARAMETERS
-				memcpy(&g_EmuCDPD.HostPresentationParameters, &(g_EmuCDPD.XboxPresentationParameters), sizeof(D3DPRESENT_PARAMETERS));
```
LukeUsher, line 2249: This was removed due to it having the potential to cause issues if the Windows structure ever changes in layout.

Also, it allows us to properly control what goes into the Host Presentation Parameters, removing many potential causes for CreateDevice failurel

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2540,20 +2475,6 @@ static DWORD WINAPI EmuCreateDeviceProxy(LPVOID)
                     free(lpCodes);						
 				}
 
-/* Disabled for now, as this has no other side-effect (other then adding a reference count that's never Released)
-                // update render target cache
-				IDirect3DSurface *pCurrentHostRenderTarget = nullptr;
```
LukeUsher, line 2545: Removed for same reason as above: Xbox directx manages this, we shouldn't need to know or care at CreateDevice time.

PR #1891, opened 2020-05-07
One more Wine-related edit

Serentty:
Somehow I missed the very first line of the readme, which also makes it sound Windows-exclusive, in my last PR. Sorry for the spam.


Serentty:
Thanks!

PR #1901, opened 2020-05-11
Partial Implement of Xbox DSound's Class Structure

RadWolfie:
Currently, only CDirectSoundVoice data are implemented for audio format, pitch, volume, and headroom. This remove couple of placeholder from host's DirectSound buffer class and more accessible for both xbox executable and future audio plugins. However I did add other xbox structure placeholders which does need more research from LLE's end, for time being this is what I have.

With this fix, it resolve WWE RAW 2's hidden bug cause invalid pitch getter value. Which should produce a fatal crash instead. All of this was discovered by loader project plus kernel console enabled.


gandalfthewhite19890404:
Game now has correct pitch in voices.
Had got a one crash in menu, cannot reproduce it again.
And if you have enabled loader with logging level INFO - game will crash at pop-up windows appearing moment.

PatrickvL:
Could you rename `XbInternalStruct` into something a little more sound-specific?

RadWolfie:
As JayFoxRox would say something like, use full path is all you need. `#include "core\hle\DSOUND\common\XbInternalStruct.hpp"` would be easy to understand. And it is unnecessary to do such as `DSOUND\DSOUND.hpp` when you could use `DSOUND\manager.hpp`, `DSOUND\modules.hpp`, or the like. In this case, it is currently combined internal structures within DirectSound class.

Even if you add another `XbInternalStruct.hpp` in a different file path, they won't conflict each other.

Beside, I do need to relocate structures into their own header file.

src/core/hle/DSOUND/XbDSoundTypes.h
```diff
@@ -369,6 +369,30 @@ struct X_DSVOICEPROPS {
...
 };
+
+// Convert frequency to pitch helper
+static inline int32_t converter_freq2pitch(uint32_t freq) {
+    // NOTE: pitch = 0 is equal to 48 KHz.
+    /* For research purpose of how to convert frequency to pitch and back to frequency.
+    // Edit hertz variable to see the result.
+    float hertz = 12000.0f;
+
+    float hertzRatio = 48000.0f;
```
PatrickvL, line 380: this 48 kHz is actually the *base frequency*

src/core/hle/DSOUND/XbDSoundTypes.h
```diff
@@ -369,6 +369,30 @@ struct X_DSVOICEPROPS {
...
+    float pitchRatio = 4096.0f;
```
PatrickvL, line 381: this 4096 is the pitch per octave

src/core/hle/DSOUND/XbDSoundTypes.h
```diff
@@ -369,6 +369,30 @@ struct X_DSVOICEPROPS {
...
+
+    // Convert hertz to pitch
+    float pitch = log2(hertz / hertzRatio) * pitchRatio;
```
PatrickvL, line 384: The `log2(hertz / hertzRatio)` intermediate result could be put in a `float octave` variable.

Then pitch could be declared as : `uint32_t pitch = trunc(pitch_per_octave * octave);`

src/core/hle/DSOUND/common/XbInternalDSVoice.cpp
```diff
@@ -0,0 +1,213 @@
...
+        funcs.SetFormat = reinterpret_cast<pSetFormat>(::SetFormat_4034_lower<settings_template>);
+        funcs.GetFrequencyDefault = reinterpret_cast<pGetUint32>(::GetFrequencyDefault_4034_lower<settings_template>);
+        funcs.GetPitch = reinterpret_cast<pGetInt32>(::GetPitch<settings_template>);
+        funcs.SetPitch = reinterpret_cast<pSetInt32>(::SetPitch<settings_template>);
+        funcs.GetVolume = reinterpret_cast<pGetInt32>(::GetVolume<settings_template>);
+        funcs.SetVolume = reinterpret_cast<pSetInt32>(::SetVolume<settings_template>);
+        funcs.GetHeadroom = reinterpret_cast<pGetUint32>(::GetHeadroom<settings_template>);
+        funcs.SetHeadroom = reinterpret_cast<pSetUint32>(::SetHeadroom<settings_template>);
+    }
+    else if (g_LibVersion_DSOUND == 4039) {
```
PatrickvL, line 173: Judging by the else block that uses 4134, either this if needs to accept version 4039 up to (but excluding) 4134, OR the else block should use `_r4040_and_up` (or a combination of the above if there's another crossover point)

src/core/hle/DSOUND/common/XbInternalStruct.hpp
```diff
@@ -0,0 +1,244 @@
...
+};
+
+struct CUnknownGenericManager {
+    // construct vtable (or grab ptr to existing)
+    CUnknownGenericManager() : ref_count(1) {}
+
+    // all virtual functions are stored in local offset 0x00's pointer
+    virtual ~CUnknownGenericManager() {};                   // 0x00
+    virtual uint32_t WINAPI AddRef() { return ++ref_count; };      // 0x04
+    virtual uint32_t WINAPI Release() { return --ref_count; };     // 0x08
```
PatrickvL, line 54: Surely when ref_count hits zero, some form of release should take place, to avoid memory leaks?!?

src/core/hle/DSOUND/common/XbInternalStruct.hpp
```diff
@@ -0,0 +1,244 @@
...
+    uint32_t                cbSize;
+    int32_t                 nSamplesPerSec;
+    uint32_t                bitsPerSample;
+};
+
+struct CUnknownGenericManager {
+    // construct vtable (or grab ptr to existing)
+    CUnknownGenericManager() : ref_count(1) {}
+
+    // all virtual functions are stored in local offset 0x00's pointer
```
PatrickvL, line 51: Better call this 0x00 member by it's name : the VMT (Virtual Method Table) pointer.

src/core/hle/DSOUND/common/XbInternalStruct.hpp
```diff
@@ -0,0 +1,244 @@
...
+    virtual ~CUnknownGenericManager() {};                   // 0x00
+    virtual uint32_t WINAPI AddRef() { return ++ref_count; };      // 0x04
```
PatrickvL, line 53: What are the end-of-line comments indicating?
That the destructor is the first member in the VMT?
and that AddRef is second and Release is third?

In that case, be sure people don't mistakenly assume the last one (`//0x08`) relates to the static_assert below!

src/core/hle/DSOUND/common/XbInternalStruct.hpp
```diff
@@ -0,0 +1,244 @@
...
+    virtual uint32_t WINAPI Release() { return --ref_count; };     // 0x08
+
+    uint32_t ref_count;                                     // 0x04
+};
+// Require to verify there is no other unknown additional data by compiler itself.
+static_assert(sizeof(CUnknownGenericManager) == 0x08);
```
PatrickvL, line 59: Best indicate that this data is comprised of the VMT pointer (for 4 bytes) and the ref_count field (also for 4 bytes).

PatrickvL: Same comments apply to all similar constructs in this Pull Request

src/core/hle/DSOUND/common/XbInternalStruct.hpp
```diff
@@ -0,0 +1,244 @@
...
+
+struct CMcpxVoiceClient: CUnknownTemplate {
+    CMcpxVoiceClient() : settings(default_settings) {};
+    // all virtual functions are stored in local offset 0x00's pointer
+    virtual ~CMcpxVoiceClient() {};
+
+    // CUnknownTemplate                                     // 0x00 - ???
+    struct _settings
+    {
+        uint32_t            Unknown2[(0x300-8)/4];          // 0x08 - ???
```
PatrickvL, line 87: Maybe it's better to declare all unknown data as `uint8_t` arrays, so that the division by 4 isn't needed.
This would also have the benefit it could be used to cover non-power-of-4 sized unknowns.

src/core/hle/DSOUND/common/windows/WFXformat.hpp
```diff
@@ -332,5 +334,13 @@ static CODEC_FORMAT WFXformat_SyncHostFormat(
             WFXformat_GeneratePCMFormat(2, 44100, 16, Host_wfxFormat);
         }
     }
+    // Forward xbox format to internal XTL::CDirectSoundVoice class.
+    xb_format.audio_codec = (codec_format_ret == CF_XADPCM ? WAVE_FORMAT_XBOX_ADPCM : WAVE_FORMAT_PCM);
+    xb_format.nChannels = Host_wfxFormat->Format.nChannels;
+    xb_format.cbSize = (codec_format_ret == CF_XADPCM ? 4 : 0);
```
PatrickvL, line 340: Since some xdk versions seem to store up to 32 bits in `cbSize`, it'd be good to explain (through a comment or constant symbol) what this literal numeric value of 4 means.

src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
```diff
@@ -230,12 +228,16 @@ HRESULT WINAPI XTL::EMUPATCH(DirectSoundCreateStream)
             DSBufferDesc.dwFlags |= DSBCAPS_CTRLPAN;
         }
 
+        // TODO: Garbage Collection
```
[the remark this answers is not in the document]
RadWolfie: Good question, yet I don't have the answer. I'm not sure what the original creator meant by this. Perhaps for `CDirectSoundStream_Release` patch or in this constructor patch? Beside, there are still more work to do with `X_CDirectSoundStream` into what I had done for `EmuDirectSoundBuffer`. Actually... I didn't make a destructor to handle internally.

src/core/hle/DSOUND/common/XbInternalStruct.hpp
RadWolfie: It's only a placeholder. Yet does need to add interface to the parent's virtual function AddRef/Release. I'm not sure how to do this task at the moment. Yet it only will happen in a child class. Also, none of us haven't experience this type of crash yet.

Perhaps I should add a TODO messages here.

src/core/hle/DSOUND/XbDSoundTypes.h
```diff
@@ -369,6 +369,30 @@ struct X_DSVOICEPROPS {
     LONG lI3DL2DirectVolume;
     LONG lI3DL2RoomVolume;
 };
+
+// Convert frequency to pitch helper
```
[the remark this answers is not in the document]
RadWolfie: Here and below are only a move outside of DirectSound plugin.

Although, I should move this into common folder. 🤔

src/core/hle/DSOUND/common/XbInternalStruct.hpp
RadWolfie: Odd, Visual Studio says `_vfptr` while debugging the program.

src/core/hle/DSOUND/common/XbInternalStruct.hpp
RadWolfie: Which do you rather have?
1) `uint8_t` reader which can make nonsense values. **or**
2) `uint32_t` reader which can be easier to read from and know what value it may actually means.

Also, most of the time they are 4 bytes size. It's better to follow CPU architecture max size for integer's and pointer's size wise.

src/core/hle/DSOUND/common/windows/WFXformat.hpp
RadWolfie: Direct translation from reverse engineered functions. And Xbox does not store wave extensible in audio_format structure. Wave extensible format does call xadpcm and pcm functions to do the translation to audio_format structure.

src/core/hle/DSOUND/common/XbInternalStruct.hpp
PatrickvL: That's still the VMT : https://stackoverflow.com/questions/14648393/what-does-type-mean-for-vfptr

src/core/hle/DSOUND/common/XbInternalStruct.hpp
RadWolfie: Huh? It's common sense. We don't need to add "Offset 0x__" or "VMT offset 0x__". And the first comment has explain it.

src/core/hle/DSOUND/XbDSoundTypes.h
RadWolfie: Pitch is not unsigned type. It is a signed type by the way.

RadWolfie: And beside, the documentation is accurate for optimized method.

src/core/hle/DSOUND/common/XbInternalStruct.hpp
PatrickvL: I wouldn't have commented if it was obvious. But if you want to leave it like this, fine.

src/core/hle/DSOUND/common/windows/WFXformat.hpp
PatrickvL: Okay, so you're saying we don't know what the 4 means? Could it be the size of a header or something like that?

src/core/hle/DSOUND/XbDSoundTypes.h
PatrickvL: Again, I only comment with the intend to help get things clarified even more. If you think it's fine like this, then just ignore these remarks.

src/core/hle/DSOUND/common/XbInternalStruct.hpp
PatrickvL: Agreed on 2 when unknown members are generally 32 bit and aligned like that.
Since this will probably be the most common case, leaving it at an `uint32_t` array is probably best, until we hit a case where alignment mismatches.

src/core/hle/DSOUND/common/windows/WFXformat.hpp
RadWolfie: `sizeof(XTL::XBOXADPCMWAVEFORMAT) - sizeof(WAVEFORMATEX)` is always 2 to be insert into cbSize value from title's end. Doesn't matter which xdk revisions was used. Internal function did this: (relative to DSOUND section virtual address base)
`0000919B - C7 40 08 04000000     - mov [eax+08],00000004        (is set to audio_format.cbSize)`

I'm only carry over what values were set. Compare to PCM function does:
```
000090FA - 0FB7 51 0E            - movzx edx,word ptr [ecx+0E]  (get wfx_format.cbSize)
000090FE - 89 50 08              - mov [eax+08],edx             (is set to audio_format.cbSize)
```

cbSize might be a debug usage for Xbox dev team. I'm not sure entirely since we don't have a functional LLE audio support. Beside, PCM format only support 0 for cbSize. I'm using default for that. Since occasionally wfx_format from titles aren't always fill in every member values. ***And internal functions are only nitpicking certain member values, not all of them, that's it.***

src/core/hle/DSOUND/common/XbInternalDSVoice.cpp
RadWolfie: For now, I'll leave it as-is since 4040 - 4133 revisions are remain unknown. Unless there are titles or such dashboard has it.

src/core/hle/DSOUND/common/XbInternalStruct.hpp
RadWolfie: Added todo message. However, it is still not resolved until the later pull request.

PR #1905, opened 2020-05-13
Separate Invalid Xbe Signature popup from Admin Privilege Config Bypass

RadWolfie:
Fixed pull request #1523's incomplete work.

I have compiled, verified UI is display correctly, and toggle check with modified xbe file. Plus checked settings.ini file is saved/load properly.

Preview:
image


PatrickvL:
Would there be a version upgrade required for the change in binary format of our settings?

RadWolfie:
Nope, as you can see in EmuShared.h header file. GUI settings are not stored in shared memory nor is a requirement between reboots.
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/ba980d249df1154acf97a48e0350ff3d0bc0ad30/src/common/win32/EmuShared.h#L290-L296

PatrickvL:
> Nope, as you can see in EmuShared.h header file. GUI settings are not stored in shared memory nor is a requirement between reboots.

Okay, great. Thanks. Then I'll merge this.

_Do be aware that now that GUI and emulation can be separate (EXE vs DLL, through the loader approach), both need to be compiled with the same code, otherwise they would see different compiled versions of the settings data._

RadWolfie:
> Do be aware that now that GUI and emulation can be separate (EXE vs DLL, through the loader approach), both need to be compiled with the same code, otherwise they would see different compiled versions of the settings data.

Um... it doesn't. Structure does not change between optimized and non-optimized. Only custom paddings if we manually set it from compiler's end. The class type itself do change and is not compatible between processes (exes + dlls).

PatrickvL:
What I meant was, that if someone would extend the settings like you did now, then only compiled the GUI EXE, *doesn't* compile the emulation DLL, and would then emulate a title using the DLL, the DLL would still use it's prior definition of the settings, while the GUI would be using an updated definition. 
In some cases, that could lead to using misaligned settings data.

We've been through this more times than I can count in the past (when we only had the Cxbx-inherited EXE+DLL architecture).

Anyway, maybe this doesn't apply to the settings changed in this PR, and it doesn't warrant a whole discussion. It was just a friendly reminder of a potential issue we need to be aware of, until we solve the whole inter-process settings debacle reliably once and for all ;)

PR #1911, opened 2020-05-25
Implement ExReadWriteLock Kernel functions

Fisherman166:
I wrote kernel tests for these functions in the test suite, and then implemented the functions in CXBX-R. See this PR for the kernel tests: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/67. The ReadWriteLockExclusive test passes on CXBX-R with this PR, but ReadWriteLockShared does not pass due to KeReleaseSempahore not being implemented.

With the exception of ReadWriteLockShared requiring KeReleaseSempahore to be implemented before it can fully be enabled, the rest of the functionality is 100% there. Once KeReleaseSempahore is implemented the #if 0 can be removed from ReadWriteLockShared to have full functionality.

See this logfile for the current state of the test suite on CXBX-R: [KrnlDebug.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/4679182/KrnlDebug.txt)


src/core/kernel/exports/EmuKrnlEx.cpp
```diff
@@ -144,34 +144,82 @@ static bool eeprom_data_is_valid(xboxkrnl::XC_VALUE_INDEX index)
...
+			RestoreInterruptMode(interrupt_mode);
+#if 0 //FIXME - Enable once KeReleaseSempahore is implemented (used in ExFreeReadWriteLock for Sharedlocks).
+			KeWaitForSingleObject(
+				&ReadWriteLock->ReaderSemaphore,
+				Executive,
+				0,
+				0,
+				0
+			);
+#endif
```
PatrickvL, line 197: These exact same statements also appears in ànother code path.

Perhaps you could refactor this implementation using return statements, so that shared code paths are condensed into one?

src/core/kernel/exports/EmuKrnlEx.cpp
```diff
@@ -144,34 +144,82 @@ static bool eeprom_data_is_valid(xboxkrnl::XC_VALUE_INDEX index)
...
+		}
+		else {
+			if (ReadWriteLock->WritersWaitingCount == 0) {
+				ReadWriteLock->ReadersEntryCount++;
+				RestoreInterruptMode(interrupt_mode);
```
PatrickvL, line 202: These exact same statements also appears in ànother code path.

Perhaps you could refactor this implementation using return statements, so that shared code paths are condensed into one?

src/core/kernel/exports/EmuKrnlEx.cpp
```diff
@@ -144,34 +144,82 @@ static bool eeprom_data_is_valid(xboxkrnl::XC_VALUE_INDEX index)
...
+#if 0 //FIXME - Enable once KeReleaseSempahore is implemented (used in ExFreeReadWriteLock for Sharedlocks).
+				KeWaitForSingleObject(
+					&ReadWriteLock->ReaderSemaphore,
+					Executive,
+					0,
+					0,
+					0
+				);
+#endif
```
PatrickvL, line 215: This is the second appearance of the same code. Consider refactoring so it'll only appear once.

src/core/kernel/exports/EmuKrnlEx.cpp
```diff
@@ -144,34 +144,82 @@ static bool eeprom_data_is_valid(xboxkrnl::XC_VALUE_INDEX index)
...
+			}
+		}
+	}
+	else {
+		ReadWriteLock->ReadersEntryCount++;
+		RestoreInterruptMode(interrupt_mode);
```
PatrickvL, line 221: This is the second appearance of the same code. Consider refactoring so it'll only appear once.

src/core/kernel/exports/EmuKrnlEx.cpp
```diff
@@ -553,16 +601,46 @@ XBSYSAPI EXPORTNUM(27) xboxkrnl::VOID NTAPI xboxkrnl::ExRaiseStatus
...
+				RestoreInterruptMode(interrupt_mode);
+				KeReleaseSemaphore(&ReadWriteLock->ReaderSemaphore, 1, (BOOLEAN)temp_readers_waiting, 0);
+			}
+		}
+		else {
+			ReadWriteLock->ReadersEntryCount--;
+			if(ReadWriteLock->ReadersEntryCount == 0) {
+				ReadWriteLock->WritersWaitingCount--;
+				RestoreInterruptMode(interrupt_mode);
+				KeSetEvent(&ReadWriteLock->WriterEvent, 1, 0);
```
PatrickvL, line 633: This is the second appearance of the same code. Consider refactoring so it'll only appear once.

src/core/kernel/exports/EmuKrnlEx.cpp
```diff
@@ -553,16 +601,46 @@ XBSYSAPI EXPORTNUM(27) xboxkrnl::VOID NTAPI xboxkrnl::ExRaiseStatus
...
-
-	RETURN(S_OK);
+	bool interrupt_mode = DisableInterrupts();
+	ReadWriteLock->LockCount--;
+	if (ReadWriteLock->LockCount >= 0) {
+		if(ReadWriteLock->ReadersEntryCount == 0) {
+			if(ReadWriteLock->ReadersWaitingCount == 0) {
+				ReadWriteLock->WritersWaitingCount--;
+				RestoreInterruptMode(interrupt_mode);
+				KeSetEvent(&ReadWriteLock->WriterEvent, 1, 0);
```
PatrickvL, line 618: These exact same statements also appears in ànother code path.

Perhaps you could refactor this implementation using return statements, so that shared code paths are condensed into one?

PatrickvL, line 618: About the call to `KeSetEvent`: There's another call identical to this, in `RtlLeaveCriticalSection` where it's written as :
`KeSetEvent((PRKEVENT)CriticalSection, (KPRIORITY)1, (BOOLEAN)0);`
(Whereby `(BOOLEAN)0` could also be changed into `FALSE`).

Would it be possible to declare `KPRIORITY` 1 (one) as an enumerate type?
Or alternatively, maybe abstract away similar calls into a define, something like `SetKernelEvent`?

Fisherman166: This one I've had some heart ache about. I've been trying to figure out a good way to write it better than it is but the modification of ReaderEntryCount-- makes it a bit tough. Do you have a suggestion on how you'd like to see it modified with returns? I'll keep thinking in the meantime.

PatrickvL: If you would consider using a return statement, then a first step to simplify this, would be to invert the first if statement, move the final else block inside is true path, and end with a return statement, so that the rest of the code can be moved one indentation to the left.

Then, all remaining paths that steer unique could end with a return statement as well, allowing the duplicate code to appear only once at the bottom of this implementation

LanHikariDS:
Last night's commit broke the FMV in Halo, now only showing a single color from the FMV across the entire screen.

RadWolfie:
@LanHikariDS Sounds like you're comparing this pull request against draft pull request. Please verify [releases page](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/releases) does match your comparison build's title hash.

Otherwise, we do not know what you meant by "Last night's commit" since it can be anything and we're all living in different timezone. Plus last pre-release is 9 days ago. I am leaning on false positive report right now.

LanHikariDS:
@RadWolfie I apologize, I posted on the wrong page, and even then, the information was incorrect and I thought I deleted it. Post has now been deleted.

src/core/kernel/exports/EmuKrnlEx.cpp
Fisherman166: Looking at ReactOs and official Microsoft documentation, there is no names for the KPRIORITY values. ReactOs uses magic numbers for this field so I don't feel like there is a meaningful enum to create here unfortunately.

Fisherman166: I used your suggestion to rewrite the function with returns and I think it hurts the readability of the code. Even as the person who reverse engineered the call from my kernel dump, my original code is MUCH easier to follow in my opinion. I will push my refactored function if you prefer it though.

```
bool interrupt_mode = DisableInterrupts();

	ReadWriteLock->LockCount--;
	if (ReadWriteLock->LockCount == -1) {
		ReadWriteLock->ReadersEntryCount = 0;
		RestoreInterruptMode(interrupt_mode);
		return;
	}

	if ((ReadWriteLock->ReadersEntryCount == 0) && (ReadWriteLock->ReadersWaitingCount != 0)) {
		ULONG temp_readers_waiting = ReadWriteLock->ReadersWaitingCount;
		ReadWriteLock->ReadersEntryCount = ReadWriteLock->ReadersWaitingCount;
		ReadWriteLock->ReadersWaitingCount = 0;
		RestoreInterruptMode(interrupt_mode);
		KeReleaseSemaphore(&ReadWriteLock->ReaderSemaphore, 1, (BOOLEAN)temp_readers_waiting, 0);
		return;
	}

	// This has to come after the if statement from above.
	if (ReadWriteLock->ReadersEntryCount != 0) {
		ReadWriteLock->ReadersEntryCount--;
	}

	if (ReadWriteLock->ReadersEntryCount != 0) {
		RestoreInterruptMode(interrupt_mode);
		return;
	}

	ReadWriteLock->WritersWaitingCount--;
	RestoreInterruptMode(interrupt_mode);
	KeSetEvent(&ReadWriteLock->WriterEvent, 1, 0);
```

PatrickvL: In your rewritten code, the second test on ReadersEntryCount alone can be moved inside the first, making this quite a concise implementation already. You could also split the if on ReadersEntryCount and ReadersWaitingCount in two, allowing an else block that won't need the test on ReadersEntryCount anymore, reducing the number of if statements even more.

PatrickvL: It would look a bit like this (taking the liberty to also use inline expressions:
```
	bool interrupt_mode = DisableInterrupts();

	if (--ReadWriteLock->LockCount == -1) {
		ReadWriteLock->ReadersEntryCount = 0;
		RestoreInterruptMode(interrupt_mode);
		return;
	}

	if (ReadWriteLock->ReadersEntryCount == 0) {
		if (ReadWriteLock->ReadersWaitingCount != 0)) {
			ULONG temp_readers_waiting = ReadWriteLock->ReadersWaitingCount;
			ReadWriteLock->ReadersEntryCount = ReadWriteLock->ReadersWaitingCount;
			ReadWriteLock->ReadersWaitingCount = 0;
			RestoreInterruptMode(interrupt_mode);
			KeReleaseSemaphore(&ReadWriteLock->ReaderSemaphore, 1, (BOOLEAN)temp_readers_waiting, 0);
			return;
		}
	} else {
		if (--ReadWriteLock->ReadersEntryCount != 0) {
			RestoreInterruptMode(interrupt_mode);
			return;
		}
	}

	ReadWriteLock->WritersWaitingCount--;
	RestoreInterruptMode(interrupt_mode);
	KeSetEvent(&ReadWriteLock->WriterEvent, 1, 0);
```

PatrickvL: Note, how this also avoids a bug!

The previous `if (ReadWriteLock->LockCount >= 0)` check would always evaluate to `true`, when `LockCount` is declared as an unsigned `LONG` (which it is)!

PatrickvL: Okay, magic number it'll be ;)

But what about applying a `SetKernelEvent` short-hand everywhere this appears?

`#define SetKernelEvent(e) KeSetEvent((PRKEVENT)e, (KPRIORITY)1, (BOOLEAN)0)`

Allows writing just `SetKernelEvent(CriticalSection)` here, and in other places where the same construct appears.

import/OpenXDK/include/xboxkrnl/xboxkrnl.h
```diff
@@ -1279,8 +1279,8 @@ EVENT_BASIC_INFORMATION, *PEVENT_BASIC_INFORMATION;
 // ******************************************************************
 typedef struct _KSEMAPHORE
 {
-	DISPATCHER_HEADER Header;
-	LONG Limit;
+	DISPATCHER_HEADER Header; // 0x00
+	LONG Limit;               // 0x10
 }
```
PatrickvL, line 1284: Consider adding a size indicator at the end of the struct as well.

import/OpenXDK/include/xboxkrnl/xboxkrnl.h
```diff
@@ -1310,12 +1310,12 @@ MUTANT_BASIC_INFORMATION, *PMUTANT_BASIC_INFORMATION;
...
-	ULONG ReadersEntryCount;
-	KEVENT WriterEvent;
-	KSEMAPHORE ReaderSemaphore;
+	LONG LockCount;             // 0x00
+	ULONG WritersWaitingCount;  // 0x04
+	ULONG ReadersWaitingCount;  // 0x08
+	ULONG ReadersEntryCount;    // 0x0C
+	KEVENT WriterEvent;         // 0x10
+	KSEMAPHORE ReaderSemaphore; // 0x20
 }
```
PatrickvL, line 1319: Consider adding a size indicator at the end of the struct as well.

src/core/kernel/exports/EmuKrnlEx.cpp
ergo720: I don't think it matters too much if ReactOS or the official MS documentation use magic numbers here, we don't have to strictly follow them. Below is a possible implementation of the enumerator:
```
enum class KPRIORITY : LONG {
	zero,
	one,
};
```

ergo720: Also, `KeSetEvent` is used here as well https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/e6b2ed88af52207e3df544b692ee7acb2ef521f3/src/core/hle/XAPI/Xapi.cpp#L800

PatrickvL: Actually, that call to `KeSetEvent` is passing in a NULL to the `Increment` argument, but that argument isn't being used yet (there's so many unfinished kernel API implementations, it's a wonder things work as well as they do right now).

Fisherman166: @ergo720 - What does creating that enum gain other than code bloat? Writing 'one' is no more descriptive than '1'. It still does not tell the reader what a KPRIORITY of 1 means.

I think the NULL turns into a 0 for the increment value. KPRIORITY is a LONG under the hood so that piece of code should change the NULL to a 0.

PatrickvL: This discussion it only slightly related to this Pull Request, so let's try to remember it, and call this one finished!

ergo720: KPRIORITY is the thread priority. Source https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-kequeryprioritythread?redirectedfrom=MSDN "KeQueryPriorityThread returns the current priority of the specified thread." and also https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-kesetevent "Increment: Specifies the priority increment to be applied if setting the event causes a wait to be satisfied."

PatrickvL:
Thanks! I hope you'll consider implementing more API's and tests ?

PatrickvL:
One more thing : This PR is ready to merge, but it's probably wise to await confirmation this doesn't cause regressions in any important software title...

medievil1:
noticed zero differences in the games I currently have ripped

PR #1913, opened 2020-05-31
Improve SEH/VEH support for compilers

### LINKED ISSUE #1860: Replace usage of SEH
Several places in the emulation rely on SEH exception handling (`__try` / `__except`) for things like catching access violations for MMIO, but these aren't really supported on anything other than Windows+MSVC.

If they can't be replaced by more apropos Win32 API calls, perhaps use something like [AddVectoredExceptionHandler](https://docs.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-addvectoredexceptionhandler) for them instead.

RadWolfie:
close #1860 

There's only two codebase does require SEH unfortunately. However, I have implement macro check to enable support for other compilers. Other compilers using VEH for error report system could make false positive report which is untested. If other compilers do produce false positive report, then we need to disable it.

And other SEH requirement is to allow Windows' users without winpcap driver to able run emulation. It may be possible to implement VEH support. Currently, I don't have have a method for it.

---
Since Windows' C++ exception is using SEH, the only solution without apply per thread's SEH method is to use `SetUnhandledExceptionFilter` function. Plus there is a chance error report system does not get caught in unhandled exception filter. With VEH testing, it does get caught every time and working as intended except for Windows' C++ exception's SEH usage conflict.

### DIFF
--- a/src/devices/x86/EmuX86.cpp
+++ b/src/devices/x86/EmuX86.cpp
@@ -57,6 +57,8 @@ extern std::atomic_bool g_bEnableAllInterrupts;
 
 static int field_pin = 0;
 
+static thread_local bool g_tls_isEmuX86Managed;
+
 uint32_t EmuX86_IORead(xbaddr addr, int size)
 {
 	switch (addr) {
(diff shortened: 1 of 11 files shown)


src/devices/x86/EmuX86.cpp
```diff
@@ -197,10 +187,9 @@ uint32_t EmuX86_Read(xbaddr addr, int size)
 			return value;
 		}
 
-		//pass the memory-access through to normal memory :
-		value = EmuX86_Mem_Read(addr, size);
-
-		EmuLog(LOG_LEVEL::DEBUG, "Read(0x%08X, %d) = 0x%08X", addr, size, value);
+		// Never allow direct access to host memory, instead should be handle as fatal crash.
```
LukeUsher, line 190: What about when a privileged instruction is used to read or write from a hardware register to/from host allocated memory? This would trigger an exception due to the privilege violation, but should be allowed to succeed.

There's also the possibility of memory->memory copies such as memcpy that could copy/paste from host memory to hardware device. 

(These are why forwarding to host reads/writes was done in the first place)

RadWolfie: I cba to fix them. There cannot be exception within exception. Otherwise we will have infinity exceptions loop in this current pull request state. imo, it is wrong to "retry" direct access to host memory for same exact address within triggered exception.

If you truly believe direct access to host's exact memory address should be allow to repeat after exception is triggered. Then we can't even remove Windows' specific feature which is `__try/__except`. What that means we can't provide support for non-Windows platform. Unless someone provide a hack for VEH's method, which wouldn't help at all either for same exact reason aka infinity exception for same address. With `__try/__except` within exception's EmuX86_Mem_Read/EmuX86_Mem_Write directly to host memory, all it does is hide bugs we never even knew about.

TLDR: In current upstream, there are serious design flaws with the LLE's exception handling. Which cause to hide bugs we never knew about. This pull request is only cover some of serious design flaws. If we have virtual CPU emulation, then we wouldn't have most of these design flaws. Last resort to use EmuX86_Mem_Read/EmuX86_Mem_Write is plain wrong. Again, it hide bugs we never knew about. By remove EmuX86_Mem_Read/EmuX86_Mem_Write as last resort, we will know what the flaws we have in our own codebase.

RadWolfie:
Only perform rebase and update `update Windows' SEH, fix overflow exception....` commit. Which has better detail in comment section, break down `EmuX86_Read` for each if statement to have their own return, and rename `lle_opcode_managed` to `g_tls_isEmuX86Managed` variable.

PR #1916, opened 2020-06-06
Couple DirectSound Fixes

RadWolfie:
**low risk**

---
Tester reported Sonic Mega Collection Plus' issue with "DirectSoundBuffer Lock Failed!" popup message. I took a look closer and found out `dwBytes` is way too large which made application perform miscalculation for buffer size to read from. It turn out application doesn't always check `ppvAudioPtr2` for null pointer. With commit 503ce9b fix, title is able to continue further.

Delta Force - Black Hawk Down produce an invalid parameter error from host's sound buffer creation.
Second commit 1a83cc7 resolve invalid parameter since it was missing a check for 3D audio buffer usage which only allow mono channel.

### DIFF
--- a/src/core/hle/DSOUND/XbDSoundFuncs.hpp
+++ b/src/core/hle/DSOUND/XbDSoundFuncs.hpp
@@ -1,0 +1,42 @@
+// ******************************************************************
+// *
+// *  This file is part of the Cxbx project.
+// *
+// *  Cxbx and Cxbe are free software; you can redistribute them
+// *  and/or modify them under the terms of the GNU General Public
+// *  License as published by the Free Software Foundation; either
+// *  version 2 of the license, or (at your option) any later version.
+// *
+// *  This program is distributed in the hope that it will be useful,
+// *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+// *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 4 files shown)


src/core/hle/DSOUND/XbDSoundFuncs.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include <cstdint>
+#include "XbDSoundTypes.h"
+
+namespace XTL {
+    namespace DSOUND {
+
+    static inline uint16_t PresetChannelDefault(uint32_t Xb_flags)
+    {
+        return ((Xb_flags & XTL_DSBCAPS_CTRL3D) ? 1 : 2);
```
Fisherman166, line 35: I'm not the biggest fan of ternary conditionals like this. I'd rather see:

if(Xb_flags & XTL_DSBCAPS_CTRL3D) {
    return 1;
}
return 2;

Either way I like that you made this into a function instead of having it hardcoded everywhere!

RadWolfie: Hmm, originally I was using ternary condition which then pastes couple times. Since there's so many, I decided to use inline function method will be benefit for future plugin be able to reuse the code.

I will correct it soon.

RadWolfie: resolved

PR #1917, opened 2020-06-06
Rewrite HalReadSMCTrayState to match original kernel implementation.

Fisherman166:
As always I used my kernel dump as reference to create the function. I left the TRAY_OPEN hack in place to be fixed in the future once the Hal library is more robust.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlHal.cpp
+++ b/src/core/kernel/exports/EmuKrnlHal.cpp
@@ -72,10 +72,17 @@ xboxkrnl::LIST_ENTRY ShutdownRoutineList = { &ShutdownRoutineList , &ShutdownRou
 // ******************************************************************
 xboxkrnl::KPCR* WINAPI KeGetPcr();
 
+#define TRAY_CLOSED_MEDIA_PRESENT 0x60
+#define TRAY_CLOSED_NO_MEDIA 0x40
+#define TRAY_OPEN 0x10
+#define INIT_TRAY_STATE 0xFFFFFFFF
+static DWORD g_TrayState = INIT_TRAY_STATE;
+static DWORD g_TrayCount = 0;
+
 // ******************************************************************
 // * 0x0009 - HalReadSMCTrayState()
 // ******************************************************************
-XBSYSAPI EXPORTNUM(9) xboxkrnl::VOID NTAPI xboxkrnl::HalReadSMCTrayState
+XBSYSAPI EXPORTNUM(9) xboxkrnl::NTSTATUS NTAPI xboxkrnl::HalReadSMCTrayState
 (
 	DWORD*	State,
 	DWORD*	Count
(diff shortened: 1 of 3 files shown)


src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -86,22 +93,36 @@ XBSYSAPI EXPORTNUM(9) xboxkrnl::VOID NTAPI xboxkrnl::HalReadSMCTrayState
...
+	if (TrayState == INIT_TRAY_STATE) {
+		ret = HalReadSMBusValue(SMBUS_ADDRESS_SYSTEM_MICRO_CONTROLLER, SMC_COMMAND_TRAY_STATE, 0, &TrayState);
+		// If bit 31 = 1, then there is an error so do not execute this code.
+		if ((ret >> 31) == 0) {
+			TrayState &= 0x70;
+			if ((TrayState != TRAY_CLOSED_MEDIA_PRESENT) && (TrayState != TRAY_CLOSED_NO_MEDIA)) {
+				TrayState = TRAY_OPEN;
+			}
+			orig_irql = KeRaiseIrqlToDpcLevel();
+			if (orig_TrayCount != g_TrayCount) {
```
RadWolfie, line 111: I don't see increment for `g_TrayCount` anywhere in this pull request. How does it supposedly work?

src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -86,22 +93,36 @@ XBSYSAPI EXPORTNUM(9) xboxkrnl::VOID NTAPI xboxkrnl::HalReadSMCTrayState
...
-	// TODO: Make this configurable?
-	// TODO: What is the count parameter for??
-
-	if (Count)
-		*Count = 1;
-
-	// Pretend the tray is open
+	UCHAR orig_irql = KeRaiseIrqlToDpcLevel();
+	DWORD TrayState = g_TrayState;
+	DWORD orig_TrayCount = g_TrayCount;
```
RadWolfie, line 98: Are there other functions using `g_TrayState` and `g_TrayCount`? Or only in this function? If it is only in this function, then we could use `static` variable in this function to store local value whenever get call again in the future.

src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -86,22 +93,36 @@ XBSYSAPI EXPORTNUM(9) xboxkrnl::VOID NTAPI xboxkrnl::HalReadSMCTrayState
...
+	KfLowerIrql(orig_irql);
+
+	NTSTATUS ret = STATUS_SUCCESS;
+	if (TrayState == INIT_TRAY_STATE) {
```
RadWolfie, line 102: LikeI asked about `g_TrayCount`'s increment. Do we reset `g_TrayState` to `INIT_TRAY_STATE` value outside of this function scope? Since I don't think any drivers does this way. Except for update value state instead of reset to initialize tray state.

That's my opinion/theory how drivers work.

RadWolfie:
I want to say thank you for working on tray's reader behavior implementation. Having accurate tray state reader functional will assist me on emulate how xbox hardware will behave. (soon™️) 🙂

src/core/kernel/exports/EmuKrnlHal.cpp
Fisherman166: There is another subroutine in the kernel that writes to each of the variables. I tried tracing up to figure out which kernel function calls that subroutine but there are so many levels of subroutines in between that I gave up.

image

src/core/kernel/exports/EmuKrnlHal.cpp
Fisherman166: Yes. In another subroutine there is an OR statement that resets g_TrayState to INIT_TRAY_STATE.

src/core/kernel/exports/EmuKrnlHal.cpp
Fisherman166: Another subroutine increments this. Its the same subroutine that modifies g_TrayState.

PatrickvL: Still, compared to previous code, the count is now zero instead of one - are we sure that doesn't cause problems?

Fisherman166: Unclear. I was able to play through the first level of MechAssault with no change in functionality.

PatrickvL: @LukeUsher do you know of any test case for this, so that we can merge this with confidence...

src/core/kernel/exports/EmuKrnlHal.cpp
RadWolfie: Okay, thanks for conforming this. 🙂

src/core/kernel/exports/EmuKrnlHal.cpp
RadWolfie: Default value would be likely zero. Although, I think it's better to remain include with tray's open state force set hack. Then it will be easier to remove hack in the same location without concern for possible breakage.

PR #1918, opened 2020-06-06
Update FscSetCacheSize to match original kernel

Fisherman166:
I used my kernel dump to rewrite this function to have a better skeleton. The underlying functions that do all the real work are still missing but those can be added in the future.


src/core/kernel/exports/EmuKrnlFs.cpp
```diff
@@ -70,6 +70,13 @@ XBSYSAPI EXPORTNUM(36) xboxkrnl::VOID NTAPI xboxkrnl::FscInvalidateIdleBlocks()
 	LOG_UNIMPLEMENTED();
 }
 
+static xboxkrnl::KEVENT FscCacheEvent;
```
RadWolfie, line 73: Missing `g_` prefix by the way.

PR #1923, opened 2020-06-18
Load the most appropriate xinput dll based on the Windows version

ergo720:
Only tested with Win10, since I don't have Win7/8. Someone should test those as well.

### DIFF
--- a/src/common/input/XInputPad.cpp
+++ b/src/common/input/XInputPad.cpp
@@ -101,14 +101,21 @@ namespace XInput
 	{
 		std::unique_lock<std::mutex> lck(Mtx);
 
+		// Load the most appropriate version of the xinput library depending on which Windows OS version we are running on.
+		// We will try the 9.1.0 version as a last resort since that's the only one that doesn't provide support for the guide button.
+		// For more info, see MS documentation at https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-versions.
+
 		// will only work on Win8/10; provides guide button info
-		hXInput = ::LoadLibrary(TEXT("xinput1_4.dll"));
+		std::string xinput_dll_name = "xinput1_4.dll";
+		hXInput = ::LoadLibrary(TEXT(xinput_dll_name.c_str()));
 		if (!hXInput) {
 			// will only work if the user has installed the June 2010 release of the DirectX SDK; provides guide button info
-			hXInput = ::LoadLibrary(TEXT("xinput1_3.dll"));
+			xinput_dll_name = "xinput1_3.dll";
+			hXInput = ::LoadLibrary(TEXT(xinput_dll_name.c_str()));
 			if (!hXInput) {
 				// will work on Win7; does NOT provide guide button info
-				hXInput = ::LoadLibrary(TEXT("xinput9_1_0.dll"));
+				xinput_dll_name = "xinput9_1_0.dll";
+				hXInput = ::LoadLibrary(TEXT(xinput_dll_name.c_str()));
 				if (!hXInput) {
 					EmuLog(LOG_LEVEL::ERROR2, "Failed to initialize XInput subsystem!");
 					XInputInitStatus = XINPUT_INIT_ERROR;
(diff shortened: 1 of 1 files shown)


src/common/input/XInputPad.cpp
```diff
@@ -106,19 +101,19 @@ namespace XInput
 	{
 		std::unique_lock<std::mutex> lck(Mtx);
 
-		// Because we use _WIN32_WINNT=0x0601, this will load the XInput 9.1.0 library
-		// The 9.1.0 (Win7) doesn't provide support for the guide button, but the legacy 1.3 does. Unfortunately, it requires
-		// that the user has installed the June 2010 release of the DirectX SDK to be available. If they didn't, we will default
-		// to the 9.1.0 instead of failing the initialization
```
PatrickvL, line 112: Could you try and keep the informational parts of the comment that have now been removed?

* mention of _WIN32_WINNT version
* mapping between DLL version and OS version
* the fact 9.1.0 would be loaded by default if no version was specified
* the fact 9.1.0 is lacking support for the guide button
* the fact 1.3 is a legacy version

src/common/input/XInputPad.cpp
```diff
@@ -106,19 +101,19 @@ namespace XInput
...
-		if (std::string(XINPUT_DLL).compare(TEXT("xinput9_1_0.dll")) == 0) {
-			hXInput = ::LoadLibrary(TEXT("xinput1_3.dll"));
-		}
+		// will only work on Win8/10; provides guide button info
+		hXInput = ::LoadLibrary(TEXT("xinput1_4.dll"));
 		if (!hXInput) {
-			hXInput = ::LoadLibrary(XINPUT_DLL);
```
PatrickvL, line 117: Perhaps also mention we're ignoring the `XINPUT_DLL` symbol

src/common/input/XInputPad.cpp
```diff
@@ -106,19 +101,19 @@ namespace XInput
...
+			// will only work if the user has installed the June 2010 release of the DirectX SDK; provides guide button info
+			hXInput = ::LoadLibrary(TEXT("xinput1_3.dll"));
 			if (!hXInput) {
-				EmuLog(LOG_LEVEL::ERROR2, "Failed to initialize XInput subsystem!");
-				XInputInitStatus = XINPUT_INIT_ERROR;
-				return;
+				// will work on Win7; does NOT provide guide button info
+				hXInput = ::LoadLibrary(TEXT("xinput9_1_0.dll"));
```
PatrickvL, line 111: Or should we use `XINPUT_DLL` symbol here?

src/common/input/XInputPad.cpp
```diff
@@ -106,19 +101,19 @@ namespace XInput
...
+				if (!hXInput) {
```
PatrickvL, line 112: When 9.1.0 is loaded, log that there's no support for guide button

src/common/input/XInputPad.cpp
ergo720: The _WIN32_WINNT version is already documented in this comment here https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/66bccac3cf52bf5575f82a26afced6a8f77d2c69/src/common/input/XInputPad.cpp#L50.
The DLL - OS version is documented by MS at https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-versions so I would just add the link.
All in all, I'd write something like this:
```
Load the most appropriate version of the xinput library depending on which Windows OS version we are running on. We will try the 9.1.0 version as a last resort since that's the only one that doesn't provide support for the guide button. For more info, see MS documentation at https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-versions. 
```

src/common/input/XInputPad.cpp
ergo720: I removed the XINPUT_DLL macro because it's Windows specific, so I decided to specify the xinput dll names directly as strings. Doing this allows to remove the MinGW workaround that was added in a previous commit.

src/common/input/XInputPad.cpp
ergo720: We sould not, see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1923#discussion_r442326075

PR #1925, opened 2020-06-23
Show a popup when xbe sections are corrupted + print section corruption status on xbe dumps as well

ergo720:
This will show a popup warning message when cxbxr detects that one or more xbe section(s) are corrupted, instead of just silently logging them in the kernel dumps. Note that this check is different than the existing xbe signature check, which only verifies the xbe header. In addition, it also prints the section checks in the xbe dumps as well. Finally, a new option was added to the gui to allow users to also ignoring this new xbe section check.


src/common/Logging.cpp
```diff
@@ -278,7 +278,7 @@ PopupReturn PopupCustomEx(const void* hwnd, const CXBXR_MODULE cxbxr_module, con
     UINT uType = MB_TOPMOST | MB_SETFOREGROUND;
 
 	// Make assert whenever the format string is null pointer which isn't allow in here.
-	assert(!message);
+	assert(message != nullptr);
```
RadWolfie, line 281: @ergo720 found a bug which require to use new change for asserts. I was not aware of this issue. 👍

src/gui/WndMain.cpp
```diff
@@ -2001,25 +2009,46 @@ void WndMain::OpenXbe(const char *x_filename)
...
+				errorMsg += "- One or more XBE section(s) are corrupted!\n";
+
+				// if we find a corrupted section, we won't bother checking the remaining sections since we know
+				// already at this point that the xbe is invalid
+				break;
+			}
+		}
+	}
+
+	if (invalidXbe) {
```
RadWolfie, line 2034: Is it really necessary for invalidXbe? My proposal is check errorMsg's size other than zero. Then display the popup message.

RadWolfie:
Preview:
* XBE signature
  image
* XBE section corrupted
  image
* both
  image

PR #1928, opened 2020-06-28
Update issue template + add action to automatically close issues not following the template

ergo720:
This adds a github action which compares the issue body with the template and automatically closes it if it does not follow the template and adds the invalid label. ~~It can also automatically reopen the issue if this is edited later and now follows the template.~~ The intent of this is to reduce the burden on the maintainers of manually closing invalid issues (which happens fairly often on this repository).

### DIFF
--- a/.github/workflows/autoclose.yml
+++ b/.github/workflows/autoclose.yml
@@ -1,0 +1,16 @@
+on:
+  issues:
+    types: [opened, edited]
+
+jobs:
+  auto_close_issues:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/autoclose.yml
```diff
@@ -0,0 +1,16 @@
+on:
+  issues:
+    types: [opened, edited]
```
RadWolfie, line 3: I noticed one thing we had talked in the past. "edited" part should be use in the main repo for time being.

PR #1932, opened 2020-07-30
Migrate Customs in One Batch

### LINKED ISSUE #1038: Implement version checking
It was mentioned on discord that it isn't clear that newer releases are available.

Unlike #1032, this issue is about adding functionality to check for newer development builds.

One way would be to use the public GitHub API to compare the git version hash (we should it on the window titlebar) - see https://developer.github.com/v3/repos/commits/#compare-two-commits

Either comparing actual builds

> https://api.github.com/repos/Cxbx-Reloaded/Cxbx-Reloaded/compare/GIT_HASH...master

Or just checking the latest build

> https://api.github.com/repos/Cxbx-Reloaded/Cxbx-Reloaded/commits/master

This would involve adding some sort of HTTP library and JSON parser - which are pretty heavyweight dependencies :truck:

### LINKED ISSUE #1886: Generate Version Hash for Every Codebase Change Occur
Since it is troublesome for local build not having versioning updated because of CMake does not detect source file changes in order to regenerate versioning hash. Plus posting screenshots of local build can produce false positive info with wrong version hash.

RadWolfie:
Following changes are:
- New project, misc-batch, to enforce always run on any build.
- Move git versioning into misc batch project.
  - Plus add dirty flag for aware of difference of some changes are not committed.
- Move duplicate glew32.dll copy into misc batch project.
  - Plus copy only if file is different. Since we're using pre-compiled build, it will only copy once than repeat per cxbx and cxbxr-emu projects.

close #1886

Thanks to @LukeUsher for git versioning's fix suggestion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,30 +29,12 @@ endif()
 # Cxbx-Reloaded projects
 set(CXBXR_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR})
 
-add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/projects/libtom")
-
-find_package(Git)
-
-if(Git_FOUND)
- message("Git found: ${GIT_EXECUTABLE}")
-
- execute_process(
-  COMMAND git describe --always --tags --first-parent
-  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-  OUTPUT_VARIABLE _GIT_VERSION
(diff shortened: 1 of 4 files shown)


gandalfthewhite19890404:
Cxbxr-debugger does not start at all after this PR was merged:
CI-4d06712 (PR 1929) - good
CI-71dc7e9 (this PR) - bad

PR #1934, opened 2020-08-15
kernel: avoid potential buffer overflow on DbgPrint

LukeUsher:
There are likely more occurrences of this in the code-base, but I ran into this one while working on Insignia.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlDbg.cpp
+++ b/src/core/kernel/exports/EmuKrnlDbg.cpp
@@ -106,16 +106,31 @@ XBSYSAPI EXPORTNUM(8) xboxkrnl::ULONG _cdecl xboxkrnl::DbgPrint
 		LOG_FUNC_END;
 
 	if (Format != NULL) {
-		char szBuffer[MAX_PATH];
-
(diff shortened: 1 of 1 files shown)


src/core/kernel/exports/EmuKrnlDbg.cpp
```diff
@@ -106,16 +106,31 @@ XBSYSAPI EXPORTNUM(8) xboxkrnl::ULONG _cdecl xboxkrnl::DbgPrint
...
+        if (buffer == nullptr) {
+            // Prevent a crash if we can't allocate enough memory
+            // We want this to be transparent to the running Xbox application
+            // Hence, we return success.
+            return STATUS_SUCCESS;
+        }
+
+        // Perform the actual print operation
+		vsnprintf(buffer, size, Format, argp);
+		free(buffer);
```
RadWolfie, line 129: I notice buffer variable is used later after freed from memory. Please fix this, thanks.

LukeUsher: Thanks!

PR #1935, opened 2020-08-15
hle: Disable RaiseException patch, let the kernel handle it.

LukeUsher:
I am not aware of any regressions, but this should be tested with a large number of games before merging.


literalmente-game:
44 games tested, no regressions (or improvements) noticed

RadWolfie:
FYI: both XTL::RaiseException and xboxkrnl::RtlRaiseException are not implemented.

PR #1942, opened 2020-08-21
Introduce xbox namespace + deduplicate xbox types

ergo720:
This PR removes both xboxkrnl and XTL namespaces and replaces them with a single xbox namespace. In addition, it also introduces a xbox_types header file which contains all xbox types from both the kernel and XTL, most of which were duplicated inside the old namespaces. Other than that, it should have no functional changes.


RadWolfie:
I had performed squash & merge to make others easier to able fix their rebase conflict once than multiple times.

PR #1955, opened 2020-08-27
Fix Xbox DSound Namespace

RadWolfie:
The changes correction will become easier to follow base on xbox namespace directly without dependent on specific platform library. Plus one possible fix may resolve some audio loudness issue.

---
~~I am currently requesting testers to verify any audio loudness issue may be either resolved or regress.~~
Tester reported no noticeable change for 15 titles.


src/common/xbox_types.h
```diff
@@ -70,6 +70,9 @@ namespace xbox
 	typedef unsigned __int64    QUAD; // 8 byte aligned 8 byte long
 	typedef int                 BOOL;
 	typedef LONG                HRESULT;
+	typedef float               FLOAT;
+	#define XCALLBACK           __stdcall
+	#define XBOXAPI             __stdcall
```
ergo720, line 75: Is this distinction necessary? Can't we just use `XBOXAPI` everywhere in the code instead of having two macros which both define `__stdcall`? At the moment, `XCALLBACK` seems to only be used in `LPFNXMOCALLBACK`.

RadWolfie: It's to tell the difference between host and xbox because of `#define` is not confined within namespace. Can we use typedef instead? Nope, compiler doesn't allow it when I tried it before.

I could, should, had used XBOXAPI in XCALLBACK define.

Xbox's DirectSound has several callbacks internally and I think from one or two interfaces as well. Which I would use in the structure for readability.

src/common/xbox_types.h
```diff
@@ -107,4 +110,21 @@ namespace xbox
...
 
 	typedef /*_Null_terminated_*/ const WCHAR *LPCWSTR, *PCWSTR;
+
+	// ******************************************************************
+	// Misc
+	// ******************************************************************
+	typedef struct _XD3DVECTOR {
+		float x;
+		float y;
+		float z;
```
ergo720, line 120: Nit: should these use the `FLOAT` type defined above?

src/common/xbox_types.h
RadWolfie: After reading @ergo720's pm, I understand calling convention will no longer become depend on is once we have lib86cpu implemented. 👍

ergo720 (approved):

Issue #1059, opened 2018-04-13
Fullscreen should go to 4:3 with Blackbars


JackaJacka123:
As I was experimenting with JSRF today, I ended up playing in fullscreen. As I own a 16:9 monitor, The pixels stretched out and it looked a lot worse than it needed too. When going fullscreen, Cxbx-R should check the aspect ratio and place black bars on the sides like how emulation on the Xbox 360 and Xbox One is done. 

This is definitely a QOL change.

BenNottelling:
What's. QOL? Quality of life? It's a dumb question, I know lol. I think if you install the dashboard and then then go into the dashboard settings you should be able to set it to a 16:9 aspect ratio and then go back and try the game again. I'm not sure if it will work though

Margen67:
@BenNottelling Yes. That option doesn't work for all games, though.

Esppiral made a cheat table for the PAL version of JSRF: https://cdn.discordapp.com/attachments/293096509019258880/424253354500161556/JET_SET_RADIO_FUTURE_PAL.CT (requires [Cheat Engine](http://cheatengine.org/))

For other games: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/wiki/Frequently-Asked-Questions-(FAQ)#how-do-i-run-games-in-a-higher-resolution-such-as-720p-1080p-or-4k

You can also try changing the aspect ratio option in the dashboard.

As a workaround you can make a custom 4:3 resolution like 1440x1080 in either your graphics driver settings or [CRU](https://www.monitortests.com/forum/Thread-Custom-Resolution-Utility-CRU), and set scaling in your graphics driver and/or monitor settings to preserve the aspect ratio. You can use a lower 4:3 resolution like 1280x960 which won't need any custom resolutions to be created, but it might be more blurry.

Also, if Cxbx-Reloaded had a Retroarch core AR scaling would be possible. (#372)

JackaJacka123:
@BenNottelling @Margen67 I think the point is so that you don't have to deal with finicky custom settings in an Intel/AMD/Nvidia control panel to reset the resolution. I think core AR scaling sounds like a good solution. If that doesn't pan out, it could just be internally be done by cxbx-r.

Margen67:
If Cxbx-Reloaded had real (exclusive) fullscreen then graphics driver/monitor settings would work.
Currently Cxbx does the scaling so it's either stretched or tiny depending on the game's resolution and Cxbx's settings.

[...]

PR #1956, opened 2020-08-28
d3d: apply aspect ratio correction to the backbuffer

LukeUsher:
This pull-request implements aspect ratio correction while also correctly handling the case where Xbox games render 16:9 content to a 4:3 buffer (anamorphic widescreen)

NOTE: When using 'Automatic' resolution setting, scaling may still be incorrect. This will be solved in a different pull-request because it requires an overhaul of how our Window creation works. In the interim, for correct aspect ratio scaling, please select a suitable display resolution that matches your screen aspect ratio.

Dashboard in 4:3, within a 16:9 window
image

Dashboard in 16:9, within a 16:9 window
image

Dashboard in 16;9 within a 4:3 window
image

Jet Set Radio Future (a 4:3 only game, in a 16:9 window)
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -108,6 +108,8 @@ static bool                         g_bHack_UnlockFramerate = false; // ignore t
 static bool                         g_bHasDepth = false;    // Does device have a Depth Buffer?
 static bool                         g_bHasStencil = false;  // Does device have a Stencil Buffer?
 static DWORD						g_dwPrimPerFrame = 0;	// Number of primitives within one frame
+static unsigned int                 g_AspectBackbufferWidth = 0; // Used for aspect ratio correction
+static unsigned int                 g_AspectBackbufferHeight = 0;// Used for aspect ratio correction
 
 static Settings::s_video            g_XBVideo;
 
@@ -2953,6 +2955,22 @@ void Direct3D_CreateDevice_End()
         g_Xbox_D3DDevice = (DWORD*)it->second;
     }
 #endif
+
+    // Cache default backbuffer width and height for aspect ratio scaling purposes
+    g_AspectBackbufferHeight = g_EmuCDPD.XboxPresentationParameters.BackBufferHeight;
+    g_AspectBackbufferWidth = g_EmuCDPD.XboxPresentationParameters.BackBufferWidth;
+
+    // NOTE: Some games use anamorphic widesceen (expecting a 4:3 surface to be displayed at 16:9)
+    // For those, we *lie* about the default width, for the scaler
+    // 720p / 1080i are *always* widescreen, and will have the correct backbuffer size, so we only
+    // apply this 'hack' for non-hd resolutions
+    if (g_EmuCDPD.XboxPresentationParameters.Flags & X_D3DPRESENTFLAG_WIDESCREEN && g_AspectBackbufferHeight < 720) {
+        // Lie and pretend we are 1280x720, this works because this ratio is only used in calculations
+        // and not used as actual raw input values
+        g_AspectBackbufferWidth = 1280;
+        g_AspectBackbufferHeight = 720;
+    }
+
     // If the Xbox version of CreateDevice didn't call SetRenderTarget, we must derive the default backbuffer ourselves
     // This works because CreateDevice always sets the current render target to the Xbox Backbuffer
     // In later XDKs, it does this inline rather than by calling D3DDevice_SetRenderTarget
@@ -5008,13 +5026,27 @@ DWORD WINAPI xbox::EMUPATCH(D3DDevice_Swap)
 
 		auto pXboxBackBufferHostSurface = GetHostSurface(g_pXbox_BackBufferSurface, D3DUSAGE_RENDERTARGET);
 		if (pXboxBackBufferHostSurface) {
-			// Blit Xbox BackBuffer to host BackBuffer
-			// TODO: Respect aspect ratio
+            // Calculate the aspect ratio scale factor
+            const auto imageAspect = (float)g_AspectBackbufferWidth/ (float)g_AspectBackbufferHeight;
+            const auto screenAspect = (float)BackBufferDesc.Width / (float)BackBufferDesc.Height;
+            const auto scaleFactor = screenAspect > imageAspect ? (float)BackBufferDesc.Height / (float)g_AspectBackbufferHeight : (float)BackBufferDesc.Width / (float)g_AspectBackbufferWidth;
+
+            const auto width = g_AspectBackbufferWidth * scaleFactor;
+            const auto height = g_AspectBackbufferHeight * scaleFactor;
+
+            // Caclulate the centered rectangle
+            RECT dest{};
+            dest.top = (BackBufferDesc.Height - height) / 2;
+            dest.left = (BackBufferDesc.Width - width) / 2;
+            dest.right = dest.left + width;
+            dest.bottom = dest.top + height;
+
+            // Blit Xbox BackBuffer to host BackBuffer
             hRet = g_pD3DDevice->StretchRect(
                 /* pSourceSurface = */ pXboxBackBufferHostSurface,
                 /* pSourceRect = */ nullptr,
                 /* pDestSurface = */ pCurrentHostBackBuffer,
-                /* pDestRect = */ nullptr,
+                /* pDestRect = */ &dest,
                 /* Filter = */ LoadSurfaceFilter
             );
 		
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -108,9 +108,10 @@ static bool                         g_bHack_UnlockFramerate = false; // ignore t
 static bool                         g_bHasDepth = false;    // Does device have a Depth Buffer?
 static bool                         g_bHasStencil = false;  // Does device have a Stencil Buffer?
 static DWORD						g_dwPrimPerFrame = 0;	// Number of primitives within one frame
-static unsigned int                 g_AspectBackbufferWidth = 0; // Used for aspect ratio correction
-static unsigned int                 g_AspectBackbufferHeight = 0;// Used for aspect ratio correction
-
+static float                        g_AspectRatioScale = 1.0f;
+static UINT                         g_AspectRatioScaleWidth = 0;
+static UINT                         g_AspectRatioScaleHeight = 0;
+static D3DSURFACE_DESC              g_HostBackBufferDesc;
 static Settings::s_video            g_XBVideo;
 
 // D3D based variables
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5008,13 +5026,27 @@ DWORD WINAPI xbox::EMUPATCH(D3DDevice_Swap)
 
 		auto pXboxBackBufferHostSurface = GetHostSurface(g_pXbox_BackBufferSurface, D3DUSAGE_RENDERTARGET);
 		if (pXboxBackBufferHostSurface) {
-			// Blit Xbox BackBuffer to host BackBuffer
-			// TODO: Respect aspect ratio
+            // Calculate the aspect ratio scale factor
+            const auto imageAspect = (float)g_AspectBackbufferWidth/ (float)g_AspectBackbufferHeight;
+            const auto screenAspect = (float)BackBufferDesc.Width / (float)BackBufferDesc.Height;
+            const auto scaleFactor = screenAspect > imageAspect ? (float)BackBufferDesc.Height / (float)g_AspectBackbufferHeight : (float)BackBufferDesc.Width / (float)g_AspectBackbufferWidth;
```
PatrickvL, line 5032: Maybe change this into a regular if else block, allowing breakpoints on either branch?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2953,6 +2955,22 @@ void Direct3D_CreateDevice_End()
         g_Xbox_D3DDevice = (DWORD*)it->second;
     }
 #endif
+
+    // Cache default backbuffer width and height for aspect ratio scaling purposes
+    g_AspectBackbufferHeight = g_EmuCDPD.XboxPresentationParameters.BackBufferHeight;
+    g_AspectBackbufferWidth = g_EmuCDPD.XboxPresentationParameters.BackBufferWidth;
```
PatrickvL, line 2961: IF non-standard, custom xbox backbuffer sizes exist (I'm not sure), then instead we would need to read VGA/CRT registers or something - what's your take on that possibility?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2953,6 +2955,22 @@ void Direct3D_CreateDevice_End()
...
+
+    // NOTE: Some games use anamorphic widesceen (expecting a 4:3 surface to be displayed at 16:9)
+    // For those, we *lie* about the default width, for the scaler
+    // 720p / 1080i are *always* widescreen, and will have the correct backbuffer size, so we only
+    // apply this 'hack' for non-hd resolutions
+    if (g_EmuCDPD.XboxPresentationParameters.Flags & X_D3DPRESENTFLAG_WIDESCREEN && g_AspectBackbufferHeight < 720) {
+        // Lie and pretend we are 1280x720, this works because this ratio is only used in calculations
+        // and not used as actual raw input values
+        g_AspectBackbufferWidth = 1280;
```
PatrickvL, line 2970: Here, it's assumed anamorphic widescreen targets 1280x720, but have people ever reported on how the image comes out when connecting an Xbox to ultra-wide or portrait displays?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5008,13 +5026,27 @@ DWORD WINAPI xbox::EMUPATCH(D3DDevice_Swap)
...
 		if (pXboxBackBufferHostSurface) {
-			// Blit Xbox BackBuffer to host BackBuffer
-			// TODO: Respect aspect ratio
+            // Calculate the aspect ratio scale factor
+            const auto imageAspect = (float)g_AspectBackbufferWidth/ (float)g_AspectBackbufferHeight;
+            const auto screenAspect = (float)BackBufferDesc.Width / (float)BackBufferDesc.Height;
+            const auto scaleFactor = screenAspect > imageAspect ? (float)BackBufferDesc.Height / (float)g_AspectBackbufferHeight : (float)BackBufferDesc.Width / (float)g_AspectBackbufferWidth;
+
+            const auto width = g_AspectBackbufferWidth * scaleFactor;
+            const auto height = g_AspectBackbufferHeight * scaleFactor;
```
PatrickvL, line 5035: This may produce pixel-drift, caused by rounding issues - are there methods to avoid that, somehow? This would be especially noticeable when ratio's should be a whole rounded number, but aren't.... Not sure this will ever happen, but a float is only 32 bits, which offers a mere 28 bits of accuracy - it should be enough, but if issues arise, perhaps using a double (64 bit float) could solve some of this (if it's a problem at, of course).

PatrickvL (commented):
Some considerations I'd be interested in to read your take on...

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: The aspect ratio is not handled by the CRT registers: for anamorphic widescreen the CRT would still receive a 4:3 signal, and the user would usually have to press a button on the TV remove to stretch to 16:9.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: Like above: anamorphic widescreen is a 16:9 aspect image packed into a 4:3 framebuffer. For correct display, it needs stretching back out to a 16:9 resolution.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: This is a fairly standard scaling algorithm, you’d be hard pressed to find something better,

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: I don’t see the point of allowing them to be individually breakpointed, if I’m completely honest

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: To clarify; any 16:9 resolution would work, what matters is the scale ratio, not the actual resolution involved.

PatrickvL:
Thanks for addressing my questions. I'll take your word for it all. Nice work BTW!

RadWolfie:
I can confirm aspect ratio is respected in 2560x1080 resolution (UWHD).
<details>
<summary>16:9</summary>

image
</details>

<details>
<summary>4:3</summary>

image
</details>

However, I notice there's 1 pixel linear doesn't match with JSRF intro fade in/out on the left and top edges. This seems only apply with render resolution higher than default. Display resolution other than default with render resolution set to default works fine.
See screenshot below for reference:
image

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5008,13 +5026,27 @@ DWORD WINAPI xbox::EMUPATCH(D3DDevice_Swap)
...
+
+            // Caclulate the centered rectangle
+            RECT dest{};
+            dest.top = (BackBufferDesc.Height - height) / 2;
+            dest.left = (BackBufferDesc.Width - width) / 2;
+            dest.right = dest.left + width;
+            dest.bottom = dest.top + height;
```
RadWolfie, line 5042: Following 4 lines above produced conversion warnings:
```
Cxbx-Reloaded\src\core\hle\D3D8\Direct3D9\Direct3D9.cpp(5039,57): warning C4244: '=': conversion from 'float' to 'LONG', possible loss of data
Cxbx-Reloaded\src\core\hle\D3D8\Direct3D9\Direct3D9.cpp(5040,56): warning C4244: '=': conversion from 'float' to 'LONG', possible loss of data
Cxbx-Reloaded\src\core\hle\D3D8\Direct3D9\Direct3D9.cpp(5041,36): warning C4244: '=': conversion from 'float' to 'LONG', possible loss of data
Cxbx-Reloaded\src\core\hle\D3D8\Direct3D9\Direct3D9.cpp(5042,36): warning C4244: '=': conversion from 'float' to 'LONG', possible loss of data
```

RadWolfie:
While the code looks good, even the hack requirement is a necessary apparently. I believe aspect code should be place into a common source file to enable accessibility in case of we have another graphic renderer.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5008,13 +5026,27 @@ DWORD WINAPI xbox::EMUPATCH(D3DDevice_Swap)
...
-			// TODO: Respect aspect ratio
+            // Calculate the aspect ratio scale factor
+            const auto imageAspect = (float)g_AspectBackbufferWidth/ (float)g_AspectBackbufferHeight;
+            const auto screenAspect = (float)BackBufferDesc.Width / (float)BackBufferDesc.Height;
+            const auto scaleFactor = screenAspect > imageAspect ? (float)BackBufferDesc.Height / (float)g_AspectBackbufferHeight : (float)BackBufferDesc.Width / (float)g_AspectBackbufferWidth;
+
+            const auto width = g_AspectBackbufferWidth * scaleFactor;
+            const auto height = g_AspectBackbufferHeight * scaleFactor;
+
+            // Caclulate the centered rectangle
```
RadWolfie, line 5037: Found a typo, Ca**c**lulate should be Cal**c**ulate.

RadWolfie:
I have finally found missing implementation for aspect ratio that causes OXM's FMVs filled whole screen no matter what resolution they are. Missing aspect ratio implement is documented here:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/8865e1187c787a2ea52b5060a44ee7a82be0f956/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp#L5210

When I copied these lines below over to missing aspect ratio implement commented, aspect ratio is now respective.
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/8865e1187c787a2ea52b5060a44ee7a82be0f956/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp#L5059-L5068

Hopefully this clear things up with OXM demos viewing videos issue.

RinMaru:
this only supposed to affect FMVs?

LukeUsher:
As of the latest commit, both Overlays and game content maintain aspect ratio.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4990,14 +5042,26 @@ DWORD WINAPI xbox::EMUPATCH(D3DDevice_Swap)
 	HRESULT hRet = g_pD3DDevice->GetBackBuffer(
 		0, // iSwapChain
 		0, D3DBACKBUFFER_TYPE_MONO, &pCurrentHostBackBuffer);
+
 	DEBUG_D3DRESULT(hRet, "g_pD3DDevice->GetBackBuffer - Unable to get backbuffer surface!");
 	if (hRet == D3D_OK) {
 		assert(pCurrentHostBackBuffer != nullptr);
 
-		// Get backbuffer dimensions; TODO : remember this once, at creation/resize time
-		D3DSURFACE_DESC BackBufferDesc;
```
LukeUsher, line 4998: This code was removed because I applied the optimization suggested in the TODO.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4990,14 +5042,26 @@ DWORD WINAPI xbox::EMUPATCH(D3DDevice_Swap)
...
-		pCurrentHostBackBuffer->GetDesc(&BackBufferDesc);
-
+        // Clear the backbuffer surface
+        IDirect3DSurface* pExistingRenderTarget = nullptr;
+        hRet = g_pD3DDevice->GetRenderTarget(0, &pExistingRenderTarget);
+        if (hRet == D3D_OK) {
```
LukeUsher, line 5053: This ensures the backbuffer is cleared before drawing to it, which prevents artifacts when changing from 16:9 to 4:3

Margen67:
#1059

medievil1:
KOTOR I and II are fixed...video scales properly now...

LanHikariDS:
This might be partly considered user-error, but this does create an issue where if a user goes full-screen with Automatic (Xbox Default) as their display resolution, the emulator will overcorrect the aspect ratio on a 16:9 monitor. 
image 
image

LukeUsher:
> This might be partly considered user-error, but this does create an issue where if a user goes full-screen with Automatic (Xbox Default) as their display resolution, the emulator will overcorrect the aspect ratio on a 16:9 monitor.
> image
> image

Yeah that makes sense. It scales to the “display” resolution which does not change when you hit alt + enter to full screen. That’s a different problem entirely that needs a separate solution (destroying and re-creating the backbuffer on window resize)

LanHikariDS:
> Yeah that makes sense. It scales to the “display” resolution which does not change when you hit alt + enter to full screen. That’s a different problem entirely that needs a separate solution (destroying and re-creating the backbuffer on window resize)

I figured it was something to that effect, hence why I mentioned it being user error to a degree, and I could see that behavior for the static display sizes, but given that the default is labled "Automatic", it makes it sound like it will (or should) shift as need be.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie: In my opinion, it is important to have ^ comment in the codebase to explain the reason behind of "clear the backbuffer surface".

GXTX:
I didn't comment on this, but I believe we should have a option setting to *disable* this fix. On hardware you would expect to see this "bug".

RadWolfie:
An option to disable aspect ratio will be eventually be feature soon™. Currently, Luke is focusing on fix relative to faux fullscreen.

medievil1:
if it is just 4:3 stretch stuff people want... use a lower output resolution (720x480 for example) and it will stretch when you go full screen... 1280x960 will do it automatically as soon as you start cxbx.. while 16:9 resolutions will offer 4:3 correction

PR #1964, opened 2020-09-19
Re-enable Mouse Support

ergo720:
This PR adds a new KeyboardMouse device which allows cxbxr to receive input from both the keyboard and mouse simultaneously. Mouse bindings will use the "Axis X/Y/Z +/-" and "Click 0/1/2" control names in the input gui when bound. Existing keyboard profiles are retained and converted to use the new KeyboardMouse device.

### DIFF
--- a/src/common/IPCHybrid.hpp
+++ b/src/common/IPCHybrid.hpp
@@ -52,6 +52,6 @@ typedef enum class _IPC_UPDATE_KERNEL {
 	, CONFIG_INPUT_SYNC
 } IPC_UPDATE_KERNEL;
 
-void ipc_send_kernel_update(IPC_UPDATE_KERNEL command, const unsigned int value, const unsigned int hwnd);
+void ipc_send_kernel_update(IPC_UPDATE_KERNEL command, const int value, const unsigned int hwnd);
 
 #endif
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -923,7 +923,7 @@ void Settings::RemoveLegacyConfigs(unsigned int CurrentRevision)
 
 		for (unsigned port_num = 0; port_num < 4; ++port_num) {
 			std::string current_section = std::string(section_input_port) + std::to_string(port_num);
-			std::string device_name = m_si.GetValue(current_section.c_str(), sect_input_port.device);
+			std::string device_name = m_si.GetValue(current_section.c_str(), sect_input_port.device, "");
 
 			// NOTE: with C++20, this can be simplified by simply calling device_name.ends_with()
 			if (device_name.length() >= kb_str.length()) {
@@ -940,7 +940,7 @@ void Settings::RemoveLegacyConfigs(unsigned int CurrentRevision)
 				break;
 			}
 
-			std::string device_name = m_si.GetValue(current_section.c_str(), sect_input_profiles.device);
+			std::string device_name = m_si.GetValue(current_section.c_str(), sect_input_profiles.device, "");
 
 			// NOTE: with C++20, this can be simplified by simply calling device_name.ends_with()
 			if (device_name.length() >= kb_str.length()) {
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -62,8 +62,9 @@ uint16_t g_LibVersion_DSOUND = 0;
 // * 5: (ergo720),   added new input gui settings and revision to core
 // * 6: (RadWolfie), added loader executable member to core, only for clean up loader expertimental setting
 // * 7: (RadWolfie), fix allowAdminPrivilege not align with other boolean members
+// * 8: (ergo720),   added general input settings
 ///////////////////////////
-const unsigned int settings_version = 7;
+const unsigned int settings_version = 8;
 
 Settings* g_Settings = nullptr;
 
@@ -125,15 +126,21 @@ static struct {
 	const char* adapter_name = "adapter_name";
 } sect_network_keys;
 
+static const char *section_input_general = "input-general";
+static struct {
+	const char *mo_axis_range = "MouseAxisRange";
+	const char *mo_wheel_range = "MouseWheelRange";
+} sect_input_general;
+
 static const char* section_controller_dinput = "controller-dinput";
 static const char* section_controller_port = "controller-port";
 
-static const char* section_input = "input-port-";
+static const char* section_input_port = "input-port-";
 static struct {
 	const char* type = "Type";
 	const char* device = "DeviceName";
 	const char* config = "ProfileName";
-} sect_input;
+} sect_input_port;
 
 static const char* section_input_profiles = "input-profile-";
 static struct {
@@ -366,6 +373,10 @@ bool Settings::LoadConfig()
 
 	// ==== Core End ============
 
+	// Delete/update legacy configs from previous revisions
+	RemoveLegacyConfigs(m_core.Revision);
+	m_core.Revision = settings_version;
+
 	// ==== Hack Begin ==========
 
 	m_hacks.DisablePixelShaders = m_si.GetBoolValue(section_hack, sect_hack_keys.DisablePixelShaders, /*Default=*/false);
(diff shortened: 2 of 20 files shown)


src/common/IPCHybrid.hpp
```diff
@@ -52,6 +52,6 @@ typedef enum class _IPC_UPDATE_KERNEL {
 	, CONFIG_INPUT_SYNC
 } IPC_UPDATE_KERNEL;
 
-void ipc_send_kernel_update(IPC_UPDATE_KERNEL command, const unsigned int value, const unsigned int hwnd);
```
RadWolfie, line 55: For what reason unsigned part is removed? It must be either signed or unsigned. Not omit either one out. Beside, it should be unsigned anyway.

ergo720: Because I need to pass the value `PORT_INVALID`, which is -1, and using an unsigned argument causes an unwanted  conversion. Also, the function is passing the value as `LPARAM` via `SendMessage`, which is a signed integer as well (it's `LONG_PTR`), so the passed unsigned value will be converted to signed anyway.

RadWolfie: Then simply replace it to `signed`.

literalmente-game:
6afcd6a insta-crashes for me

ergo720:
@literalmente-game  Yeah, I can confirm that as well. I have only tested by upgrading from an existing installation, which worked fine for me.

src/common/Settings.cpp
```diff
@@ -893,4 +917,38 @@ void Settings::RemoveLegacyConfigs(unsigned int CurrentRevision)
 	default:
 		break;
 	}
+
+	if (CurrentRevision < 8) {
+		const std::string kb_str = "Keyboard";
+
+		for (unsigned port_num = 0; port_num < 4; ++port_num) {
+			std::string current_section = std::string(section_input_port) + std::to_string(port_num);
+			std::string device_name = m_si.GetValue(current_section.c_str(), sect_input_port.device);
```
ergo720, line 926: Problem found, it's crashing in this line

src/gui/resource/Cxbx.rc
```diff
@@ -589,10 +603,10 @@ BEGIN
         MENUITEM "&Open Xbe...",                ID_FILE_OPEN_XBE,MFT_STRING,MFS_ENABLED
         MENUITEM "Open D&ashboard...\tF7",      ID_FILE_OPEN_DASHBOARD,MFT_STRING,MFS_ENABLED
         MENUITEM "&Close Xbe",                  ID_FILE_CLOSE_XBE,MFT_STRING,MFS_ENABLED
-        MENUITEM "", -1, MFT_SEPARATOR
```
RadWolfie, line 592: I know Visual Studio tend to remove optional fields. However it is a requirement for Clang.

ergo720:
Commit 01382d7 now allows users to lock the mouse cursor to the cxbxr rendering window when they press the F3 key. This is useful in windowed mode to avoid the cursor drifting to other windows while playing with the mouse. To release the cursor, simply press F3 again (it works like a toggle).

literalmente-game:
From @LukeUsher on Discord 

> The problem with these toggles is they get forgotten when a new xbe is launched. In morrowind for example that constantly reboots itself
As a result, much of these toggle should store their state in emushared

ergo720:
Commit dd24225 drops keyboard/mouse input when the mouse cursor is outside of the rendering window. This is again useful when playing with the mouse in windowed mode. This way, the game won't receive bogus input from the keyboard/mouse when the user is selecting/working on other windows.

src/gui/resource/Cxbx.rc
```diff
@@ -27,6 +27,10 @@ LANGUAGE LANG_NEUTRAL, SUBLANG_NEUTRAL
 #ifdef APSTUDIO_INVOKED
 GUIDELINES DESIGNINFO
 BEGIN
+    IDD_INPUT_CFG, DIALOG
```
RadWolfie, line 30: @ergo720 Did you add a new dialog then decided to revert the change?
Same thing for `IDD_INPUT_CFG AFX_DIALOG_LAYOUT` below.

src/gui/resource/ResCxbx.h
```diff
@@ -304,6 +304,9 @@
 #define IDC_RUMBLE_TEST                 1302
 #define IDC_NETWORK_ADAPTER             1303
 #define IDC_LOG_POPUP_TESTCASE          1304
+#define IDC_INPUT_OPTIONS               1305
+#define IDC_MOUSE_RANGE                 1306
+#define IDC_WHEEL_RANGE                 1307
```
RadWolfie, line 309: For some reason these additions did not update `_APS_NEXT_CONTROL_VALUE` value to 1308.

src/core/kernel/init/CxbxKrnl.h
```diff
@@ -200,9 +200,11 @@ bool CxbxIsElevated();
 extern uint32_t CxbxKrnl_KernelThunkTable[379];
 
 extern bool g_bIsWine;
-
+extern bool g_bClipCursor;
 extern bool g_CxbxPrintUEM;
 extern ULONG g_CxbxFatalErrorCode;
+extern bool g_bIsTrackingMoLeave;
```
RadWolfie, line 206: IMHO, these global should be confine in `DInputKeyboardMouse` files or in general input file.

RadWolfie:
I have tested and confirmed mouse cursor is working as intended in-game. I'm not sure if Turok Evolution's menu selection with mouse is intended design or not. Since in-game does behave correctly, it may be just the game design.

I also have tested and confirmed mouse cursor does stay within render window in exclusive full screen, faux full screen, and windowed modes. Including multiple reboots.
👍

I am waiting on review remarks to be discuss or resolve them.

src/gui/resource/Cxbx.rc
ergo720: I didn't, I simply added the new options to the existing input gui. According to [here](https://stackoverflow.com/a/49444451), `DESIGNINFO` fields are only needed at design-time and don't change anything at runtime so this is a non-issue. BTW, `AFX_DIALOG_LAYOUT` (which also has a similar addition) is documented by Microsoft [here](https://docs.microsoft.com/en-us/cpp/mfc/dynamic-layout?view=vs-2019#to-set-the-dynamic-layout-programmatically-from-a-resource-file).

RadWolfie: Thanks for clarification. Sometime I see resource editor randomly place something like this in the past, either at the top or at the bottom. It's hard to tell if it's intended design or not, or revert changes usually get left behind. Since they are usually empty like we see in here. 👍

LukeUsher (dismissed):
I like it!

src/gui/DlgInputConfig.cpp
```diff
@@ -183,16 +222,66 @@ INT_PTR CALLBACK DlgInputConfigProc(HWND hWndDlg, UINT uMsg, WPARAM wParam, LPAR
...
+	{
+		// Make sure we only allow decimal numbers and some special keys to delete characters
+		if (!((wParam >= '0' && wParam <= '9')
+			|| wParam == VK_CANCEL
+			|| wParam == VK_CLEAR
+			|| wParam == VK_DELETE
+			|| wParam == VK_BACK))
+		{
+			return FALSE;
+		}
```
RadWolfie, line 277: If you don't mind my asking, what was the purpose for this check?

src/gui/DlgInputConfig.cpp
```diff
@@ -144,6 +178,11 @@ INT_PTR CALLBACK DlgInputConfigProc(HWND hWndDlg, UINT uMsg, WPARAM wParam, LPAR
 				{
 				case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE): 
 				case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S): {
+					if (g_bHasOptChanges) {
+						UpdateInputOpt(hWndDlg);
+						g_InputDeviceManager.UpdateOpt(true);
+					}
```
RadWolfie, line 184: Do we need this code inside DeviceType switch statement? What about other controller types that are not standard controller? Since it is generic code. Perhaps we can check DeviceType's value itself outside of switch statement.

src/gui/DlgInputConfig.cpp
ergo720: This check is critical. The text that the user inputs for the mouse range must be a valid decimal number because this text is later converted by `std::stol`. If it is not, the function throws an exception and because it will be unhanded, it will cause cxbxr to crash.

src/core/kernel/init/CxbxKrnl.h
```diff
@@ -200,7 +200,7 @@ bool CxbxIsElevated();
 extern uint32_t CxbxKrnl_KernelThunkTable[379];
 
 extern bool g_bIsWine;
-
+extern bool g_bClipCursor;
```
RadWolfie, line 203: TODO: Eventually move this extern global clip cursor once move some unrelate functions outside of Direct3D9 files. Such as `EmuMsgProc` function.

src/gui/DlgInputConfig.cpp
ergo720: The purpose of this code is to synchronize the dinput mouse ranges with those that the user has chosen in the input gui. Without this, if the user has changed the ranges then, when they open the controller configuration, it will still use the old values, that is, those that were set when the input gui was first opened. This should be done regardless of the DeviceType, so that every xbox device under configuration will use the updated mouse ranges.

RadWolfie: I don't think you're getting the message. Couple lines down `DialogBoxParam` opens up duke configuration. Chihiro, Steel Battalion, dance pad, etc are not using standard controller. Which require a separate case with a different `DialogBoxParam` call to entirely different input configuration. That means it will require duplicate code you have here to the other controller type.

src/gui/DlgInputConfig.cpp
RadWolfie: 👌
If I look at `EditControlSubclassProc` function and look at the rest of the code within this function. I would not know what exactly this function serve the purpose. Perhaps rename it to have mouse word in it or add a comment above it will help clarification.

LukeUsher: Why not try/catch around stol, and handle the exception gracefully? (Unless it's performance-critical)

ergo720: In my opinion preventing the problem from happening in the first place is cleaner than trying to handle it later. Also, what should happen if the user inputs a garbage (non-integer) value here? Open a MessageBox describing the problem? That's more bloated than simply forbidding anything but integers.

RadWolfie: Actually, that wouldn't work. This function is done identical to eeprom's UI which only accept certain characters. There should be input listener, which is what we have here, to accept certain characters than hexadecimal/paste random string. Without the input listener, users can insert anything they want. Which can cause invalid input from what we expect from users. I don't think there is a way around this foolproof design.

RadWolfie (approved):
All review remarks are resolve and had verified it is working as intended.

PR #1966, opened 2020-09-25
Move xxHash to Submodule

RadWolfie:
Moved xxHash to submodule handler and update to latest release, v0.8.0, commit.


LukeUsher (approved):

RadWolfie:
**NOTICE: Since XXH3 is out of beta, it may be faster than crc32c. We're in need of benchmark testing to verify results.**

PR #1968, opened 2020-09-25
Fix for starting with debugger

x1nixmzeng:
The debugger wasn't launching from the main menu due to how it was handling the new CLI options passed to the loader


RadWolfie:
I downloaded CI's build, it still doesn't launch debugger properly for me.

x1nixmzeng:
I found the cmake script was adding the embedded resources and properties as sources - which the compiler was happy with - the app would even run, but it failed setting up the main form and immediately closed.

This is now fixed (moving the properties out of the sources group)

RadWolfie (approved):
Tested and confirmed back to working state.

PR #1969, opened 2020-09-27
Use fixed width types for xbox types + restructured kernel header files

ergo720:
This PR updates the xbox types to use fixed width standard types, as opposed to the previous variable length types, which is beneficial for future 64 bit builds and cpu emulation. I also moved the kernel header files from the OpenXDK folder to the cxbxr source tree and restructured the kernel header files. This change also fixed all the bogus errors reported by intellisense in the kernel files. Other than that, this should not have any functional changes.


ergo720:
Note: I changed the xbox type `WCHAR` from `wchar_t` to `char16_t`. The kernel function `RtlInitUnicodeString` was previously using `wcslen` on the `WCHAR` argument, but that breaks with `char16_t` types, so I replaced `wcslen` with `std::u16string(SourceString).length()` which does the same thing. I tested this with the kernel test suite (which has a test for `RtlInitUnicodeString`) and I confirmed that it still succeeds, so this change should be fine.

src/core/kernel/exports/EmuKrnlKi.h
```diff
@@ -170,7 +165,7 @@ extern xbox::KI_TIMER_LOCK KiTimerMtx;
         (_Object_)->OwnerThread = (_Thread_);                                \
         if ((_Object_)->Abandoned == TRUE) {                                 \
             (_Object_)->Abandoned = FALSE;                                   \
-            (_Thread_)->WaitStatus = STATUS_ABANDONED;                       \
+            (_Thread_)->WaitStatus = xbox::status_abandoned;                       \
```
RadWolfie, line 168: nit: fix slash alignment, same thing for another one below.

RadWolfie (approved):

PR #1970, opened 2020-10-01
Allow users to disable aspect ratio correction

LukeUsher:
This replaces the unused 'Hardware YUV overlay' option in the video settings.


PatrickvL (dismissed):
LGTM

RadWolfie:
Tested confirmed working as intended. The code is perfect except it is missing removal of HardwareYUV option from settings.ini file and update settings revision.
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/ea7007090c0d1084cf9e49ae3516e18291020122/src/common/Settings.cpp#L65-L67
and
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/ea7007090c0d1084cf9e49ae3516e18291020122/src/common/Settings.cpp#L904
For RemoveLegacyConfigs, simply use if statement for check below version to remove unused HardwareYUV option.

RadWolfie:
For clarification, using if statements is easier to handle and allow us to add at the end of the function. Any time we update versioning and something get remove. There shouldn't be any **breaks** at all in switch statement. Reason for that is... HardwareYUV exists way before version 6. Plus missing version 7 and 8 from the list.

I originally thought using switch statement would be cleaner environment. It is, except it doesn't have support such as "`case 7..9:`"

PatrickvL:
`case 7: FALL_THROUGH; `case 8: FALL_THROUGH; case 9:`

RadWolfie:
@PatrickvL Not exactly what I meant. 😛 What I meant is instead of repeat every numbers, use something like `case 1: .. case 9:`. D language has this type of support.

RadWolfie (approved):

PR #1975, opened 2020-10-05
Add an rdtsc false positive check for Group S Challenge [CC-005] [1.05]

CookiePLMonster:
Group S Challenge [1.03] was going ingame for years, but [1.05] crashed on startup. Turns out, it's due to a very funny `rdtsc` false positive which detected the instruction in the middle of a call instruction:

`E8 0F 31 01 00`

This PR adds a check for this false positive for a 0x01 rdtsc pattern (triggered also by the offset, as evidenced).

Group S Challenge [1.05] now goes into menus:
image

### DIFF
--- a/src/core/kernel/init/CxbxKrnl.cpp
+++ b/src/core/kernel/init/CxbxKrnl.cpp
@@ -492,7 +492,7 @@ const uint8_t rdtsc_pattern[] = {
     0xEB,
     0xF6,
     0xA1,
-    0x01
+    0x01                       // one false positive in Group S Challenge [1.05] .text E8 0F 31 01 00
 };
 const int sizeof_rdtsc_pattern = sizeof(rdtsc_pattern);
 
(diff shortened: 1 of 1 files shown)


LukeUsher:
I’ll approve this for now, but we really need a better solution for handling false positives than a blacklist. 

Perhaps we could disassemble the following instructions to check if the code is valid (if we hit an invalid opcode before a ret for example we can be sure it’s not code!)
we could also specifically check for call/jmp bytes

LukeUsher (approved):

CookiePLMonster:
> (if we hit an invalid opcode before a ret for example we can be sure it’s not code!)
easier said than done with x86 IMO, as even some gibberish can be a string of valid instructions.

> we could also specifically check for call/jmp bytes
could add a check for "is this an offset of a call/jump" check, but then you would again risk hitting false positives if e.g. data 2-5 bytes before `rdtsc` had `E8`/`E9` in it. It's a very tricky problem to solve without resorting to full disassembly, and also the reason why x86 is really bad when it comes to reading without "context" :|

PatrickvL:
When disassembling-validating the next few instructions, any absolute memory-reference outside of the XBE range could be dismissed as an invalid access, which might help a little...

PR #1976, opened 2020-10-05
Add several D3D8LTCG function variations for xdk-3911

CookiePLMonster:
Requires https://github.com/Cxbx-Reloaded/XbSymbolDatabase/pull/117 - as far as I understand, it must be merged first, and then this PR has to be rebased.

This variation puts Stage on the stack and pTexture in `eax`. The existing implementation was shared with xdk-4721 with a wrong assumption that both functions are the same, but they were not.

Test case: NASCAR Heat 2002

Gets NASCAR Heat 2002 from instantly crashing on startup to reaching invisible intro movies and menus.
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -4318,7 +4318,6 @@ xbox::void_xt __fastcall xbox::EMUPATCH(D3DDevice_SetVertexShaderConstantNotInli
 
 // LTCG specific D3DDevice_SetTexture function...
 // This uses a custom calling convention where parameter is passed in EAX
-// TODO: XB_trampoline plus Log function is not working due lost parameter in EAX.
 // Test-case: Metal Wolf Chaos
 xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetTexture_4)
 (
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
I will have more fixes for xdk-3911 D3D8LTCG coming (at least two more functions with incorrect definitions and correct signatures), so it might make sense to hold this PR off until I got those implemented. Marking as draft.

CookiePLMonster:
Updated with the following:

* Fixed `D3DDevice_SetVertexShaderConstant_8`. It's an LTCG variant which shuffles arguments around and puts `ConstantCount` in EDX, but the existing implementation was mediocre and never worked correctly. Replaced it with a proper implementation.
* Added `D3DDevice_SetStreamSource ` for xdk-3911, which passes the first argument in EDX.

This gets NASCAR Heat 2002 ingame and even past menus!
image

CookiePLMonster:
Updated with added implementations of `D3DDevice_DrawVerticesUP_12` and `D3DDevice_SetVertexShader_0`. Both those were patched in their non-LTCG variations, but those xdk-3911 only LTCG variations were not.

NASCAR Heat 2002 now displays intro movies:
image

CookiePLMonster:
Updated the PR with the latest XbSymbolDatabase update. It directly ties to this PR, so it makes sense to treat them as one chunk of work.

LukeUsher (approved):

PR #1977, opened 2020-10-07
Add a new rdtsc pattern based on NASCAR Heat 2002 logs

CookiePLMonster:
Based off

```
[0x0B04] INFO : INIT    Skipped potential rdtsc: Unknown opcode pattern  0xA3, @ 0x00103A18
```

being a sequence as follows:

```
.text:00103A18 010 0F 31                                   rdtsc
.text:00103A1A 010 A3 20 65 17 00                          mov     dword_176520, eax
```

**I have not verified this against false positives in other games so it would be good to check with with at least a handful.**


LukeUsher (approved):
nice job!

PR #1978, opened 2020-10-08
MSAA improvements basing on xdk-3911 behaviour

CookiePLMonster:
Introduces those changes to MSAA scaling:

* Scaling should be performed on `Clear` calls too.
* Scaling factors should be recalculated when binding a render target, and should be derived from multisampling type only when binding the back buffer.

The code contains a code path checking for `X_D3DMULTISAMPLE_SAMPLING_MULTI` sampling type and divides scaling by 2 if that is the case. **This is how xdk-3911 runtime does it**, but it seems to result in half-width rendering at the moment (possibly due to something else assuming that multisampling and supersampling are the same), so I left it commented out. It can be re-enabled once that issue is identified.

When combined with #1976, makes rendering in NASCAR Heat 2002 close to perfect:
image

Known regressions:
- [x] ~~Star Wars - Knights of the Old Republic [LA-003] - yellow screen; seems to happen only with VS2019 and not VS2017, so possibly a bad codegen and not a functional regression; needs further investigation.~~ Unrelated bug, fixed by #1981

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -173,6 +173,8 @@ static xbox::X_D3DSurface           *g_pXbox_DefaultDepthStencilSurface = xbox::
        xbox::X_D3DSurface           *g_pXbox_RenderTarget = xbox::zeroptr;
 static xbox::X_D3DSurface           *g_pXbox_DepthStencil = xbox::zeroptr;
        xbox::X_D3DMULTISAMPLE_TYPE   g_Xbox_MultiSampleType = xbox::X_D3DMULTISAMPLE_NONE;
+static float                         g_Xbox_MultiSampleXScale = 1.0f;
+static float                         g_Xbox_MultiSampleYScale = 1.0f;
 
        xbox::X_VERTEXSHADERCONSTANTMODE g_Xbox_VertexShaderConstantMode = X_D3DSCM_192CONSTANTS; // Set by D3DDevice_SetShaderConstantMode, TODO : Move to XbVertexShader.cpp
 static xbox::dword_xt                   g_Xbox_BaseVertexIndex = 0; // Set by D3DDevice_SetIndices, read by D3DDevice_DrawIndexedVertices : a value that's effectively added to every vertex index (as stored in an index buffer) by multiplying this by vertex stride and added to the vertex buffer start (see BaseVertexIndex in CxbxDrawIndexed)
@@ -2995,15 +2997,10 @@ static inline void GetMultiSampleOffset(float& xOffset, float& yOffset)
 	xOffset = yOffset = GetMultiSampleOffsetDelta();
 }
 
-static inline void GetMultiSampleScale(float& xScale, float& yScale)
-{
-	xScale = (float)CXBX_D3DMULTISAMPLE_XSCALE(g_Xbox_MultiSampleType);
-	yScale = (float)CXBX_D3DMULTISAMPLE_YSCALE(g_Xbox_MultiSampleType);
-}
-
 void GetMultiSampleOffsetAndScale(float& xScale, float& yScale, float& xOffset, float& yOffset)
 {
-	GetMultiSampleScale(xScale, yScale);
+	xScale = g_Xbox_MultiSampleXScale;
+	yScale = g_Xbox_MultiSampleYScale;
 	GetMultiSampleOffset(xOffset, yOffset);
 }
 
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
Rebased on the latest master to fix a "regression", which was an unrelated bug uncovered by this PR. Should avoid some potential other false positives when testing the relevant changes.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4099,10 +4114,8 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetViewport)
 	}
 
 	// Apply MSAA scale and offset
-	float xScale, yScale;
-	GetMultiSampleScale(xScale, yScale);
-	HostViewPort.Width *= (DWORD)xScale;
-	HostViewPort.Height *= (DWORD)yScale;
+	HostViewPort.Width *= (DWORD)g_Xbox_MultiSampleXScale;
```
PatrickvL, line 4117: Review note, also discussed in discord: Since the scales are floating point, all multiplications over ordinal values should done by casting to floating point first and cast back to ordinal after the multiplication

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4955,12 +4968,12 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_Clear)
...
     if (pRects != nullptr) {
-        // Scale the fill based on our scale factor
-        D3DRECT rect = *pRects;
-        rect.x1 *= g_RenderScaleFactor;
-        rect.x2 *= g_RenderScaleFactor;
-        rect.y1 *= g_RenderScaleFactor;
-        rect.y2 *= g_RenderScaleFactor;
+        // Scale the fill based on our scale factor and MSAA scale
+        D3DRECT rect;
+        rect.x1 = static_cast<LONG>(pRects->x1 * g_Xbox_MultiSampleXScale * g_RenderScaleFactor);
```
PatrickvL, line 4973: Ditto

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -480,10 +480,10 @@ const int X_D3DMULTISAMPLE_9_SAMPLES_MULTISAMPLE_GAUSSIAN          = 0x1233;
 const int X_D3DMULTISAMPLE_9_SAMPLES_SUPERSAMPLE_GAUSSIAN          = 0x2233;
 
 // Multisample masks (Cxbx additions)
-const int X_D3DMULTISAMPLE_YSCALE_MASK                             = 0x0003;
+const int X_D3DMULTISAMPLE_YSCALE_MASK                             = 0x000F;
```
PatrickvL, line 483: Could you please add a rationally in comments on why these masks are now set to 0xF, because as far as I remember, the maximum scale Xbox supports uses 2 bits maximum?

PatrickvL (changes requested):

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: This one is fine, since `g_Xbox_MultiSampleXScale` is a float everything will be promoted to a float - hence the cast :)

src/core/hle/D3D8/XbD3D8Types.h
CookiePLMonster: The MSAA scale computation from xdk-3911 as well as whatever XDK KoToR uses have this as their baseline for scaling:
```cpp
    v3 = (double)((unsigned __int8)g_Xbox_MultiSampleType >> 4);
    v1 = (double)(g_Xbox_MultiSampleType & 0xF);
```

PatrickvL: With all due respect, that code doesn't necessary imply all 4 bits are meant to be used. From what I remember, 3 was the maximum. But if we can find a test case that exceeds 2 bits, and your wider mask produces better results than the old 2 bits mask, then we'd have proof. So, maybe a test case could be added for scales above 3?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Ah, implicit behaviour - those are the best ;)

src/core/hle/D3D8/XbD3D8Types.h
CookiePLMonster: Test case sounds good! Will add a `scaleX > 3.0f || scaleY > 3.0f` test case.

CookiePLMonster:
Test case added, made scaling calculated on floats where applicable. As far as I can tell all instances of scaling are now covered.

PatrickvL (approved):
Nicely done, thanks!

PR #1981, opened 2020-10-09
Actually zero g_InlineVertexBuffer_Table[0]

CookiePLMonster:
`operator=` invoked default constructors where available, and a default constructor for `D3DXVECTOR4` does nothing. `memset`/`memcpy` zeroes and copies the entire structure for real.

Fixes a festival of colours in Star Wars: Knights of the Old Republic depending on random factors such as unrelated code changes, Debug/Release configuration and optimization options (in practice, relying on stale stack values):

Before:
image
image

After:
image

PR #1982, opened 2020-10-10
Use std::vector for g_InlineVertexBuffer_Table instead of realloc

ergo720:
This PR was triggered after a lengthy discussion among the developers about https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1981. In short, we don't want to use `memcpy/memset` for `g_InlineVertexBuffer_Table` and instead, we'd prefer a C++ way to do it. This PR turns the `g_InlineVertexBuffer_Table` pointer to a `std::vector` and implements the constructor and the assignment operator for `_D3DIVB`, which removes the need to call `memcpy/memset`. I tested this with Constantine (which also calls `D3DDevice_SetVertexData4f`) and I confirmed that `g_InlineVertexBuffer_Table[0] = {};` indeed sets the first vertex back to zero after a call to `D3DDevice_Begin`. Please also check that Star Wars: Knights of the Old Republic still works correctly, since I don't have that game.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -4606,20 +4606,30 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetVertexData4f)
 
 	// Grow g_InlineVertexBuffer_Table to contain at least current, and a potentially next vertex
 	if (g_InlineVertexBuffer_TableLength <= g_InlineVertexBuffer_TableOffset + 1) {
+		UINT InlineVertexBuffer_TableLength_Original = g_InlineVertexBuffer_TableLength;
 		if (g_InlineVertexBuffer_TableLength == 0) {
 			g_InlineVertexBuffer_TableLength = PAGE_SIZE / sizeof(struct _D3DIVB);
 		} else {
 			g_InlineVertexBuffer_TableLength *= 2;
 		}
 
-		g_InlineVertexBuffer_Table = (struct _D3DIVB*)realloc(g_InlineVertexBuffer_Table, sizeof(struct _D3DIVB) * g_InlineVertexBuffer_TableLength);
-		EmuLog(LOG_LEVEL::DEBUG, "Reallocated g_InlineVertexBuffer_Table to %d entries", g_InlineVertexBuffer_TableLength);
+		for (unsigned i = 0; i < (g_InlineVertexBuffer_TableLength - InlineVertexBuffer_TableLength_Original); ++i) {
+			g_InlineVertexBuffer_Table.emplace_back(_D3DIVB{});
+		}
+
+		EmuLog(LOG_LEVEL::DEBUG, "Expanded g_InlineVertexBuffer_Table to %u entries", g_InlineVertexBuffer_TableLength);
+
+		// Sanity check: ensure that g_InlineVertexBuffer_Table is not growing indefinetly. This can happen if D3DDevice_Begin and D3DDevice_End
+		// are not patched, since they both reset g_InlineVertexBuffer_TableOffset back to zero, thus preventing further growth
+		if (g_InlineVertexBuffer_TableLength > 50000) {
+			LOG_TEST_CASE("g_InlineVertexBuffer_TableLength > 50000! This probably means that g_InlineVertexBuffer_Table is growing indefinitely.");
+		}
 	}
 
 	// Is this the initial call after D3DDevice_Begin() ?
 	if (g_InlineVertexBuffer_FVF == 0) {
 		// Set first vertex to zero (preventing leaks from prior Begin/End calls)
-		memset(&g_InlineVertexBuffer_Table[0], 0, sizeof(g_InlineVertexBuffer_Table[0]));
+		g_InlineVertexBuffer_Table[0] = {};
 
 		// Handle persistent vertex attribute flags, by resetting non-persistent colors
 		// to their default value (and leaving the persistent colors alone - see the
@@ -4692,7 +4702,7 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetVertexData4f)
 			// Start a new vertex
 			g_InlineVertexBuffer_TableOffset++;
 			// Copy all attributes of the previous vertex (if any) to the new vertex
-			memcpy(&g_InlineVertexBuffer_Table[g_InlineVertexBuffer_TableOffset], &g_InlineVertexBuffer_Table[o], sizeof(g_InlineVertexBuffer_Table[o]));
+			g_InlineVertexBuffer_Table[g_InlineVertexBuffer_TableOffset] = g_InlineVertexBuffer_Table[o];
 	
 			break;
 		}
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -4614,7 +4614,7 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetVertexData4f)
 		}
 
 		for (unsigned i = 0; i < (g_InlineVertexBuffer_TableLength - InlineVertexBuffer_TableLength_Original); ++i) {
-			g_InlineVertexBuffer_Table.emplace_back(_D3DIVB{});
+			g_InlineVertexBuffer_Table.emplace_back();
 		}
 
 		EmuLog(LOG_LEVEL::DEBUG, "Expanded g_InlineVertexBuffer_Table to %u entries", g_InlineVertexBuffer_TableLength);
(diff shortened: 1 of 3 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4649,7 +4659,7 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetVertexData4f)
 			// Start a new vertex
 			g_InlineVertexBuffer_TableOffset++;
 			// Copy all attributes of the previous vertex (if any) to the new vertex
-			memcpy(&g_InlineVertexBuffer_Table[g_InlineVertexBuffer_TableOffset], &g_InlineVertexBuffer_Table[o], sizeof(g_InlineVertexBuffer_Table[o]));
+			g_InlineVertexBuffer_Table[g_InlineVertexBuffer_TableOffset] = g_InlineVertexBuffer_Table[o];
```
CookiePLMonster, line 4662: If `g_InlineVertexBuffer_TableOffset` is used only to index the "head" of the vector, why not go one step further and remove it, instead relying on `emplace_back`?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4563,20 +4563,30 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetVertexData4f)
...
 			g_InlineVertexBuffer_TableLength = PAGE_SIZE / sizeof(struct _D3DIVB);
 		} else {
 			g_InlineVertexBuffer_TableLength *= 2;
 		}
 
-		g_InlineVertexBuffer_Table = (struct _D3DIVB*)realloc(g_InlineVertexBuffer_Table, sizeof(struct _D3DIVB) * g_InlineVertexBuffer_TableLength);
-		EmuLog(LOG_LEVEL::DEBUG, "Reallocated g_InlineVertexBuffer_Table to %d entries", g_InlineVertexBuffer_TableLength);
+		for (unsigned i = 0; i < (g_InlineVertexBuffer_TableLength - InlineVertexBuffer_TableLength_Original); ++i) {
+			g_InlineVertexBuffer_Table.emplace_back(_D3DIVB{});
+		}
```
CookiePLMonster, line 4575: If this needs to stay (see my comment below), it should just become a `.resize(g_InlineVertexBuffer_TableLength, _D3DIVB{})`.

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -69,6 +69,58 @@ uint32_t GetPixelContainerWidth(xbox::X_D3DPixelContainer* pPixelContainer);
...
 void ApplyXboxMultiSampleOffsetAndScale(float& x, float& y);
 
+_D3DIVB::_D3DIVB()
+{
+	Position.x = 0.0f;
+	Position.y = 0.0f;
+	Position.z = 0.0f;
+	Rhw = 0.0f;
+	for (unsigned i = 0; i < 4; ++i) {
+		Blend[i] = 0.0f;
```
CookiePLMonster, line 79: Could use `std::fill(std::begin(Blend), std::end(Blend), 0.0f);`

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -69,6 +69,58 @@ uint32_t GetPixelContainerWidth(xbox::X_D3DPixelContainer* pPixelContainer);
...
+	Normal.x = 0.0f;
+	Normal.y = 0.0f;
+	Normal.z = 0.0f;
+	Diffuse = 0u;
+	Specular = 0u;
+	Fog = 0.0f;
+	BackDiffuse = 0u;
+	BackSpecular = 0u;
+	for (unsigned i = 0; i < 4; ++i) {
+		TexCoord[i].x = 0.0f;
```
CookiePLMonster, line 90: Ditto

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -69,6 +69,58 @@ uint32_t GetPixelContainerWidth(xbox::X_D3DPixelContainer* pPixelContainer);
...
+}
+
+struct _D3DIVB &_D3DIVB::operator=(const struct _D3DIVB &Val)
+{
+	Position.x = Val.Position.x;
+	Position.y = Val.Position.y;
+	Position.z = Val.Position.z;
+	Rhw = Val.Rhw;
+	for (unsigned i = 0; i < 4; ++i) {
+		Blend[i] = Val.Blend[i];
```
CookiePLMonster, line 104: Could use `std::copy` in the same fashion as fill in the constructor.

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -69,6 +69,58 @@ uint32_t GetPixelContainerWidth(xbox::X_D3DPixelContainer* pPixelContainer);
...
+	Normal.x = Val.Normal.x;
+	Normal.y = Val.Normal.y;
+	Normal.z = Val.Normal.z;
+	Diffuse = Val.Diffuse;
+	Specular = Val.Specular;
+	Fog = Val.Fog;
+	BackDiffuse = Val.BackDiffuse;
+	BackSpecular = Val.BackSpecular;
+	for (unsigned i = 0; i < 4; ++i) {
+		TexCoord[i].x = Val.TexCoord[i].x;
```
CookiePLMonster, line 115: Ditto

medievil1:
verified, KOTOR works(no yellow issue)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Chiming in; I'm unfamiliar with emplace_back, but if it grows the vector memory, it can't be used since the storage space for g_InlineVertexBuffer_Table is meant to be reused. If vector does facilitate this use case, and emplace_back uses an internal index that's completely separate from the allocated size, then it's fine

CookiePLMonster: It can, doesn't have to. However since a call to `.clear()` [does not have to free the underlying memory](http://www.cplusplus.com/reference/vector/vector/clear/), then I see no reason not to use regular vector semantics for this.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
ergo720: If I read the standard correctly, `resize()` will append copy constructed elements to the vector, while `emplace_back()` will construct in place by forwarding the arguments to the constructor. So I'd rather use the latter instead.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
ergo720: Since I'm not too familiar with this area of the code, I'd rather keep the existing logic of `g_InlineVertexBuffer_Table` intact to be honest.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: But all you forward here is `_D3DIVB{}` itself, and thus it won't make any noticeable difference. In this case `resize()` more reflects the intent.

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -69,6 +69,44 @@ uint32_t GetPixelContainerWidth(xbox::X_D3DPixelContainer* pPixelContainer);
...
+	std::fill(std::begin(Blend), std::end(Blend), 0.0f);
+	Normal.x = 0.0f;
+	Normal.y = 0.0f;
+	Normal.z = 0.0f;
+	Diffuse = 0u;
+	Specular = 0u;
+	Fog = 0.0f;
+	BackDiffuse = 0u;
+	BackSpecular = 0u;
+	std::fill_n(std::begin(TexCoord), 4, D3DXVECTOR4{ 0.0f , 0.0f, 0.0f, 0.0f });
```
CookiePLMonster, line 87: `TexCoord` is a normal array just like `Blend`, so can just use a normal fill
```suggestion
	std::fill(std::begin(TexCoord), std::end(TexCoord), D3DXVECTOR4{ 0.0f , 0.0f, 0.0f, 0.0f });
```

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -69,6 +69,44 @@ uint32_t GetPixelContainerWidth(xbox::X_D3DPixelContainer* pPixelContainer);
...
+	std::copy(std::begin(Val.Blend), std::end(Val.Blend), std::begin(Blend));
+	Normal.x = Val.Normal.x;
+	Normal.y = Val.Normal.y;
+	Normal.z = Val.Normal.z;
+	Diffuse = Val.Diffuse;
+	Specular = Val.Specular;
+	Fog = Val.Fog;
+	BackDiffuse = Val.BackDiffuse;
+	BackSpecular = Val.BackSpecular;
+	std::copy_n(std::begin(Val.TexCoord), 4, std::begin(TexCoord));
```
CookiePLMonster, line 105: Ditto
```suggestion
	std::copy(std::begin(Val.TexCoord), std::end(Val.TexCoord), std::begin(TexCoord));
```

LukeUsher (approved):

PatrickvL:
Here a write-up on the background story of this change (and it's preceding pull request) : https://cookieplmonster.github.io/2020/10/20/cxbx-r-kotor-random-bug/

PR #1984, opened 2020-10-11
IDirectSoundBuffer_SetBufferData: Stop the sound before waiting for finish

CookiePLMonster:
I checked `CDirectSoundBuffer_SetBufferData` implementations in games using dsound `3925` and `5455` and they both send a stop request before waiting for the buffer to stop playing.

This resolves a sound related deadlock in NASCAR Heat 2002, since it was waiting for a looped sound to finish (which, naturally, never happened).

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
@@ -756,9 +756,11 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_SetBufferData)
     HRESULT hRet = DSERR_OUTOFMEMORY;
     DWORD dwStatus;
 
-    // force wait until buffer is stop playing.
+    // stop and force wait until the buffer has stopped playing.
+    pThis->EmuDirectSoundBuffer8->Stop();
     pThis->EmuDirectSoundBuffer8->GetStatus(&dwStatus);
-    while ((dwStatus & DSBSTATUS_PLAYING) > 0) {
+    while ((dwStatus & DSBSTATUS_PLAYING) != 0) {
+        // TODO: Add a timeout, like on xbox
         SwitchToThread();
         pThis->EmuDirectSoundBuffer8->GetStatus(&dwStatus);
     }


src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
```diff
@@ -756,9 +756,11 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_SetBufferData)
     HRESULT hRet = DSERR_OUTOFMEMORY;
     DWORD dwStatus;
 
-    // force wait until buffer is stop playing.
+    // stop and force wait until buffer is stop playing.
```
Margen67, line 759:
```suggestion
    // stop and force wait until buffer stops playing.
```

PatrickvL: or, "has stopped playing"

src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
```diff
@@ -756,9 +756,11 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_SetBufferData)
...
+    pThis->EmuDirectSoundBuffer8->Stop();
     pThis->EmuDirectSoundBuffer8->GetStatus(&dwStatus);
-    while ((dwStatus & DSBSTATUS_PLAYING) > 0) {
+    while ((dwStatus & DSBSTATUS_PLAYING) != 0) {
```
PatrickvL, line 762: Since the calculation results in an unsigned integral value, there's no functional difference between `>0` and `!= 0`, so it wasn't really necessary do change this, right?

PatrickvL (approved):
Great to have yet another deadlock fixed!

src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
CookiePLMonster: It's not of course, just personal preference given it's a bit mask.

RadWolfie: It's fine != operator, which sort of should had been done in first place. 😛

RadWolfie:
To be clear, internal stop function was called in a different child function.

RadWolfie (approved):

PR #1985, opened 2020-10-11
Make RDTSC and ACPI timers stateful to fix overflows

CookiePLMonster:
**This PR needs some testing. Pay close attention to games running too slow or too fast with those changes.**

Stateless RDTSC and ACPI timers were ticking relative to the host QPC and multiplied to nanoseconds. This resulted in values so huge they would overflow since 20-30 minutes.

Introducing state allows to multiply to nanoseconds only over a delta value, which should be reasonably small in almost all cases.

I also proofed these calculations against data races by putting them under a mutex, but I only lock it if "Run Xbox threads on all cores" is enabled. Without this hack it **should** not be required.

Confirmed fixes:
* NASCAR Heat 2002 - fixes game freezing after X minutes of gameplay (from a few minutes, up to ~30 minutes)

### DIFF
--- a/src/core/kernel/exports/EmuKrnlKe.cpp
+++ b/src/core/kernel/exports/EmuKrnlKe.cpp
@@ -324,43 +324,64 @@ void InitDpcThread()
 	g_DpcData.DpcEvent = CreateEvent(/*lpEventAttributes=*/nullptr, /*bManualReset=*/FALSE, /*bInitialState=*/FALSE, /*lpName=*/nullptr);
 }
 
-#define XBOX_TSC_FREQUENCY 733333333 // Xbox Time Stamp Counter Frequency = 733333333 (CPU Clock)
-#define XBOX_ACPI_FREQUENCY 3375000  // Xbox ACPI frequency (3.375 mhz)
-ULONGLONG NativeToXbox_FactorForRdtsc;
-ULONGLONG NativeToXbox_FactorForAcpi;
+static constexpr uint32_t XBOX_TSC_FREQUENCY = 733333333; // Xbox Time Stamp Counter Frequency = 733333333 (CPU Clock)
+static constexpr uint32_t XBOX_ACPI_FREQUENCY = 3375000;  // Xbox ACPI frequency (3.375 mhz)
+static constexpr uint32_t SEC_TO_NSEC = 1000000000; // For seconds -> nanoseconds conversions
+static uint64_t NativeToXbox_FactorForRdtsc = 0, NativeToXbox_FactorForAcpi = 0;
 
-ULONGLONG CxbxGetPerformanceCounter(bool acpi) {
-	LARGE_INTEGER tsc;
-	ULARGE_INTEGER scaledTsc;
+// State for CxbxGetPerformanceCounter - concurrent access should be next to non-existent, but secure against it anyway
+static std::mutex RdtscLock, AcpiLock;
+static LARGE_INTEGER LastRdtscQPC, LastAcpiQPC;
+static ULONGLONG CurrentRdtsc = 0, CurrentAcpi = 0;
+static ULONGLONG CurrentRdtscRemainder = 0, CurrentAcpiRemainder = 0;
 
-	QueryPerformanceCounter(&tsc);
+ULONGLONG CxbxGetPerformanceCounter(bool acpi)
+{
+	if (acpi == false && NativeToXbox_FactorForRdtsc != 0) {
+		std::lock_guard<std::mutex> lock(RdtscLock);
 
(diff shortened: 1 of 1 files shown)


src/core/kernel/exports/EmuKrnlKe.cpp
```diff
@@ -324,43 +324,75 @@ void InitDpcThread()
...
 }
 
-#define XBOX_TSC_FREQUENCY 733333333 // Xbox Time Stamp Counter Frequency = 733333333 (CPU Clock)
-#define XBOX_ACPI_FREQUENCY 3375000  // Xbox ACPI frequency (3.375 mhz)
-ULONGLONG NativeToXbox_FactorForRdtsc;
-ULONGLONG NativeToXbox_FactorForAcpi;
+static constexpr uint32_t XBOX_TSC_FREQUENCY = 733333333; // Xbox Time Stamp Counter Frequency = 733333333 (CPU Clock)
+static constexpr uint32_t XBOX_ACPI_FREQUENCY = 3375000;  // Xbox ACPI frequency (3.375 mhz)
+static constexpr uint32_t SEC_TO_NSEC = 1000000000; // For seconds -> nanoseconds conversions
+static uint64_t NativeToXbox_FactorForRdtsc;
```
PatrickvL, line 330: Since these are checked against zero, and there's currently only one function that does an assignment, it's probably best to initialise these variables to zero.

PatrickvL (changes requested):

src/core/kernel/exports/EmuKrnlKe.cpp
CookiePLMonster: Done!

PatrickvL (dismissed):
LGTM, thanks.

Once enough testing has been done and no regressions (and possibly some improvements) have been reported in here, it's good to merge.

literalmente-game:
Tested (with no changes):
- Arctic Thunder
- Castlevania - Curse Of Darkness
- Constantine
- Darkwatch
- Dino Crisis 3
- Doom 3
- Fatal Frame
- Lotus Challenge
- Silent Hill 2
- Teen Titans

Not tested yet (but could be probably fixed by this PR):
- Total Overdose (needs to be tested 1+ hour to be sure anything actually improved)

src/core/kernel/exports/EmuKrnlKe.cpp
```diff
@@ -324,43 +324,74 @@ void InitDpcThread()
...
+	return lock;
+}
 
-ULONGLONG CxbxGetPerformanceCounter(bool acpi) {
-	LARGE_INTEGER tsc;
-	ULARGE_INTEGER scaledTsc;
+ULONGLONG CxbxGetPerformanceCounter(bool acpi)
+{
+	if (acpi == false && NativeToXbox_FactorForRdtsc != 0) {
+		auto lock = LockIfNeeded(RdtscLock);
```
CookiePLMonster, line 351: Note to self, it needs to lock even when the hack is disabled because a context switch can happen here. Consider doing it lock-free.

CookiePLMonster:
As discussed on Discord, made the mutex lock at all times, since even without the multicore hack context switch could happen in the middle of the function.

literalmente-game:
Retested:
- Arctic Thunder: no changes
- Castlevania - Curse Of Darkness: no changes
- Constantine: no changes
- Darkwatch: no changes
- Dino Crisis 3: no changes
- Doom 3: no changes
- Fatal Frame: no changes
- Lotus Challenge: no changes
- Silent Hill 2: no changes
- Teen Titans: didn't changed, crashed at the same point

Newly tested:
- GTA 3: improved. Latest known behavior was some visual issues covering the entire screen after being in-game between 20 and 40 minutes. Game was just tested for 2 hours without being able to repro said issue. Bad news: game crashed ("Sofware stopped working") anyway, but now probably related to something else.

Not tested yet (but could be probably fixed by this PR):
- Aliens vs Predator Extinction (always freezes after some time)
- Mafia (reportedly it freezes after some time)
- Total Overdose (needs to be tested 1+ hour to be sure anything actually improved)
- GTA Vice City (reportedly it freezes after some time)

PatrickvL (approved):

CookiePLMonster:
Idea - calculations could be reduced to further reduce the likelihood of overflow by dividing the factors by `GCD(1000000000, NativeToXbox_FactorForRdtsc);` and `GCD(1000000000, NativeToXbox_FactorForAcpi);`. In my test it yields 1 for rdtsc and 2 for acpi, but I don't see any reason not to go for it anyway - thoughts?

PatrickvL:
Adding more complexity for just one more bit looks not really worth the effort. Creative thinking though.

PR #1986, opened 2020-10-11
Update cxbx-debugger to support switching XBEs during emulation

### LINKED ISSUE #1428: Cxbx-Debugger does not work with multi-xbe titles
Cxbx-Debugger does not work with multi-xbe titles - it just ends debugging session when game switches to another xbe.
Test-case: The Elder Scrolls III - Morrowind.

x1nixmzeng:
Updated the debugger project to use MDI forms to allow multiple windows inside the main app

The change needed in the cxbx project is trivial - see `EmuKrnlHal.cpp`

Fixes #1428


RadWolfie:
Is there a reason for debugger window being so extra large? Plus there are lot of wasted space for having it being extra large.
ref: (top reworked window; bottom upstream window)
image

x1nixmzeng:
I'm using a new monitor that requires scaling to 150% but it doesn't play well with Visual Studio - fixed!

LukeUsher (approved):

PR #1991, opened 2020-10-14
DSound improvements

CookiePLMonster:
**Needs testing. Test games especially if:**
* **Some sounds stopped playing after #1983**
* **Sounds do not pause/stop/mute when they should have and keep playing infinitely in a loop**

Mutually exclusive with #1993 

Introduces these changes:

* Fixes a copypaste typo in GetFormat_4034_lower and changes the signature of `GetFormat()` to be easier to use
* Refactors `HybridDirectSoundBuffer_SetMixBinVolumes_8` - removes redundant nested loops and adds a special case for mono sounds with left/right speaker mixbins muted. This might be DSP specific (that's why a test case has been added), but mono sounds with muted left/right speaker but not the center speaker seem to use that volume instead. The test case is there to verify if any titles break because of this.
* Fixes a bug in `IDirectSoundBuffer_SetBufferData` where submitting 0 byte long buffer did not stop the audio. Looking at the log and existing bugs of NASCAR Heat 2002, it's obvious that games can call `SetBufferData(NULL, 0);` to stop sounds. The Xbox implementation does not verify arguments against 0/null specifically either (at least not before stopping the current sound).

Fixes two bugs in NASCAR Heat 2002:
* Muted spotter audio
* Sounds "lingering" in pause menu and during the race

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <mutex>
+#include <optional>
 
 #include "common/XADPCM.h"
 #include "core/hle/DSOUND/XbDSoundTypes.h"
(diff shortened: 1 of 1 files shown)


LanHikariDS:
Mentioned this in Discord, but for the sake of documenting, this doesn't correct the issue with Burnout 3 where music plays during the intros, causing songs to overlap before a profile is selected, while also incorrectly looping the title theme- https://www.youtube.com/watch?v=Hu6jhXwwf6c

CookiePLMonster:
#1993 was picked, so this PR goes.

PR #1993, opened 2020-10-15
DSound improvements (alternative volume heuristics)

CookiePLMonster:
PR mutually exclusive to #1991, submitted so people can test and pick which sounds better.

Brings the same changes as the other PR, but with different heuristics for picking the buffer volume - this PR picks the maximum channel volume from all speakers (except for the LFE speaker).

Since the current code picks a single volume level for all channels, it will never be perfect - but those two PRs exist so we can test which of the two approaches provides better volume levels.

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
@@ -27,6 +27,7 @@
 #pragma once
 
 #include <mutex>
+#include <optional>
 
 #include "common/XADPCM.h"
 #include "core/hle/DSOUND/XbDSoundTypes.h"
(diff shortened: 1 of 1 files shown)


LanHikariDS:
This PR does not fix the volume issue for sound effects in Jedi Outcast, unfortunately

RadWolfie:
I am currently in favor of get this pull request merge since it is more logical than the other one. Plus what I had seen in several titles do have several channels involve center channel mixbin volume.

@literalmente-game Had you made any tests with this pull request for any regression/improvement?

literalmente-game:
Couldn't find any difference in all games I've checked (around 20 games). Burnout engine sound it's _probably_ slightly higher with this change, but nothing very noticeable

RadWolfie (approved):

gandalfthewhite19890404:
Black is completely silent with this PR.

PatrickvL:
@CookiePLMonster stated that this PR causes a regression : Main menu BGM is gone in The Warriors

PR #1994, opened 2020-10-17
Treat identical X_D3DResource objects with a different address as one

CookiePLMonster:
This fixes issues where resource objects got relocated in memory or reused without having been released properly first.

Fixes black cars in NASCAR Heat 2002 caused by the game bug - texture resources were not refcounted properly before being released (thus this is **not** a memory leak), and later "reappeared" as reused under a different memory address.
image


LukeUsher (approved):
This is an obvious oversight/issue with our resource management code, as such I'm willing to merge rather quickly.

PatrickvL:
Loving the way this fix also resulted in less state and code! Well done!

PR #1997, opened 2020-10-17
emux86: fix bad bios rom mapping

LukeUsher:
Bios ROM region reads were incorrectly based at 0xFFF00000 instead of 0xFF000000

Prevents a crash in Shenmue II

### DIFF
--- a/src/devices/x86/EmuX86.cpp
+++ b/src/devices/x86/EmuX86.cpp
@@ -149,6 +149,10 @@ uint32_t EmuFlash_Read32(xbox::addr_xt addr) // TODO : Move to EmuFlash.cpp
 {
 	uint32_t r;
 
+    // Some games attempt to read from the BIOS ROM directly, for purposes of hardware-version detection
+    // And other (currently unknown) reasons. We can't include the entire bios, and we don't want to require it
+    // So we specifiy the specific values that games rely on.
+
 	switch (addr) {
     case 0x08: // ? Test Case - Shenmue II attempts to read this address but never uses the resulting value? 
         r = 0x2B16D065;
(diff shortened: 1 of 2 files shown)


src/devices/x86/EmuX86.cpp
```diff
@@ -176,8 +179,8 @@ uint32_t EmuX86_Read(xbox::addr_xt addr, int size)
 
 	uint32_t value;
 
-	if (addr >= XBOX_FLASH_ROM_BASE) { // 0xFFF00000 - 0xFFFFFFF
-		return EmuFlash_Read32(addr - XBOX_FLASH_ROM_BASE); // TODO : Make flash access size-aware
+	if (addr >= FLASH_DEVICE1_BASE) { // 0xFFF00000 - 0xFFFFFFF
```
PatrickvL, line 182: Comment still mentions 0xFFF

src/devices/x86/EmuX86.cpp
```diff
@@ -150,6 +150,9 @@ uint32_t EmuFlash_Read32(xbox::addr_xt addr) // TODO : Move to EmuFlash.cpp
 	uint32_t r;
 
 	switch (addr) {
+    case 0x08: // ? Test Case - Shenmue II attempts to read this address but never uses the resulting value? 
+        r = 0x2B16D065;
```
PatrickvL, line 154: What's the meaning of this magic value?

PatrickvL (commented):
2 small comments

src/devices/x86/EmuX86.cpp
LukeUsher: No idea: this is the value that exists in that address on a real Xbox. (It’s in rom, so it’s constant, and it doesn’t change between bios revisions either)

PatrickvL: Maybe mention that in a comment, then?

LukeUsher: I updated comment at the top of the function to explain it's purpose.

PatrickvL (approved):

PR #1511, opened 2018-11-28
Debugging hangs

LukeUsher:
PR provided to allow easy testing by those affected by the new hanging issues. (AppVeyor autobuilds)

Please don't merge yet

Issue #857, opened 2018-01-14
PsCreateSystemThreadEx regression


PatrickvL:
Since https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/853 went in yesterday, some titles hang instead of continuing;

Confirmed cases:
https://github.com/Cxbx-Reloaded/game-compatibility/issues/542#issuecomment-357476308
https://github.com/Cxbx-Reloaded/game-compatibility/issues/62#issuecomment-357476876

Possibly:
https://github.com/Cxbx-Reloaded/game-compatibility/issues/118#issuecomment-357470829

Instead of reverting the entire commit (and losing the safeguards that got added) the solution is probably to move this line `SetEvent(hStartedEvent);` down to where it was before, under the `callComplete:` label.

It's strange though. Restoring that previous behaviour will make the thread calling `PsCreateSystemThreadEx` wait until the new thread is done executing. You would expect it to return once the new thread is created (especially so if the new thread is initially created to be in suspended state!).

Thus, I suspect there's something else wrong instead of this. But that will have to be researched later. First get this regression out if the way. Then, create an issue to look into this strange behaviour. (It might even be that the caller *has* to wait, but then we should verify and document that.)

LukeUsher:
PR #858 solves this issue in Call of Duty, but I'm not entirely sure why.

Thread creation is in a poor state at the moment, so much so that all Debug executables assert immediately due to incorrect memory layout (mostly regarding TLS), this is just another issue that furthers the need for revisiting these kernel APIs and creating a better implementation,

ergo720:
All three games reported above are now confirmed as booting and no longer hang. Thus, it's very likely this was fixed during the past years and can probably be closed now.

PatrickvL:
See #1998

PR #1998, opened 2020-10-17
Simplify thread creation logic, remove hardcoded delays and tighten affinity changes

CookiePLMonster:
**This PR needs extensive testing. It touches one of the areas of emulation which frequently had regressions resulting in games freezing (notably, #1511 and #857). Below I have collected a list of games which were mentioned in such issues in the past. Please verify those games for regressions or improvements:**

- ❔ Blood Wake (known past regression, later fixed)
- ❔ [Call of Duty: Finest Hour](https://github.com/Cxbx-Reloaded/game-compatibility/issues/542) (known past regression, later fixed)
- ➖ [Far Cry Instincts](https://github.com/Cxbx-Reloaded/game-compatibility/issues/62) (known past regression, later fixed) **- no change**
- ❔ Grand Theft Auto III (froze during intro movies - fixed for me, but please verify)
- ❔ Red Faction 2 (known past regression, later fixed)
- ❔ The Simpsons: Road Rage (known past regression, later fixed)
- ❔ The Warriors (crashed on startup - fixed for me, but please verify)

This PR simplifies the thread creation logic and removes hardcoded "magic" sleeps in favour of tighter affinity changes. Those "magic" sleeps were introduced to work around games hanging due to timing issues, but it appears those hangs would happen, because child threads would initially execute on any core before changing their own affinity.

Changes introduced in this PR:
* Changed `ThreadId` parameter type to `PWORD`.
* Removed `hStartedEvent` since it doesn't seem to protect anything meaningful anymore, and only messed with thread scheduling.
* Starts child threads suspended and finalize their initialization before resuming - this closes a possible race condition where a child thread uses `ThreadHandle` or `dwThreadId` and starts before `_beginthreadex` even returns. It also allowed me to remove "magic" sleeps by ensuring that the child thread really never starts on a wrong CPU core.

Gets The Warriors in-game. Before this PR, a race condition in the original game code was hit - the game spawns a thread which waits on the event. However, the event in question is created **after** spawning that thread. Previous thread creation timings always caused the child thread to try and wait on the event before it was created.
image

**Previous PR description below, not valid anymore - the hack was now replaced with a proper change.**

~~This PR adds a similar hack to the one found in Xenia: https://github.com/xenia-project/xenia/blob/8123978c498257ad9d65353bc59af3e544b4fc27/src/xenia/kernel/xthread.cc#L502~~

~~Delay the thread startup a bit to "simulate" slow startup and to encourage the scheduler not to switch to the newly created thread right away. This **is** a hack, but it's also the only way to accommodate for buggy games which initialize resources after spawning a thread which uses them.~~

~~Gets The Warriors in-game. It needs such a delay because it spawns a thread which waits on the event. However, the event in question is created **after** spawning the thread, so context switching to that thread right after creating it resulted in a spurious wakeup and in consequence a crash.~~

### DIFF
--- a/src/core/kernel/common/ps.h
+++ b/src/core/kernel/common/ps.h
@@ -25,7 +25,7 @@ namespace xbox
 XBSYSAPI EXPORTNUM(254) ntstatus_xt NTAPI PsCreateSystemThread
 (
 	OUT PHANDLE         ThreadHandle,
-	OUT PHANDLE         ThreadId OPTIONAL,
+	OUT PDWORD          ThreadId OPTIONAL,
 	IN  PKSTART_ROUTINE StartRoutine,
 	IN  PVOID           StartContext,
 	IN  boolean_xt         DebuggerThread
@@ -40,7 +40,7 @@ XBSYSAPI EXPORTNUM(255) ntstatus_xt NTAPI PsCreateSystemThreadEx
 	IN  ulong_xt           ThreadExtensionSize,
 	IN  ulong_xt           KernelStackSize,
 	IN  ulong_xt           TlsDataSize,
-	OUT PHANDLE         ThreadId OPTIONAL,
+	OUT PDWORD          ThreadId OPTIONAL,
 	IN  PKSTART_ROUTINE StartRoutine,
 	IN  PVOID           StartContext,
 	IN  boolean_xt         CreateSuspended,
--- a/src/core/kernel/exports/EmuKrnlPs.cpp
+++ b/src/core/kernel/exports/EmuKrnlPs.cpp
@@ -166,7 +166,7 @@ void PspSystemThreadStartup
 XBSYSAPI EXPORTNUM(254) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThread
 (
 	OUT PHANDLE         ThreadHandle,
-	OUT PHANDLE         ThreadId OPTIONAL,
+	OUT PDWORD          ThreadId OPTIONAL,
 	IN  PKSTART_ROUTINE StartRoutine,
 	IN  PVOID           StartContext,
 	IN  boolean_xt         DebuggerThread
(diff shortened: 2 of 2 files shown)


CookiePLMonster:
Side note: In Debug mode I still tend to crash on this race condition, maybe it needs more than 10ms sleep.

gandalfthewhite19890404:
Tested couple of games which requires all core hacks enabled - **GTA San Andreas** and **Lord of the Rings: Third Age**- they both are same.

LanHikariDS:
Star Wars: Republic Commando still crashes on startup, too

CookiePLMonster:
Replaced the previous PR with a new, better approach which removes hacks instead of adding additional ones. Please retest your games!

Luffykun007:
Far cry instincts still freezes at the same point for me with this PR.

Capturar4

PatrickvL (approved):
Looking good, let's await test results; No regressions expected, hopefully a few improvements!

medievil1:
Curse: the eye of isis V1.04 now boot and plays
image
image

literalmente-game:
Games tested with no changes:
- ATV Quad Power Racing 2				
- American McGee Presents Scrapland 		
- BMX XXX						
- Animaniacs - The Great Edgar Hunt		
- Backyard Wrestling				
- Beyond Good and Evil				
- Blitz The League				
- BloodyRoar Extreme				
- Breakdown					
- Brute Force					
- Call of Cthulhu - Dark corners of the Earth	
- Jet Set Radio Future				
- Lego Star Wars					
- Murakumo Renegade Mech Pursuit			
- Pitfall - The Lost Expedition			
- Chronicles of Riddick				
- F1 2002						
- Gun						
- Star Wars - Republic Commando			
- Destroy all Humans 2				
- Tak - The Great Juju Challenge			
- Serious Sam					
- Hunter The Reckoning: Redeemer			
- Stake - Fortune Fighters			
- I-Ninja						
- Reign Of Fire					
- The Thing					
- Sudeki						
- Dino Crisis 3					
- Fatal Frame					
- Land of the Dead - Road to Fiddler's Green
- Quantum Redshift				
- RalliSport Challenge				
- Rally Fusion - Race Of Champions		
- Return to Castle Wolfenstein			
- Sonic Heroes					
- Tork - Prehistoric Punk				
- Volvo - Drive for Life				
- Xyanide

LukeUsher:
After plenty of test results from @literalmente-game  (no regressions found so far), I think we're good to merge now!

RadWolfie:
Titles has been tested with no regression:
* Blood Wake
* Call of Duty: Finest Hour
* Constantine
* Crimson Skies
* Grand Theft Auto III
* Haunted Mansion
* Red Faction 2
* The Simpsons: Road Rage

PR #1999, opened 2020-10-18
Pixel combiner fixes

CookiePLMonster:
**Needs testing, especially with Rayman games since "Rayman" was mentioned in the code I modified!**

1. Fixes `Inserted to avoid LRP parameters referencing the output register` pass relying on a wrong register. It changed the destination register address, but not the type.
2. Implemented better Blue-to-Alpha handling by fixing the output mask (previously erroreously only updated the B channel) and inserting an additional `mov rX.a, rX.b` opcode.

Fixes GPU decoding of BINK videos - my test case was The Warriors, but it might fix more games using a pixel combiner to decode BINK.
image

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -5166,6 +5166,7 @@ bool  PSH_XBOX_SHADER::SimplifyLRP(PPSH_INTERMEDIATE_FORMAT Cur, int index)
           {
               insert = true;
               Cur->Parameters[p].Address = output;
+              Cur->Parameters[p].Type = PARAM_R;
           }
       }
       if (insert)
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -2089,17 +2089,15 @@ bool PSH_INTERMEDIATE_FORMAT::Decode(DWORD aCombinerStageNr, DWORD PSInputs, DWO
       // And if this is true, how do the blue-to-alpha flags behave if present on both AB and CD?
 
       // TODO : Rayman does this in some shaders, requires a fixup (as output.b is incorrect and not allowed)
+      // TODO: Above may not be valid anymore, needs testing
       Output[0].Modifiers = Output[0].Modifiers | (1 << ARGMOD_BLUE_REPLICATE);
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -4545,9 +4543,31 @@ bool PSH_XBOX_SHADER::FixInstructionModifiers()
...
+                Ins.Output[0].Address = Ins.Parameters[0].Address = output.Address;
+                Ins.Output[0].Mask = MASK_A;
+                Ins.Parameters[0].Mask = MASK_B;
+                Ins.CommentString = "; Inserted Blue-to-Alpha";
+                EmuLog(LOG_LEVEL::DEBUG, "; Inserted Blue-to-Alpha");
+                insert = true;
+            }
+
+            if (insert)
+            {
```
PatrickvL, line 4566: Here, the `insert` boolean doesn't need to be used, as leaving out these marked lines would lead to functionally identical code. (The `insert` variable used in the preceding switch/cases, is just used there to skip the `InsertIntermediate` call when the default statement block was hit. After that, the rest of the function still needed to continue, so `return` couldn't be used.)

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -4545,9 +4543,31 @@ bool PSH_XBOX_SHADER::FixInstructionModifiers()
...
             Cur->Modifier = INSMOD_NONE;
-            InsertIntermediate(&Ins, InsertPos);
+            InsertIntermediate(&Ins, InsertPos++);
             Result = true;
         }
+
+        // Handle blue-to-alpha which is technically an instruction modifier, but operates on arguments
+        for (int i = 0; i < PSH_OPCODE_DEFS[Cur->Opcode]._Out; i++) {
+            auto& output = Cur->Output[i];
+            insert = false;
```
PatrickvL, line 4553: This entire assignment line can be removed as well, if & when the suggestion below is applied.

PatrickvL (changes requested):
Great work! Just one suggestion for a small simplification.

Also, maybe mention required re-testing of Rayman in pull request description?

CookiePLMonster:
Made the changes and added a note about Rayman (no idea which it is about though).

PatrickvL (approved):

PR #2001, opened 2020-10-22
Pixel shader fix XFC SUM 

PatrickvL:
Fix how our current pixel shader conversion calculates the final combiner special purpose register 'sum' : it was accidentally multiplying instead of adding it's arguments!

Pixel shaders that use this 'sum' register will show more accurate output because of this change.

This is a low-risk, possibly high-impact change.

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -3925,8 +3925,8 @@ void PSH_XBOX_SHADER::ConvertXFCToNative(int i)
 
   if (NeedsSum)
   {
-    // Add a new opcode that calculates r0*v1 :
-    Ins.Initialize(PO_MUL);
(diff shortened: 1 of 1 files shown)


medievil1:
tested about 10 games so far....just quickly, ones that could have potential pixel shader issues, not seen any difference.....I suppose the ones I have just have other issues....

literalmente-game:
Some improvements found:
- Group S Challenge:
Before:
before
After: 
after
- Disney Extreme Skate Adventure:
Before:

After:

fuhuan416:
Wow！I can't wait to test Iron Phoenix.

LukeUsher (approved):
Good spot, thanks!

PatrickvL:
The over-saturation in Disney Extreme Skate Adventure could be because we don't yet clamp negative final combiner inputs to zero.

PR #2002, opened 2020-10-24
Simplfy pixel shader constant handling;

PatrickvL:
Since we've ported over to Direct3D 9, and we're using pixel shader version 1.4, we've got more than enough constants available to remove the need for constant packing.

Also, there was a left-over patch on SetPixelShaderConstant which must no longer be applied, since nowadays we read constant values straight from their corresponding render state slots.
This also implies we no longer need to declare the final combiner constants as part of the shader assembly, because these 2 are also read from their corresponding xbox render state slots, and thus can be transferred to host on each update.

This will likely improve the output of pixel shaders which stay otherwise unchanged but rely on changing constant values.


src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -6051,7 +5960,7 @@ void DxbxUpdateActivePixelShader() // NOPATCH
     // as these could have been updated via SetRenderState or otherwise :
     for (i = 0; i < PSH_XBOX_CONSTANT_MAX; i++)
 	{
-      if (RecompiledPixelShader->ConstInUse[i])
+      // Asume all constants are in use (this is much easier than tracking them for no other purpose than to skip a few here)
```
CookiePLMonster, line 5963: Nit: Typo

CookiePLMonster (approved):

LukeUsher (approved):

PR #2003, opened 2020-10-24
Fix APU timer ticking at wrong frequency

CookiePLMonster:
When looking for a different issue I spotted this abomination in the APU timer code:

```cpp
// Measure current host performance counter and frequency
QueryPerformanceCounter(&APUInitialPerformanceCounter);
NativeToXboxAPU_FactorForPerformanceFrequency = (double)APU_TIMER_FREQUENCY / APUInitialPerformanceCounter.QuadPart;
```

This is obviously wrong and probably pretty inaccurate due to the usage of floating point variables, so I replaced this code with an adaptation of RDTSC ticking code I wrote for #1985

No known improvements, but might improve the accuracy of whatever is relying on the APU timer for some sound timings.

### DIFF
--- a/src/common/Timer.cpp
+++ b/src/common/Timer.cpp
@@ -170,3 +170,37 @@ void Timer_Init()
 #error "Unsupported OS"
 #endif
 }
+
+// ******************************************************************
+
+void ScaledPerformanceCounter::Reset(uint32_t frequency)
+{
+	std::lock_guard<std::mutex> lock(m_mutex);
+
+	m_frequencyFactor = Muldiv64(HostClockFrequency, SCALE_S_IN_NS, frequency);
+
+	LARGE_INTEGER tsc;
+	QueryPerformanceCounter(&tsc);
+	m_lastQPC = tsc.QuadPart;
+
+	m_currentCount = 0;
+	m_currentRemainder = 0;
+}
+
+uint64_t ScaledPerformanceCounter::Tick()
+{
+	std::lock_guard<std::mutex> lock(m_mutex);
+
+	LARGE_INTEGER qpc;
+	QueryPerformanceCounter(&qpc);
+
+	int64_t lastQpc = std::exchange(m_lastQPC, qpc.QuadPart);
+	qpc.QuadPart -= lastQpc;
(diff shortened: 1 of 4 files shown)


PatrickvL (commented):
What if you'd move all implementations into `CxbxGetPerformanceCounter`, and select the required factors using an enum argument? That way there's just one place where accurate timer adjustments & drift handling takes place...

CookiePLMonster:
> What if you'd move all implementations into `CxbxGetPerformanceCounter`, and select the required factors using an enum argument? That way there's just one place where accurate timer adjustments & drift handling takes place...

That actually was my original plan, but I encountered issues with including `EmuKrnlKe.h` from other compilation units! If you think this is the best way forward I can revisit this idea and try to get it right this time.

PatrickvL:
> > What if you'd move all implementations into `CxbxGetPerformanceCounter`, and select the required factors using an enum argument? That way there's just one place where accurate timer adjustments & drift handling takes place...
> 
> That actually was my original plan, but I encountered issues with including `EmuKrnlKe.h` from other compilation units! If you think this is the best way forward I can revisit this idea and try to get it right this time.

Is it possible to move these functions to `Timer.h/cpp`?

RadWolfie:
I agree with @PatrickvL. Although, if header file is going to be in C++. I rather use `.hpp` than `.h`. Since imho, `.h` is intended for external libraries provide support with C/C++.

CookiePLMonster:
How about a "compromise" in form of factorizing these calculations and state to a `HighPrecisionTimer` class? This way it'd be instantiated in respective files like

`HighPrecisionTimer APUTimer(APU_TIMER_FREQUENCY);`

and "ticked" on `Get()` accordingly. This would also have an added benefit of reduced code duplication inside `CxbxGetPerformanceCounter`.

PatrickvL:
Amaze us!

CookiePLMonster:
Done! I made it a separate commit so it's clear that it's only a refactor and that a "fix APU" commit doesn't touch TSC/ACPI timers.

PatrickvL (approved):
Beautifully done!

PR #2004, opened 2020-10-24
Dsstream pause fixes

CookiePLMonster:
This PR fixes a bug potentially introduced by #1853. It changes two things:

1. Partially implements `X_DSSPAUSE_PAUSENOACTIVATE` because it's identical to `X_DSSPAUSE_PAUSE` in the case where the stream is not playing yet. The other case is still unimplemented.
2. Marks flushed sound streams as paused, as flushing is intended to stop the stream. Removing the pause flag made games keep pushing data packets to the stream after flushing, resulting in them not stopping.

**Needs testing with Obscure,  since fixes from #1853 targetted it!**
**Needs testing with Manhunt - it reportedly had the same bug as The Warriors (shared codebase) but I did not verify that.**
**Needs testing with games which would not pause/stop sounds and/or music as intended.**

Fixes cutscene audio not stopping when skipping cutscenes in The Warriors.
Fixes EA Trax playing in the background when launching Burnout 3. Main menu theme loops after a few seconds, but this was also the case before this PR and is an unrelated bug.

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
@@ -372,7 +372,7 @@ bool DSStream_Packet_Flush(
     }
     // Clear flags and set status to zero.
     DSStream_Packet_FlushEx_Reset(pThis);
-    pThis->EmuFlags &= ~DSE_FLAG_PAUSE;
+    pThis->EmuFlags |= DSE_FLAG_PAUSE;
     pThis->Xb_Status = 0;
     return false;
 }
--- a/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
@@ -854,10 +854,10 @@ static inline HRESULT HybridDirectSoundBuffer_Pause(
          // TODO: NOTE: If stream is playing, it perform same behavior as pause flag. If it is not played, it act as a queue until trigger to play it.
         case X_DSSPAUSE_PAUSENOACTIVATE:
             if ((dwEmuFlags & DSE_FLAG_PAUSE) != 0) {
-                // TODO: Start queueing
+                // TODO: Start queueing, might need the use of a DSE_FLAG_IS_ACTIVATED flag
                 break;
             }
-            // Intentional fallthrough
+            [[fallthrough]];
         case X_DSSPAUSE_PAUSE:
             pDSBuffer->Stop();
             DSoundBufferSynchPlaybackFlagRemove(dwEmuFlags);
--- a/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
@@ -851,6 +851,13 @@ static inline HRESULT HybridDirectSoundBuffer_Pause(
             dwEmuFlags &= ~DSE_FLAG_PAUSE;
             Xb_rtTimeStamp = 0;
             break;
+         // TODO: NOTE: If stream is playing, it perform same behavior as pause flag. If it is not played, it act as a queue until trigger to play it.
+        case X_DSSPAUSE_PAUSENOACTIVATE:
+            if ((dwEmuFlags & DSE_FLAG_PAUSE) != 0) {
+                // TODO: Start queueing
+                break;
+            }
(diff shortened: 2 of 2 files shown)


src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
```diff
@@ -372,7 +372,7 @@ bool DSStream_Packet_Flush(
     }
     // Clear flags and set status to zero.
     DSStream_Packet_FlushEx_Reset(pThis);
-    pThis->EmuFlags &= ~DSE_FLAG_PAUSE;
+    pThis->EmuFlags |= DSE_FLAG_PAUSE;
```
RadWolfie, line 375: If I recall it right, Flush caller never did add pause flag. And it will affect how samples read from GetStatus' result. What really stop backend buffer process is `Host_isProcessing` as long as host's buffer made a full stop first. I'll try make further investigation in our current codebase.

In summary, Flush call basically make halt audio and remove all of packets. Since the next packet will trigger to resume. It is something I had verified on xbox hardware with the sample.

src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
```diff
@@ -851,6 +851,13 @@ static inline HRESULT HybridDirectSoundBuffer_Pause(
             dwEmuFlags &= ~DSE_FLAG_PAUSE;
             Xb_rtTimeStamp = 0;
             break;
+         // TODO: NOTE: If stream is playing, it perform same behavior as pause flag. If it is not played, it act as a queue until trigger to play it.
+        case X_DSSPAUSE_PAUSENOACTIVATE:
+            if ((dwEmuFlags & DSE_FLAG_PAUSE) != 0) {
+                // TODO: Start queueing
+                break;
+            }
+            // Intentional fallthrough
```
RadWolfie, line 860: In C++17 feature, it is `[[FALL_THROUGH]];`.

src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
```diff
@@ -851,6 +851,13 @@ static inline HRESULT HybridDirectSoundBuffer_Pause(
             dwEmuFlags &= ~DSE_FLAG_PAUSE;
             Xb_rtTimeStamp = 0;
             break;
+         // TODO: NOTE: If stream is playing, it perform same behavior as pause flag. If it is not played, it act as a queue until trigger to play it.
+        case X_DSSPAUSE_PAUSENOACTIVATE:
+            if ((dwEmuFlags & DSE_FLAG_PAUSE) != 0) {
+                // TODO: Start queueing
```
RadWolfie, line 857: TODO: Likely need interaction with `DSE_FLAG_IS_ACTIVATED` check.

RadWolfie (commented):
In my opinion, it needs more thorough investigation such as should internal Flush clear pause flag or not? I'll investigate on origin of cause since `X_DSSPAUSE_PAUSENOACTIVATE` wasn't implement.

src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
CookiePLMonster: That is correct. However, after calling Flush, GetStatus is **not** supported to report "playing" status because the stream had stopped. Because we don't differentiate "paused" and "stopped", I figured this is the best way to handle it.

src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
CookiePLMonster: I am aware of it, but I forgot VS2017 supports it too, thanks.

literalmente-game:
Obscure tested against latest master (https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/97bf1d91696f254911b8697fb4c4b7bd96130986). Music on the menus and themes in-game (combat music or themes of any kind) don't play with this PR. Sounds are fine

RadWolfie:
Currently convert to draft pull request since there are stages in need to go through fixing. There are some difference between before 4134 and later. Plus is a chance of SetFormat function doesn't reset pause flag which is calling our internal Flush function.

CookiePLMonster:
@RadWolfie Given that your latest changes to `XbSymbolDatabase` have been merged, what is the way forward with this PR?

PR #2005, opened 2020-10-24
Improve reverse screenspace transformation precision

CookiePLMonster:
**This PR needs testing with games which:**
* **Have Z fighting anywhere**
* **Have interactive menus which don't show**

When a D24 depth buffer was used, its viewport Z scale was `16777215.0f`. This number is high enough to cause precision issues in the reverse screenspace transformation calculation for some elements. In my case, it shifted the Z to slightly above 1.0f, causing it to get clipped.

Basing on some tests, pre-dividing the divisor increases the numerical stability of those calculations and allows to retain better precision. **WARN:** pre-multiplying instead of pre-dividing might seem like a better idea, but I found it did **not** resolve the precision issues!

Also adds an "unrelated" change of better batching for viewport scale/offset constant submission calls.

Makes Burnout 3 display menus properly:
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
@@ -261,9 +261,13 @@ float4 reverseScreenspaceTransform(float4 oPos)
 	// mad oPos.xyz, r12, r1.x, c-37
 	// where c-37 and c-38 are reserved transform values
 
+	// oPos.w and xboxViewportScale.z might be VERY big when a D24 depth buffer is used
+	// and multiplying oPos.xyz by oPos.w may cause precision issues.
+	// Pre-divide them to help keep the values reasonably small.
+	// Test case: Burnout 3
+	float3 divisor = xboxViewportScale.xyz / oPos.w;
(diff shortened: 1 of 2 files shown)


src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -261,9 +261,9 @@ float4 reverseScreenspaceTransform(float4 oPos)
 	// mad oPos.xyz, r12, r1.x, c-37
 	// where c-37 and c-38 are reserved transform values
 
+	float3 divisor = xboxViewportScale.xyz / oPos.w;
```
PatrickvL, line 264: Please add a comment to explain why it's done this way, otherwise a future maintainer might accidentally 'optimize' the variable away.

PatrickvL (changes requested):
Nice work, as this avoids changing any value, and addresses just the precision issue!

Just one request, for some more comments....

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
CookiePLMonster: Done, also noted the test case.

PatrickvL (approved):

PR #2008, opened 2020-10-25
Revert "hle: Disable RaiseException patch, let the kernel handle it."

LukeUsher:
This reverts commit de92e5048b2967ea1e6ca16d02f799f51c9f6f0d.

With this, Auto Modelista works again.


LukeUsher:
My bad, I was wrong, this used to fix Auto Modelesta in earlier testing several weeks back, but it seems that it doesn't anymore.
It's likely the only real fix is to properly implement Exception support.

PR #2007, opened 2020-10-25
Changes to D3D patches

CookiePLMonster:
* Patches `D3D_BlockOnTime_4` just like its non-LTCG equivalent.
* Unpatches `D3DDevice_GetTransform` and trampolines `SetTransform`/`MultiplyTransform` to the guest code. This fixes cases where games tried to use a potentially unpatched or inlined `GetTransform` and/or read the D3D state directly.

Fixes Burnout 3 freezing past the car selection screen and makes it render geometry properly. Currently this is not yet visible because of an unhandled "binding surface as a texture" case, but rendering now **does** happen.

image

PR #2011, opened 2020-10-26
Guard against nested SetTransform/MultiplyTransform calls

CookiePLMonster:
`MultiplyTransform` can be implemented in means of doing the multiplication internally and `SetTransform` which corrupted the host's internal state (host `SetTransform` called with a multiplied state, followed by a host `MultiplyTransform` resulting in a double multiplication).

This PR introduces a guard variable to ensure we call to host only once per the patch chain and keep the internal state pristine. I facilitated this by introducing a `NestedPatchCounter` class which acts as a scoped counter - when used on a variable shared between potentially conflicting/nested patches, this class allows functions to skip parts of the host set-up if it is either redundant or harmful (like in this case).

Fixes a regression in 25 to Life introduced by #2007

image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -6459,40 +6459,66 @@ void CxbxImpl_SetTransform
         LOG_FUNC_ARG(pMatrix)
         LOG_FUNC_END;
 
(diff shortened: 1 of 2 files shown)


LukeUsher (approved):

romaxy151:
I don't know about you, but I have 25 to Life issues an error after the last commits.
Снимок экрана 2020-10-27 094708

PR #2012, opened 2020-10-26
Interlocked lockcounts

CookiePLMonster:
Makes `LockCount` for RW locks and critical sections atomic, as previously those two primitives were (ironically) thread unsafe.

No known change in behaviour, but on paper it might improve the stability with the All cores hack.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlEx.cpp
+++ b/src/core/kernel/exports/EmuKrnlEx.cpp
@@ -148,8 +148,7 @@ XBSYSAPI EXPORTNUM(12) xbox::void_xt NTAPI xbox::ExAcquireReadWriteLockExclusive
 	LOG_FUNC_ONE_ARG(ReadWriteLock);
 
 	bool interrupt_mode = DisableInterrupts();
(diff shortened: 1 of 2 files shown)


PatrickvL (approved):
Looks okay. Not sure it's actually needed, but since it probably won't add much overhead, and it may help stabilize the multi-core hack somewhat, let just merge this. Thanks.

CookiePLMonster:
Regarding possible improvements for the multi-core hack, I'm sure we are on the same page and it's not even up for debate.

I believe this will benefit single-core use cases too though, because we don't control context switches. Most notably, `x++; if (x == 0)` is a series of separate operations:
* read x
* increment
* write to x
* read x
* compare against 1

Even with a single core, context switch can happen between **any** of those operations (even if it's unlikely), breaking the integrity of the reference counter. An interlocked operation ensures atomicity of this entire chain of commands, without the risk of the context switch breaking anything.

I agree it's very much an edge case, but I **have** seen like this break in production.

PatrickvL:
I fully agree

PR #2014, opened 2020-10-28
HLSL fixed function implementation

NZJenkins:
Should behave more-or-less the same as D3D9 fixed function mode
But allow us to handle Xbox extensions and idiosyncrasies better e.g. backface lighting

F2 toggles back to D3D9 fixed function rendering

Co-authored with @PatrickvL

Opening as draft in case anyone feels like looking for regressions. Everything should generally look exactly the same


projects/cxbx/CMakeLists.txt
```diff
@@ -189,3 +189,18 @@ if(${CMAKE_GENERATOR} MATCHES "Visual Studio ([^9]|[9][0-9])")
...
+set_source_files_properties(
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl"
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl" 
+  PROPERTIES VS_TOOL_OVERRIDE "None"
+)
+
+# Copy shader files to hlsl directory
+add_custom_target(copy_hlsl ALL
+ COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:cxbx>/hlsl
+ COMMAND ${CMAKE_COMMAND} -E copy "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShaderState.cpp" $<TARGET_FILE_DIR:cxbx>/hlsl
```
PatrickvL, line 204: maybe change extension into `.hlsli` ?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6314,6 +6322,142 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
...
+		UpdateFixedFunctionShaderLight(d3d8LightState.EnabledLights[i], &ffShaderState.Lights[i], &LightAmbient, rowMajorViewTransform);
+	}
+
+	ffShaderState.AmbientPlusLightAmbient = Ambient + LightAmbient;
+	ffShaderState.BackAmbientPlusLightAmbient = BackAmbient + LightAmbient;
+
+	// Write fixed function state to shader constants
+	const int slotSize = 16;
+	const int fixedFunctionStateSize = (sizeof(FixedFunctionVertexShaderState) + slotSize - 1) / slotSize;
+	auto hRet = g_pD3DDevice->SetVertexShaderConstantF(0, (float*)&ffShaderState, fixedFunctionStateSize);
```
PatrickvL, line 6458: Log when `!SUCCEEDED(hRet)`?

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,600 @@
...
+#ifdef CXBX_ALL_TEXCOORD_INPUTS
+	return xIn.v[index];
+#else
+    // switch statements inexplicably don't work here
+    if(index == position) return xIn.pos;
+    if(index == weight) return xIn.bw;
+    if(index == normal) return xIn.normal;
+    if(index == diffuse) return xIn.color[0];
+    if(index == specular) return xIn.color[1];
+    //if(i == fogCoord) xIn.fog = value;
```
PatrickvL, line 60: Although commented out (why?), best `return` a value

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,600 @@
...
+    // Specular
+    if (state.Modes.SpecularEnable)
+    {
+        // Blinn-Phong
+        // https://learnopengl.com/Advanced-Lighting/Advanced-Lighting
+        float3 halfway = normalize(toViewerVN + toLightVN);
+        float NdotH = dot(View.Normal, halfway);
+
+        float3 frontSpecular = pow(abs(NdotH), powers[0]) * lightSpecular.rgb;
+        float3 backSpecular = pow(abs(NdotH), powers[1]) * lightSpecular.rgb;
```
PatrickvL, line 126: All `powers` variables passed along are used only here, for specular light calculation. Therefore, best rename it into `specular_powers` everywhere.

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,600 @@
...
+        // Generate texture coordinates
+		float3 reflected = reflect(normalize(View.Position.xyz), View.Normal);
+        
+		if (tState.TexCoordIndexGen == TCI_CAMERASPACENORMAL)
+			texCoord = float4(View.Normal, 1);
+		else if (tState.TexCoordIndexGen == TCI_CAMERASPACEPOSITION)
+			texCoord = View.Position;
+		else if (tState.TexCoordIndexGen == TCI_CAMERASPACEREFLECTIONVECTOR)
+			texCoord.xyz = reflected;
+        // else if TCI_OBJECT TODO is this just model position?
```
PatrickvL, line 428: How could unhandled cases like this be detected? Make it somehow visually stand out? But how?

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,600 @@
...
+            xIn.v[i] = value;
+        #else
+	        // switch statements inexplicably don't work here
+	        if(i == position) xIn.pos = value;
+	        if(i == weight) xIn.bw = value;
+	        if(i == normal) xIn.normal = value;
+	        if(i == diffuse) xIn.color[0] = value;
+	        if(i == specular) xIn.color[1] = value;
+	        //if(i == fogCoord) xIn.fog = value;
+	        //if(i == pointSize) xIn.psize = value;
```
PatrickvL, line 494: Why are these commented out, again?

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,600 @@
...
+    // Initialize input registers from the vertex buffer data
+    // Or use the register's default value (which can be changed by the title)
+    for (uint i = 0; i < 16; i++) {
+        float4 value = lerp(Get(xInput, i), vRegisterDefaultValues[i], vRegisterDefaultFlags[i]);
+        #ifdef CXBX_ALL_TEXCOORD_INPUTS
+            xIn.v[i] = value;
+        #else
+	        // switch statements inexplicably don't work here
+	        if(i == position) xIn.pos = value;
+	        if(i == weight) xIn.bw = value;
```
PatrickvL, line 489: Are the else statements left out for efficiency? Perhaps explain in a comment.

src/core/hle/D3D8/FixedFunctionState.cpp
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            EmuLog(LOG_LEVEL::INFO, "Attempted to disable light %d but it was not enabled", index);
+            return;
+        }
+    }
+
+    // Replace the oldest element and move to end
+    EnabledLights[0] = index;
+    std::rotate(std::begin(EnabledLights), std::begin(EnabledLights) + 1, std::end(EnabledLights));
```
PatrickvL, line 82: Why the shuffling around of lights? If it's a requirement so that the fixed function shader can access just the enabled lights in a linear fashion, then that rationally should probably be explained in a comment somewhere.

src/core/hle/D3D8/XbVertexShader.h
```diff
@@ -107,6 +107,18 @@ extern D3DVERTEXELEMENT *EmuRecompileVshDeclaration
     CxbxVertexDeclaration *pCxbxVertexDeclaration
 );
 
+enum class VertexShaderMode {
+	FixedFunction,
+	// TODO implement the FixedFunction vs. Passthrough distinction
+	// When titles use Xbox fixed function with pre-transformed vertices
+	// it actually uses a special "passthrough" shader program
+	// Passthrough,
+	ShaderProgram
```
PatrickvL, line 116: To avoid symbol conflicts, it may be good to prefix these with lowercase `vsm` or something like that.

PatrickvL (changes requested):
A few questions / remarks, but VERY COOL to have this!!!

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShaderState.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#define float4x4 D3DMATRIX
+#define float4 D3DXVECTOR4
+#define float3 D3DVECTOR
+#define float2 D3DXVECTOR2
+#define arr(name, type, length) std::array<type, length> name
+
+#else
+// HLSL
+#define arr(name, type, length) type name[length]
```
PatrickvL, line 18: Somewhat silly, but when these two defines are swapped, defining `arr` last, it would mirror prettier with the undef's at the end.

gandalfthewhite19890404:
**Regression** - LEGO Star Wars 2 The original trilogy does not boot, hits test-case popup and crashes
LOG_TEST_CASE: Couldn't assemble vertex shader
In CompileHlsl (D:\a\Cxbx-Reloaded\Cxbx-Reloaded\src\core\hle\D3D8\Direct3D9\VertexShader.cpp line 266)

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,600 @@
...
+
+    // We always send four coordinates
+    // If we are supposed to send less than four
+    // then copy the last coordinate to the remaining coordinates
+    // For D3DTTFF_PROJECTED, the value of the *last* coordinate is important
+    // Test case: ProjectedTexture sample, which uses 3 coordinates
+	if (countFlag == 1)
+		texCoord.yzw = texCoord.x;
+	else if (countFlag == 2)
+		texCoord.zw = texCoord.y;
```
PatrickvL, line 452: Should the source be `.yy`?

projects/cxbx/CMakeLists.txt
```diff
@@ -189,3 +189,18 @@ if(${CMAKE_GENERATOR} MATCHES "Visual Studio ([^9]|[9][0-9])")
...
+# HLSL compilation
+# Don't build HLSL files, though include them in Visual Studio
+set_source_files_properties(
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl"
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl" 
+  PROPERTIES VS_TOOL_OVERRIDE "None"
+)
+
+# Copy shader files to hlsl directory
+add_custom_target(copy_hlsl ALL
```
RadWolfie, line 202: Redundant code, see pull request #1932 for better solution using `file(COPY ...)` command. (1)

projects/cxbxr-emu/CMakeLists.txt
```diff
@@ -162,3 +162,18 @@ target_link_libraries(cxbxr-emu
...
+# HLSL compilation
+# Don't build HLSL files, though include them in Visual Studio
+set_source_files_properties(
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl"
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl" 
+  PROPERTIES VS_TOOL_OVERRIDE "None"
+)
+
+# Copy shader files to hlsl directory
+add_custom_target(copy_hlsl-emu ALL
```
RadWolfie, line 175: Redundant code, see pull request #1932 for better solution using `file(COPY ...)` command. (2)

projects/cxbx/CMakeLists.txt
```diff
@@ -189,3 +189,18 @@ if(${CMAKE_GENERATOR} MATCHES "Visual Studio ([^9]|[9][0-9])")
 endif()
 
 add_dependencies(cxbx cxbxr-ldr cxbxr-emu misc-batch)
+
+# HLSL compilation
+# Don't build HLSL files, though include them in Visual Studio
+set_source_files_properties(
```
RadWolfie, line 195: IMHO, create a list in root CMakeLists.txt and refer it here will remove unnecessary redundant list. (1)

projects/cxbxr-emu/CMakeLists.txt
```diff
@@ -162,3 +162,18 @@ target_link_libraries(cxbxr-emu
 )
 
 add_dependencies(cxbxr-emu cxbxr-ldr misc-batch)
+
+# HLSL compilation
+# Don't build HLSL files, though include them in Visual Studio
+set_source_files_properties(
```
RadWolfie, line 168: IMHO, create a list in root CMakeLists.txt and refer it here will remove unnecessary redundant list. (2)

projects/cxbx/CMakeLists.txt
```diff
@@ -189,3 +189,18 @@ if(${CMAKE_GENERATOR} MATCHES "Visual Studio ([^9]|[9][0-9])")
...
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl"
+  "${CXBXR_ROOT_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl" 
+  PROPERTIES VS_TOOL_OVERRIDE "None"
```
RadWolfie, line 198: Had you tried with `HEADER_FILE_ONLY ON` instead of `VS_TOOL_OVERRIDE "None"`? If it works, then we could apply for the same with NV2A source files.

RadWolfie (changes requested):

projects/cxbx/CMakeLists.txt
NZJenkins: makes sense, will do

projects/cxbx/CMakeLists.txt
NZJenkins: Yes good idea

projects/cxbx/CMakeLists.txt
NZJenkins: Yes `HEADER_FILE_ONLY` used to work, but VS seems to have started building it regardless...
Followed this answer suggesting `VS_TOOL_OVERRIDE` https://stackoverflow.com/a/46740948

projects/cxbx/CMakeLists.txt
NZJenkins: I assume you mean I can both deduplicate the hlsl copy code, and use `file(COPY ...)` with `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}`?
Will look into it

src/core/hle/D3D8/FixedFunctionState.cpp
NZJenkins: Does the comment in the header file help? I should add some explanation here too
We are holding onto the 8 most recently enabled lights. So if a title activates 9 lights we can disable the oldest one (at index 0)
IIRC that's what D3D9 does anyway, not sure of a test case though.

Oops, I think the enable case rotates the wrong way!

src/core/hle/D3D8/XbVertexShader.h
NZJenkins: `VertexShaderMode::` prefix is required because of `enum class` syntax

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: vaguely remember Xbox FVF didn't take FOG or PSIZE (but host D3D9 FVF supports PSIZE?)
will check against the vertex pr and see what we were doing

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: True - wonder if it's worth bundling `toViewerVN`, `powers`, and `lightSpecular` into a `SpecularInfo` or something
I think they're all specular specific

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,600 @@
...
+		if (tState.TexCoordIndexGen == TCI_CAMERASPACENORMAL)
+			texCoord = float4(View.Normal, 1);
+		else if (tState.TexCoordIndexGen == TCI_CAMERASPACEPOSITION)
+			texCoord = View.Position;
+		else if (tState.TexCoordIndexGen == TCI_CAMERASPACEREFLECTIONVECTOR)
+			texCoord.xyz = reflected;
+        // else if TCI_OBJECT TODO is this just model position?
+		else if (tState.TexCoordIndexGen == TCI_SPHERE)
+		{
+            // TODO verify
```
NZJenkins, line 431: Pretty sure this is verified... list test case/s and remove TODO

gandalfthewhite19890404:
Artifacts does not includes hlsl folder, it is causing crashes with every titles, already discussed and verified on Discord.

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: In this particular case it should be apparent since it's texture coordinates

projects/cxbx/CMakeLists.txt
RadWolfie: Correct, see `projects/misc/batch.cmake` file from that pull request or in master branch. Basically your files you need to copy into build folder should be done inside batch.cmake file. You may not remember our previous discussion for this type of solution.

RadWolfie:
@NZJenkins Some time in the future, I will have "install" functionality in CMake scripts into much generic usage. Since I have discover a way to do it with XbSymbolDatabase repo's CI action. Then every contiguous integration service doesn't need another update, **even per platform doesn't need extra work as well**. In fact, I should create a ticket for this feature.

PatrickvL:
This is why I was trying to avoid separate HLSL files

projects/cxbx/CMakeLists.txt
NZJenkins: I hit the issue again after switching between branches - it might be a Visual Studio + CMake issue.
Will restore `HEADER_FILE_ONLY` and see if I can get it to build again.

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: but that's extra typing!
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-per-component-math
> f_4D.xzyw = pos.w; // write one component to more than one component

src/core/hle/D3D8/FixedFunctionState.cpp
PatrickvL: The comment in the header states what it does, but not why it's required to be that way

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,616 @@
...
+    if(index == texcoord0) return xIn.texcoord[0];
+    if(index == texcoord1) return xIn.texcoord[1];
+    if(index == texcoord2) return xIn.texcoord[2];
+    if(index == texcoord3) return xIn.texcoord[3];
+    return 1;
+#endif
+}
+
+// Output registers
+struct VS_OUTPUT
```
PatrickvL, line 73: For pixel shader HLSL, this exact output structure would need to be available as input, could this be moved into it's own file so it can be shared? Or shall that be postponed until the need arises?

Same for generic vertex shader HLSL btw...

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
```diff
@@ -0,0 +1,616 @@
...
+uniform float4 vRegisterDefaultValues[16] : register(c192);
+uniform float4 vRegisterDefaultFlagsPacked[4] : register(c208);
+static  bool  vRegisterDefaultFlags[16];
+
+uniform FixedFunctionVertexShaderState state : register(c0);
+
+#undef CXBX_ALL_TEXCOORD_INPUTS // Enable this to disable semantics in VS_INPUT (instead, we'll use an array of generic TEXCOORD's)
+
+// Input registers
+struct VS_INPUT
```
PatrickvL, line 13: Generic vertex shader HLSL probably overlaps with this. Could it be refactored a bit to avoid duplication?

gandalfthewhite19890404:
What do you think about function rendering mode GUI indication?

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: Hit an issue today where `if` and `if else` can be the difference between the shader compiling and `X4505: maximum temp register index exceeded` (in unrelated parts of the shader!)
So maybe the compiler doesn't like `else`

Anyway, nothing to explain here, other than `switch` didn't seem to work at the time
You would think, branching on a single variable, there would be no difference between `if`, `if else`, and `switch` 🤔

medievil1:
azurik(MS-007 V1.01) not happy with the current  state...lol 
image

medievil1:
azurik is better in game, start menu has the offset center picture though (issue popped up before) 
image

medievil1:
and ummmm....lol
image

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1788,40 +1789,29 @@ void CxbxImpl_DeleteVertexShader(DWORD Handle)
...
+	// Write Vertex Shader constants in nv2a
+	extern float* HLE_get_NV2A_vertex_constant_float4_ptr(unsigned const_index); // TMP glue
+	float* constant_floats = HLE_get_NV2A_vertex_constant_float4_ptr(Register);
+	memcpy(constant_floats, pConstantData, ConstantCount * sizeof(float) * 4);
+
+	// Mark the constant as dirty, so that CxbxUpdateHostVertexShaderConstants will pick it up
+	extern NV2ADevice* g_NV2A; // TMP glue
+	auto nv2a = g_NV2A->GetDeviceState();
+	for (int i = 0; i < ConstantCount; i++) {
+		nv2a->pgraph.vsh_constants_dirty[Register + i] = true;
```
PatrickvL, line 1814: `CxbxUpdateHostVertexShaderConstants` doesn't check dirty anymore, so this could be removed or disabled..

NZJenkins: yep left the dirty tracking in even though we're not reading it - wasn't worried about that. But now the comment is misleading so should probably fix one or the other

medievil1:
azurik decided it all of a sudden in mid step wants to go black screen now.... 
image

medievil1:
KOTOR crashes after skipping video(immediately crashes) with two fatal errors 
`[0x2084] MAIN: Received Exception (Code := 0xC0000005)

 EIP := 0x658F82DA
 EFL := 0x00010206
 EAX := 0xE5AC3404 EBX := 0x00000000 ECX := 0x65A4AB20 EDX := 0x65ADCFE4
 ESI := 0x65AA983C EDI := 0x00000001 ESP := 0x0D73E5AC EBP := 0x0D73E5B0
 CR2 := 0x00000000

[0x2084] FATAL: X86     Popup : Received Exception Code 0xC0000005 @ EIP := 0x658F82DA

  Press "OK" to terminate emulation.
  Press "Cancel" to debug.
[0x31C8] MAIN: Received Exception (Code := 0xC0000005)
[0x3A94] MAIN: Received Exception (Code := 0xC0000005)

 EIP := 0x701B7BC3
 EFL := 0x00010297
 EAX := 0x3F800000 EBX := 0x65AC40A8 ECX := 0xCEC655EF EDX := 0x00000000
 ESI := 0x00000000 EDI := 0x701B7EB0 ESP := 0x1684FEB8 EBP := 0x1684FEDC
 CR2 := 0x00000000

[0x3A94] FATAL: X86     Popup : Received Exception Code 0xC0000005 @ EIP := 0x701B7BC3

  Press "OK" to terminate emulation.
  Press "Cancel" to debug.
`

medievil1:
> 
> 
> @medievil1 I saw the Azurik issue happen, I suspect it's a random pre-existing issue. Did you try it more than once?
> Fixed the KOTOR issue and cleaned up the fix commits.
> Still some tricky regressions remaining

yup, just tried again even on the latest and yea, it still goes black 
image

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
PatrickvL: Okay, so then it should probably read : `assert(i != fogCoord); // fogCoord is never part of any Xbox FVF` and similar for `pointSize`. (Not sure how to implement `assert`, but even having it stubbed would still improve clarity.)

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: Added log test cases when the state is set. Previously we logged

jenci8888:
Regression intro screen:
image

I've checked against to master build (after 1894 PR)
image

projects/cxbx/CMakeLists.txt
NZJenkins: have redone the cmake part of the PR which should resolve this

projects/cxbx/CMakeLists.txt
NZJenkins: using HEADER_FILE_ONLY again

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: just removed this since its not part of the inputs.
the reason being its not used in DoFog - since the fog value is stored in the specular alpha

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: lighting calculations including DoSpecular were consolidated into one method, so variables aren't needlessly passed round any more

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl
NZJenkins: unused and not part of VS_INPUT

src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShaderState.cpp
NZJenkins: iirc there was one `#undef` for both, which prevented mirroring - its now duplicated so both def and undef sections mirror

RadWolfie:
Waiting on pull request #2076 to be merge first. Then changes for include files will need to be revert and shift hlsl files to install command via `install(FILES ...)` usage.

medievil1:
2076 has been merged :P looking forward to this getting merged too

CMakeLists.txt
```diff
@@ -407,6 +412,12 @@ install(FILES ${cxbxr_INSTALL_files}
   DESTINATION bin
 )
 
+install(FILES
+	"${CMAKE_SOURCE_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShaderState.hlsli"
```
RadWolfie, line 416: No tabs anywhere in CMake relative files please. Use single space as indent.

RadWolfie (dismissed):

RadWolfie:
LTGM for CMake relative changes. 👍

PatrickvL (approved):
Nzjenkins ftw!

LukeUsher (approved):

PR #1894, opened 2020-05-08
Vertex declaration refactoring

PatrickvL:
This pull request unpatches a few D3D API's and implements a few others, so that vertex declarations are now more reliable (all paths now go through the Xbox vertex attribute format structure).

Vertex declaration caching is also improved, possibly bringing a little performance benefit.

Memory leaks might occur - we're still pondering how to address those.

Once merged, this work allows fixed function HLSL to get finished up, as well, so it's rather important to get this in.

Please report the new test-case on this issue (mentioning "SetVertexShaderInput"), or any related others; Once we collected enough, the popups will be removed before merge.

---
2020-11-06 Status : Replaced by a squashed version of the same work, rebased onto latest master. ( The original commits are preserved in : https://github.com/PatrickvL/Cxbx-Reloaded/commits/vertex_declaration_refactoring_backup for future reference and/or potential cherry-picking )

2020-09-04 Status : Finalizing, summarizing regressions, fixing final issues. (Was : work in progress)

Previously reported regressions (checked when verified to be fixed now) :
* :heavy_check_mark: GTA 3 : ~~Vertex explosions~~
* :heavy_check_mark: GTA Vice City
* :heavy_check_mark: Indiana Jones : ~~Just the character is exploding (fine until Indy appeared)~~
* :grey_question: Sphinx and the Cursed Mummy : LESS broken (before exploded vertex, now correct) but rainbow textures
* :x: Worms 3D SE : vertices are exploded, shows a "Failed to create vertex buffer" fatal error after a few seconds
* :heavy_check_mark: Buffy : ~~just Buffy's head is exploded~~
* :heavy_check_mark: Sonic heroes : huge vertex explosion
* :heavy_check_mark: SpongeBob SP: BOBB game
* :grey_question: Star Wars Obi wan : comes up with an error, that was not present in 0f853f3
* :grey_question: Ty the tazmanin tiger 1, 2 and 3 : rendering errors
* :heavy_check_mark: Indiana Jones :  ~~black screen from the last video for all of the menu section until in game~~
* :grey_question: Crazy Taxi 3 : flickers graphics on the character select screen
* :grey_question: TimeSplitters: Future Perfect : vertex explosions everywhere, screen constantly flickers (though this PR is the first time the game is actually rendering anything, so it could be unrelated.)
* :grey_question: Steel Battalion: Line of Contact : no longer reaching menu (when it did in this PR previously)
* :grey_question: Battle for Bikini Bottom : not rendering the titles screen or some textures
[...]

PR #2022, opened 2020-10-30
Miscellaneous Direct3D LTCG fixes

CookiePLMonster:
**This PR needs testing with LTCG games, mainly those having a `Direct3D_CreateDevice_16` symbol!**

Marking as draft for now because this PR depends on https://github.com/Cxbx-Reloaded/XbSymbolDatabase/pull/121 and https://github.com/Zeex/subhook/pull/50. Once both are merged, I'll fixup this PR to point at the actual merge commits and un-draft.

A next batch of Direct3D fixes targeted at LTCG games:
* Patches `D3DDevice_SetRenderTarget_0` properly and accounts for a "CreateDevice did not call SetRenderTarget" test case - now this case only calls an internal implementation function, rather than trampolining to (sometimes non-existant) guest function.
* Patches `D3D_CommonSetRenderTarget` as a possibly recursive patch (called by `SetRenderTarget`). Even though it's an internal D3D function, some games (my test case: **Midtown Madness 3**) call it directly, presumably due to code inlining.
* Fixes `D3DDevice_SetPixelShader_0` corrupting the stack pointer. This is likely the reason which made **Metal Wolf Chaos** crash when using a Debug version of the emulator.
* Refactors all LTCG variations of `Direct3D_CreateDevice` - using naked functions where applicable and splitting `CreateDevice_16` into two separate functions, as OOVPAs matched the same symbol to functions with different calling conventions; we now properly preserve and pass arguments those functions, which may improve games using this symbol. For the sake of readable logs (and because naked functions cannot contain objects on their stack), I have created private overloaded functions with same names as the patch functions and placed the logging functions there. The end result are logs indistinguishable from the case where logging functions reside straight in the patch function.
* Implements `D3DDevice_DeleteVertexShader_0`. This untested function was forwarding to a non-LTCG function and thus promptly crashing, because a non-LTCG trampoline of course does not exist. Now it calls the internal function followed by its own trampoline.
* Bonus nitpick: `D3DDevice_Swap` was leaking a backbuffer handle. It's unlikely to change anything (it's a backbuffer after all), but it's also an one-line fix.

The combination of those changes allows **Midtown Madness 3** not to crash on startup and start presenting frames, but it will not show visuals until #1894 is finalized and merged.


src/core/hle/Patches.cpp
```diff
@@ -187,10 +188,12 @@ std::map<const std::string, const xbox_patch_t> g_PatchTable = {
 	PATCH_ENTRY("D3D_LazySetPointParams", xbox::EMUPATCH(D3D_LazySetPointParams), PATCH_HLE_D3D),
 	PATCH_ENTRY("D3D_SetCommonDebugRegisters", xbox::EMUPATCH(D3D_SetCommonDebugRegisters), PATCH_HLE_D3D),
 	PATCH_ENTRY("Direct3D_CreateDevice", xbox::EMUPATCH(Direct3D_CreateDevice), PATCH_HLE_D3D),
-	PATCH_ENTRY("Direct3D_CreateDevice_16", xbox::EMUPATCH(Direct3D_CreateDevice_16), PATCH_HLE_D3D),
+	PATCH_ENTRY("Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ebx_ppReturnedDeviceInterface", xbox::EMUPATCH(Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ebx_ppReturnedDeviceInterface), PATCH_HLE_D3D),
+	PATCH_ENTRY("Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface", xbox::EMUPATCH(Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface), PATCH_HLE_D3D),
 	PATCH_ENTRY("Direct3D_CreateDevice_4", xbox::EMUPATCH(Direct3D_CreateDevice_4), PATCH_HLE_D3D),
 	PATCH_ENTRY("Lock2DSurface", xbox::EMUPATCH(Lock2DSurface), PATCH_HLE_D3D),
 	PATCH_ENTRY("Lock3DSurface", xbox::EMUPATCH(Lock3DSurface), PATCH_HLE_D3D),
+	PATCH_ENTRY("D3D_CommonSetRenderTarget", xbox::EMUPATCH(D3D_CommonSetRenderTarget), PATCH_HLE_D3D),
```
RadWolfie, line 196: This line should be move above D3D_BlockOnTime's patch entry. We're using alphabet order by the way.

ESPPIRAL:
Improved

image

CookiePLMonster:
subbook PR has been merged to upstream and I updated the PR to point at the merge commit.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3143,10 +3148,25 @@ void Direct3D_CreateDevice_End()
...
+(
+    xbox::dword_xt                BehaviorFlags,
+    xbox::X_D3DPRESENT_PARAMETERS *pPresentationParameters,
+    IDirect3DDevice               **ppReturnedDeviceInterface
+)
+{
+	LOG_FUNC_BEGIN
+		LOG_FUNC_ARG(BehaviorFlags)
+		LOG_FUNC_ARG(pPresentationParameters)
+		LOG_FUNC_ARG(ppReturnedDeviceInterface)
```
CookiePLMonster, line 3162: Note to self: Turn those into `LOG_FUNC_ARG_OUT`.

CookiePLMonster:
Fixed rebase errors and slightly improved logging.

gandalfthewhite19890404:
World Series Baseball 2k3 now able get to menus, with flickering screen so if you have epilepsy - be careful:
<details>

image
</details>

gandalfthewhite19890404:
I have found another title: **Pro Race Driver**.
Master recognized Direct3D_CreateDevice_16 = 0x1f9200
and it will crash with this
`Received Exception Code 0xC0000005 @ EIP := 0x001F92A0(=Direct3D_CreateDevice_16+0xa0)`

PR2022 recognized Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface = 0x1f9200
and it will crash with this
`Received Exception Code 0xC0000005 @ EIP := 0x6F2116D3`

Here is PR2022 logfile with debug level for d3d8 functions - 
[KrnlDebug_PRD1.zip](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/5470342/KrnlDebug_PRD1.zip)

CookiePLMonster:
Note to self - subhook has a new version, since we need to update the submodule either way update it to v0.8 before un-drafting!

literalmente-game:
Kelly Slater's Pro Surfer
Master: crashes
PR: reaches in-game
Symbol cache (PR): [Kelly Slater's Pro Surfer-637751d7.txt](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/files/5474338/Kelly.Slater.s.Pro.Surfer-637751d7.txt)

CookiePLMonster:
Both submodule PRs are now merged, so this PR is ready for review/merge.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3155,58 +3175,178 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(Direct3D_CreateDevice_4)
...
+	__asm {
+		push ebp
+		mov  ebp, esp
+		sub  esp, __LOCAL_SIZE
+		mov  BehaviorFlags, eax
+		mov  ppReturnedDeviceInterface, ecx
+	}
+
+	// Log
+	Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface(Adapter, DeviceType, hFocusWindow, BehaviorFlags, pPresentationParameters, ppReturnedDeviceInterface);
```
PatrickvL, line 3254: If the `if(g_bPrintfOn)` check can be moved into a `LOG_CHECK_ENABLED_EX`, then a bit of overhead could be prevented by inserting `LOG_CHECK_ENABLED(LOG_LEVEL::DEBUG)` before the logging call, which makes it more clear the call is a logging call anyway, so the `// Log` comment line can be left out as well.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7300,8 +7441,10 @@ void CxbxImpl_InsertCallback
 	}
 }
 
-xbox::void_xt __declspec(noinline) D3DDevice_SetPixelShaderCommon(xbox::dword_xt Handle)
+xbox::void_xt D3DDevice_SetPixelShaderCommon(xbox::dword_xt Handle)
```
PatrickvL, line 7444: Better rename this into `CxbxImpl_SetPixelShader`

PatrickvL: (unless you want to keep the name for logging purposes, but then warn against renaming via a comment that describes WHY the name is specifically chosen this way.)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: Function call overhead is negligible especially since the compiler is free to come up with a custom calling convention, but a valid point nonetheless.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: I didn't put much thought into it, just noticed a `noinline` statement which was likely a workaround to some past bug and now is just noise.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7856,6 +7991,105 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetRenderTarget)
...
+	NestedPatchCounter call(setRenderTargetCount);
+
+	XB_TRMP(D3DDevice_SetRenderTarget)(pRenderTarget, pNewZStencil);
+
+	CxbxImpl_SetRenderTarget(pRenderTarget, pNewZStencil);
+}
+
+// LTCG specific D3DDevice_SetRenderTarget function...
+// Passes pRenderTarget in ecx and pNewZStencil in eax
+static void D3DDevice_SetRenderTarget_0_Inner
```
PatrickvL, line 8017: Perhaps the `_Inner` suffix can be dropped to avoid logging confusion? (There'll be no naming conflict, because the other is wrapped in a `EMUPATCH` macro.)

CookiePLMonster: Good idea, I handled the other cases but missed this one. Even without the `EMUPATCH` macro there would be no conflict because of different amount of arguments :)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: With me being unfamiliar with the logging system, how do you suggest this to be fixed?

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1602,7 +1602,7 @@ HRESULT CxbxImpl_CreateVertexShader(CONST DWORD *pDeclaration, CONST DWORD *pFun
 
 void CxbxImpl_DeleteVertexShader(DWORD Handle)
 {
-	LOG_INIT // Allows use of DEBUG_D3DRESULT
+	LOG_FUNC_ONE_ARG(Handle);
```
PatrickvL, line 1605: The `CxbxImpl_` prefix will show up in the log; Not sure we should fix/avoid that by jumping through hoops (like done above), but perhaps there's a solution that can avoid this issue entirely? Perhaps a change to the `LOG_INIT` macro that allows a custom function name replacing `__func__`?

PatrickvL (changes requested):
Some thoughts on logging. Otherwise again great work!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: In Logging.h, change :
```
#define LOG_CHECK_ENABLED_EX(cxbxr_module, level) \
	if (g_EnabledModules[to_underlying(cxbxr_module)] && to_underlying(level) >= g_CurrentLogLevel)
```
into 
```
#define LOG_CHECK_ENABLED_EX(cxbxr_module, level) \
	if (g_bPrintfOn && g_EnabledModules[to_underlying(cxbxr_module)] && to_underlying(level) >= g_CurrentLogLevel)
```

and change:
```
#define LOG_FUNC_BEGIN_NO_INIT \
	do { if(g_bPrintfOn) { \
```
into 
```
#define LOG_FUNC_BEGIN_NO_INIT \
	do { { \
```

then, in here, change : 
```
	// Log
	Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface(Adapter, DeviceType, hFocusWindow, BehaviorFlags, pPresentationParameters, ppReturnedDeviceInterface);
```
into 
```
	LOG_CHECK_ENABLED(LOG_LEVEL::DEBUG)
		Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface(Adapter, DeviceType, hFocusWindow, BehaviorFlags, pPresentationParameters, ppReturnedDeviceInterface);
```

Optionally, in Logging.h introduce : 
```
#define IF_LOG_ENABLED() LOG_CHECK_ENABLED(LOG_LEVEL::DEBUG)
```
Then it can become :
```
	IF_LOG_ENABLED()
		Direct3D_CreateDevice_16__LTCG_eax_BehaviorFlags_ecx_ppReturnedDeviceInterface(Adapter, DeviceType, hFocusWindow, BehaviorFlags, pPresentationParameters, ppReturnedDeviceInterface);
```

CookiePLMonster:
Applied all review comments except for the one we discussed on Discord which I consider too risky for this PR.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Let's shelve this for later

PatrickvL (approved):
Splendid!

gandalfthewhite19890404:
Lord of the Rings The Third Age crashes since PR was merged.

PR #2028, opened 2020-11-10
Update subhook

CookiePLMonster:
Treating this as a hotfix because updating to subhook v0.8 caused me to hit https://github.com/Zeex/subhook/issues/51 with at least one game. With this in mind, we should have reasons to believe regressions (startup crashes) might have occurred for some users with some games.

PR updates subhook to v0.8.1 which hotfixed this crash.


PatrickvL (approved):
I'll approve blindly, but if you could add a comment or two on which cases were affected, that'd be good for completeness. Thanks.

CookiePLMonster:
> I'll approve blindly, but if you could add a comment or two on which cases were affected, that'd be good for completeness. Thanks.

I had a startup crash in Project Gotham Racing 2 with a few extra hooks I added to get the game's output. After updating to 0.8.1, those crashes are gone.

PR #2031, opened 2020-11-15
Misc fixes after the vertex declaration changes

CookiePLMonster:
This PR fixes a few regressions after the vertex declaration changes have been merged:

* Fixes `D3DDevice_SetVertexShader_0` not calling to a trampoline - after recent changes calling to guest code here became instrumental for correct rendering but only the non-LTCG variation had a trampoline.
* Fixes `X_VERTEXSHADER_FLAG_PROGRAM` flag for xdk-3948. In this XDK version the flag appears to have a value of 4, not 16. Adjusted the code accordingly and added test cases to verify that assumption, also removed the previous test case (with a xdk-3948 game as a case), because now we have a more specific test case added instead.
* Fixes a merge error in `CxbxImpl_SetViewPort` - causing `X`/`Y` values to be multiplied by MSAA scaling twice.

Fixes regressions in **NASCAR Heat 2002**, revealing that the vertex declaration changes helped this game a lot, fixing shadowing issues and making self-shadowing work correctly!
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -3061,6 +3061,8 @@ void Direct3D_CreateDevice_Start
 	xbox::X_D3DPRESENT_PARAMETERS     *pPresentationParameters
 )
 {
+    CxbxVertexShaderSetFlags();
+
     if (!XboxRenderStates.Init()) {
         CxbxKrnlCleanup("Failed to init XboxRenderStates");
     }
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -4304,8 +4304,6 @@ void CxbxImpl_SetViewPort(xbox::X_D3DVIEWPORT8* pViewport)
 	HostViewPort.Y = static_cast<DWORD>(HostViewPort.Y * g_Xbox_MultiSampleYScale);
 	HostViewPort.Width = static_cast<DWORD>(HostViewPort.Width * g_Xbox_MultiSampleXScale);
 	HostViewPort.Height = static_cast<DWORD>(HostViewPort.Height * g_Xbox_MultiSampleYScale);
-	HostViewPort.X = static_cast<DWORD>(HostViewPort.X * g_Xbox_MultiSampleXScale);
-	HostViewPort.Y = static_cast<DWORD>(HostViewPort.Y * g_Xbox_MultiSampleYScale);
 	// Since Width and Height are DWORD, adding GetMultiSampleOffset 0.0f or 0.5f makes no sense
 
 	HRESULT hRet = g_pD3DDevice->SetViewport(&HostViewPort);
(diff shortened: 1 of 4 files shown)


LukeUsher:
nice find! Excellent work, as always

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3060,6 +3061,8 @@ void Direct3D_CreateDevice_Start
 	xbox::X_D3DPRESENT_PARAMETERS     *pPresentationParameters
 )
 {
+	CxbxVertexShaderSetFlags();
```
LukeUsher, line 3064: Indenting seems off here

LukeUsher (changes requested):
just a minor indentation fix, otherwise good

PatrickvL (dismissed):
LGTM

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1465,9 +1487,6 @@ void CxbxImpl_SetVertexShader(DWORD Handle)
 			CxbxSetVertexShaderPassthroughProgram();
 			g_Xbox_VertexShader_IsFixedFunction = false;
 			g_Xbox_VertexShader_IsPassthrough = true;
-		} else if (pXboxVertexShader->Flags & X_VERTEXSHADER_FLAG_UNKNOWN) {
-			// Test-case : Amped
```
PatrickvL, line 1469: Perhaps mention this removed test-case above, near `LOG_TEST_CASE("Unknown vertex shader flag")`

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1408,7 +1425,12 @@ void CxbxImpl_SetVertexShader(DWORD Handle)
 
 	xbox::X_D3DVertexShader* pXboxVertexShader = CxbxGetXboxVertexShaderForHandle(Handle);
 	g_Xbox_VertexShader_IsPassthrough = false;
-	if (pXboxVertexShader->Flags & X_VERTEXSHADER_FLAG_PROGRAM) {
+
+	if ((pXboxVertexShader->Flags & g_X_VERTEXSHADER_FLAG_VALID_MASK) != pXboxVertexShader->Flags) {
+		LOG_TEST_CASE("Unknown vertex shader flag");
+	}
+
+	if (pXboxVertexShader->Flags & g_X_VERTEXSHADER_FLAG_PROGRAM) {
```
PatrickvL, line 1433: Perhaps refer to `CxbxVertexShaderSetFlags` here (otherwise someone might think this is a typo)

src/core/hle/D3D8/XbVertexShader.cpp
CookiePLMonster: I don't think it's necessary, as it's effectively the same test case, but more specific.

PatrickvL: Sure. Maybe we should try to get confirmation that Amped won't trigger the non-valid check now ;)

LukeUsher (approved):

medievil1:
fixes NFS HP2 (EA 040 V1.02) rearview mirror and I think road shadows looks better, car road shadows look inverted though, lighter where they should be darker, darker where they should be lighter 
image

src/core/hle/D3D8/XbVertexShader.cpp
CookiePLMonster: I don't see how it would - in order to hit this code path `X_VERTEXSHADER_FLAG_PROGRAM` **has to** be false (so our new test case will **not** trigger). Therefore, it behaves the same way as NASCAR Heat 2002 does, both titles using an old enough XDK.

PatrickvL: Right. I had to wrap my head around the set/unset bits. No verification needed, indeed.

PatrickvL (approved):

escape209:
Fixes shadows in Tony Hawk's Pro Skater 2X (previously invisible)
image

ObiKKa:
> This PR fixes a few regressions after the vertex declaration changes have been merged:
The original poster above is alluding to PR #1894 created on May 9, 2020 by both PatrickvL and NZJenkins together (The sole commit in the PR seems to have come from a rebase, but there was an older list of many commits by both coders.). It had been in draft stage for a few months and was then merged by Luke Usher on November 15, 2020.

Sorry to bump this thread up.

PR #2032, opened 2020-11-15
Fix CDirectSoundStream_GetStatus for difference output over time

RadWolfie:
From titles built with XDK 3911 to 4039, I notice `CDirectSoundStream_GetStatus` API only has limit to one flag output. Then 4134 and later has extend to include more flags. This pull request will resolve the issue.


CookiePLMonster:
The few test cases we briefly discussed:

* NASCAR Heat 2002 (XDK 3911) - no changes, audio is fine
* Burnout 3 (XDK 5849) - no changes, EA Trax still plays and doesn't stop as the game starts
* The Warriors (XDK 5849) - no changes, cutscene audio still doesn't stop after skipping cutscenes

RadWolfie:
For clarification, titles below XDK 4134 build only check one value which is ready for input packets or not. There could be a chance some titles may not check the flag instead check the whole integer when not supposed to. In my opinion, this is critical fix to cancel any false positive reports.

medievil1:
I've tested this with all I have ripped and not found any issues, good or bad

PatrickvL (commented):
LGTM

PatrickvL (approved):

PR #2033, opened 2020-11-16
Improve viewport calculations

### LINKED ISSUE #1927: Improve viewport scale and offset emulation
My understanding of viewport transforms. Probably some incorrect details/terminology in here

#### Xbox vs D3D9

In D3D9, vertices are output in clipspace (?) (screen X and Y range -1 to 1)
On Xbox, vertices are output in screenspace (screen X and Y range from 0 to render target width/height)

Xbox vertex shader programs generally contain the following instructions, to apply the transformation from clip to screenspace:
```hlsl
mul oPos.xyz, r12, c-38 // Multiple by scale factor
+rcc r1.x, r12.w // compute 1 / w for division by w
mad oPos.xyz, r12, r1.x, c-37 // divide by w and add offset
```
where `c-38` is the viewport scale and `c-37` the viewport offset constant registers.

By default, the Xbox D3D8 library will set the viewport scale and offset constants, but titles can:
* Tell Xbox D3D8 not to set the reserved viewport registers
* Use vertex shader programs that ignore the viewport registers, whether they are getting set by Xbox D3D8 or not

To handle these different cases, we can reverse the screenspace transformation and output normalized coordinates.
Ideally we could get the real viewport and offset coordinates from Xbox D3D8 - otherwise we can compute it.

Note in the cases where the conventional instructions are used, it doesn't matter if our scale and offset computations are correct.
But in the cases where the conventional registers and instructions are not used, we can get a mismatch between the transform that was done and the transform that we reverse.

#### The viewport offset and scale constants are affected by:
* Rendertarget size
* The viewport

where there rendertarget size is affected by
* Backbuffer scale
* Multisample mode
Supersampling affects the scale constant, but multisampling does not

 #### Test cases for some viewport scale and offset related effects
* Need for Speed: Hot Pursuit 2 (fantastic test case!)
  * Uses multiple different viewports for different menu / 2d elements
  * Uses backbuffer scaling (IIRC parts of in-game HUD use 0.75x)
  * 2x MSAA quincunx
  * Passthrough vertex shader program (rather than using FVF XYZRHW passthrough mode)
* Antialias sample
  * All antialias modes
  * Expects different coordinates for HUD vs 3d (multiple different sized rendertargets?)
* DoA 3
  * Uses viewport offset and scale for character portraits in character select
* JSRF
  * Multiple viewports for the character select screen

*Experimental* viewport work https://github.com/NZJenkins/Cxbx-Reloaded/commits/e/viewport

#### TODO
relationship between clipping and viewport handling?

NZJenkins:
- SetViewport tracks the Xbox viewport values instead of setting the viewport,
so we can calculate what we need from them
- Rework GetViewportOffsetAndScale and scaling methods
- Vertex shader programs do not depend on D3D9 viewport to transform verts
- Account for backbuffer scale
- Add some D3DSWAP flags and LOG_TEST_CASE for unhandled usage

- [x] Clean up repetitive scale calculations
- [x] Investigate Antialias and backbufferscale samples

Closes #1927


src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -271,20 +271,22 @@ float4 reverseScreenspaceTransform(float4 oPos)
...
-	oPos.xyz -= xboxViewportOffset.xyz; // reverse offset
-	oPos.xyz *= oPos.w; // reverse perspective divide
-	oPos.xyz *= xboxViewportScaleInverse.xyz; // reverse scale
-
+	// Reverse screenspace offset
+	oPos += xboxViewportOffset;
+	// Reverse screenspace scale
+	oPos *= xboxViewportScaleInverse;
+	// Flip Y
+	oPos.y *= -1;
```
PatrickvL, line 283: Flipping Y could be done on CPU, by flipping the sign of the `.y` component of the `xboxViewportScaleInverse` constant

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4240,74 +4239,60 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_SetViewport)
...
+		// Use default values
 
-	// Apply MSAA scale and offset
-	HostViewPort.X = static_cast<DWORD>(HostViewPort.X * g_Xbox_MultiSampleXScale);
-	HostViewPort.Y = static_cast<DWORD>(HostViewPort.Y * g_Xbox_MultiSampleYScale);
-	HostViewPort.Width = static_cast<DWORD>(HostViewPort.Width * g_Xbox_MultiSampleXScale);
-	HostViewPort.Height = static_cast<DWORD>(HostViewPort.Height * g_Xbox_MultiSampleYScale);
-	// Since Width and Height are DWORD, adding GetMultiSampleOffset 0.0f or 0.5f makes no sense
 
-	HRESULT hRet = g_pD3DDevice->SetViewport(&HostViewPort);
```
PatrickvL, line 4309: If this was the last appearance of host `SetViewport()`, should or shouldn't that still be done once, in `Direct3D_CreateDevice_End` for instance?

PatrickvL: Nevermind - I see there's a new call to host SetViewport below...

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7121,6 +7106,65 @@ void CxbxUpdateHostVertexShaderConstants()
 	CxbxUpdateHostViewPortOffsetAndScaleConstants();
 }
 
+void CxbxUpdateHostViewport() {
+	// We don't have a fixed function shader so we rely on D3D9 fixed function mode
```
PatrickvL, line 7110: (yet)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7121,6 +7106,65 @@ void CxbxUpdateHostVertexShaderConstants()
...
+
+		D3DVIEWPORT hostViewport;
+		hostViewport.X = 0;
+		hostViewport.Y = 0;
+		hostViewport.Width = HostRenderTarget_Width;
+		hostViewport.Height = HostRenderTarget_Height;
+		hostViewport.MinZ = 0.0f;
+		hostViewport.MaxZ = 1.0f;
+
+		g_pD3DDevice->SetViewport(&hostViewport);
```
PatrickvL, line 7154: Okay, so when Fixed Function HLSL is completed, this bit of code /should/ go to `Direct3D_CreateDevice_End`, right? Or can `HostRenderTarget_Width` and/or `HostRenderTarget_Height` change over time? Then maybe it couldd be worthwhile to see if we could update vertex shaders such, that host viewport would always stay the same (even when render target dimensions change), if that's possible?

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -382,15 +382,17 @@ float4 reverseScreenspaceTransform(float4 oPos)
...
-
-	oPos.xyz -= xboxViewportOffset.xyz; // reverse offset
-	oPos.xyz *= oPos.w; // reverse perspective divide
-	oPos.xyz *= xboxViewportScaleInverse.xyz; // reverse scale
+	// Reverse screenspace offset
+	oPos += xboxViewportOffset;
+	// Reverse screenspace scale
+	oPos *= xboxViewportScaleInverse;
+	// Flip Y
+	oPos.y *= -1;
```
PatrickvL, line 390: Flipping Y could be done on CPU, by flipping the sign of the .y component of the xboxViewportScaleInverse constant

PatrickvL (commented):
A few considerations & questions

PatrickvL:
Before, current master :
image
After : 
image

escape209:
This fixes some 2D elements being missing in SpongeBob SquarePants: Battle for Bikini Bottom.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4135,73 +4140,80 @@ float GetZScaleForSurface(xbox::X_D3DSurface* pSurface)
...
+	float scaledWidth = g_Xbox_Viewport.Width * aaScaleX * g_BackbufferScaleX;
+	float scaledHeight = g_Xbox_Viewport.Height * aaScaleY * g_BackbufferScaleY;
 
-	// Offset
-	vOffset[0] = multisampleOffsetX + xboxScreenWidth * 0.5f;
-	vOffset[1] = multisampleOffsetY + xboxScreenHeight * 0.5f;
-	vOffset[2] = 0.0f; //offsetZ;
-	vOffset[3] = 0.0f;
+	// D3D9 viewport info
+	https://docs.microsoft.com/en-us/windows/win32/direct3d9/viewports-and-clipping
```
mzji, line 4164: Lacks of `//`

gandalfthewhite19890404:
### Improvements:
**Black** now renders graphics - FMVs and actual gameplay.

gandalfthewhite19890404:
### Regressions:
**Loons: The Fight for Fame** crashed with Unknown primitive pop-up fatal error. BUT I tried latest master - it does not crashed, so I decided to test 2033 again to get error screenshot and it does not crashed. Why?
With PR game looks much worse - missing graphics and exploded polygons:
image


PS
should I edit previous post or I may just add them?

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
NZJenkins: yes, I can't remember why I made a point of doing it separately here
though AFAIK it should be optimized away since it is constant over the render, same with the inverse thing 🤷‍♂️ 

Viewport isn't really an accurate description of this variable now.
I'm thinking update the name to `xboxScreenspaceScale/Offset` or something, and remove the inverse so we can easily read what we thought the render target width/height was for debuggability?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: I don't know what the significance of `Direct3D_CreateDevice_End` is

`ValidateRenderTargetDimensions` implementation suggests the rendertarget size can change at any time
I removed one of the calls to it so need to add that back in
I guess in `UpdateNativeD3DResources` makes the most sense, so the viewport stuff will probably have to stay there

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: nice catch

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: These variables don't even contain Xbox screenspace directly, they contain the adjustments so that output becomes normalized (and converted from OpenGL to Direct3D orthogonality).

As for the multiplication by inverse instead of a division, wasn't that also to retain accurately?

NZJenkins: True it's not screenspace.
It is the rendertarget "space" not including subpixels (but because rendertarget sizes are larger when using subpixel MSAA it is a bit confusing?)
Yes we just want to normalize all the things...`xboxNormalizingScale`? idk
Uh I don't remember I thought I just read somewhere divisions should be avoided (but since it is a constant I don't think it matters)

medievil1:
ummm... KOTOR (LA-003 V1.9526529) menu is kinda jacked...lol
image

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
NZJenkins: added some helper functions that calculate various dimensions side-by-side, hopefully that makes it clearer what the differences are

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4135,73 +4172,69 @@ float GetZScaleForSurface(xbox::X_D3DSurface* pSurface)
...
+	float xboxRendertargetHeight;
+	GetRendertargetOutputDimensions(xboxRendertargetWidth, xboxRendertargetHeight);
+
+	// Xbox outputs vertex positions in the rendertarget pixel coordinate space, with non-normalized Z
+	// We want to scale it back to normalized device coordinates i.e. XY are (-1, +1) and Z is (0, 1)
+	float normalizeScale[4] = { xboxRendertargetWidth / 2,  -xboxRendertargetHeight / 2, vScaleOffset[0][2], 1 };
+	float normalizeOffset[4] = { -xboxRendertargetWidth / 2, -xboxRendertargetHeight / 2, -vScaleOffset[1][2], 0 };
+	g_pD3DDevice->SetVertexShaderConstantF(CXBX_D3DVS_NORMALIZE_SCALE_BASE, normalizeScale, CXBX_D3DVS_NORMALIZE_SCALE_SIZE);
+	g_pD3DDevice->SetVertexShaderConstantF(CXBX_D3DVS_NORMALIZE_OFFSET_BASE, normalizeOffset, CXBX_D3DVS_NORMALIZE_OFFSET_SIZE);
+	g_pD3DDevice->SetVertexShaderConstantF(CXBX_D3DVS_IS_RHW_TRANSFORMED_POSITION_BASE, isRHWTransformedPosition, CXBX_D3DVS_IS_RHW_TRANSFORMED_POSITION_SIZE);
```
PatrickvL, line 4237: Because `xboxIsRHWTransformedPosition` is removed from HLSL, this `CXBX_D3DVS_IS_RHW_TRANSFORMED_POSITION_BASE` constant seems no longer needed...

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3056,6 +3060,38 @@ void CalculateMultiSampleScaleForRenderTarget(xbox::X_D3DSurface* renderTarget)
...
+void GetRendertargetBaseDimensions(float& x, float& y) {
+	GetRendertargetRawDimensions(x, y);
+
+	x /= g_Xbox_MultiSampleXScale;
+	y /= g_Xbox_MultiSampleYScale;
+}
+
+// Get rendertarget output dimensions as far as the GPU is concerned
+// This is affected by the backbuffer scale and multisampling type
+void GetRendertargetOutputDimensions(float& x, float& y) {
```
PatrickvL, line 3079: nit : uppercase T : Render**T**arget

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4135,73 +4172,69 @@ float GetZScaleForSurface(xbox::X_D3DSurface* pSurface)
...
+	// Test case: GTA III
 
-	// Scale
-	vScale[0] = xboxScreenWidth * 0.5f;
-	vScale[1] = xboxScreenHeight * -0.5f;
-	vScale[2] = scaleZ; // ?
-	vScale[3] = 1.0f; // ?
+	auto zRange = g_Xbox_Viewport.MaxZ - g_Xbox_Viewport.MinZ;
+	vScale[0] = scaledWidth / 2;
+	vScale[1] = -scaledHeight / 2;
```
PatrickvL, line 4202: Please add a comment about intentional Flipping of Y here (and perhaps also explain why)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: It seems that for now, host Viewport must change along with RenderTarget dimensions. Optimizing this update to happen only when it changes can be done later.

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1025,13 +1025,13 @@ typedef DWORD X_VERTEXSHADERCONSTANTMODE;
...
-#define CXBX_D3DVS_VIEWPORT_OFFSET_MIRROR_BASE     (CXBX_D3DVS_VIEWPORT_SCALE_INVERSE_BASE + CXBX_D3DVS_VIEWPORT_SCALE_INVERSE_SIZE)
-#define CXBX_D3DVS_VIEWPORT_OFFSET_MIRROR_SIZE     1
+#define CXBX_D3DVS_NORMALIZE_OFFSET_BASE     (CXBX_D3DVS_NORMALIZE_SCALE_BASE + CXBX_D3DVS_NORMALIZE_SCALE_SIZE)
+#define CXBX_D3DVS_NORMALIZE_OFFSET_SIZE     1
 
-#define CXBX_D3DVS_TEXTURES_SCALE_BASE             (CXBX_D3DVS_VIEWPORT_OFFSET_MIRROR_BASE + CXBX_D3DVS_VIEWPORT_OFFSET_MIRROR_SIZE)
+#define CXBX_D3DVS_TEXTURES_SCALE_BASE             (CXBX_D3DVS_NORMALIZE_OFFSET_BASE + CXBX_D3DVS_NORMALIZE_OFFSET_SIZE)
 #define CXBX_D3DVS_TEXTURES_SCALE_SIZE             4
 
 #define CXBX_D3DVS_IS_RHW_TRANSFORMED_POSITION_BASE (CXBX_D3DVS_TEXTURES_SCALE_BASE + CXBX_D3DVS_TEXTURES_SCALE_SIZE)
```
PatrickvL, line 1037: `CXBX_D3DVS_IS_RHW_TRANSFORMED_POSITION_BASE` can be removed (see above)

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -271,20 +271,20 @@ float4 reverseScreenspaceTransform(float4 oPos)
...
 	// and multiplying oPos.xyz by oPos.w may cause precision issues.
 	// Test case: Burnout 3
 
-	oPos.xyz -= xboxViewportOffset.xyz; // reverse offset
-	oPos.xyz *= oPos.w; // reverse perspective divide
-	oPos.xyz *= xboxViewportScaleInverse.xyz; // reverse scale
+	// Reverse screenspace offset
+	oPos += xboxNormalizeOffset;
+	// Reverse screenspace scale
+	oPos /= xboxNormalizeScale;
```
PatrickvL, line 281: Debugging advantage aside, replacing this divide by a multiplication of the inverse scale would avoid division by zero.

Also, didn't Silent/CookieMonster submit a PR on this very calculation a while ago about accuracy?

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: Nice, thanks.

PatrickvL (changes requested):
A few remarks, still ;)

gandalfthewhite19890404:
Mafia (map and letters) and Pirates of a Caribbean(dialogues) are fixed now.

MrDinkle:
Mortal Kombat: Deception and Mortal Kombat: Armageddon now both render all of the ui-related elements correctly.
<details>
[CXBX_Deception](https://user-images.githubusercontent.com/43343017/99454467-72516400-28f4-11eb-8c6c-c9753a4971e6.PNG)

[CXBX_Armageddon](https://user-images.githubusercontent.com/43343017/99454464-71b8cd80-28f4-11eb-8ab8-1f2ad6a75f44.PNG)
</details>

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
NZJenkins: we still would divide by zero, just not as part of the shader
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2005 got clobbered by #1894 I believe

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: I don't know why.. it is what the XDK (and D3D9) does, isn't it just a convention?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4135,73 +4172,68 @@ float GetZScaleForSurface(xbox::X_D3DSurface* pSurface)
...
+	vOffset[3] = 0;
 }
 
 void CxbxUpdateHostViewPortOffsetAndScaleConstants()
 {
 	extern bool g_Xbox_VertexShader_IsPassthrough;
 
     float vScaleOffset[2][4]; // 0 - scale 1 - offset
-    GetViewPortOffsetAndScale(vScaleOffset[1], vScaleOffset[0]);
+	float isRHWTransformedPosition[4] = { 0 };
```
PatrickvL, line 4216: `isRHWTransformedPosition` can also be removed

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4135,73 +4172,68 @@ float GetZScaleForSurface(xbox::X_D3DSurface* pSurface)
...
+    GetViewportOffsetAndScale(vScaleOffset[1], vScaleOffset[0]);
+
+	if (g_Xbox_VertexShader_IsPassthrough) {
+		isRHWTransformedPosition[0] = 1.0f;
```
PatrickvL, line 4220: `isRHWTransformedPosition` can also be removed

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: Indeed - and my conflict resolve is to blame for that;

As for what was done in #2005, it's comment (which is still present a few lines up) read :
````
 	// oPos.w and xboxViewportScale.z might be VERY big when a D24 depth buffer is used
	// and multiplying oPos.xyz by oPos.w may cause precision issues.
	// Pre-divide them to help keep the values reasonably small.
	// Test case: Burnout 3
````
And it did this by changing this old code :
````
	oPos.xyz *= oPos.w; // reverse perspective divide
	oPos.xyz /= xboxViewportScale.xyz; // reverse scale
````
Into this alternative :
````
	float3 divisor = xboxViewportScale.xyz / oPos.w;
	oPos.xyz /= divisor; // reverse scale and perspective divide
````

So, is appears that the division you now have here might avoid the precision issue as well (assuming that your `xboxNormalizeScale` divisor has the same accuracy when now pre-calculated on CPU compared to the GPU accuracy of the `divisor` in #2005).

Perhaps someone could confirm Burnout 3 is still fine with this PR?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Just add a comment stating : `// Flip Y to convert Xbox (OpenGL) to our D3D coordinate system` or something along those lines.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
medievil1: image

medievil1: image

medievil1: whatever's on top doesn't clear though....tracers left behind 
image

PatrickvL: Thanks @medievil1! @CookiePLMonster confirmed on discord that this is how Burnout 3 looked before for him as well :
    good menus and UI, black screen in races because of them using an inlined IVB to "blit" it
So, this confirms this has not regressed.
Closing discussion.

CookiePLMonster:
Fixes cut off menus in DRIV3R:

Before:
image

After:
image

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: Note in this code, _we_ are not flipping Y, it is what the XDK does. Ideally we are not doing anything here, since we want to match XDK behaviour.
It is the same as is documented in D3D9 so I don't think it's of note or there's anything much to add

Will comment to make it clearer that we are just imitating XDK, and we should be getting this from NV2A LLE instead

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
NZJenkins: I flip-flopped back to calling it screenspace because I couldn't think of a good name 😖

PatrickvL (approved):
LGTM!

PR #2039, opened 2020-11-23
Another Round of DSound's Stream fixes

RadWolfie:
After going though series of hardware sampling tests with DirectSoundStream class. Both hardware and this pull request's log are exactly the same.

With testcase, Burnout 3, mentioned by @CookiePLMonster give me a theory of what's happening inside stream's internal flush function. I had to rerun extended sampling test to fix final expectation result from cxbxr's end.

Recommended to test against:
* Burnout 3 ✅ (resolve in commit: `dsound: don't process flush if there's no packets`)
* The Warrior ✅ (resolve in commit: `dsound: fix status return when envelope is complete`)
* NASCAR HEAT 2002 ✅ (no change, not affected)
* Obscure ❌ (Currently still using hack for now, functions need to break down for improve understanding how things work.)
* ???

Obscure is still using same hack by call internal flush function directly once FlushEx is called with async and immediate flags. It will be investigate further down the path.

---

P.S. I also fix an inaccurate log output result. And to my surprising, on stream's release does call flush. Another hidden bug I was not aware of.

close #2004


src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
```diff
@@ -218,6 +211,16 @@ bool DSStream_Packet_Process(
         return 0;
     }
 
+    // Do not allow to process when voice is not activate.
```
CookiePLMonster, line 214:
```suggestion
    // Do not allow to process when the voice is not activated.
```

src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
```diff
@@ -354,6 +347,13 @@ bool DSStream_Packet_Flush(
     xbox::X_CDirectSoundStream* pThis
     )
 {
+
+    // TESTCASE: Burnout 3 set stream to pause state then call SetFormat without process any packets.
```
CookiePLMonster, line 351:
```suggestion
    // TESTCASE: Burnout 3 sets stream to pause state then calling SetFormat without processing any packets.
```

src/core/hle/DSOUND/DirectSound/DSStream_PacketManager.cpp
```diff
@@ -354,6 +347,13 @@ bool DSStream_Packet_Flush(
...
+    // Which then doesn't need to flush any packets.
+    if (pThis->Host_BufferPacketArray.size() == 0) {
```
CookiePLMonster, line 353:
```suggestion
    if (pThis->Host_BufferPacketArray.empty()) {
```

src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
```diff
@@ -549,14 +549,16 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(CDirectSoundStream_GetStatus__r2)
 
     // Convert host to xbox status flag.
     if (hRet == DS_OK) {
-        if (pThis->Host_isProcessing && !(dwStatusXbox & X_DSSSTATUS_PAUSED)) {
+        if (pThis->Host_BufferPacketArray.size() != 0 && (pThis->EmuFlags & (DSE_FLAG_PAUSE | DSE_FLAG_PAUSENOACTIVATE)) == 0) {
```
CookiePLMonster, line 552:
```suggestion
        if (!pThis->Host_BufferPacketArray.empty() && (pThis->EmuFlags & (DSE_FLAG_PAUSE | DSE_FLAG_PAUSENOACTIVATE)) == 0) {
```

src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
```diff
@@ -549,14 +549,16 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(CDirectSoundStream_GetStatus__r2)
...
             dwStatusXbox |= X_DSSSTATUS_PLAYING;
 
         }
         if (pThis->Host_BufferPacketArray.size() != pThis->X_MaxAttachedPackets) {
             dwStatusXbox |= X_DSSSTATUS_READY;
         }
-
+        if (pThis->Host_BufferPacketArray.size() != 0 && (pThis->EmuFlags & DSE_FLAG_PAUSE) != 0) {
```
CookiePLMonster, line 559:
```suggestion
        if (!pThis->Host_BufferPacketArray.empty() && (pThis->EmuFlags & DSE_FLAG_PAUSE) != 0) {
```

RadWolfie:
I will resolve review remarks above and I made a discovery of `DSStream_Packet_Starved` has paused flag which doesn't get clear out. **Please do not merge until it is fixed.**

CookiePLMonster:
Tested a few games:
* **NASCAR Heat 2002** - no changes, still all good
* **Burnout 3** - EA Trax does not play in the menu anymore. Music in main menu still loops, but that's an unrelated issue
* **The Warriors** - **not fixed, cutscene audio still plays when skipping the cutscene**

LanHikariDS:
Not sure what this does as opposed to Silent's DSound PR?

RadWolfie:
@LanHikariDS Partial answer is:
> After going though series of hardware sampling tests with DirectSoundStream class. Both hardware and this pull request's log are exactly the same.

Although, I'll extend the reason of not merge Silent's DSound pull request.

Base on my experience through trials and errors and how hardware testing result are. Set stream class to pause, with voice activated, state with a packet after it will always reset pause flag when SetFormat/Flush is called. Once title submit new packet will trigger play the packet instead of remain pause. However, "pause with no voice activated" flag remain always paused state no matter how many time it has call SetFormat/Flush whenever there's no packets input. But... if the stream is paused with packets and has voice active. Then get called again with "pause with no voice activated" flag will resume stream.

Silent's DSound pull request contain always pause the stream which isn't accurate compare to hardware and will cause a lot of regressions.

Beside, this pull request is intended to implement "pause with no activated" flag base on how it should behave compare from hardware testing.

RadWolfie:
@CookiePLMonster One thing I notice from your The Warrior's log file using `CDirectSoundStream_FlushEx` calls. What you're experiencing audio still playing from cutscene still remain unknown cause. It should be terminate at some point.

RadWolfie:
Ready for review/merge. It's unlikely able to cause regression with this current state since there are numerous of improvements. Except do need to break them down into separate functions for better understanding as in general overall.

RadWolfie:
rebased + remove forced pause removal in patched FlushEx function

CookiePLMonster:
Fixes cutscene skipping not stopping the audio in **The Warriors** at the expense of the main menu music and in-game BGM not working anymore.

PatrickvL:
Caused, not paused? As in : "regressed in"?

RadWolfie:
> Caused, not paused? As in : "regressed in"?

caused*, I'm very sure is what he meant. Instead of revert the change, it's better to investigate how XDK handle the mixbin system. Then make a proper implement yet still follow with highest volume value usage.

PatrickvL (approved):

Issue #1658, opened 2019-07-11
Popup "Existing RenderTarget width/height changed"


gandalfthewhite19890404:
LOG_TEST_CASE: Existing RenderTarget width/height changed
In ValidateRenderTargetDimensions (C:\projects\cxbx-reloaded\src\core\hle\D3D8\Direct3D9\Direct3D9.cpp line 3227)
Titles:
- WWE RAW 2

infinitephantasm:
Jet Set Radio Future triggers this popup when boost dashing.

ObiKKa:
This game, Cold Fear, consists of 8 different test case messages according to an April 2020 update in its issue.

- [Cold Fear](https://github.com/Cxbx-Reloaded/game-compatibility/issues/728#issuecomment-615327344)

doug-clayton:
This is also appearing in Dead or Alive 3.

literalmente-game:
Also present on Conker: Live & Reloaded

escape209:
Need for Speed: Carbon displays this on startup.

fantesykikachu:
Ninja Gaiden: Black also shows this message, when after starting the first mission.

ObiKKa:
A test report from one user at PR #2073 (Pixel Shader HLSL [Draft]), dated December 12, 2020 and initially started in the draft stage, showed that several Need for Speed games did progress a good way and mostly into the in-game stage and, as a result, presented this issue's popup message. Exciting times! :sweat_smile: :raised_hands: :man_mechanic: :cityscape: :city_sunset: :motorway: :vertical_traffic_light: :red_car: :oncoming_automobile: :blue_car: :checkered_flag: :triangular_flag_on_post: 

Of course, one tester from two comments above did confirm that two of the NFS games showed this same popup message.

- [Need for Speed: Underground](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2073#issuecomment-743695466) (2004)
- [Need for Speed: Most Wanted](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2073#issuecomment-743695466) (2005)
- [Need for Speed: Carbon](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2073#issuecomment-743695466) (2006)

PR #2040, opened 2020-11-24
Z fixes

NZJenkins:
- Calculate Z matching D3D9 docs (combined with Xbox depth scaling)
- Normalize Z from either (0, 1) or (0, zbuffer depth)

~~Haven't found any test case for this yet...~~
> The Rendertarget commit corrected aboth JSRF test cases, as well as the GunValkyrie one.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -4862,12 +4862,14 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_Clear)
         // Scale the fill based on our scale factor and MSAA scale
 		float aaX, aaY;
 		GetMultiSampleScaleRaw(aaX, aaY);
+		auto rtUpscale = GetRenderTargetUpscaleFactor();
+
         std::vector<D3DRECT> rects(Count);
         for (DWORD i = 0; i < Count; i++) {
-            rects[i].x1 = static_cast<LONG>(pRects[i].x1 * aaX * GetRenderTargetUpscaleFactor());
-            rects[i].x2 = static_cast<LONG>(pRects[i].x2 * aaX * GetRenderTargetUpscaleFactor());
-            rects[i].y1 = static_cast<LONG>(pRects[i].y1 * aaY * GetRenderTargetUpscaleFactor());
-            rects[i].y2 = static_cast<LONG>(pRects[i].y2 * aaY * GetRenderTargetUpscaleFactor());
+            rects[i].x1 = static_cast<LONG>(pRects[i].x1 * aaX * rtUpscale);
+            rects[i].x2 = static_cast<LONG>(pRects[i].x2 * aaX * rtUpscale);
+            rects[i].y1 = static_cast<LONG>(pRects[i].y1 * aaY * rtUpscale);
+            rects[i].y2 = static_cast<LONG>(pRects[i].y2 * aaY * rtUpscale);
 		}
         hRet = g_pD3DDevice->Clear(Count, rects.data(), HostFlags, Color, Z, Stencil);
     } else {
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2274,7 +2274,7 @@ static DWORD WINAPI EmuCreateDeviceProxy(LPVOID)
 
             if (g_EmuCDPD.bCreate) {
 				// Apply render scale factor for high-resolution rendering
-				g_RenderScaleFactor = g_XBVideo.renderScaleFactor;
+				g_RenderUpscaleFactor = g_XBVideo.renderScaleFactor;
 
 				// Setup the HostPresentationParameters
 				{
@@ -3075,6 +3075,14 @@ void GetRenderTargetBaseDimensions(float& x, float& y) {
 
 	x /= aaX;
 	y /= aaY;
+}
+
+// Get the host upscale factor for the current rendertarget
+// Currently, only the host backbuffer is upscaled
+int GetRenderTargetUpscaleFactor() {
+	return g_pXbox_RenderTarget == g_pXbox_BackBufferSurface
+		? g_RenderUpscaleFactor
+		: 1;
 }
 
 void Direct3D_CreateDevice_Start
@@ -4112,15 +4120,13 @@ void ValidateRenderTargetDimensions(DWORD HostRenderTarget_Width, DWORD HostRend
     // Because of this, we need to validate that the associated host resource still matches the dimensions of the Xbox Render Target
     // If not, we must force them to be re-created
     // TEST CASE: Chihiro Factory Test Program
-    DWORD XboxRenderTarget_Width_Scaled = XboxRenderTarget_Width * g_RenderScaleFactor;
-    DWORD XboxRenderTarget_Height_Scaled = XboxRenderTarget_Height * g_RenderScaleFactor;
+    DWORD XboxRenderTarget_Width_Scaled = XboxRenderTarget_Width * GetRenderTargetUpscaleFactor();
+    DWORD XboxRenderTarget_Height_Scaled = XboxRenderTarget_Height * GetRenderTargetUpscaleFactor();
     if (HostRenderTarget_Width != XboxRenderTarget_Width_Scaled || HostRenderTarget_Height != XboxRenderTarget_Height_Scaled) {
         LOG_TEST_CASE("Existing RenderTarget width/height changed");
 
-        if (g_pXbox_RenderTarget == g_pXbox_BackBufferSurface) {
-            FreeHostResource(GetHostResourceKey(g_pXbox_RenderTarget)); g_pD3DDevice->SetRenderTarget(0, GetHostSurface(g_pXbox_RenderTarget, D3DUSAGE_RENDERTARGET));
-            FreeHostResource(GetHostResourceKey(g_pXbox_DepthStencil)); g_pD3DDevice->SetDepthStencilSurface(GetHostSurface(g_pXbox_DepthStencil, D3DUSAGE_DEPTHSTENCIL));
-        }
+        FreeHostResource(GetHostResourceKey(g_pXbox_RenderTarget)); g_pD3DDevice->SetRenderTarget(0, GetHostSurface(g_pXbox_RenderTarget, D3DUSAGE_RENDERTARGET));
+        FreeHostResource(GetHostResourceKey(g_pXbox_DepthStencil)); g_pD3DDevice->SetDepthStencilSurface(GetHostSurface(g_pXbox_DepthStencil, D3DUSAGE_DEPTHSTENCIL));
     }
 }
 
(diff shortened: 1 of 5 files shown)


LanHikariDS:
The Rendertarget commit corrected aboth JSRF test cases, as well as the GunValkyrie one.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4110,15 +4118,13 @@ void ValidateRenderTargetDimensions(DWORD HostRenderTarget_Width, DWORD HostRend
...
         LOG_TEST_CASE("Existing RenderTarget width/height changed");
 
-        if (g_pXbox_RenderTarget == g_pXbox_BackBufferSurface) {
-            FreeHostResource(GetHostResourceKey(g_pXbox_RenderTarget)); g_pD3DDevice->SetRenderTarget(0, GetHostSurface(g_pXbox_RenderTarget, D3DUSAGE_RENDERTARGET));
-            FreeHostResource(GetHostResourceKey(g_pXbox_DepthStencil)); g_pD3DDevice->SetDepthStencilSurface(GetHostSurface(g_pXbox_DepthStencil, D3DUSAGE_DEPTHSTENCIL));
-        }
+        FreeHostResource(GetHostResourceKey(g_pXbox_RenderTarget)); g_pD3DDevice->SetRenderTarget(0, GetHostSurface(g_pXbox_RenderTarget, D3DUSAGE_RENDERTARGET));
+        FreeHostResource(GetHostResourceKey(g_pXbox_DepthStencil)); g_pD3DDevice->SetDepthStencilSurface(GetHostSurface(g_pXbox_DepthStencil, D3DUSAGE_DEPTHSTENCIL));
     }
 }
```
NZJenkins, line 4129: Because non-backbuffer rendertargets aren't upscaled, the host rendertarget would never be the size of (xboxrendertarget * scale)
In that case we'd incorrectly hit the test case when upscaling was enabled

Titles from https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1658 are likely to be affected by the same regression in JSRF that this commit addresses

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4857,10 +4862,10 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_Clear)
 		GetMultiSampleScaleRaw(aaX, aaY);
         std::vector<D3DRECT> rects(Count);
         for (DWORD i = 0; i < Count; i++) {
-            rects[i].x1 = static_cast<LONG>(pRects[i].x1 * aaX * g_RenderScaleFactor);
-            rects[i].x2 = static_cast<LONG>(pRects[i].x2 * aaX * g_RenderScaleFactor);
-            rects[i].y1 = static_cast<LONG>(pRects[i].y1 * aaY * g_RenderScaleFactor);
-            rects[i].y2 = static_cast<LONG>(pRects[i].y2 * aaY * g_RenderScaleFactor);
+            rects[i].x1 = static_cast<LONG>(pRects[i].x1 * aaX * GetRenderTargetUpscaleFactor());
```
PatrickvL, line 4865: Perhaps it's faster too perfect the upscale factor before this loop, just like you did elsewhere in this commit?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: This remark deserves to be put in a code comment, as it's not immediately obvious at first sight!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Actually, perhaps the render target specific upscale factor could be put into it's own separate global variable, and only be updated in CbxbImpl_SetRenderTarget

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie: I can confirm following titles no longer show the testcase...
* WWE RAW
  * No change.
* Jet Set Radio Future
  * Improved, both character selection and boost dash no longer make the view elsewhere.
* Grand Theft Auto: San Andreas
  * No change.

However, table has rotate. I found a title produce the testcase.
* Halo

NZJenkins: @PatrickvL 
A post to the issue should do it - it's more of an issue tracking thing than anything to do with the code.

@RadWolfie 
I'll see what's Halo's doing

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: Fair, I've updated to fetch once in this spot.
For now not too worried if we re-resolve it - then we don't have to worry about it being dirty if we move stuff around

Ideally this value wouldn't change at all - but we don't upscale rendertargets consistently yet 🤔

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4110,15 +4118,13 @@ void ValidateRenderTargetDimensions(DWORD HostRenderTarget_Width, DWORD HostRend
     // Because of this, we need to validate that the associated host resource still matches the dimensions of the Xbox Render Target
     // If not, we must force them to be re-created
     // TEST CASE: Chihiro Factory Test Program
-    DWORD XboxRenderTarget_Width_Scaled = XboxRenderTarget_Width * g_RenderScaleFactor;
-    DWORD XboxRenderTarget_Height_Scaled = XboxRenderTarget_Height * g_RenderScaleFactor;
+    DWORD XboxRenderTarget_Width_Scaled = XboxRenderTarget_Width * GetRenderTargetUpscaleFactor();
+    DWORD XboxRenderTarget_Height_Scaled = XboxRenderTarget_Height * GetRenderTargetUpscaleFactor();
```
PatrickvL, line 4122: Here too the factor is determined twice.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -3075,6 +3075,14 @@ void GetRenderTargetBaseDimensions(float& x, float& y) {
...
 }
 
+// Get the host upscale factor for the current rendertarget
+// Currently, only the host backbuffer is upscaled
+int GetRenderTargetUpscaleFactor() {
+	return g_pXbox_RenderTarget == g_pXbox_BackBufferSurface
+		? g_RenderUpscaleFactor
+		: 1;
+}
+
```
PatrickvL, line 3085: Given how `g_ZScale` is already set only once in `CxbxImp_SetRenderTarget`, and that function also handles setting `g_pXbox_BackBufferSurface` *and* `g_pXbox_RenderTarget `, it makes sense to remove `GetRenderTargetUpscaleFactor` altogether and introduce a global variable that gets updated in there as well, and used everywhere you currently call `GetRenderTargetUpscaleFactor`

PatrickvL: PS: We have a `extern float g_ZScale; // TMP glue` line somewhere in code that introduces `g_ZScale` but isn't used, so that line could be removed as well while you're at it.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Hope you don't mind, but I fixed it for you ;)

RadWolfie:
FYI: Jet Set Radio Future made a regression after @NZJenkins made a fix when render resolution is 2x+ selected.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie: @PatrickvL, seems your proposal latest change resolve Halo's testcase which no longer popup as of commit 82da26e.

CC: @NZJenkins

PatrickvL: That's ... unexpected... As far as I intended, I didn't change any functionality. Perhaps the rebase onto Luke's surface fixes helped here?

RadWolfie: I think it's because you move GetRenderTargetUpscaleFactor function into CxbxImpl_SetRenderTarget as central update once instead of doing extra check.

RadWolfie: Hmm... I tested by remove newest commits to the point of NZJenkins commit and the previous. I do not get the testcase popup. Even download the older artifact builds from NZJenkins relative to those commits. I cannot reproduce the testcase. I guess it was false positive somehow.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -155,6 +155,8 @@ static std::queue<s_Xbox_Callback>  g_Xbox_CallbackQueue;
 static bool                         g_bHack_DisableHostGPUQueries = false; // TODO : Make configurable
 static IDirect3DQuery              *g_pHostQueryWaitForIdle = nullptr;
 static IDirect3DQuery              *g_pHostQueryCallbackEvent = nullptr;
+static int                          g_RenderUpscaleFactor = 1;
+static int                          g_RenderTargetUpscaleFactor = 1;
```
NZJenkins, line 159: nit: these vars have just about the same name, we should add a comment to disambiguate

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -4862,14 +4858,15 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_Clear)
         // Scale the fill based on our scale factor and MSAA scale
 		float aaX, aaY;
 		GetMultiSampleScaleRaw(aaX, aaY);
-		auto rtUpscale = GetRenderTargetUpscaleFactor();
+		aaX *= g_RenderTargetUpscaleFactor;
+		aaY *= g_RenderTargetUpscaleFactor;
```
NZJenkins, line 4862: nit: `aaScale` is a misleading. Maybe `scale` will do

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: I agree. Perhaps also rename the first into simply `g_ScaleFactor` so that the other one is clearly solely meant for RenderTarget.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7131,13 +7135,16 @@ void CxbxUpdateHostViewport() {
 		LOG_TEST_CASE("Could not get rendertarget dimensions while setting the viewport");
 	}
 
+	aaScaleX *= g_RenderTargetUpscaleFactor;
+	aaScaleY *= g_RenderTargetUpscaleFactor;
```
PatrickvL, line 7139: These too could be renamed into `scaleX` (and Y)

PatrickvL: Perhaps even XScale (so dimension in front), in line with existing `g_ZScale`?

LukeUsher (approved):

PR #2045, opened 2020-11-25
Palette fixes

CookiePLMonster:
This PR consists of 3 parts:

* The first commit fixes a stupid copypaste error in `D3DDevice_SetRenderTarget_0`
* Fixes `D3DDevice_SetPalette` and `D3DDevice_SetPalette_4` not calling to guest code - because palettes are regular D3D resources, not calling to guest potentially resulted in broken reference counting and leaking of guest palette resources. This should fix guest memory leaks in games using palettes.
* In **DRIV3R** I observed a case where the game had a palettized texture bound, but did not bind a palette. This caused a crash on "resolving" palettes, so I added a test case for such occurrences and fill the texture with zeroes - without a palette it is not possible to guess what colors should have been used (AFAIK), so filling with zeroes is the safest choice.


PatrickvL (approved):

gandalfthewhite19890404:
I should check WWF RAW, it was crashed due nulled palette.

PR #2047, opened 2020-11-26
This restores two slight differences I made to Luke's original code;

PatrickvL:
* Only create depth stencil when render target flag is not set
* call GetContainer on IDirect3DTexture (not IDirect3DBaseTexture)

Both my previous (code-review) changes may have caused some of the artifacts that were reported after merging, so here a fix if so necessary.

(Reject if not required)

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -5691,7 +5691,8 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 		// Create the surface/volume/(volume/cube/)texture
 		switch (XboxResourceType) {
 		case xbox::X_D3DRTYPE_SURFACE: {
-			if (D3DUsage & D3DUSAGE_DEPTHSTENCIL) {
+			// Note : Create a depth stencil only when D3DUSAGE_DEPTHSTENCIL is set and D3DUSAGE_RENDERTARGET is not!
+			if (D3DUsage & (D3DUSAGE_RENDERTARGET | D3DUSAGE_DEPTHSTENCIL) == D3DUSAGE_DEPTHSTENCIL) {
 				hRet = g_pD3DDevice->CreateDepthStencilSurface(dwWidth * g_RenderScaleFactor, dwHeight * g_RenderScaleFactor, PCFormat,
 					g_EmuCDPD.HostPresentationParameters.MultiSampleType,
 					0, // MultisampleQuality
@@ -6968,8 +6969,8 @@ IDirect3DBaseTexture* CxbxConvertXboxSurfaceToHostTexture(xbox::X_D3DBaseTexture
 		return nullptr;
 	}
 
-	IDirect3DBaseTexture* pNewHostBaseTexture = nullptr;
-	auto hRet = pHostSurface->GetContainer(IID_PPV_ARGS(&pNewHostBaseTexture));
+	IDirect3DTexture* pNewHostTexture = nullptr;
+	auto hRet = pHostSurface->GetContainer(IID_PPV_ARGS(&pNewHostTexture));
     DEBUG_D3DRESULT(hRet, "pHostSurface->GetContainer");
 
 	if (FAILED(hRet)) {
@@ -6977,7 +6978,7 @@ IDirect3DBaseTexture* CxbxConvertXboxSurfaceToHostTexture(xbox::X_D3DBaseTexture
 		return nullptr;
 	}
 
-	return pNewHostBaseTexture;
+	return static_cast<IDirect3DBaseTexture*>(pNewHostTexture);
 }
 
 void CxbxUpdateHostTextures()


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5691,7 +5691,8 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 		// Create the surface/volume/(volume/cube/)texture
 		switch (XboxResourceType) {
 		case xbox::X_D3DRTYPE_SURFACE: {
-			if (D3DUsage & D3DUSAGE_DEPTHSTENCIL) {
+			// Note : Create a depth stencil only when D3DUSAGE_DEPTHSTENCIL is set and D3DUSAGE_RENDERTARGET is not!
+			if (D3DUsage & (D3DUSAGE_RENDERTARGET | D3DUSAGE_DEPTHSTENCIL) == D3DUSAGE_DEPTHSTENCIL) {
```
RadWolfie, line 5695: Huh? This only perform a check for `D3DUSAGE_DEPTHSTENCIL`. Are you sure this is intended change?

PatrickvL: Yes. Previously, Luke had a code block for D3DUSAGE_RENDERTARGET which was identical to the `else` block.
So, I folded it into one, but doing that, the D3DUSAGE_DEPTHSTENCIL case now took precedence.
Therefor, I now skip the D3DUSAGE_DEPTHSTENCIL code block when D3DUSAGE_RENDERTARGET is set as well,
because D3DUSAGE_RENDERTARGET really has to be handled separate from D3DUSAGE_DEPTHSTENCIL.

(Without this new check, cases where both D3DUSAGE_RENDERTARGET  and D3DUSAGE_DEPTHSTENCIL are set, would be creating a depth stencil, instead of a texture, like Luke did originally.)

RadWolfie: Ah, I see your point.

PatrickvL: This also means this PR shouldn't be deleted right away, because there can be cases where the above described scenario would come into play, and this would be the fix.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6968,16 +6969,16 @@ IDirect3DBaseTexture* CxbxConvertXboxSurfaceToHostTexture(xbox::X_D3DBaseTexture
 		return nullptr;
 	}
 
-	IDirect3DBaseTexture* pNewHostBaseTexture = nullptr;
-	auto hRet = pHostSurface->GetContainer(IID_PPV_ARGS(&pNewHostBaseTexture));
+	IDirect3DTexture* pNewHostTexture = nullptr;
```
RadWolfie, line 6972: I'm not sure if this change is necessary since it's @CookiePLMonster proposal change. 🤔

medievil1:
breaks crimson skies 
image

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6968,16 +6969,16 @@ IDirect3DBaseTexture* CxbxConvertXboxSurfaceToHostTexture(xbox::X_D3DBaseTexture
...
+	auto hRet = pHostSurface->GetContainer(IID_PPV_ARGS(&pNewHostTexture));
     DEBUG_D3DRESULT(hRet, "pHostSurface->GetContainer");
 
 	if (FAILED(hRet)) {
 		LOG_TEST_CASE("Failed to get Texture from Surface");
 		return nullptr;
 	}
 
-	return pNewHostBaseTexture;
+	return static_cast<IDirect3DTexture*>(pNewHostTexture);
```
PatrickvL, line 6981: Duh. I intended to cast to a base texture here! Better get some sleep first...

medievil1: sleep stupor...lol.. but wasn't it Einstein who hardly ever slept so stayed in a stupor and was considered a genius??..lol

Margen67:
> 
> 
> breaks crimson skies
> image

This PR actually breaks everything.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: This change was indeed safe, for two reasons:
1. Technically, `IDirect3DTexture*` is a child of `IDirect3DBaseTexture*` so as per COM rules this is impossible to fail.
2. Practically, I've been running locally with this change way before Luke PR'd it and it was all fine.

PatrickvL: Still, there's reports of failed `CreateImageSurface` with this (and the above) change (see https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2047#issuecomment-734008646) 

I already updated this commit to return a `IDirect3DBaseTexture`, as that's the stated return type. But that shouldn't really change anything.

I'm stumped...

medievil1:
still the same failed to create error....

PatrickvL:
> still the same failed to create error....

Actually, the `CreateImageSurface` error should be preceded by a `g_pD3DDevice.CreateDepthStencilSurface` or `g_pD3DDevice.CreateTexture` D3D8 DEBUG log message, and a few lines after that as well - could you paste that part of the log here as well?

medievil1:
> 
> 
> > still the same failed to create error....
> 
> Actually, the `CreateImageSurface` error should be preceded by a `g_pD3DDevice.CreateDepthStencilSurface` or `g_pD3DDevice.CreateTexture` D3D8 DEBUG log message, and a few lines after that as well - could you paste that part of the log here as well?

it didn't give much info...

> [0x48C8] DEBUG: D3D8    CreateHostResource : Successfully created Surface (0x00212604, 0x14E66620)
[0x48C8] DEBUG: D3D8    CreateHostResource : Width : 1280, Height : 480, Format : 22
[0x48C8] D3D8    CreateHostResource(
   pResource            : 0x0021264C -> X_D3DResource* {
   .Common               : 10D0002 :
    RefCount             : 2
    Type                 : (X_D3DCOMMON_TYPE)0x00050000 = X_D3DCOMMON_TYPE_SURFACE
    IntRefCount          : 1
    ?D3DCreated          : 1
   .Data                 : 720000
   .Lock                 : 0}
   iTextureStage        : FFFFFFFF
   dwSize               : 0
   ResourceTypeName     : Surface
);
[0x48C8] D3D8    CreateHostResource : g_pD3DDevice->CreateTexture D3D error (0x8876086C: D3DERR_INVALIDCALL: Invalid Call)
[0x48C8] WARN : D3D8    CreateDepthStencilSurface Failed

Error: D3DERR_INVALIDCALL
Desc: Invalid call
[0x48C8] WARN : D3D8    Trying Fallback
[0x48C8] FATAL: D3D8    Popup : Received Fatal Message:

* CreateImageSurface Failed!

Error: D3DERR_INVALIDCALL
Desc: Invalid call

[0x48C8] INFO : INIT    MAIN: Terminating Process

PatrickvL:
It seems this isn't needed after all. Closing.

PR #2048, opened 2020-11-26
Optimize vertex stream cache for partial buffer access

CookiePLMonster:
**IMPORTANT NOTE: This PR has potentially huge implications. I only tested it with a handful of games and spotted no regressions, only performance improvements. These changes WILL affect most of the games and on paper they can all see performance improvements, although some of the improvements may not be measurable. Test with ALL your games if possible.**

Drastically reduces the amount of copied and converted data in cases where any of the following apply:
* `BaseVertexIndex` is not 0
* `LowIndex` is not 0
* `dwStartVertex` is not 0

Also potentially fixes an issue where data past an indexed vertex buffer was hashed and converted, leading to potential false positives on dirty checks, maybe more.

Improves in-race performance in **Group S Challenge** by over 10 times (3 FPS -> ~40 FPS), and makes the emulator not run out of memory after a few seconds into the race. The same is still too slow and buggy to be playable.
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -6864,7 +6864,7 @@ void CxbxDrawIndexed(CxbxDrawContext &DrawContext)
 	// for an explanation on the function of the BaseVertexIndex, MinVertexIndex, NumVertices and StartIndex arguments.
 	HRESULT hRet = g_pD3DDevice->DrawIndexedPrimitive(
 		/* PrimitiveType = */EmuXB2PC_D3DPrimitiveType(DrawContext.XboxPrimitiveType),
-		BaseVertexIndex,
+		/* BaseVertexIndex, = */-CacheEntry.LowIndex, // Base vertex index has been accounted for in the stream conversion, now we need to "un-offset" the index buffer
 		/* MinVertexIndex = */CacheEntry.LowIndex,
 		/* NumVertices = */(CacheEntry.HighIndex - CacheEntry.LowIndex) + 1,
 		/* startIndex = DrawContext.dwStartVertex = */0,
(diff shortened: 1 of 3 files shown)


PatrickvL:
Amazing how one single negative index applied at just the right location can avoid so much overhead and complexity! Well done!

PatrickvL (approved):
Even most comments have been kept and the changes are actually very minimal. Loving it!

LukeUsher:
Decent performance increase across the board in my testing, great job!

Margen67:
The minimum fps in Test Drive Off-Road: Wide Open during gameplay has increased from <1 to 6.
Blinx now goes ingame:
image

CookiePLMonster:
Pushed a minor fix to whitespace, no code changes.

escape209:
WWF RAW is now playable
image

Margen67:
Reduces the length of stutters in Battle for Bikini Bottom, especially during pausing/unpausing.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7553,8 +7550,8 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_DrawVertices)
 				LOG_TEST_CASE("X_D3DPT_LINELOOP"); // TODO : Text-cases needed
 
 				assert(DrawContext.dwBaseVertexIndex == 0); // if this fails, it needs to be added to LowIndex and HighIndex :
-				INDEX16 LowIndex = (INDEX16)DrawContext.dwStartVertex;
-				INDEX16 HighIndex = (INDEX16)(DrawContext.dwStartVertex + DrawContext.dwHostPrimitiveCount);
+				INDEX16 LowIndex = 0;
```
PatrickvL, line 7553: Did you test this change by forcing X_D3DPT_LINELOOP? (There's no known test cases for that, so must be faked)

CookiePLMonster: I don't know how the correct input data for this looks, so I just mirrored the other changes (i.e. removing the `startVertex` offset).

PatrickvL: Yeah, it does look okay. Maybe one of the XDK samples, like the stencil test, draws lines loops (or just  lines, which could be hacked to draw line loops). If that still looks okay, it'll be fine

LanHikariDS:
No noticeable regressions from my library, nothing but improvements

LukeUsher (approved):

Issue #1900, opened 2020-05-11
Pop-up "Applying 3925 fog disable hack"


gandalfthewhite19890404:
LOG_TEST_CASE: Applying 3925 fog disable hack
In ApplyDeferredRenderState (D:\a\Cxbx-Reloaded\Cxbx-Reloaded\src\core\hle\D3D8\Direct3D9\RenderStates.cpp line 337)

Titles:

- Airforce Delta Storm

doug-clayton:
This also appears for Dead or Alive 3.

Issue #1685, opened 2019-07-29
Investigate and remove 3925 fog hack


PatrickvL:
Since [this commit](https://github.com/LukeUsher/Cxbx-Reloaded/commit/b62b8341cc2965c2069c197c6c0060ecff8cf2ac), fog is disabled for 3925 XDK titles (like Halo and Silent Hill 2).

For this issue:
* research why fog causes a problem when the hack is disabled
* document the findings
* determine if a fix can be made that would avoid the hack
* implement possible improvements
* test!

RadWolfie:
Popup test case relative to this ticket: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1900

jeltaqq:
Halo Japanese version was built by XDK 3948. Have a same fog problem.

PR #2052, opened 2020-11-28
remove 3925 fog/alpha hacks

LukeUsher:
This PR removes the Fog and Alpha disable hacks for 3925 XDK.

We don't yet understand why these hacks were needed, and while they have been present, a solution has not been investigated.

This *will* cause apparent regressions in some titles using this XDK (Silent Hill 2, perhaps Halo), however, while this hack is present, we cannot understand and fix the proper behavior.

There's also a chance that there are some 3925 games this hack hinders, rather than helps. 

I don't think it's productive to have such a massive hack in the codebase.

ergo720 EDIT:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1900 because it removes the corresponding test case.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/RenderStates.cpp
+++ b/src/core/hle/D3D8/Direct3D9/RenderStates.cpp
@@ -278,14 +278,6 @@ void XboxRenderStateConverter::ApplySimpleRenderState(uint32_t State, uint32_t V
             Value = EmuXB2PC_D3DSTENCILOP(Value);
             break;
         case xbox::X_D3DRS_ALPHATESTENABLE:
-            if (g_LibVersion_D3D8 == 3925) {
-                // HACK: Many 3925 have missing polygons when this is true
-                // Until  we find out the true underlying cause, and carry on
-                // Test Cases: Halo, Silent Hill 2.
-                LOG_TEST_CASE("Applying 3925 alpha test disable hack");
-                Value = false;
-            }
-            break;
         case xbox::X_D3DRS_ALPHABLENDENABLE:
         case xbox::X_D3DRS_BLENDCOLOR:
         case xbox::X_D3DRS_ALPHAREF: case xbox::X_D3DRS_ZWRITEENABLE:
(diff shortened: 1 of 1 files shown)


PatrickvL (commented):
So is there an issue that describes the problem(s)? Otherwise this removal of all relevent details would mean someone, somewhere, at some point in the future, would have to discover this all over again

LukeUsher:
> So is there an issue that describes the problem(s)? Otherwise this removal of all relevent details would mean someone, somewhere, at some point in the future, would have to discover this all over again

The problem is we don’t know or understand what the problems are. Disabling fog and alpha makes some missing elements visible, but that’s clearly a nasty hack and may have no direct relation to the underlying problem 

It could be incorrect Xdk specific tables, it could be because the games use functionality we don’t emulate yet (for example, alphakill) or it could be pixel or vertex shader problems. 

We know to little to create a sufficient issue and I believe having this hack in place has prevented investigation into the underlying causes.

PatrickvL:
> The problem is we don’t know or understand what the problems are. Disabling fog and alpha makes some missing elements visible, but that’s clearly a nasty hack and may have no direct relation to the underlying problem
> 
> It could be incorrect Xdk specific tables, it could be because the games use functionality we don’t emulate yet (for example, alphakill) or it could be pixel or vertex shader problems.
> 
> We know too little to create a sufficient issue and I believe having this hack in place has prevented investigation into the underlying causes.

Oh, I agree about all of this. It's just that if no issue is created for this now, we might forget, and newcomers won't have anything to go on, so a lot of work might get wasted, while now there's the opportunity to preserve the current knowledge.

ergo720:
@PatrickvL There is already this issue which tracks this problem https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1685

ergo720 (approved):

PatrickvL:
Lol, forgot all about it! Thanks for looking it up and linking it here. Cheers!

PR #2053, opened 2020-11-28
rework texturestate mapping to be more sane + fix unmapped get

LukeUsher:
This reworks the TextureState mapping table indexing to be a little saner/easier to grasp what's happening.

It also fixes an oversight where TextureState reads were not taking mapping into account.


LukeUsher:
I noticed some weird artefacts ended up in the file. I’ll fix it after dinner

PatrickvL (approved):
Much better, thanks!

PR #2055, opened 2020-11-29
Preserve original IO status block and context for ApcRoutine

CookiePLMonster:
Fixes cases where guest's APC routine/completion callback does not receive the original IO status block variable, but it assumed it did (it's allowed as per WinAPI docs).

In the MSDN docs for [ReadFileEx](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfileex) it is mentioned that "user data" may be stashed in the `OVERLAPPED` structure:

> The ReadFileEx function ignores the OVERLAPPED structure's hEvent member. An application is free to use that member for its own purposes in the context of a ReadFileEx call. ReadFileEx signals completion of its read operation by calling, or queuing a call to, the completion routine pointed to by lpCompletionRoutine, so it does not need an event handle.

For reasons I can't currently explain, `NtReadFile` does not always respect this and the `ApcContext` returned back to the game isn't always the same. I can only imagine that since `NtReadFile` and the APC fields in particular are **undocumented**, we are hitting some non-contractual behaviour which would not be visible if `ReadFileEx` was used directly (I poked around in the debugger and our "original" pointer was stashed 16 bytes after the one our APC routine received, so **I think** it was a stack of some worker thread).

Either way, my fix should be safe regardless of what Windows truly is doing under the hood.

Fixes **Project Gotham Racing** deadlocking after intro movies unless `silence.wma` was removed. I think it also makes it load main menu music properly, but I did not verify that for certain.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlNt.cpp
+++ b/src/core/kernel/exports/EmuKrnlNt.cpp
@@ -1648,10 +1648,17 @@ XBSYSAPI EXPORTNUM(219) xbox::ntstatus_xt NTAPI xbox::NtReadFile
 		CxbxDebugger::ReportFileRead(FileHandle, Length, Offset);
 	}
 
+	if (ApcRoutine != nullptr) {
+		// Pack the original parameters to a wrapped context for a custom APC routine
+		CxbxIoDispatcherContext* cxbxContext = new CxbxIoDispatcherContext(IoStatusBlock, ApcRoutine, ApcContext);
+		ApcRoutine = CxbxIoApcDispatcher;
+		ApcContext = cxbxContext;
+	}
+
 	NTSTATUS ret = NtDll::NtReadFile(
 		FileHandle,
 		Event,
-		(PVOID)ApcRoutine,
+		ApcRoutine,
 		ApcContext,
 		IoStatusBlock,
 		Buffer,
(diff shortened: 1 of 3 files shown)


src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -2043,8 +2050,8 @@ XBSYSAPI EXPORTNUM(232) xbox::void_xt NTAPI xbox::NtUserIoApcDispatcher
 		dwErrorCode = RtlNtStatusToDosError(IoStatusBlock->Status);
 	}
 
-	LPOVERLAPPED_COMPLETION_ROUTINE CompletionRoutine = (LPOVERLAPPED_COMPLETION_ROUTINE)ApcContext;
-	LPOVERLAPPED lpOverlapped = (LPOVERLAPPED)CONTAINING_RECORD(IoStatusBlock, OVERLAPPED, Internal);
+	LPOVERLAPPED_COMPLETION_ROUTINE CompletionRoutine = reinterpret_cast<LPOVERLAPPED_COMPLETION_ROUTINE>(ApcContext);
+	LPOVERLAPPED lpOverlapped = reinterpret_cast<LPOVERLAPPED>(IoStatusBlock);
```
PatrickvL, line 2054: Why did you change this line, putting another value into the variable?

PatrickvL (commented):
Nice work. Just one question

src/core/kernel/exports/EmuKrnlNt.cpp
CookiePLMonster: I don't think the original line was correct conceptually. `IoStatusBlock` **is** `OVERLAPPED` here, it doesn't "contain" it.

LukeUsher (approved):

PR #2059, opened 2020-12-04
Refactor D3D device creation

CookiePLMonster:
This PR lays the foundation for migration to Direct3D9Ex by cleaning up and improving the correctness of D3D device creation:
* "Synchronization" via `volatile bool` variables has been completely removed - either by making it useless or by replacing with an event.
* `EmuCreateDeviceProxy` has been removed - now the entire process of device creation is done inline, with the exception of a call to `CreateDevice` - as it must be done on a window message thread (see the link below), it is now sent to an appropriate thread (in a blocking manner) via a newly added `RunOnWndMsgThread` function.

On threading considerations when creating, releasing or resetting D3D devices:
https://docs.microsoft.com/en-us/windows/win32/direct3d9/multithreading-issues

No changes are expected, but since it makes D3D9 usage more correct anything can happen - performance improvements are nearly impossible, but it **may** fix e.g. some lockups on startup. **Keep this open for a few days to have it tested, I'm especially curious about results of testing it on AMD GPUs without DXVK.**

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -87,21 +87,14 @@ bool                                g_bClipCursor  = false; // indicates that th
 IDirect3DDevice                    *g_pD3DDevice   = nullptr; // Direct3D Device
 
 // Static Variable(s)
-static IDirectDrawSurface7         *g_pDDSPrimary  = nullptr; // DirectDraw7 Primary Surface
-static IDirectDrawClipper          *g_pDDClipper   = nullptr; // DirectDraw7 Clipper
-static IDirectDraw7                *g_pDD7          = nullptr; // DirectDraw7
-static HMONITOR                     g_hMonitor      = NULL; // Handle to DirectDraw monitor
-static GUID                         g_ddguid = { 0 }; // DirectDraw driver GUID
-static DDCAPS                       g_DriverCaps = { 0 };
-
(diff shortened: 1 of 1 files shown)


PatrickvL (commented):
Why remove the default assignment to only these global variables, when there's many more globals that could be treated similarly?

Either keep the initial assignments (which would communicate better what happens), or remove them for all global variables (although then this PR would contain unrelated code changes)?

medievil1:
halo v1.09 and v1.01 black screen, just audio fro videos and wrong audio in menu
popup about timestamp (though I think that has been present for a while)
image

CookiePLMonster:
> Why remove the default assignment to only these global variables, when there's many more globals that could be treated similarly?
> 
> Either keep the initial assignments (which would communicate better what happens), or remove them for all global variables (although then this PR would contain unrelated code changes)?

I'd say it's personal preference, for two reasons:
1. In this context, I'd rather communicate that there is "no default value" for those arrays (we know that on the binary level it's not true, but that's not my point), as we fill it with valid data on device creation.
2. `array[x] = { false };` creates a false impression that this `false` applies to each element of the array, it doesn't! Of course, in the case of `0`/`false` it doesn't matter, but it was `true` then it would.

I personally think most of those static variables should have their default assignments removed if not needed, but in this PR I only touched the init code for these particular format arrays.

PatrickvL (approved):

LukeUsher (approved):
Works fine for me, however I didn't notice any changes, AMD is still slow without dxvk

Issue #1718, opened 2019-09-08
NtDuplicateObject creates invalid handles for NtReadFile


ergo720:
Test cases:
* Project Zero 1
* Project Zero 2

Here is my original message on the issue for Project Zero 2:
>I understood the problem: it's an issue with ```NtDuplicateObject```. The offending handle was first created with a call to ```NtCreateFile``` for the file "bin\\data\\tim\\font.pk2", which does exist in the game files and the function succeeds. Then it calls ```NtDuplicateObject``` on this handle, which succeeds as well and creates a duplicate of it (note that ```IsEmuHandle``` returns false for this call). After that it calls ```ObReferenceObjectByHandle``` on the duplicate and this one fails with ```STATUS_INVALID_HANDLE```, which is expected as our ```NtDuplicateObject``` simply forwards to its Windows counterpart ```Ntdll::NtDuplicateObject```, and thus our kernel Ob functions will never know about the duplicate. Finally, it calls ```NtReadFile``` on the duplicate (which simply forwards to ```Ntdll::NtReadFile```), which forever fails and hence the hang. So, it seems that the duplicate handle created by ```Ntdll::NtDuplicateObject``` is rejected by ```Ntdll::NtReadFile``` for some reason.

And here are the relevant kernel function calls extracted form the kernel log:
```
[0x0950] NT      NtCreateFile forwarding to "IoCreateFile"...
[0x0950] IO      IoCreateFile(
 OUT FileHandle         : 0x0B32FDFC
   DesiredAccess        : 80100080
   ObjectAttributes     : 0x0B32FDE0 -> OBJECT_ATTRIBUTES* {
   .RootDirectory        : FFFFFFFD
   .ObjectName           : 0x0B32FDF4 -> STRING* {
   .Length               : 18
   .MaximumLength        : 19
   .Buffer               : (char *)0x00362900 = "D:\\bin\\data\\tim\\font.pk2"}
   .Attributes           : 40}
 OUT IoStatusBlock      : 0x0B32FDEC
   AllocationSize       : 0x00000000
   FileAttributes       : 0
   ShareAccess          : 1
   Disposition          : (CREATE_DISPOSITION)0x00000001 = FILE_OPEN
   CreateOptions        : (CREATE_OPTION)0x00000048 = FILE_NO_INTERMEDIATE_BUFFERING|FILE_NON_DIRECTORY_FILE
   Options              : 0
);
[...]

PR #2060, opened 2020-12-07
Fix incorrect flags in NtDuplicateObject

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1718. The problem was that we were passing an access mask of zero to `NtDuplicateObject`, which meant that the duplicated handle would have no access rights at all, which explains why all subsequent file access operations on it would inevitably fail. With this change, both Project Zero 1 and 2 successfully reach the title screens.

ff1

ff2

### DIFF
--- a/src/core/kernel/exports/EmuKrnlNt.cpp
+++ b/src/core/kernel/exports/EmuKrnlNt.cpp
@@ -702,7 +702,10 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
 	}
 	else
 	{
-		// On the xbox, the duplicated handle always has the same access rights of the source handle
+		// On the xbox, the duplicated handle always has the same access rigths of the source handle
+		const ACCESS_MASK DesiredAccess = 0;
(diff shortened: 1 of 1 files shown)


src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -702,19 +702,17 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
 	}
 	else
 	{
-		// TODO : What arguments should we use?
-		const ACCESS_MASK DesiredAccess = 0;
-		const ULONG Attributes = 0;
+		// On the xbox, the duplicated handle always has the same access rights of the source handle
```
PatrickvL, line 705: This explains a lot, but instead of removing the consts, you could restrict to only removing the TODO, put this comment above them, and put the update of Options under them. That way, the call doesn't need updating.

LukeUsher (approved):

PatrickvL:
rigthly done ;)

PR #2067, opened 2020-12-09
Fix Audio Overall Volume Not Included From Input Sets of Volume

RadWolfie:
Since @CookiePLMonster notice an issue with The Warrior's background music being missing from main menu. After debugging and found the cause of the issue. Certain titles do input specific mixbins' volume, however we exclude the other volumes for maximum volume result. With this fix, The Warrior's background music is back to audible.

fixes #1993 regression

---

Also add X_DSVOICEPROPS to log output for future debugging.


src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
```diff
@@ -1298,26 +1298,28 @@ static inline HRESULT HybridDirectSoundBuffer_SetMixBinVolumes_8(
...
+        }
 
-            // Since we cannot set per-channel volumes, we want to pick "dominant" volume
-            if (it_in.dwMixBin != XDSMIXBIN_LOW_FREQUENCY && it_in.dwMixBin < XDSMIXBIN_SPEAKERS_MAX) {
-                if (it_in.lVolume > maxVolume) {
-                    maxVolume = it_in.lVolume;
+        // Since we cannot set per-channel volumes, we want to pick "dominant" volume
+        LONG maxVolume = DSBVOLUME_MIN;
+        for(unsigned i = 0; i < Xb_VoiceProperties.dwMixBinCount; i++) {
+            auto it_out = Xb_VoiceProperties.MixBinVolumePairs[i];
```
CookiePLMonster, line 1319:
```suggestion
            auto& it_out = Xb_VoiceProperties.MixBinVolumePairs[i];
```

CookiePLMonster (changes requested):
Only one nitpick, will approve after that since it looks good

src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
RadWolfie: I can add `const`, as for ampersand... it will require change others to a pointer. Is that's part of your nitpick too?

CookiePLMonster: Why pointer? It's only about not copying the contents of `Xb_VoiceProperties.MixBinVolumePairs[i]` to the stack.

CookiePLMonster (dismissed):

src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
```diff
@@ -1298,26 +1298,28 @@ static inline HRESULT HybridDirectSoundBuffer_SetMixBinVolumes_8(
...
-            // Since we cannot set per-channel volumes, we want to pick "dominant" volume
-            if (it_in.dwMixBin != XDSMIXBIN_LOW_FREQUENCY && it_in.dwMixBin < XDSMIXBIN_SPEAKERS_MAX) {
-                if (it_in.lVolume > maxVolume) {
-                    maxVolume = it_in.lVolume;
+        // Since we cannot set per-channel volumes, we want to pick "dominant" volume
+        LONG maxVolume = DSBVOLUME_MIN;
+        for(unsigned i = 0; i < Xb_VoiceProperties.dwMixBinCount; i++) {
+            const auto& it_out = Xb_VoiceProperties.MixBinVolumePairs[i];
+            if (it_out.dwMixBin != XDSMIXBIN_LOW_FREQUENCY && it_out.dwMixBin < XDSMIXBIN_SPEAKERS_MAX) {
+                if (it_out.lVolume > maxVolume) {
```
PatrickvL, line 1321: IMHO `if (maxVolume < it_out.lVolume)` would be nicer, since that would put the symbols in exactly the same order as in the assignment line below.

RadWolfie: True, also... eventually, inline functions will be slowly phase out into relative class in a similar concept of how XDK did in child classes. Including this block of code into its own inline function maybe.

Although, I will update it in this pull request anyway. 😉

RadWolfie: Done!

PatrickvL (approved):

CookiePLMonster (approved):

PR #1196, opened 2018-05-22
Add controller port configuration GUI and support virtual SB controller with feedback status

jackchentwkh:
Using dialog to let user assign which host controller he want to assign to each xbox port.
add preliminary support for virtual SteelBatallion controller. input mapped to 1st Xinput and directinput, 
key mapping will be provided later.
also add a new dialog via CxbxVSBC.dll for virtual SteelBatallion controller feedback status.
later I will add GUI input support via the same dialog.

besides the virtual SteelBatallion controller, the user configuration for host controller assignment is a major change. 
I also add some new structures together with my new findings for the DeviceTable and DeviceInfo of xbox devices.

**EDIT by RadWolfie:**
Here's the new change to GUI. (Used to have option toggle XInput option.)
image
Only with Steel Batallion title, controller feedback status will open.
9bb0a29ed23adfbc

PR #2068, opened 2020-12-10
WIP: Reimplement Steel Battalion controller

LanHikariDS:
The end-goal should be self-explanatory, cleaning up the old VSBC code and updating it for the new input system
Still a work in progress, as I can't figure out how to set everything properly (InputManager.h being the current issue). Also my first time working in quite a few different areas (Github, Visual Studio, and (to my knowledge) C++), so there are bound a few mistakes and superfluous commits


src/common/input/InputManager.cpp
```diff
@@ -60,7 +60,7 @@ int Gui2XboxPortArray[4] = {
 int dev_num_buttons[to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX)] = {
 	XBOX_CTRL_NUM_BUTTONS, // MS_CONTROLLER_DUKE
 	XBOX_CTRL_NUM_BUTTONS, // MS_CONTROLLER_S
-	0,
+	XBOX_CTRL_NUM_BUTTONS, // STEEL_BATTALION_CONTROLLER This probably isn't correct?? Let's see what happens
```
RadWolfie, line 63: Please move this change down to line 67 (index 6 which is STEEL_BATTALION_CONTROLLER originally located).

src/common/input/InputManager.cpp
```diff
@@ -347,16 +347,16 @@ bool InputDeviceManager::UpdateXboxPortInput(int usb_port, void* Buffer, int Dir
 				switch (xid_type)
 				{
 				case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE):
-                case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S): {
+                case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S):
+				case to_underlying(XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER): {
```
RadWolfie, line 351: I am unsure about this change since maybe there is a way to able dynamic support different virtual controller within `UpdateInputXpad`? If not, it will need a brand new function such as `UpdateInputSBC`.

src/common/input/InputManager.h
```diff
@@ -62,11 +63,25 @@ struct XpadInput {
 	int16_t sThumbRY;
 };
 
+struct SBCInput { //I'm not quite sure how this works, but it should come out to 0x18
```
RadWolfie, line 66: Please move SBCInput down to before SBCOutput structure is at or the other way around for SBCOutput structure move up. We're not sorting base on Input/Output, we're sorting base on controller type listing.

src/devices/usb/XidGamepad.cpp
```diff
@@ -242,7 +274,7 @@ XboxDeviceState* XidGamepad::ClassInitFn()
 		m_pPeripheralFuncStruct->handle_destroy = std::bind(&XidGamepad::UsbXid_HandleDestroy, this);
 		m_pPeripheralFuncStruct->handle_attach  = std::bind(&XidGamepad::UsbXid_Attach, this, _1);
 		m_pPeripheralFuncStruct->product_desc   = dev->ProductDesc.c_str();
-		m_pPeripheralFuncStruct->usb_desc       = &desc_xbox_gamepad;
+		m_pPeripheralFuncStruct->usb_desc       = &desc_SBC_gamepad; //Put split for SBC controller type
```
RadWolfie, line 277: Since this is only for testing, please remember to revert this change.

RadWolfie:
**Friendly reminder for anyone trying out this pull request build(s):**

It is intended for ongoing experiment to add Steel Battalion controller support for specific titles that support it. It is very likely other titles will not work as intended. Please refrain report issues with other titles until pull request has move out of draft status.
Thanks!

src/common/input/InputManager.cpp
ergo720: The correct number to use is documented in EmuShared.h
```
char m_DeviceControlNames[4][XBOX_CTRL_NUM_BUTTONS][30]; // macro should be num of buttons of dev with highest num buttons
```

src/common/input/InputManager.cpp
ergo720: This code change is wrong: `UpdateInputXpad` is specific for xbox duke and S controllers. The xboxdevwiki documents that the steel battalion controller uses totally different buttons, so the function can't work with those and thus this requires a new function specifically for this controller type.

src/common/input/InputManager.h
ergo720: This struct should represent the number of input buttons that the steel battalion controller uses. These are documented in the xboxdevwiki in the section "Controller to Xbox" and you can see that they don't correspond to those used by standard duke and S controller, which is what this struct, as written here, is using. Thus this needs changing as well.

src/common/input/InputManager.h
```diff
@@ -62,11 +63,25 @@ struct XpadInput {
...
+	int16_t sThumbRY;
+};
+
 struct XpadOutput {
 	uint16_t left_actuator_strength;
 	uint16_t right_actuator_strength;
 };
 
+struct SBCOutput { //Likewise, with 0x14
+	uint16_t left_actuator_strength;
```
ergo720, line 81: These should be the output controls of the steel battalion controller, which are documented in the xboxdevwiki in the section "Xbox to Controller".

src/common/input/InputManager.h
LanHikariDS: > This struct should represent the number of input buttons that the steel battalion controller uses. These are documented in the xboxdevwiki in the section "Controller to Xbox" and you can see that they don't correspond to those used by standard duke and S controller, which is what this struct, as written here, is using. Thus this needs changing as well.
I'm aware this is incorrect, I asked about it on Discord and got no response, so I wasn't sure how to define them, hence the comment

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -80,7 +80,8 @@ bool operator==(xbox::PXPP_DEVICE_TYPE XppType, XBOX_INPUT_DEVICE XidType)
 	switch (XidType)
 	{
 	case XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE:
-    case XBOX_INPUT_DEVICE::MS_CONTROLLER_S: {
+    case XBOX_INPUT_DEVICE::MS_CONTROLLER_S:
+    case XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER: {
```
ergo720, line 84: The device type used by the steel battalion controller is different than the one used by xbox gamepads. We are deriving it in `SetupXboxDeviceTypes` at line 227:
```
                case XINPUT_DEVTYPE_STEELBATALION:
                    printf("XDEVICE_TYPE_STEELBATALION)\n");
                    break;
```

So, this required a new global xpp type for this controller type.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -140,7 +140,17 @@ bool ConstructHleInputDevice(int Type, int Port)
     }
 	break;
 	case to_underlying(XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER): {
-		// TODO
+		g_XboxControllerHostBridge[Port].XboxPort = Port;
+        g_XboxControllerHostBridge[Port].XboxType = XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER;
+        g_XboxControllerHostBridge[Port].InState = new XpadInput();
```
ergo720, line 145: `XpadInput` is specific is for xbox controllers, so this should use `SBCInput` instead.

src/devices/usb/XidGamepad.cpp
LanHikariDS: I don't know how to reference the current controller's type in this part of the code

src/devices/usb/XidGamepad.cpp
```diff
@@ -203,6 +203,38 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+		0x0a7b,            // idVendor
+		0xd000,            // idProduct
+		0x0100,            // bcdDevice
+		STR_MANUFACTURER,  // iManufacturer
+		STR_PRODUCT,       // iProduct
+		STR_SERIALNUMBER   // iSerialNumber
+	},
+	&desc_device_SBC_gamepad
+};
+
```
ergo720, line 237: Ni: the USB descriptors are only necessary for LLE USB, you don't need those for HLE XInput so these additions are unnecessary but probably fine to have.

src/devices/usb/XidGamepad.cpp
ergo720: `XidGamepad` is a class which implements an LLE xbox duke controller, so this change is wrong and should be reverted.

src/common/input/InputDevice.h
```diff
@@ -58,11 +58,11 @@ typedef enum class _XBOX_INPUT_DEVICE : int {
 	DEVICE_INVALID = -1,
 	MS_CONTROLLER_DUKE,
 	MS_CONTROLLER_S,
+	STEEL_BATTALION_CONTROLLER,
```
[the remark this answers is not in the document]
LanHikariDS: I didn't see any need to revert it since I'd already changed it, but sure

src/gui/DlgInputConfig.cpp
```diff
@@ -188,6 +188,12 @@ INT_PTR CALLBACK DlgInputConfigProc(HWND hWndDlg, UINT uMsg, WPARAM wParam, LPAR
 				}
 				break;
 
+				case to_underlying(XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER): {
+					DialogBoxParam(GetModuleHandle(nullptr), MAKEINTRESOURCE(IDD_VIRTUAL_SBC_FEEDBACK), hWndDlg, DlgXidControllerConfigProc,
+						(DeviceType << 8) | port);
```
ergo720, line 193: `DlgXidControllerConfigProc` will open the input dialog for an xbox controller, while I imagine you want to open the old input dialog for the steel battalion controller. Thus this requires a new function to handle the old input dialog.

src/devices/usb/XidGamepad.cpp
RadWolfie: Yet it will be great to have SB controller prepped for LLE USB support. 😉 Beside, I do recall Luke need go down full LLE USB in order to emulate Chihiro system. Or isn't using our LLE USB manager for variant of controllers on retail Xbox's end.

ergo720: Except that I just remembered that XidGamepad is specific for duke controllers only, so the descriptors for the steel battalion should really go in its own file which implements the corresponding LLE device. Thus, putting them here feels incorrect to me.

LanHikariDS: > 
> 
> Nit: the USB descriptors are only necessary for LLE USB, you don't need those for HLE XInput so these additions are unnecessary but probably fine to have.

Noted, not quite sure what to do with it, then...

src/gui/DlgInputConfig.cpp
LanHikariDS: But I've already got the old SBC input dialogue opening. Of course, the old dialogue isn't the final solution, as I'll need to redo it for bindings and such, but is that the only issue with this section?

RadWolfie:
@LanHikariDS Would you like me to send you a pull request with new files for you to work on? There may will be more than one pull request to assist you on the right path. Though, I'm not able to assist right now. Some time after 24 hours from now I will.

src/core/hle/XAPI/Xapi.cpp
ergo720: To be honest, I'm not totally sure if the xpp type of the steel battalion controller is different than that of xbox gamepads. If it is the same, then `g_DeviceType_Gamepad` can be reused here. However, even in that case, the current code in `SetupXboxDeviceTypes` simply logs the type but does nothing to it, which means that `g_DeviceType_Gamepad` will stay as nullptr. So, I still suggest the below change if the xpp type is the same:
```
case XINPUT_DEVTYPE_STEELBATALION:
g_DeviceType_Gamepad = deviceTable[i]->XppType;
printf("XDEVICE_TYPE_STEELBATALION)\n");
break;
```

LanHikariDS:
> @LanHikariDS Would you like me to send you a pull request with new files for you to work on? There may will be more than one pull request to assist you on the right path. Though, I'm not able to assist right now. Some time after 24 hours from now I will.

That would probably be helpful, as I'm currentlystruggling with SBCInput and SBCOutput, as XpadInput doessn't match XboxDevWiki, and SBCOutput has some tags on XboxDevWiki that cause issues when compiling

ergo720:
>XpadInput doessn't match XboxDevWiki

It does, look carefully at the offsets and masks reported in the wiki. It says that the first field starts at offset 2, and that's because the first two fields are `bReportId` and `bLength`, which are only relevant for LLE USB and thus are ignored in `XpadInput`. All the following 8 buttons start at the same offset 2 and occupy 2 bytes (because the first button following them starts at offset 4). The masks for these buttons tell you that they are represented by a single bit, and that's because they are digital buttons, that is, they can only be on or off. The following 8 buttons are one byte each and are analog instead, and their intensity values can range between 0 - 255 (0 - 0xFF). Finally, the analog sticks use two bytes each, and their state is represented by a signed 16 bit value (because the wiki says they can have negative values).

LanHikariDS:
> It does, look carefully at the offsets and masks reported in the wiki. It says that the first field starts at offset 2, and that's because the first two fields are `bReportId` and `bLength`, which are only relevant for LLE USB and thus are ignored in `XpadInput`. All the following 8 buttons start at the same offset 2 and occupy 2 bytes (because the first button following them starts at offset 4). The masks for these buttons tell you that they are represented by a single bit, and that's because they are digital buttons, that is, they can only be on or off. The following 8 buttons are one byte each and are analog instead, and their intensity values can range between 0 - 255 (0 - 0xFF). Finally, the analog sticks use two bytes each, and their state is represented by a signed 16 bit value (because the wiki says they can have negative values).

So the actual names (bAnalogButtons[8]) aren't actually important, and don't need to be anything specific?

src/core/hle/XAPI/Xapi.cpp
ObiKKa: Remember the spelling (double T's): "STEELBATTALION"

ergo720:
Using an array is done just for brevity: this way, I don't need to add 8 separate struct members. Also, the SB input struct has already being implemented for you in `Xapi.h` at line 148:
```
typedef struct _X_SBC_GAMEPAD {
    xbox::word_xt    wButtons[3];
    xbox::short_xt   sAimingX;
    xbox::short_xt   sAimingY;
    xbox::short_xt   sRotationLever;//maybe only high byte was used.
    xbox::short_xt   sSightChangeX;
    xbox::short_xt   sSightChangeY;
    xbox::word_xt    wLeftPedal;//maybe only high byte was used.
    xbox::word_xt    wMiddlePedal;//maybe only high byte was used.
    xbox::word_xt    wRightPedal;//maybe only high byte was used.
    xbox::uchar_xt   ucTunerDial;//low nibble, The 9 o'clock postion is 0, and the 6 o'clock position is 12. The blank area between the 6 and 9 o'clock positions is 13, 14, and 15 clockwise.
    xbox::uchar_xt   ucGearLever;//GearLever 1~5 for gear 1~5, 7~13 for gear R,N,1~5, 15 for gear R. we use the continues range from 7~13
}
X_SBC_GAMEPAD, *PX_SBC_GAMEPAD;
```
The function `EmuSBCGetState` in `Xapi.cpp` at line 559 also implements filling this input state from the host controller input state, but please don't just copy it because, like this comment https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1196#issuecomment-390973441 says, it's using hard-coded bindings, while we want to let users set arbitrary bindings. You can check the function `UpdateInputXpad` to see how I did that for xbox controllers.

Regarding the output state of `SBCOutput`: this represents the LED state of certain buttons of the SB controller. However, to my knowledge, XInput, DInput and SDL don't support LED states on buttons (assuming the host controller even has LEDs in the first place), so you can't really implement the SB output state and so you can just ignore it and only implement the input state.

One thing I notice in the `XInputGetState` patch: it will break with the SB device because the `pState` buffer used by it will not be `Gamepad` but a SB specific buffer, and thus the `memcpy` operation below will be incorrect (`xbox::XINPUT_GAMEPAD` is 18 bytes while `xbox::X_SBC_GAMEPAD` is 24 bytes instead). To handle that, I suggest the below code change:
```
memcpy((void *)&pState->Gamepad, g_XboxControllerHostBridge[Port].InState, g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucInputStateSize);
```

src/core/hle/XAPI/Xapi.cpp
ergo720: Ok, I checked with the game and I can confirm that the xpp type for the SB controller is different, so you need a new global variable here.
```
g_DeviceType_SBC = deviceTable[i]->XppType;
```

@ObiKKa That's an existing typo in the code, I merely copied it as is, but yes it needs to be fixed as well.

RadWolfie:
> Regarding the output state of `SBCOutput`: this represents the LED state of certain buttons of the SB controller. However, to my knowledge, XInput, DInput and SDL don't support LED states on buttons (assuming the host controller even has LEDs in the first place), so you can't really implement the SB output state and so you can just ignore it and only implement the input state.

Seems only accessible through USB level base on google search. SDL apparently only has one LED support depending on how it's implement.

LanHikariDS:
> Regarding the output state of `SBCOutput`: this represents the LED state of certain buttons of the SB controller. However, to my knowledge, XInput, DInput and SDL don't support LED states on buttons (assuming the host controller even has LEDs in the first place), so you can't really implement the SB output state and so you can just ignore it and only implement the input state.
Just trying to have it match up so the Feedback size comes out the same, rather than actually work, as according to Jackchen, it needs to be 0x14. As for the Input struct, currently just trying to get the game to _recognize_ the controller before worry about the buttons working and such, but my end goal is indeed for them to be remappable, not just using hardcoded binds

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -80,14 +81,20 @@ bool operator==(xbox::PXPP_DEVICE_TYPE XppType, XBOX_INPUT_DEVICE XidType)
...
-    case XBOX_INPUT_DEVICE::MS_CONTROLLER_S:
-    case XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER: {
+    case XBOX_INPUT_DEVICE::MS_CONTROLLER_S: {
 		if (XppType == g_DeviceType_Gamepad) {
 			return true;
 		}
 	}
 	break;
 
+	case XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER: {
```
RadWolfie, line 91: Great finding! Although, it is currently not connected to `SetupXboxDeviceTypes` function. You should be able to find keyword `XINPUT_DEVTYPE_STEELBATTALION`. Plus possibility replace `break;` to `continue;` for allow iterate whole table.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -61,6 +61,7 @@ PFARPROC1 fnCxbxVSBCOpen;
 //typedef DWORD(*fnCxbxVSBCSetState)(UCHAR *);
 //typedef DWORD(*fnCxbxVSBCGetState)(UCHAR *);
 xbox::PXPP_DEVICE_TYPE g_DeviceType_Gamepad = nullptr;
+xbox::PXPP_DEVICE_TYPE g_DeviceType_SBC = nullptr; //deviceTable[i]->XppType?
```
RadWolfie, line 64: Oh for your questionable comment. Yes, that is correct as I referred to previous comment.

src/core/hle/XAPI/Xapi.cpp
ergo720: No need for `continue`, the whole table will still be iterated even with the `break` inside the `switch`.

RadWolfie: Ah, I didn't see the switch statement. 😅

src/common/input/InputManager.h
```diff
@@ -67,6 +68,74 @@ struct XpadOutput {
 	uint16_t right_actuator_strength;
 };
 
+struct SBCInput1 { //I'm not quite sure how this works, but it should come out to 0x18
```
ergo720, line 71: This struct is wrong and should just be deleted.

src/common/input/InputManager.h
```diff
@@ -67,6 +68,74 @@ struct XpadOutput {
...
+	uint16_t wButtons;
+	uint8_t bAnalogButtons[8];
+	int16_t sThumbLX;
+	int16_t sThumbLY;
+	int16_t sThumbRX;
+	int16_t sThumbRY;
+};
+
+typedef struct SBCInput { //Was told not to copy this, did it anyways for the sake of having the proper input size and potentially modifying it's bindings from there.
```
ergo720, line 80: No need for `typedef` here, you can just write `struct SBCInput` + you can remove the `X_SBC_GAMEPAD, * PX_SBC_GAMEPAD;` below. Also, this is the input struct as used by the input system. As such it should not use the xbox types, but just the host types instead. word_xt -> uint16_t, short_xt -> int16_t and uchar_xt -> uint8_t.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -140,7 +147,17 @@ bool ConstructHleInputDevice(int Type, int Port)
...
+		g_XboxControllerHostBridge[Port].XboxPort = Port;
+        g_XboxControllerHostBridge[Port].XboxType = XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER;
+        g_XboxControllerHostBridge[Port].InState = new SBCInput();
+        g_XboxControllerHostBridge[Port].bPendingRemoval = false;
+        g_XboxControllerHostBridge[Port].bSignaled = false;
+        g_XboxControllerHostBridge[Port].bIoInProgress = false;
+        g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucType = XINPUT_DEVTYPE_STEELBATTALION;
+        g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucSubType = XINPUT_DEVSUBTYPE_GC_GAMEPAD_ALT;
+        g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucInputStateSize = sizeof(SBCInput);
+        g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucFeedbackSize = sizeof(XpadOutput);
```
ergo720, line 159: This should be 0x14 instead of taking the size of the output xbox controller struct.

src/common/input/InputManager.h
```diff
@@ -67,6 +68,74 @@ struct XpadOutput {
...
+	xbox::short_xt   sSightChangeY;
+	xbox::word_xt    wLeftPedal;//maybe only high byte was used.
+	xbox::word_xt    wMiddlePedal;//maybe only high byte was used.
+	xbox::word_xt    wRightPedal;//maybe only high byte was used.
+	xbox::uchar_xt   ucTunerDial;//low nibble, The 9 o'clock postion is 0, and the 6 o'clock position is 12. The blank area between the 6 and 9 o'clock positions is 13, 14, and 15 clockwise.
+	xbox::uchar_xt   ucGearLever;//GearLever 1~5 for gear 1~5, 7~13 for gear R,N,1~5, 15 for gear R. we use the continues range from 7~13
+}
+X_SBC_GAMEPAD, * PX_SBC_GAMEPAD;
+
+struct SBCOutput { //Likewise, with 0x14
```
ergo720, line 95: Since we are not going to implement the SB output state, this struct is unnecessary and can be deleted. Unless you want to have it for documentation purposes? But there's the xbox wiki already for that, so I'm not sure.

src/common/input/InputManager.h
LanHikariDS: This was my old attempt at the input struct from scratch, kept because I assumed we wouldn't _actually_ use the copied one in the long run

src/core/hle/XAPI/Xapi.cpp
LanHikariDS: Pretty sure I tried this and it just wouldn't compile, hence why I went through with SBCOutput, but I'll try it again, and resolve your Line 95 comment if it works

src/common/input/InputManager.h
LanHikariDS: If I can get it to compile and work without it, I don't really see it as valuable for documentation, as it's likely incorrect

LanHikariDS:
Why did this close...?

RadWolfie:
I reopened the pull request for you.

RadWolfie:
Github CI fail to compile, see error message below.
> D:\a\Cxbx-Reloaded\Cxbx-Reloaded\src\core\hle\XAPI\Xapi.cpp(92,26): error C2196: case value '_XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER' already used [D:\a\Cxbx-Reloaded\Cxbx-Reloaded\build\projects\cxbxr-emu\cxbxr-emu.vcxproj]

LanHikariDS:
Yeah, and I see why. the SBC definition got moved, by the old one wasn't cleared

src/core/hle/XAPI/Xapi.cpp
LanHikariDS: Yes, but I'm not sure what you meant by 'deviceTable[i]->XppType'

src/core/hle/XAPI/Xapi.cpp
LanHikariDS: It does indeed compile, but the game still doesn't recognize the controller...

src/core/hle/XAPI/Xapi.cpp
LanHikariDS: Holy shit, this did it. This finally fucking did it.

LanHikariDS:
For some odd reason, both Steel Battalion games crash when IDC_SET_LEFT_NEGX is added to the input mapping dialog. I tried correcting this by giving it a SB-specific case instead, but that doesn't work either.

src/common/input/InputManager.cpp
LanHikariDS: Was this corrected by your PR, or is this still an issue?

src/gui/DlgInputConfig.cpp
LanHikariDS: I've been having no issues with the UI, but maybe this was an issue before I'd run into it that's now resolved, or do I still need a new function?

LanHikariDS:
Something got fuckered, current build doesn't start

src/common/input/InputManager.cpp
LanHikariDS: Changing this to SBC_NUM_BUTTONS causes the emulator to crash immediately upon starting up, as seen in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2068/commits/0af6f46ade4e6bb98da3590b3f9f5fbd457b2404

ergo720: Then you'll have to debug and see why it crashes.

src/gui/DlgInputConfig.cpp
ergo720: This now calls `DlgSBControllerConfigProc`, which is the correct thing to do. Before, you were trying to call `DlgXidControllerConfigProc` for the SBC, which was wrong.

src/common/input/InputManager.cpp
ergo720: No, it wasn't and this is still wrong. You need a new function `UpdateInputSBC` specific for the SBC.

src/gui/DlgSBControllerConfig.cpp
```diff
@@ -0,0 +1,176 @@
...
+		int port_num = lParam & 0xFF;
+		int dev_type = (lParam & 0xFF00) >> 8;
+
+		// Ensure that port_num is a valid xbox port
+		assert(port_num >= PORT_1 && port_num <= PORT_4);
+
+		// Ensure that the controller type is valid
+		assert(dev_type == to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE) ||
+			dev_type == to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S) ||
+			dev_type == to_underlying(XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER));
```
ergo720, line 51: `DlgSBControllerConfigProc` should only assert that the type is "SBC" and not "xbox controller".

src/gui/DlgDukeControllerConfig.cpp
```diff
@@ -50,7 +50,8 @@ INT_PTR CALLBACK DlgXidControllerConfigProc(HWND hWndDlg, UINT uMsg, WPARAM wPar
 
 		// Ensure that the controller type is valid
 		assert(dev_type == to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE) ||
-			dev_type == to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S));
+			dev_type == to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_S) ||
+			dev_type == to_underlying(XBOX_INPUT_DEVICE::STEEL_BATTALION_CONTROLLER));
```
ergo720, line 54: `DlgXidControllerConfigProc` should only assert that the type is "xbox controller" and not "SBC", so please revert this change.

src/gui/DlgSBControllerConfig.cpp
LanHikariDS: So lines 50 and 51 should be removed from this file, and like 54 should be removed from DlgDukeControllerConfig.cpp, yes?

RadWolfie: @LanHikariDS, "xbox controller" are default type of controller which are Duke and S controllers. Only thing should be in here is "Steel Battalion Controller". Plus Xbox's Duke controller file at line 54 need to be remove, aka revert, yes.

Duke controller is at line 49. Plus S controller is at line 50.

Asserts are only there to verify the code should be working as intended design in debug build.

LanHikariDS: These have been corrected on my local end, will commit and push later tonight after rebasing

LanHikariDS:
Aaaaand, I forgot to commit _before_ pushing the rebase and new changes, so it closed again

ergo720:
Superseded by https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2134

PR #2072, opened 2020-12-11
Add warning to untested CIs

Margen67:


RadWolfie:
You are aware it will affect people using non-github hosting site? Such as GitLab, GitBucket, etc. Or perhaps any of us decide to move to different hosting site. imho, output a message of deprecated CI ymls aren't worth it. Plus other hosting site can act as a mirror as well.

Margen67:
> 
> 
> You are aware it will affect people using non-github hosting site? Such as GitLab, BitBucket, etc. Or perhaps any of us decide to move to different hosting site. imho, output a message of deprecated CI ymls aren't worth it. Plus other hosting site can act as a mirror as well.

They're not maintained so if we absolutely have to keep them we should at least add a disclaimer.

.azure-pipelines.yml
```diff
@@ -46,6 +46,11 @@ jobs:
         configuration: Debug
 
   steps:
+  - script: |
+      echo "This CI isn't tested against master, and therefore, isn't guaranteed to work. Pull requests are welcome."
+      echo "If it doesn't work and you'd rather not fix it, it's recommended to use GitHub Actions CI instead."
+    displayName: 'Deprecation warning'
```
RadWolfie, line 52: I suggest rename title to "Third-Party CI Warning" or the like.

RadWolfie (changes requested):

RadWolfie (approved):

PR #2078, opened 2020-12-12
HACK: Add DSP hack for force failure to avoid access invalid pointer

RadWolfie:
With this hack, Otogi is able to progress further which is part of hidden bug.

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
@@ -538,7 +538,8 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSound_DownloadEffectsImage)
 
     LOG_NOT_SUPPORTED();
 
-    return S_OK;
+    // HACK: Return as failure workaround to allow titles to progress further.
+    return E_FAIL;
 }
 
 // ******************************************************************
--- a/src/core/hle/DSOUND/DirectSound/XFileMediaObject.cpp
+++ b/src/core/hle/DSOUND/DirectSound/XFileMediaObject.cpp
@@ -117,7 +117,8 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(XAudioDownloadEffectsImage)
 
 	LOG_NOT_SUPPORTED();
 
-    return S_OK;
+    // HACK: Return as failure workaround to allow titles to progress further.
+    return E_FAIL;
 }
 
 // ******************************************************************


CookiePLMonster:
My suggestion would be to return failure only if `ppImageDesc` is non-NULL, because then it means the game might want to read the DSP program data back (e.g. PGR2 does it). In other cases it will probably not care, and so then it may be better to "lie" to the game IMO.

PatrickvL:
Best also mention known test cases in code

medievil1:
halo 2 MS-100 V1.03 boots further with this, plays the first video (though black screen) before hitting the pop up error

LukeUsher (changes requested):
As per @CookiePLMonster's note: Only return E_FAIL when ppImageDesc is set.

When ppImageDesc is set to null, the games will not read back the program, and it's better to allow it to succeed.

RadWolfie:
> As per @CookiePLMonster's note: Only return E_FAIL when ppImageDesc is set.
> 
> When ppImageDesc is set to null, the games will not read back the program, and it's better to allow it to succeed.

Base on our chat, Silent is assuming. I had not heard back about actual result for the difference. And I had not thoroughly check with internal functions to see if other parameters actually do write back to title's input pointers.

@PatrickvL I'll referred to ticket of titles affecting the issue.

LukeUsher:
> > As per @CookiePLMonster's note: Only return E_FAIL when ppImageDesc is set.
> > When ppImageDesc is set to null, the games will not read back the program, and it's better to allow it to succeed.
> 
> Base on our chat, Silent is assuming. I had not heard back about actual result for the difference. And I had not thoroughly check with internal functions to see if other parameters actually do write back to title's input pointers. If other parameters do write back, we need to update them in some way with a successful return. Otherwise, it's pointless to return successful when nothing is set.
> 
> @PatrickvL I'll referred to ticket of titles affecting the issue.

Sure. There’s just a very height chance returning fail is what causes the regression mentioned above

RadWolfie:
> And I have to call it a regression, SW Jedi knight, no audio after the initial video's at all now... before it was just other characters being so low you could not hear them, now it is complete silence...

Yeah, it needs some investigation. Although main purpose of this pull request is to allow titles progress further than trying to access data from non-existence pointer which can affect other contributors investigate unrelative issue to this. DSP isn't my field of expertise.

RadWolfie:
Closing as there's no one willing to test either solution nor have time to reverse engineer internal functions relative to patches in this pull request.

PR #2104, opened 2020-12-15
Move to Direct3D 9Ex

CookiePLMonster:
For a full list of improvements see:
https://docs.microsoft.com/en-us/windows/win32/direct3darticles/direct3d-9ex-improvements

For us the most important aspect is no more device loss, which never existed on Xbox. This allows exclusive fullscreen to work properly,

In the future might also potentially allow for resizing the backbuffer at runtime.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -84,7 +84,7 @@ using namespace std::literals::chrono_literals;
 // Global(s)
 HWND                                g_hEmuWindow   = NULL; // rendering window
 bool                                g_bClipCursor  = false; // indicates that the mouse cursor should be confined inside the rendering window
-IDirect3DDevice                    *g_pD3DDevice   = nullptr; // Direct3D Device
+IDirect3DDevice9Ex                 *g_pD3DDevice   = nullptr; // Direct3D Device
 
 // Static Variable(s)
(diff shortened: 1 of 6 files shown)


PatrickvL (approved):

PatrickvL:
Although the changes look fine, it does break with the convention that we're aliasing away the exact Direct3D version we use. A future port to newer versions would require us the revisit the currently touched symbols again.

Now, obviously that will happen a lot anyway, but there's a chance that we could use the same mechanism to try yet another port to even newer versions, so the aliassing approach might then still prove useful. Hence, perhaps it's wise the keep using that for now?

CookiePLMonster:
I don't see any value in that because D3D9 and D3D11 don't share any API elements at all, unlike D3D8 and D3D9. Back then maybe it made sense, but moving forward it doesn't IMO.

PatrickvL:
Doesn't D3D 12 look a lot like 11? If so, that might be a good enough reason to keep using aliasing?

CookiePLMonster:
No, they look nothing alike, 12 isn't an "upgrade" over 11 - they exist simultaneously (both are still maintained) and offer two different approaches to D3D.

PatrickvL:
Ah, right, good, then I'm totally fine with dropping aliasing.

If ever we think there would be any benefit in a rendering abstraction layer, it can be done then and there. No need for YAGNI features.

medievil1:
going out the door for work, just wanted to report one possible issue, seems SOME games, now, won't run with higher resolution output (1920x1080 for example) with exclusive fullscreen selected.. unselect it and they run fine at that resolution

CookiePLMonster:
I guess the games which restart themselves may have problems with exclusive fs.

PR #2106, opened 2020-12-19
Fix Halo fog issue

NZJenkins:
Halo sets the specular alpha via the `SetVertexData` API - in the fixed-function world this corresponds to the fog factor
This PR copies the specular alpha value to the fog register
With Halo's fog renderstate configuration this essentially disables fog
Not sure if the solution is 100% correct but hopefully a step in the right direction

May fix fog issues for other titles mentioned in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1685

Tested in Halo v1.01 (PAL XDK 3948)

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
@@ -288,14 +288,6 @@ float4 reverseScreenspaceTransform(float4 oPos)
 
 VS_OUTPUT main(const VS_INPUT xIn)
 {
-	// Output variables
-	float4 oPos, oD0, oD1, oB0, oB1, oT0, oT1, oT2, oT3;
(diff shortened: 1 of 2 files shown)


RadWolfie:
@NZJenkins I would like to add a remark to this pull request.

Both retail and demo versions of Halo I have does behave properly. Meaning, it is able to see broken intros, render main menu only, and navigation too.

After numerous of attempt to get in-game with demo version. I am able to see 3D environment which looks almost the same as fog hack implemented. However, the island looks little bit different from the beach while still in the pelican.

With only fog hack implemented, demo version of Halo doesn't render anything except for start of the mission. Whilst retail version is able to render almost everything.

RadWolfie:
In my opinion, it is likely in the right direction. And perhaps able to find the actual fault in current codebase.

PatrickvL:
Above comment became PR #2109

PR #2109, opened 2020-12-21
Initialize vertex shader outputs to vertex shader attribute values

NZJenkins:
Initialize vertex shader outputs to the current value of the vertex attributes
Which by the sounds of it fixes some Halo demo graphics: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2106#issuecomment-748609302

Dunno if this is right, needs testing :)


PatrickvL:
This drops the "Pre-initialize w component of outputs to 1" feature; So I wonder, now that the inputs are copied into the output registers by default, what value will the components take that weren't present in the input (like, what about z and w for a two component only)?

PatrickvL (approved):

PR #2112, opened 2021-01-01
Fixed booting games from write-protected drives

x1nixmzeng:
This fixes a bug that abused `SHCreateDirectoryEx` to use the return value **`ERROR_ALREADY_EXISTS`** to check if the directory already exists. The device `\Device\CdRom0` is used for registering the XBE file path and will usually exist. I have switched to check this using `std::filesystem` instead.

On write-protected drives `SHCreateDirectoryEx` was returning **`ERROR_WRITE_PROTECT`** which caused the device to not register, and would stop games from booting.

Note "write-protected drives" are not the same as file permissions - on Windows, read-only folder permissions are ignored

### DIFF
--- a/src/core/kernel/support/EmuFile.cpp
+++ b/src/core/kernel/support/EmuFile.cpp
@@ -553,8 +553,6 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
 
 int CxbxRegisterDeviceHostPath(std::string XboxDevicePath, std::string HostDevicePath, bool IsFile)
 {
-	int result = -1;
-
 	XboxDevice newDevice;
 	newDevice.XboxDevicePath = XboxDevicePath;
 	newDevice.HostDevicePath = HostDevicePath;
@@ -573,24 +571,17 @@ int CxbxRegisterDeviceHostPath(std::string XboxDevicePath, std::string HostDevic
 
 	// If this path is not a raw file partition, create the directory for it
 	if (!IsFile) {
-
-		if (std::filesystem::exists(HostDevicePath))
-		{
-			succeeded = true;
-		}
-		else
-		{
-			succeeded = std::filesystem::create_directory(HostDevicePath);
-		}
+		succeeded = std::filesystem::exists(HostDevicePath) || std::filesystem::create_directory(HostDevicePath);
 	}
 
-	if (succeeded)
-	{
-		Devices.emplace_back(newDevice);
-		result = Devices.size() - 1;
+	if (succeeded) {
+		Devices.push_back(newDevice);
+		return static_cast<int>(Devices.size()) - 1;
 	}
 
-	return result;
+	EmuLog(LOG_LEVEL::FATAL, "Failed to register host device (%s)\n", HostDevicePath.c_str());
+
+	return -1;
 }
 
 
--- a/src/core/kernel/support/EmuFile.cpp
+++ b/src/core/kernel/support/EmuFile.cpp
@@ -554,29 +554,39 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
 int CxbxRegisterDeviceHostPath(std::string XboxDevicePath, std::string HostDevicePath, bool IsFile)
 {
 	int result = -1;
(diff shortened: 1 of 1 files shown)


src/core/kernel/support/EmuFile.cpp
```diff
@@ -554,29 +554,39 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
-		status = xbox::status_success;
+		succeeded = true;
 	}
 
 	// If this path is not a raw file partition, create the directory for it
 	if (!IsFile) {
-		status = SHCreateDirectoryEx(NULL, HostDevicePath.c_str(), NULL);
+
+		if (std::filesystem::exists(HostDevicePath))
+		{
```
PatrickvL, line 578: nit : move { at the end of the if line

src/core/kernel/support/EmuFile.cpp
```diff
@@ -554,29 +554,39 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
+			succeeded = true;
+		}
+		else
+		{
```
PatrickvL, line 582: } else {

src/core/kernel/support/EmuFile.cpp
```diff
@@ -554,29 +554,39 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
+			succeeded = std::filesystem::create_directory(HostDevicePath);
+		}
 	}
 
-	if (status == xbox::status_success || status == ERROR_ALREADY_EXISTS) {
-		Devices.push_back(newDevice);
+	if (succeeded)
+	{
```
PatrickvL, line 588: if (succeeded) {

src/core/kernel/support/EmuFile.cpp
```diff
@@ -554,29 +554,39 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
+		Devices.emplace_back(newDevice);
```
PatrickvL, line 589: Why change `push_back` into `emplace_back`? To avoid the move intrinsic?

src/core/kernel/support/EmuFile.cpp
```diff
@@ -554,29 +554,39 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
 	if (!IsFile) {
-		status = SHCreateDirectoryEx(NULL, HostDevicePath.c_str(), NULL);
+
+		if (std::filesystem::exists(HostDevicePath))
+		{
+			succeeded = true;
+		}
+		else
+		{
+			succeeded = std::filesystem::create_directory(HostDevicePath);
```
PatrickvL, line 583: Could be written much shorter, like this  :
`succeeded = std::filesystem::exists(HostDevicePath) || std::filesystem::create_directory(HostDevicePath);`

src/core/kernel/support/EmuFile.cpp
```diff
@@ -554,29 +554,39 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
 	if (_strnicmp(XboxDevicePath.c_str(), DeviceHarddisk0PartitionPrefix.c_str(), DeviceHarddisk0PartitionPrefix.length()) == 0) {
-		std::string partitionHeaderPath = (HostDevicePath + ".bin").c_str();
-		if (!PathFileExists(partitionHeaderPath.c_str())) {
+		std::string partitionHeaderPath = HostDevicePath + ".bin";
+		if (!std::filesystem::exists(partitionHeaderPath)) {
 			CxbxCreatePartitionHeaderFile(partitionHeaderPath, XboxDevicePath == DeviceHarddisk0Partition0);
 		}
 
-		status = xbox::status_success;
+		succeeded = true;
```
PatrickvL, line 571: This serves no real purpose.

Instead, you could move the variable declaration below this code block, and initialize it to `true`, so that code behaves functionally equivalent.

PatrickvL (commented):
Nice work! Still, some small remarks - feel free to pick up or ignore ;)

PatrickvL (approved):

PR #2116, opened 2021-01-08
Add Pull Request Labeler Support

RadWolfie:
I have filled what are most likely folders and some files to be automatically label. However, there are some that do either need to relocate or rename files.

Such as dialog controllers. Currently in draft pull request for further review and possibility re-arrange files before then.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,74 @@
+#labels are sorted by alphabet order.
+
+cmake:
+  - CMake*
+  - '**/CMakeLists.txt'
+  - '**/*.cmake'
+
+cpu-emulation:
+  - src/devices/x86/**
+
+deployment:
+  - '*.yml'
+  - .github/workflows/CI.yml
+
+file-system:
+  - src/core/kernel/support/EmuFile*
+
+graphics:
+  - src/core/hle/D3D8/**
+  - src/core/hle/XGRAPHIC/**
+  - src/devices/video/**
+  - src/gui/*Video*
+
+HLE:
+  - src/core/hle/**
+  - src/core/kernel/**
+
+informational:
+  - '**/*Logging*'
+  - '**/*Logging*/**'
+
+input:
+  - src/common/input/**
+  - src/core/hle/XAPI/input/**
+  - src/devices/usb/**
+  - src/gui/controllers/**
+  - src/gui/*Input*
+
+kernel:
+  - src/core/kernel/**
+
+LLE:
+  - src/devices/**
+
+memory:
+  - src/core/kernel/memory-manager/**
+
+networking:
+  - src/core/hle/XONLINE/**
+  - src/devices/network/**
+  - src/gui/*Network*
+
+sound:
+  - src/common/audio/**
+  - src/core/hle/DSOUND/**
+  - src/core/hle/XACTENG/**
+  - src/devices/audio/**
+  - src/gui/*Audio*
+
+modules:
+  - import/**
+
+threading:
+  - src/core/kernel/support/EmuFS*
+
+timing:
+  - src/common/Timer*
+
+user interface:
+  - src/gui/**
+# - TODO: Dear Imgui location, maybe in src/core/gui path.
+
+xbdm:
+  - src/common/xbdm/**
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,14 @@
+name: Pull Request Manager
+
+on:
+  pull_request_target:
+
+jobs:
+  pr_manager:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Labeler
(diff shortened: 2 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,66 @@
...
+
+graphics:
+  - src/core/hle/D3D8/**
+  - src/core/hle/XGRAPHIC/**
+  - src/devices/video/**
+  - src/gui/*Video*
+
+HLE:
+  - src/core/hle/**
+  - src/core/kernel/**
```
PatrickvL, line 25: Best remove kernel from HLE label, it has it's own kernel label, so better not overlap them

PatrickvL (changes requested):
Nice work! One small remark (I didn't verify any label though, so it's just a cursory remark)

.github/labeler.yml
RadWolfie: As far as I understand, our own kernel are patched through title's kernel vtable are HLE and is not directly using hardware layer. Hence we don't require an actual bios to be use in Cxbx-Reloaded. So, our kernel fall under HLE category. In my honest opinion, it should be in hle folder instead of core folder.

PatrickvL: Respectfully, I still think it's better to not label any change in the kernel folder as "HLE".
Reason is, that the "kernel" label is already specifically meant for kernel changes.

Also, HLE is a term that indicates emulation by bypassing hardware. This doesn't apply to kernel code, as the Xbox kernel is not really a hardware device, it's just software, for which our kernel forms a replacement.
Hence, I believe the "HLE" label doesn't apply to our kernel code, and should thus not be assigned automatically.

RadWolfie: > Also, HLE is a term that indicates emulation by bypassing hardware.

Our own kernel do bypass hardware, only very little code portion does trigger hardware commands directly. Plus our kernel is not emulating inside "virtual machine" as well. If we have custom kernel that works on xbox hardware or our own cxbxr kernel running inside emulation instead of outside of emulation environment. Then "HLE" wouldn't apply to such condition. In other word, we're running kernel as a wrapper just like our patches (see below for further comment).

Afair, JayFoxRox did say our kernel is HLE anyway. And our patches for graphic/audio/etc are UHLE.

Anyhow, I think it's better to have vote from our team members to decide

PatrickvL: Agree on the vote. But do realize that the Xbox kernel itself doesn't address many hardware devices. It interacts with a clock, listens to a few hardware interrupts, but otherwise mainly implements hundreds of API's that user code can call for various tasks. There's not much hardware involved with that, and thus our kernel is for the biggest part just another, replacement implementation of those Xbox kernel API's.
That's why our kernel is best not described as "LLE" or "HLE", but simply as "kernel", because it's not patching Xbox software, nor emulating Xbox hardware devices, it's merely an alternative implementation.

RadWolfie: If dumped bios/custom kernel was done inside xbox emulation instead of outside, then yes I agree it doesn't fall under either HLE and LLE which then run as native software. Although, the software still access to hardware directly anyway. 😛

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,14 @@
...
+  pull_request_target:
+
+jobs:
+  pr_manager:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Labeler
+        uses: actions/labeler@v3
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
Margen67, line 13:
```suggestion
          repo-token: ${{ secrets.GITHUB_TOKEN }}
```

RadWolfie: The quotes came from Labeler's action template example, I'll fix this and the other pull request.

.github/labeler.yml
RadWolfie: Our majority agreement of kernel is HLE. However, only couple members made a vote not to include HLE label. **The rest of the members still need vote to either include or not.**

ergo720: Ok. I'm fine with using the HLE label for the kernel as well.

PatrickvL (dismissed):

PatrickvL:
Was about to merge, saw there's conflicts now... Please resolve.

RadWolfie:
> Was about to merge, saw there's conflicts now... Please resolve.

What conflict? Pull request is still in draft stage.

.github/labeler.yml
RadWolfie: Most team members voted Yes, leaving this change intact.

RadWolfie:
What's new since force pushed?
- Added two missing file paths for input label.
- Added informational label for any "Logging" keyword in file path.
- Added one missing file path for networking label.

LukeUsher (approved):

PR #2117, opened 2021-01-10
Ensure the xbox viewport is applied when Clear is called

NZJenkins:
This fixes an issue introduced in 46ec7516faca6641255246b34423013cec5bf4a1
The xbox viewport was applied in UpdateNativeD3DResources instead of the SetViewport patch, so wouldn't be applied until a Draw call.
Crash Tag Team Racing graphics would be blacked out whenever the rear view mirror appeared on the screen.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -4836,6 +4836,9 @@ xbox::void_xt WINAPI xbox::EMUPATCH(D3DDevice_Clear)
 
 	DWORD HostFlags = 0;
 
+	// Clear requires the Xbox viewport to be applied
+	CxbxUpdateNativeD3DResources();
+
     // make adjustments to parameters to make sense with windows d3d
     {
 		if (Flags & X_D3DCLEAR_TARGET) {


CookiePLMonster (approved):
LGTM and the change makes sense too, because Clear **does** respect the viewport.

PatrickvL:
Although this seems to work, it's a bit overkill. Perhaps we could refactor `CxbxImpl_SetRenderTarget`, to only let it set Cxbx state, and extract all host updating code into an `CxbxUpdateRenderTarget()` function, that could be called from `CxbxUpdateNativeD3DResources()`, _and_ **instead of** this added call to `CxbxUpdateNativeD3DResources()` (and then remove the comment about `DxbxUpdateActiveRenderTarget`)

PatrickvL:
Ehrm, good question. Possibly I've mixed up things there, because SetRenderTarget also updates the viewport (if I remember correctly). Anyway, regardless, CxbxUpdateNativeD3DResources() still does much more than strictly required for Clear() to work reliably. So the suggestion is still to try and call only those updates that are required, so that all unnecessary updating can be avoided.

LukeUsher (approved):

PatrickvL:
Oh, Luke merged & closed. Shall I create a ticket for the above suggestion, @NZJenkins  ?

PR #2122, opened 2021-01-12
Export high performance graphics hints in cxbxr-ldr.exe

CookiePLMonster:
This removes NVidia Optimus/AMD Xpress exports from `cxbxr-emu.dll` and adds them to `cxbxr-ldr.exe`. This should make the loader executable opt in to use high performance graphics, like `cxbx.exe` does already.


PatrickvL (commented):
Nice catch!

PatrickvL (approved):

RadWolfie:
I would had approved this. Although, I would had suggest shorten `CXBXR_SOURCE_EXECUTABLES` to `CXBXR_SOURCE_EXEC`. 😛

PR #2123, opened 2021-01-17
Reduce the amount of work done per draw

NZJenkins:
Certain titles split geometry into many small sections, and do a huge number of draws per frame
These are some optimizations and tweaks to code that runs per-draw

Significantly increases performance in Crash Nitro Kart

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2702,7 +2702,7 @@ ConvertedIndexBuffer& CxbxUpdateActiveIndexBuffer
 	}
 
 	// Poor-mans eviction policy : when exceeding treshold, clear entire cache :
-	if (g_IndexBufferCache.size() > 256) {
+	if (g_IndexBufferCache.size() > 10000) {
 		g_IndexBufferCache.clear(); // Note : ConvertedIndexBuffer destructor will release any assigned pHostIndexBuffer
 	}
 
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -175,6 +175,8 @@ xbox::X_D3DVIEWPORT8 g_Xbox_Viewport = { 0 };
 float g_Xbox_BackbufferScaleX = 1;
 float g_Xbox_BackbufferScaleY = 1;
 
+static constexpr size_t INDEX_BUFFER_CACHE_SIZE = 10000;
+
 /* Unused :
 static xbox::dword_xt                  *g_Xbox_D3DDevice; // TODO: This should be a D3DDevice structure
 */
@@ -2641,8 +2643,8 @@ public:
 	}
 };
 
-std::unordered_map<uint32_t, ConvertedIndexBuffer> g_IndexBufferCache;
-	
+	std::unordered_map<uint32_t, ConvertedIndexBuffer> g_IndexBufferCache;
+
 void CxbxRemoveIndexBuffer(PWORD pData)
 {
 	// HACK: Never Free
@@ -2702,7 +2704,7 @@ ConvertedIndexBuffer& CxbxUpdateActiveIndexBuffer
 	}
 
 	// Poor-mans eviction policy : when exceeding treshold, clear entire cache :
-	if (g_IndexBufferCache.size() > 10000) {
+	if (g_IndexBufferCache.size() > INDEX_BUFFER_CACHE_SIZE) {
 		g_IndexBufferCache.clear(); // Note : ConvertedIndexBuffer destructor will release any assigned pHostIndexBuffer
 	}
 
--- a/src/core/hle/D3D8/Direct3D9/TextureStates.cpp
+++ b/src/core/hle/D3D8/Direct3D9/TextureStates.cpp
@@ -75,6 +75,13 @@ TextureStateInfo CxbxTextureStateInfo[] = {
     { "D3DTSS_COLORKEYCOLOR",           false,  0 },
 };
 
+// Keep track of the last state that was set
+typedef struct {
+    bool HasBeenSet = false;
+    DWORD LastValue = 0;
+} PreviousState;
+PreviousState PreviousStates[xbox::X_D3DTS_STAGECOUNT][xbox::X_D3DTSS_LAST + 1] = {};
+
 bool XboxTextureStateConverter::Init(XboxRenderStateConverter* pState)
 {
     // Deferred states start at 0, this means that D3DDeferredTextureState IS D3D__TextureState
@@ -176,11 +183,19 @@ void XboxTextureStateConverter::Apply()
 
         for (int State = xbox::X_D3DTSS_FIRST; State <= xbox::X_D3DTSS_LAST; State++) {
             // Read the value of the current stage/state from the Xbox data structure
-            DWORD Value = Get(XboxStage, State); // OR D3D__TextureState[(XboxStage * xbox::X_D3DTS_STAGESIZE) + XboxTextureStateOffsets[State]];
+            DWORD XboxValue = Get(XboxStage, State); // OR D3D__TextureState[(XboxStage * xbox::X_D3DTS_STAGESIZE) + XboxTextureStateOffsets[State]];
+            DWORD PcValue = XboxValue;
+
+            // If the state hasn't changed, skip setting it
+            auto lastState = &PreviousStates[XboxStage][State];
+            if (lastState->HasBeenSet && lastState->LastValue == XboxValue) {
+                continue;
+            }
+
             switch (State) {
                 // These types map 1:1 but have some unsupported values
                 case xbox::X_D3DTSS_ADDRESSU: case xbox::X_D3DTSS_ADDRESSV: case xbox::X_D3DTSS_ADDRESSW:
-                    switch (Value) {
+                    switch (XboxValue) {
                         case 0: // Let's ignore zero (its no known X_D3DTADDRESS_ mode, but logging this seems useless)
                         case xbox::X_D3DTADDRESS_WRAP:        // = 1 = D3DTADDRESS_WRAP = 1,
                         case xbox::X_D3DTADDRESS_MIRROR:      // = 2 = D3DTADDRESS_MIRROR = 2,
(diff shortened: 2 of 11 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2702,7 +2702,7 @@ ConvertedIndexBuffer& CxbxUpdateActiveIndexBuffer
 	}
 
 	// Poor-mans eviction policy : when exceeding treshold, clear entire cache :
-	if (g_IndexBufferCache.size() > 256) {
+	if (g_IndexBufferCache.size() > 10000) {
```
CookiePLMonster, line 2705: Since you're changing it, I suggest changing it to a `static constexpr size_t INDEX_BUFFER_CACHE_SIZE = 10000;` somewhere on top of this file, just to have it readily available.

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
```diff
@@ -75,6 +75,13 @@ TextureStateInfo CxbxTextureStateInfo[] = {
     { "D3DTSS_COLORKEYCOLOR",           false,  0 },
 };
 
+// Keep track of the last state that was set
+typedef struct {
+    bool HasBeenSet = false;
+    DWORD LastValue = 0;
+} PreviousState;
+PreviousState PreviousStates[xbox::X_D3DTS_STAGECOUNT][xbox::X_D3DTSS_LAST + 1] = {};
```
CookiePLMonster, line 83: I believe you want a `std::optional<DWORD>` here, since that'll allow for values "not to exist".

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -648,9 +646,21 @@ class XboxVertexShaderDecoder
 
 		return VshGetField(pShaderToken, FLD_FINAL) == 0;
 	}
-
 };
 
+// Get the function size excluding the final field
+size_t GetVshFunctionSize(const xbox::dword_xt* pXboxFunction) {
+	auto curToken = (uint32_t*)pXboxFunction;
```
CookiePLMonster, line 653: casting away constness is meh. Recommend `reinterpret_cast<const uint32_t*>(pXboxFunction);` instead.

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1547,54 +1557,18 @@ void CxbxImpl_SetVertexShaderConstant(INT Register, PVOID pConstantData, DWORD C
...
-			if (!VshDecoder.VshConvertToIntermediate(pCurToken, pShader)) {
-				if (i < pShader->Header.NumInst - 1) {
-					LOG_TEST_CASE("Shader instructions after final instruction");
-				}
-				break;
-			}
-			pCurToken += X_VSH_INSTRUCTION_SIZE;
-		}
-	}
+	uint32_t* pCurToken = (uint32_t*)((uintptr_t)pXboxFunction);
```
CookiePLMonster, line 1566: A double typecast is most likely redundant here.

CookiePLMonster (commented):
Just a few nitpicks. I haven't tested it yet, but the changes look exciting! I'll be happy not to see all those redundant texture state sets in PIX captures too, so that change also improves debuggability.

CookiePLMonster:
Brings a big speedup to **NASCAR Heat 2002** - I wasn't even aware previously that it's a 60 FPS game.

Master:
image

PR:
image

**Outrun 2** now has much better framerates:
image

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
CookiePLMonster: Additional benefit of `std::optional` - comparing `optional == value` will always return false if optional doesn't have a value, so the check below can be simplified to a simple equality check and it'll implicitly check for presence of a value.

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -194,32 +194,6 @@ std::string DebugPrependLineNumbers(std::string shaderString) {
...
 
-
-extern ShaderType EmuGetShaderInfo(IntermediateVertexShader* pIntermediateShader) {
-
-	if (pIntermediateShader->Instructions.size() == 0) {
-		// Do not attempt to compile empty shaders
-		// This is a declaration only shader, so there is no function to compile
-		return ShaderType::Empty;
-	}
-
```
PatrickvL, line 205: Not having this version check anymore, makes one wonder : Are we going to forget about it, is there still other code mentioning it, or do we need to create an issue on the matter?

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -80,6 +77,8 @@ ShaderKey VertexShaderSource::CreateShader(const xbox::dword_xt* pXboxFunction,
 	}
 	else {
 		// We can't do anything with this shader
+		// Test case: ???
+		EmuLog(LOG_LEVEL::WARNING, "Empty vertex shader");
```
PatrickvL, line 81: If this is the only code left (rudimentary) recognizing state shaders and/or other unsupported versions, than indeed a LOG_TEST_CASE is needed, mentioning what versions should be handled later on.

PatrickvL (changes requested):
Great work! Relatively little code changed, but with significant results, congrats!

Please do address the review remarks, and then it'll quickly be good to get merged!

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
```diff
@@ -176,11 +183,19 @@ void XboxTextureStateConverter::Apply()
 
         for (int State = xbox::X_D3DTSS_FIRST; State <= xbox::X_D3DTSS_LAST; State++) {
             // Read the value of the current stage/state from the Xbox data structure
-            DWORD Value = Get(XboxStage, State); // OR D3D__TextureState[(XboxStage * xbox::X_D3DTS_STAGESIZE) + XboxTextureStateOffsets[State]];
+            DWORD XboxValue = Get(XboxStage, State); // OR D3D__TextureState[(XboxStage * xbox::X_D3DTS_STAGESIZE) + XboxTextureStateOffsets[State]];
+            DWORD PcValue = XboxValue;
+
+            // If the state hasn't changed, skip setting it
+            auto lastState = &PreviousStates[XboxStage][State];
```
PatrickvL, line 190: I'm not entirely sure our cached state will always reflected the actual host D3D state. There's a significant change host states could change by calls to SetRenderTarget, Clear, and possibly others.

So perhaps as detection code first to see if that indeed happens? (Detection is simply a compare between your cached value and the result of a call to host GetTexture[Sampler]State.)

Then, if internal changes are indeed detected, we can decide to either always compare against host state, or to use your cache but add case-specific host checks alongside it.

Or, we discover host never changes states internally, ever, and then we learned something valuable too ;)

ObiKKa:
...
> **Outrun 2** now has much better framerates:
...

Can you please mention your PC CPU & system memory specifications used for testing these games?

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
NZJenkins: Nice that's exactly what's needed

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: Can reinterpret_cast make a difference or it's style?

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
NZJenkins: I didn't see an indication of that in the docs?
If nothing's in the docs we may as well just wait for something to go wrong instead imo and save the churn

Previous states should probably be a class variable here - since the device can be recreated, then they might go out of sync

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
NZJenkins: Unless I'm misreading it the headers are consumed here now
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/master/src/core/hle/D3D8/XbVertexShader.cpp#L1376-L1404

src/core/hle/D3D8/XbVertexShader.cpp
CookiePLMonster: Style - I'm not as worried about a C style cast as I am about casting away constness.

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
CookiePLMonster: States will definitely desync on device reset, but we don't do that. Other than that indeed, might be a good idea to clear those on device creation to have the states sync with possible device recreation during the process lifetime.

Regarding texture states, IIRC in OpenGL they bind to textures but in D3D they bind to slots, so I don't think we should be concerned about them "desyncing".

src/core/kernel/init/CxbxKrnl.h
```diff
@@ -235,4 +235,22 @@ extern char szFilePath_Xbe[MAX_PATH*2];
 // Returns the last Win32 error, in string format. Returns an empty string if there is no error.
 extern std::string CxbxGetLastErrorString(char * lpszFunction);
 
+// The reason of having EmuLogOutputEx in LOG_TEST_CASE is to allow dump to log directly for any test cases triggered.
+// Which will make developers easier to note which applications has triggered quicker, easier, and doesn't require any individual log enabled to capture them.
```
RadWolfie, line 239: Can you add the reason for the move back to CxbxKrnl.h? Such as CxbxKrnl_Xbe->m_szAsciiTitle is dependency to this, CxbxKrnl.h, file since common Logging.h file doesn't have any dependency.

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "VertexShaderSource.h"
 
-#include "Logging.h"
+#include "core/kernel/init/CxbxKrnl.h"
```
RadWolfie, line 5: Also, did you tried include CxbxKnl.h first then Logging.h afterward?

src/core/hle/D3D8/XbVertexShader.cpp
RadWolfie: I for one rather avoid using C style and use C++ style. There's only little room in C++ standard to use C style under certain condition. The rest is better off using C++ style. Casting does have some difference between C and C++ by the way.

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
NZJenkins: It works including more than one header, but then you need to know to do that

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
PatrickvL: "The proof of the pudding, is in the eating", or : we won't know until we try.

I know for a fact that some Xbox D3D APIs change state, so it's not that far fetched to consider this might apply to host D3D as well.
Perhaps, we could extend the LOG_D3D_RESULT macro with a call to a state verification function, so that we can even see which host D3D API (if any) changes which state.

NZJenkins: IMO searching for undocumented D3D9 behaviour isn't a must have for the purposes of this PR - and if added in the most straightforward way during state updates, it would slow things down significantly

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
NZJenkins: as mentioned in another comment the headers appear handled at the `LoadVertexShader`

LukeUsher (approved):

PatrickvL (approved):

PR #2125, opened 2021-01-21
Revert "Initialize vertex shader outputs to vertex shader attribute values"

NZJenkins:
Reverts Cxbx-Reloaded/Cxbx-Reloaded#2109

Since it breaks font rendering in samples


LukeUsher (approved):

LukeUsher:
We had font rendering issues in the XDK samples before: later XDK samples abuse position shader registers to store texture-coordinates.

Since it's a 2D font, they store everything in position: pos.x = x, pos.x = y, and tex-coords go in pos.z and pos.w

PR #2127, opened 2021-01-29
vs: use model 2.a primarily, with 3.0 fallback

LukeUsher:
Based on conversations with @CookiePLMonster: This PR defaults to shader model 2.a for Vertex Shaders, rather than 3.0. 

Using Vertex Shader Model 3.0 with fixed-function or pixel shader model < 3 is *undefined* behavior and not supposed to work, yet Cxbx-R does this.

This PR avoids that undefined behavior in the default case, removing the need for the AMD hack, while still falling back to 3.0 in case of compilation failure, to allow large shaders to use their existing (undefined, but better than crashing) behavior.

This may not have *any* impact whatsoever in most games, but it also has the potential to fix weird issues that have no obvious explanation.

I'd consider this a low-risk merge, as we know vertex shader model 2.a is functional (ALL shaders use this on AMD gpus already, with,or without dxvk).

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/VertexShader.cpp
+++ b/src/core/hle/D3D8/Direct3D9/VertexShader.cpp
@@ -6,7 +6,7 @@
 
 #include <sstream>
 
-extern const char* g_vs_model = vs_model_3_0;
+extern const char* g_vs_model = vs_model_2_a;
 
 // HLSL generation
 void OutputHlsl(std::stringstream& hlsl, VSH_IMD_OUTPUT& dest)
@@ -201,6 +201,7 @@ HRESULT CompileHlsl(const std::string& hlsl, ID3DBlob** ppHostShader, const char
 	ID3DBlob* pErrors = nullptr;
 	ID3DBlob* pErrorsCompatibility = nullptr;
 	HRESULT             hRet = 0;
+	auto hlslErrorLogLevel = FAILED(hRet) ? LOG_LEVEL::ERROR2 : LOG_LEVEL::DEBUG;
 
 	UINT flags1 = D3DCOMPILE_OPTIMIZATION_LEVEL3;
 	hRet = D3DCompile(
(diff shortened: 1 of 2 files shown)


src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -216,6 +216,27 @@ HRESULT CompileHlsl(const std::string& hlsl, ID3DBlob** ppHostShader, const char
 		ppHostShader, // out
 		&pErrors // ppErrorMsgs out
 	);
+
+	// If the sharder failed in the default vertex shader model, retry in vs_model_3_0
```
CookiePLMonster, line 220:
```suggestion
	// If the shader failed in the default vertex shader model, retry in vs_model_3_0
```

CookiePLMonster, line 220:
```suggestion
	// If the shader failed in the default vertex shader model, retry in vs_model_3_0
```

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -216,6 +216,27 @@ HRESULT CompileHlsl(const std::string& hlsl, ID3DBlob** ppHostShader, const char
...
+			hlsl.length(),
+			pSourceName, // pSourceName
+			nullptr, // pDefines
+			D3D_COMPILE_STANDARD_FILE_INCLUDE, // pInclude // TODO precompile x_* HLSL functions?
+			"main", // shader entry poiint
+			vs_model_3_0, // shader profile
+			flags1, // flags1
+			0, // flags2
+			ppHostShader, // out
+			&pErrors // ppErrorMsgs out
```
CookiePLMonster, line 235: `pErrors` will likely be non-null at the time of this call, therefore causing a resource leak. Do you think it is within the scope of this PR to resolve this, preferably by using `WRL::ComPtr` in place of a naked pointer?

CookiePLMonster (changes requested):
A few nitpicks. Untested.

On top of that, what is the situation with pixel shaders? Do we ever use `ps_3_0`?

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
LukeUsher: I'd consider this out of scope because the same problem exists for the other fallback too? Saying that, if you feel like refactoring to use a ComPtr, feel free :P

LukeUsher:
> A few nitpicks. Untested.
> 
> On top of that, what is the situation with pixel shaders? Do we ever use `ps_3_0`?

Pixel Shaders (at least in current builds) are using shader model 2.0

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
LukeUsher: I updated to avoid the pErrors leak

CookiePLMonster:
First "unexplained issue" fixed by this PR - **World Racing 2** now loads to the race, it used to fail on... locking a vertex buffer.

image

CookiePLMonster (approved):

PR #2129, opened 2021-02-02
Update Prerequisites Section

RadWolfie:
Add Wine prerequisites to readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@ Cxbx-Reloaded is an emulator for running Microsoft Xbox (and eventually, Chihiro
     * Make sure to enable winpcap compatibility mode!
 
 ### Wine
-
 **NOTICE: Please use latest stable release version of Wine. If it does not work for you, then roll back to Wine 5.0.3 as last known working version.**
   * Winetricks
     * `vcrun2017` or `vcrun2019`
--- a/README.md
+++ b/README.md
@@ -10,19 +10,22 @@ Cxbx-Reloaded is an emulator for running Microsoft Xbox (and eventually, Chihiro
   * OS: Windows 7+ x64, or x86-64 Linux with Wine. 32-bit is not supported.
   * GPU: Direct3D 9.0c with Pixel Shader Model 2.x, and Vertex Shader Model 3.0.
 
-### Prerequisites
-Windows
+## Prerequisites
+### Windows
   * [32-bit (x86) Visual C++ 2019 Redistributable](https://aka.ms/vs/16/release/vc_redist.x86.exe)
   * [Npcap *(used for network emulation)*](https://nmap.org/npcap/#download)
     * Make sure to enable winpcap compatibility mode!
 
-Wine
+### Wine
+
+**NOTICE: Please use latest stable release version of Wine. If it does not work for you, then roll back to Wine 5.0.3 as last known working version.**
   * Winetricks
     * `vcrun2017` or `vcrun2019`
       * NOTE: vcrun2019 require latest winetricks script!
     * `d3dcompiler_47`
       * NOTE: May will change over time.
   * Winpcap is built-in, no installation requirement.
+
 
 ## Automated Builds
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,11 +10,22 @@ Cxbx-Reloaded is an emulator for running Microsoft Xbox (and eventually, Chihiro
...
 
-### Prerequisites
+## Prerequisites
+### Windows
   * [32-bit (x86) Visual C++ 2019 Redistributable](https://aka.ms/vs/16/release/vc_redist.x86.exe)
   * [Npcap *(used for network emulation)*](https://nmap.org/npcap/#download)
     * Make sure to enable winpcap compatibility mode!
 
+### Wine
+**NOTICE: Please use latest stable release version of Wine. If it does not work for you, then roll back to Wine 5.0.3 as last known working version.**
```
ergo720, line 20:
```suggestion
**NOTICE: Please use the latest stable release version of Wine. If it does not work for you, then roll back to Wine 5.0.3 which is the last known working version.**
```

README.md
```diff
@@ -10,11 +10,22 @@ Cxbx-Reloaded is an emulator for running Microsoft Xbox (and eventually, Chihiro
...
+  * Winetricks
+    * `vcrun2017` or `vcrun2019`
+      * NOTE: vcrun2019 require latest winetricks script!
```
ergo720, line 23:
```suggestion
      * NOTE: vcrun2019 requires the latest winetricks script!
```

README.md
```diff
@@ -10,11 +10,22 @@ Cxbx-Reloaded is an emulator for running Microsoft Xbox (and eventually, Chihiro
...
+    * `d3dcompiler_47`
+      * NOTE: May will change over time.
```
ergo720, line 25:
```suggestion
      * NOTE: May/will change over time.
```

README.md
```diff
@@ -10,11 +10,22 @@ Cxbx-Reloaded is an emulator for running Microsoft Xbox (and eventually, Chihiro
...
+  * Winpcap is built-in, no installation requirement.
```
ergo720, line 26:
```suggestion
  * Winpcap is built-in, no installation is required.
```

README.md
RadWolfie: Or better yet... **May be subject to** change over time.

README.md
RadWolfie: done

README.md
RadWolfie: done

README.md
RadWolfie: done

README.md
RadWolfie: done

ergo720 (approved):

README.md
```diff
@@ -10,11 +10,22 @@ Cxbx-Reloaded is an emulator for running Microsoft Xbox (and eventually, Chihiro
...
   * GPU: Direct3D 9.0c with Pixel Shader Model 2.x, and Vertex Shader Model 3.0.
 
-### Prerequisites
+## Prerequisites
+### Windows
   * [32-bit (x86) Visual C++ 2019 Redistributable](https://aka.ms/vs/16/release/vc_redist.x86.exe)
   * [Npcap *(used for network emulation)*](https://nmap.org/npcap/#download)
     * Make sure to enable winpcap compatibility mode!
 
+### Wine
```
PatrickvL, line 19: Above specifies the operating system Windows, so this line should probably be "Linux, using Wine" or something like that.

RadWolfie: I don't exactly agree with you. I consider Wine as its own platform since it can be use on Linux, BSD, Solaris, Android, and macOS. Maybe more that I missed.

RadWolfie: I do understand what you're trying to say. Perhaps you meant "Unix, using Wine"?

RadWolfie: Base on internal discussion, it's better to leave it as is. Since there's some variant of operating system which do not fall under Unix, such as Linux. And Wine isn't available for all POSIX operating system.

Since main developers are not on Linux full time, I forgot to add one detail under Wine section.

PR #2133, opened 2021-02-07
Fix some AA scale issues

NZJenkins:
* Fix Shenmue II rendering to half the screen
The title used SSAA with the fixed function passthrough mode. The code expected the title to scale its vertices, but apparently it doesn't have to
* Fix Max Payne 2 bullet time graphics
The title uses the backbuffer with MSAA as a texture. When normalizing texture coordinates we need to use pixel dimensions instead of the raw subpixel dimensions of the texture.

Needs testing with titles that use AA since these scale factors are pretty fiddly

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2878,6 +2878,20 @@ void GetMultiSampleScaleRaw(float& xScale, float& yScale) {
 // Titles can render pre-transformed vertices to screen space (using XYZRHW vertex position data or otherwise)
 // so we need to know the space they are in to interpret it correctly
 void GetScreenScaleFactors(float& scaleX, float& scaleY) {
+	scaleX = 1;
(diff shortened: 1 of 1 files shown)


medievil1:
the stretched out afterburner motion blur thing in crimson skies is gone completely with this pr (not rendered at all)
image

literalmente-game:
Didn't fixed WhiteOut nitro effect
image

medievil1:
ok even weirder, crimsons skies in the pr is as above, no afterburner effect, BUT on the branch the pr comes from, the distorted afterburner effect is present....

LukeUsher (approved):

RadWolfie:
@LukeUsher, I don't believe this pull request should had been merged. @literalmente-game mentioned a bug in here.
> PS: this PR fixes Max Payne 2 bullet time, but while the effect is in use the internal resolution goes back to native

crackersden:
is this somehow related with the gunvalkyrie "boost" graphic glitch?

Issue #1390, opened 2018-08-15
OG XBox Controller support


literalmente-game:
After talking with both @RadWolfie and @LukeUsher, I was asked to make an issue covering the details about what I've found when searching the SDL source code. For starts, SDL _does_ support the device (at least it seems like it), but keep in mind that the mirror of the source code in Github is outdated and doesn't have the support.
The line that implies the support is this one https://hg.libsdl.org/SDL/file/eb2820ee52ee/src/joystick/SDL_gamecontrollerdb.h#l447. Another detail is that, if you go back in this database, you can see that the support is behind a "#if defined(__LINUX__)"

The up to date library source code can be found here: https://hg.libsdl.org/SDL/
Latest build (from where I've taken this info): https://hg.libsdl.org/SDL/file/eb2820ee52ee/
Direct link to the controller database: https://hg.libsdl.org/SDL/file/eb2820ee52ee/src/joystick/SDL_gamecontrollerdb.h

LukeUsher:
On Windows, the Original Xbox controller requires installation of an unsigned driver, so that's out of the question for HLE..

However, for LLE OHCI, we can simply implement USB-passthrough, which would allow *any* xbox accessory using the controller ports to 'just work' out of the box.

ergo720:
Is this issue still relevant? There is now the [mayflash](https://www.s-config.com/xbcd-mayflash-xbox-joystick-driver/) driver which is signed, and thus solves the signed driver issue. This makes USB-passthrough unnecessary and unlikely to happen since it would require LLE OHCI.

PatrickvL:
At the very least, we should document prerequisites for this in our readme/installation guide

ergo720:
This has now been implemented in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2290 with libusb.

PR #2134, opened 2021-02-08
Add Steel Battalion controller support to input gui

ergo720:
This PR builds upon https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2068 and actually completes the implementation and adds support for the Steel Battalion controller to the input gui. I tested the functionality of the buttons with the controller test option available in the game itself and confirmed that they all work as expected and the toggle functionality of certain buttons is working as well. One thing I noticed: if you map the mouse axis to the controller sticks, they will re-center when the mouse stops moving. This is caused by the same issue that causes Virtua Cop 3 to do the same with the crosshair, and should/will be fixed in a future PR. 

sbc


src/common/Settings.cpp
```diff
@@ -471,22 +471,35 @@ bool Settings::LoadConfig()
...
+			for (int i = 0; i < num_buttons; i++) {
+				char control_name[BUF_NAME_LENGTH];
+				std::sprintf(control_name, sect_input_profiles.control, ctrl_names[i]);
+				control_names[device].push_back(control_name);
+			}
+		};
+
+		switch (num_buttons)
+		{
+		case XBOX_CTRL_NUM_BUTTONS:
```
RadWolfie, line 492: Hmm... I'm not exactly keen on using number of inputs from a device to match hardcode number. In my opinion, it would be better to check device type matrix connect to a pointer of an array list.

In summary, we still have standalone mapping when we should be using a tree-like mapping for better management altogether.

CMakeLists.txt
```diff
@@ -250,6 +248,7 @@ file (GLOB CXBXR_SOURCE_GUIv1
  "${CXBXR_ROOT_DIR}/src/gui/DlgEepromConfig.cpp"
  "${CXBXR_ROOT_DIR}/src/gui/DlgLoggingConfig.cpp"
  "${CXBXR_ROOT_DIR}/src/gui/DlgNetworkConfig.cpp"
+ "${CXBXR_ROOT_DIR}/src/gui/controllers/DlgSBControllerConfig.cpp"
```
RadWolfie, line 251: This line should be reside next to "${CXBXR_ROOT_DIR}/src/gui/controllers/DlgDukeControllerConfig.cpp". We are using sorted order (case insensitive).

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -107,14 +100,13 @@ bool operator!=(xbox::PXPP_DEVICE_TYPE XppType, XBOX_INPUT_DEVICE XidType)
 bool ConstructHleInputDevice(int Type, int Port)
 {
 	g_bIsDevicesEmulating = true;
-	bool ret = true;
+	bool ret;
```
RadWolfie, line 103: I think it's better to leave this as true. That way, we don't need to assign true for every device support.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -107,14 +100,13 @@ bool operator!=(xbox::PXPP_DEVICE_TYPE XppType, XBOX_INPUT_DEVICE XidType)
...
 
 	switch (Type)
 	{
 	case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE): {
 		g_XboxControllerHostBridge[Port].XboxPort = Port;
 		g_XboxControllerHostBridge[Port].XboxType = XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE;
-		g_XboxControllerHostBridge[Port].InState = new XpadInput();
```
RadWolfie, line 117: What is the reason for this removal? How exactly does InState work now?

RadWolfie (changes requested):

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -40,38 +40,25 @@
...
-//typedef DWORD(*fnCxbxVSBCGetState)(UCHAR *);
+
+
 xbox::PXPP_DEVICE_TYPE g_DeviceType_Gamepad = nullptr;
+xbox::PXPP_DEVICE_TYPE g_DeviceType_SBC = nullptr;
 
 // Flag is unset after initialize devices is done by simulate LLE USB thread.
 std::atomic<bool> g_bIsDevicesInitializing = true;
 std::atomic<bool> g_bIsDevicesEmulating = false;
+static CXBX_XINPUT_IN_STATE InState[4];
```
RadWolfie, line 54: Since this is global variable, I strongly advice to add "g_" prefix just like every other global variables has. Though, I'm not sure what's the reason behind having this here.

src/gui/DlgInputConfig.cpp
```diff
@@ -114,11 +116,15 @@ INT_PTR CALLBACK DlgInputConfigProc(HWND hWndDlg, UINT uMsg, WPARAM wParam, LPAR
 
 		for (int i = 0, j = 0; i != 4; i++) {
 			HWND hHandle = GetDlgItem(hWndDlg, IDC_DEVICE_PORT1 + i);
-			for (auto input : input_support_list) {
-				LRESULT index = SendMessage(hHandle, CB_ADDSTRING, 0, reinterpret_cast<LPARAM>(input.name));
-				SendMessage(hHandle, CB_SETITEMDATA, index,
-					to_underlying(input.type));
-				if (g_Settings->m_input_port[i].Type == to_underlying(input.type)) {
+			for (auto dev_type : {
+				XBOX_INPUT_DEVICE::DEVICE_INVALID,
```
RadWolfie, line 120: :/ I preferred this to stay in central location for easier update which is InputManager.h file without the need to keep on digging where device support list is locate, proper tree-like mapping, etc.

ergo720: I instead deleted it because it feels superfluous to me, because there is already the function `GetInputDeviceName` which allows you to translate an xbox device type to its name which removes the need for the `input_support_list` array. 
> digging where device support list is locate

There's also `dev_num_buttons` in `InputManager.cpp` which gives it away, since it will have to have the correct number of buttons for each supported device.

src/core/hle/XAPI/Xapi.cpp
ergo720: `InState` will now always point to `CXBX_XINPUT_IN_STATE InState[port]`, which is instead allocated statically. Doing this allows to remove the need to reallocate it every time a new device is connected.

RadWolfie: Yeah, I left this as my "todo" to ask which I found it a bit later below comment.

CMakeLists.txt
ergo720: I didn't do that, it was done in the other the PR https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2068 and ended here when I reused it...

src/gui/DlgInputConfig.cpp
RadWolfie: > There's also `dev_num_buttons` in `InputManager.cpp` which gives it away, since it will have to have the correct number of buttons for each supported device.

Which require manual check to be sure number of inputs or the name itself is in right location. Except it is standalone mapping than tree-like mapping. It is still prone for mistakes anyway.

src/core/hle/XAPI/Xapi.cpp
RadWolfie: > Though, I'm not sure what's the reason behind having this here.

Answer: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2134#discussion_r572213541

CMakeLists.txt
RadWolfie: Yeah, I think this line was the original location until I moved variant gui controller(s) into its own folder in master branch for two reasons, easier to find list of controllers and ability to add them to input label. 😅

(This is resolve, feel free to mark as resolve after reading this.)

src/common/Settings.cpp
```diff
@@ -726,11 +752,11 @@ void Settings::SyncToEmulator()
 					return false;
 				});
 			if (it != m_input_profiles[m_input_port[i].Type].end()) {
-				char controls_name[XBOX_CTRL_NUM_BUTTONS][30];
+				char controls_name[HIGHEST_NUM_BUTTONS][30];
```
RadWolfie, line 755: One thing I notice from change higher up in this code from `char control_name[30];` to `char control_name[BUF_NAME_LENGTH];`. Which you have `BUF_NAME_LENGTH`, perhaps also apply here and strncpy couple lines below.

src/gui/DlgInputConfig.cpp
```diff
@@ -63,11 +65,11 @@ void SyncInputSettings(int port_num, int dev_type, bool is_opt)
 						return false;
 					});
 				if (it != g_Settings->m_input_profiles[dev_type].end()) {
-					char controls_name[XBOX_CTRL_NUM_BUTTONS][30];
+					char controls_name[HIGHEST_NUM_BUTTONS][30];
```
RadWolfie, line 68: Also here for `BUF_NAME_LENGTH` and couple lines below too.

RadWolfie (dismissed):
Few changes for same thing. Which I hope you'll consider update them in this pull request.

src/common/Settings.cpp
ergo720: I'll do it, but keep in mind that these two buffers hold the host button names, while the static assertion are against the xbox button names. So these two buffers could have a size other than 30, they just happen to have the same size.

RadWolfie: Oh, that's interesting... I didn't actually look where BUF_NAME_LENGTH came from.

RadWolfie: Okay, I see you added it. Doing so will need to sync other 30s to BUF_NAME_LENGTH for avoid confusion.

Did you only replaced couple of 30s to BUF_NAME_LENGTH? If so, then I think it's better to either do above or revert back to original and make a separate pull request to have 30s replace to BUF_NAME_LENGTH to stay in sync management purpose.

ergo720: I introduced two new macros `XBOX_BUTTON_NAME_LENGTH` and `HOST_BUTTON_NAME_LENGTH`, to make it clear that the buffers hold different names.

src/common/input/layout_xbox_device.h
```diff
@@ -0,0 +1,241 @@
...
+	IDC_BTN_MIDDLE_PEDAL,
+	IDC_BTN_RIGHT_PEDAL,
+	IDC_RADIO_TD_UP,
+	IDC_RADIO_TD_DOWN,
+	IDC_GEAR_UP,
+	IDC_GEAR_DOWN,
+};
+#endif
+
+inline constexpr const char* button_xbox_ctrl_names[XBOX_CTRL_NUM_BUTTONS] = {
```
RadWolfie, line 123: I don't understand the extra "const" when you already have constexpr here. Technically, all variables with constexpr at compile time are const. Reference: https://docs.microsoft.com/en-us/cpp/cpp/constexpr-cpp#constexpr-variables

ergo720: This is a constant array of constant char pointers. The `constexpr` makes the array constant, while the `const char` makes the chars constant. In addition, the array is used inside the function `check_button_name_size`, which must run at compile time because it's used in a static assertion. Without the `constexpr`, the array is not guaranteed to be initialized at compile time and the compiler throws an error.

RadWolfie: > The confusion arises because constexpr is placed to the left of the pointer, instead of the right like you would do with const

I see, seems like a design flaw for human readability and a brand new playing field to understand the difference for where constexpr is placed vs where it is actually will be place in different position. *slap dislike button*

RadWolfie:
LGTM. Although, I'm requesting testers to test with their Steel Battalion titles for any regression or in need of change.

Plus general testers to verify any of their titles made any regressions.

LanHikariDS:
Aside from not being a fan of the less fancy UI (I suppose that's what I get for slacking), I do have some issues with the simplified Tuner Dial and Gear Lever. While having an easy Gear/Tuner Up/Down was requested in the Steel Battalion LoC Discord, I think the options for direct setting should be available as well to maintain compatibility with binding real SB controllers, though investigations are ongoing on the current SBC drivers, as I can't get either of those inputs to even work in the driver at the moment.

RadWolfie:
It appears Steel Battalion controller owners are using vJoy with SteelBattalion64 driver. Physical SB controller with vJoy will show device name to`SDL/0/vJoy Device`.

After reviewing from list of source files...
 - [vJoy.cs](https://sourceforge.net/p/steel-batallion-64/code/HEAD/tree/trunk/%20steel-batallion-64/Steel_Batallion_64_v2/vJoy.cs#l342) (Steel Battalion controller driver)
 - [vJoyInterface.cpp](https://github.com/shauleiz/vJoy/blob/8111294ef89b25d35fa2617a62ce96d23275ab16/apps/common/vJoyInterface.cpp#L879-L883) (vJoy kernel driver interface)
 - [vJoyDevice.cpp](https://github.com/shauleiz/vJoy/blob/8111294ef89b25d35fa2617a62ce96d23275ab16/apps/vJoyConf/vJoyDevice.cpp#L54-L55)
   - There's a limit to 2 slider support for some reason.
 - [SDL_dinputjoystick.c](https://github.com/SDL-mirror/SDL/blob/effed5b7539ffb65d9611998be1f6a7aaa1a359f/src/joystick/windows/SDL_dinputjoystick.c#L784) (SDL's windows dinput joystick)
 - [SDL_joystick.h](https://github.com/SDL-mirror/SDL/blob/master/include/SDL_joystick.h) (Official interface from SDL's joystick)
   - Although, it does have `SDL_JoystickGetAxisInitialState` function interface. Which can be useful for remember its last state value from actual Steel Battalion controller to virtual SB controller. (Our current codebase doesn't have it.)
 - [Windows' DInput const type input](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ee418634)

From what I have seen, almost everything is there except for certain limitation, from vJoy device driver, and missing get initial axis state value. Plus like @ergo720 said in original post about mouse's directional preservation needed for Steel Battalion titles and Virtua Cop 3.

src/core/hle/XAPI/Xapi.h
```diff
@@ -145,22 +146,24 @@ typedef struct _XINPUT_GAMEPAD
...
-    xbox::short_xt   sRotationLever;//maybe only high byte was used.
-    xbox::short_xt   sSightChangeX;
-    xbox::short_xt   sSightChangeY;
-    xbox::word_xt    wLeftPedal;//maybe only high byte was used.
-    xbox::word_xt    wMiddlePedal;//maybe only high byte was used.
-    xbox::word_xt    wRightPedal;//maybe only high byte was used.
-    xbox::uchar_xt   ucTunerDial;//low nibble, The 9 o'clock postion is 0, and the 6 o'clock position is 12. The blank area between the 6 and 9 o'clock positions is 13, 14, and 15 clockwise.
-    xbox::uchar_xt   ucGearLever;//GearLever 1~5 for gear 1~5, 7~13 for gear R,N,1~5, 15 for gear R. we use the continues range from 7~13
+    xbox::ushort_xt  sAimingX;
+    xbox::ushort_xt  sAimingY;
```
RadWolfie, line 154: For what reason is aiming for X and Y change to unsigned? They should be signed integer (short).

RadWolfie: > Answer: https://sourceforge.net/p/steel-batallion-64/code/HEAD/tree/trunk/%20steel-batallion-64/Steel_Batallion_64_v2/SteelBattalionController.cs#l644

Huh, that's interesting...

RadWolfie (approved):
LGTM

ergo720:
> I think the options for direct setting should be available as well to maintain compatibility with binding real SB controllers

Impossible to do with the current input system design, and for a good reason. The emulated SBC has to work with any host device the user might decide to bind, not just with a real SBC the user might happen to have. This is achieved by using the abstract class `InputDevice`, which then the API specific classes inherit from and actually implement a real host device controlled by that API. Every one of those classes have a `GetState` function to retrieve the input state of its controls, which is always reported as `double` number normalized in the range 0.0 (not pressed) and 1.0 (pressed at full intensity). This way, the emulated xbox device doesn't need to care about what is the bound host device and can just call `GetState` and convert the state to the xbox range appropriate for the bound xbox button, and it will always work regardless of the underlying host device or API being used, instead of having to have separate cases for the different host device ranges / APIs. Applying this to the SBC, it means that the gear/tuner value will be converted to a number in the range [0.0 - 1.0], which then `UpdateInputSBC` will convert to the relevant xbox range, the raw value will never be plugged in directly as is to the `SBCInput` struct. To do it, I would need to break the design and add a special case just for the real SBC, which is what the design wants to avoid. Instead, the best bet to support it directly is to implement USB pass-through support for it, like this comment [here](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1390#issuecomment-413157224) says.

PR #2135, opened 2021-02-11
Add support for mouse input as cursor position relative to the rendering window

ergo720:
This PR adds a new mouse input mode which reports its state as the cursor position relative to the rendering window, as opposed to mouse relative motion (the only mode available for the mouse right now). In the input GUI, this is represented with the "Cursor X/Y+/-" name when bound to a button, instead of the "Axis X/Y/Z+/-" name used otherwise. Users can toggle between the two modes by SHIFT - right clicking the corresponding button (will only work if the KeyboardMouse device is selected in the device list). Using this mode fixes the re-center issue of both the crosshair of Virtua Cop 3 and the sticks of Steel Battalion.

### DIFF
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -947,12 +947,9 @@ void Settings::RemoveLegacyConfigs(unsigned int CurrentRevision)
 			std::string current_section = std::string(section_input_port) + std::to_string(port_num);
 			std::string device_name = m_si.GetValue(current_section.c_str(), sect_input_port.device, "");
 
-			// NOTE: with C++20, this can be simplified by simply calling device_name.ends_with()
-			if (device_name.length() >= kb_str.length()) {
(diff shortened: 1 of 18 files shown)


legend800:
Just tried VC3 and it worked perfect. Finally a true arcade experience now. No issues for me on Win10. 

The only note is it's a bit weird to configure - you have to set up the mouse for all directions, then hold shift while right clicking on the buttons...little awkward but a one time setup.

RadWolfie:
Hmm, after testing the tool tip window. I notice it disappear quite bit too early. I took a look at the doc and notice there's a 5 seconds duration of auto popup window shown. I think we should increase this to around 15 seconds.

ref: https://docs.microsoft.com/en-us/windows/win32/controls/ttm-setdelaytime

Beside that, I notice "+ shift" is after right-click. In my opinion, it should follow like other standard use which is "<hold down key name here> + right-click".

RadWolfie (approved):

Issue #1145, opened 2018-05-09
Allocate and AllocateZeroed memory functions


ghost:
These two functions are currently used in various places in the code and I have seen that they usually request very small amounts of memory so, perhaps, the pool manager is more indicated in those cases since it's designed to allocate small amounts of memory. Also, why are they needed in the first place? They don't exist as Xbox APIs, they are a construct from us, could they be dropped now?

PatrickvL:
I agree, good thinking

LukeUsher:
They are used for when we need to allocate memory in the Xbox address space, rather than the host address space, allocations that the VM Manager can keep track of. They could be replaced with Kernel APIs, but they do need to be tracked by the VM Manager.

PR #2140, opened 2021-02-17
Removed redundant allocation functions in VMManager

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1145. This removes the redundant functions `Allocate(Zeroed)` and `Deallocate` and replaces them with calls to `ExAllocatePool` and `ExFreePool` respectively. Other than that, this should not have any effect on any games, but testing is needed to be certain of that.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -1269,7 +1269,7 @@ int GetD3DResourceRefCount(IDirect3DResource *EmuResource)
 /*
 xbox::X_D3DSurface *EmuNewD3DSurface()
 {
-	xbox::X_D3DSurface *result = (xbox::X_D3DSurface *)g_VMManager.AllocateZeroed(sizeof(xbox::X_D3DSurface));
+	xbox::X_D3DSurface *result = (xbox::X_D3DSurface *)xbox::ExAllocatePool(sizeof(xbox::X_D3DSurface));
 	result->Common = X_D3DCOMMON_D3DCREATED | X_D3DCOMMON_TYPE_SURFACE | 1; // Set refcount to 1
 	return result;
 }
--- a/src/core/hle/XACTENG/XactEng.cpp
+++ b/src/core/hle/XACTENG/XactEng.cpp
@@ -34,7 +34,6 @@
 #include "core\kernel\support\Emu.h"
 #include "EmuShared.h"
 #include "core\hle\XACTENG\XactEng.h"
-#include "core\kernel\memory-manager\VMManager.h"
 
 #include <mmreg.h>
 #include <msacm.h>
@@ -65,7 +64,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(XACTEngineCreate)
 
 	// TODO: Any other form of initialization?
 
-	*ppEngine = (X_XACTEngine*)g_VMManager.AllocateZeroed(sizeof( X_XACTEngine ));
+	*ppEngine = (X_XACTEngine*)ExAllocatePool(sizeof( X_XACTEngine ));
 
 		
 	
@@ -109,7 +108,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_RegisterWaveBank)
 
 	// TODO: Implement
 
-	*ppWaveBank = (X_XACTWaveBank*)g_VMManager.AllocateZeroed(sizeof( X_XACTWaveBank ));
+	*ppWaveBank = (X_XACTWaveBank*)ExAllocatePool(sizeof( X_XACTWaveBank ));
 
 	RETURN(S_OK);
 }
@@ -132,7 +131,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_RegisterStreamedWaveBank)
 
 	// TODO: Implement
 
-	*ppWaveBank = (X_XACTWaveBank*)g_VMManager.AllocateZeroed(sizeof( X_XACTWaveBank ));
+	*ppWaveBank = (X_XACTWaveBank*)ExAllocatePool(sizeof( X_XACTWaveBank ));
 
 	RETURN(S_OK);
 }
@@ -157,7 +156,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_CreateSoundBank)
 
 	// TODO: Implement
 
-	*ppSoundBank = (X_XACTSoundBank*)g_VMManager.AllocateZeroed(sizeof( X_XACTSoundBank ));
+	*ppSoundBank = (X_XACTSoundBank*)ExAllocatePool(sizeof( X_XACTSoundBank ));
 
 	RETURN(S_OK);
 }
@@ -203,7 +202,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_CreateSoundSource)
 		LOG_FUNC_ARG(ppSoundSource)
 		LOG_FUNC_END;
 
-	*ppSoundSource = (X_XACTSoundSource*)g_VMManager.AllocateZeroed(sizeof( X_XACTSoundSource ));
+	*ppSoundSource = (X_XACTSoundSource*)ExAllocatePool(sizeof( X_XACTSoundSource ));
 
 	RETURN(S_OK);
 }
@@ -500,7 +499,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_UnRegisterWaveBank)
 	// to IXACTWaveBank is released.
 
 //	if(pWaveBank)
-//		g_VMManager.Deallocate((VAddr)pWaveBank);
+//		ExFreePool(pWaveBank);
 
 	RETURN(S_OK);
 }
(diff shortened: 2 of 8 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1269,7 +1269,7 @@ int GetD3DResourceRefCount(IDirect3DResource *EmuResource)
 /*
 xbox::X_D3DSurface *EmuNewD3DSurface()
 {
-	xbox::X_D3DSurface *result = (xbox::X_D3DSurface *)g_VMManager.AllocateZeroed(sizeof(xbox::X_D3DSurface));
+	xbox::X_D3DSurface *result = (xbox::X_D3DSurface *)xbox::ExAllocatePool(sizeof(xbox::X_D3DSurface));
```
PatrickvL, line 1272: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

src/core/hle/XACTENG/XactEng.cpp
```diff
@@ -65,7 +64,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(XACTEngineCreate)
 
 	// TODO: Any other form of initialization?
 
-	*ppEngine = (X_XACTEngine*)g_VMManager.AllocateZeroed(sizeof( X_XACTEngine ));
+	*ppEngine = (X_XACTEngine*)ExAllocatePool(sizeof( X_XACTEngine ));
```
PatrickvL, line 67: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

src/core/hle/XACTENG/XactEng.cpp
```diff
@@ -109,7 +108,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_RegisterWaveBank)
 
 	// TODO: Implement
 
-	*ppWaveBank = (X_XACTWaveBank*)g_VMManager.AllocateZeroed(sizeof( X_XACTWaveBank ));
+	*ppWaveBank = (X_XACTWaveBank*)ExAllocatePool(sizeof( X_XACTWaveBank ));
```
PatrickvL, line 111: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

src/core/hle/XACTENG/XactEng.cpp
```diff
@@ -132,7 +131,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_RegisterStreamedWaveBank)
...
```
PatrickvL, line 134: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

src/core/hle/XACTENG/XactEng.cpp
```diff
@@ -157,7 +156,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_CreateSoundBank)
 
 	// TODO: Implement
 
-	*ppSoundBank = (X_XACTSoundBank*)g_VMManager.AllocateZeroed(sizeof( X_XACTSoundBank ));
+	*ppSoundBank = (X_XACTSoundBank*)ExAllocatePool(sizeof( X_XACTSoundBank ));
```
PatrickvL, line 159: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

src/core/hle/XACTENG/XactEng.cpp
```diff
@@ -203,7 +202,7 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IXACTEngine_CreateSoundSource)
 		LOG_FUNC_ARG(ppSoundSource)
 		LOG_FUNC_END;
 
-	*ppSoundSource = (X_XACTSoundSource*)g_VMManager.AllocateZeroed(sizeof( X_XACTSoundSource ));
+	*ppSoundSource = (X_XACTSoundSource*)ExAllocatePool(sizeof( X_XACTSoundSource ));
```
PatrickvL, line 205: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

src/core/kernel/support/EmuFile.cpp
```diff
@@ -771,7 +771,7 @@ NtDll::FILE_LINK_INFORMATION * _XboxToNTLinkInfo(xbox::FILE_LINK_INFORMATION *xb
 
 	// Build the native FILE_LINK_INFORMATION struct
 	*Length = sizeof(NtDll::FILE_LINK_INFORMATION) + convertedFileName.size() * sizeof(wchar_t);
-	NtDll::FILE_LINK_INFORMATION *ntLinkInfo = (NtDll::FILE_LINK_INFORMATION *) g_VMManager.AllocateZeroed(*Length);
+	NtDll::FILE_LINK_INFORMATION *ntLinkInfo = (NtDll::FILE_LINK_INFORMATION *)xbox::ExAllocatePool(*Length);
```
PatrickvL, line 774: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

src/core/kernel/support/EmuFile.cpp
```diff
@@ -791,7 +791,7 @@ NtDll::FILE_RENAME_INFORMATION * _XboxToNTRenameInfo(xbox::FILE_RENAME_INFORMATI
 
 	// Build the native FILE_RENAME_INFORMATION struct
 	*Length = sizeof(NtDll::FILE_RENAME_INFORMATION) + convertedFileName.size() * sizeof(wchar_t);
-	NtDll::FILE_RENAME_INFORMATION *ntRenameInfo = (NtDll::FILE_RENAME_INFORMATION *) g_VMManager.AllocateZeroed(*Length);
+	NtDll::FILE_RENAME_INFORMATION *ntRenameInfo = (NtDll::FILE_RENAME_INFORMATION *)xbox::ExAllocatePool(*Length);
```
PatrickvL, line 794: Here, I see no call to `xbox::RtlZeroMemory` applied to the return result, while that *is* done in other places; Have you verified this call to not require zeroing?

PatrickvL (commented):
Sorry for the repeats, but it's the same observation in multiple places, and I didn't want to simple say "here too".

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
ergo720: This code is disabled by a comment so it doesn’t matter right now.

src/core/hle/XACTENG/XactEng.cpp
ergo720: X_XACTEngine is an empty struct of 1 byte size so I don’t think it’s necessary.

src/core/hle/XACTENG/XactEng.cpp
ergo720: X_XACTWaveBank is an empty struct of 1 byte size so I don’t think it’s necessary.

src/core/hle/XACTENG/XactEng.cpp
ergo720: X_XACTWaveBank is an empty struct of 1 byte size so I don’t think it’s necessary.

src/core/hle/XACTENG/XactEng.cpp
ergo720: X_XACTSoundBank is an empty struct of 1 byte size so I don’t think it’s necessary.

src/core/hle/XACTENG/XactEng.cpp
ergo720: X_XACTSoundSource is an empty struct of 1 byte size so I don’t think it’s necessary.

src/core/kernel/support/EmuFile.cpp
ergo720: All members of the FILE_LINK_INFORMATION struct are assigned just below so I don’t think it’s necessary.

src/core/kernel/support/EmuFile.cpp
ergo720: All members of the FILE_RENAME_INFORMATION struct are assigned just below so I don’t think it’s necessary.

ergo720:
The memory returned by `xbox::ExAllocatePool` is not guaranteeded to be zero so you have to call `xbox::RtlZeroMemory` if you want that. In the other cases it's done already because there is a comment which explicitly says that the memory must be zeroed: `Clear, to prevent side-effects on random contents`

PatrickvL (approved):
Thanks for the clarifications.

If you think of any lingering issue in this area, please create a ticket for it, so it won't be forgotten, and can be tracked.

Thanks for the fix; Approved!

medievil1:
tested with my current rips and didn't find any difference from previous so, seems good to me

RadWolfie (approved):

PatrickvL:
> This commit seems to make Intel GPA fail to attach to CXBX for graphics debugging.
> GPA is flakey as all hell, so I don't know if this behaviour means anything about the correctness of this change

@ergo720 suggested to.... "Try to use NtAllocateVirtualMemory, that one allocates in the user region just like the old Allocate(Zero) functions"

ergo720:
Resolved in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2171

PR #2141, opened 2021-02-20
Hide Sensitive Fields from EEPROM gui

### LINKED ISSUE #2139: Feature: Hide Sensitive Fields From EEPROM gui
## Quick summary
Proposal change options for EEPROM's gui window.
1. Add an option for all sensitive fields to reveal from ONE checkbox.
2. Add checkbox next to each sensitive field. (I think all commercial products do use option 2.)
### **or**
3. Reveal field value when it is focused then hide again once unfocused. (easier to manage without need of extra toggle)

## Details
Since we have sensitive fields in EEPROM gui window, by default we don't censor them. Once user report an issue with relative gui window(s), sometimes those screenshots doesn't have censor fields. Either they do not know how to censor them out or don't know if they are meant to be viewed only from their side (without leaking).

See quick summary of proposal change options.

The following fields are critical functional to networking service through LAN or third-party online service.
- HDD Key
- Serial Number
- Online Key
- MAC address

Meanwhile Parental Control Password is not intended to be reveal on the Internet since it fall under personal information privacy.

### System Configuration:
N/A

### Additional Information (if any):
N/A

RadWolfie:
Due to common mistakes occur from community revealed their own information and some users had to blur their sensitive info which can take time. Since we already know auto-generated are working fine for ages. This pull request introduce hide sensitive info from the Internet and from our discord's help support channel. Regardless if they are auto-generated, dumped from actual hardware, or manual input of their own preference. Following fields are set as sensitive fields and automatic hidden by default.
- HDD Key
- Serial Number
- MAC address
- Online Key
- Parental Control Password (always hide)

With parental control password as always hidden, in my opinion should remain this way even on focused field. Feedback is welcome if requested otherwise for this field.

---

I had tried 0x25CF which is solid big dot except it is rendered at different character. Even after attempt to use both SendMessageA/W does not resolve the issue.

Due to funky side-effect with using solid dot for password's character. It is best to have `ES_PASSWORD` set on one of the fields then send message with `EM_GETPASSWORDCHAR` to obtain proper value. Finally, it will properly able to use the correct value with `EM_SETPASSWORDCHAR`.

resolve #2139


x1nixmzeng:
This seems completely redundant - why would you not want to show your own config options?

RadWolfie:
> This seems completely redundant - why would you not want to show your own config options?

What do you mean by redundant? The code is very neat and simple to understand how it works.

The OP and the ticket reference explained the reason.

x1nixmzeng:
Your code is fine but I think this is a non-issue. I've left a comment on the bug.

PatrickvL:
It's to protect sensitive information when people post screenshots of the EEPROM edit window.

LukeUsher (approved):

PR #2146, opened 2021-02-25
Implement IsEmuHandle by keeping a lookup of EmuHandle objects

NZJenkins:
Fixes a crash in Amped on loading a map after having exited one
Amped requests an invalid handle 0xFDFDFDFD
IsEmuHandle return a false positive and we'd crash trying to call NtClose()

### DIFF
--- a/src/core/kernel/exports/EmuKrnlNt.cpp
+++ b/src/core/kernel/exports/EmuKrnlNt.cpp
@@ -142,7 +142,7 @@ XBSYSAPI EXPORTNUM(187) xbox::ntstatus_xt NTAPI xbox::NtClose
 
 	NTSTATUS ret = xbox::status_success;
 
-	if (IsEmuHandle(Handle))
+	if (EmuHandle::IsEmuHandle(Handle))
 	{
 		// delete 'special' handles
 		EmuHandle *iEmuHandle = HandleToEmuHandle(Handle);
@@ -682,7 +682,7 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
 
 	NTSTATUS ret = xbox::status_success;
 
-	if (IsEmuHandle(SourceHandle)) {
+	if (EmuHandle::IsEmuHandle(SourceHandle)) {
 		EmuHandle* iEmuHandle = HandleToEmuHandle(SourceHandle);
 		ret = iEmuHandle->NtDuplicateObject(TargetHandle, Options);
 /*
@@ -739,7 +739,7 @@ XBSYSAPI EXPORTNUM(198) xbox::ntstatus_xt NTAPI xbox::NtFlushBuffersFile
 		LOG_FUNC_END;
 	NTSTATUS ret = xbox::status_success;
 	
-	if (IsEmuHandle(FileHandle)) 
+	if (EmuHandle::IsEmuHandle(FileHandle))
 		LOG_UNIMPLEMENTED();
 	else
 		ret = NtDll::NtFlushBuffersFile(FileHandle, (NtDll::IO_STATUS_BLOCK*)IoStatusBlock);
--- a/src/core/kernel/exports/EmuKrnlNt.cpp
+++ b/src/core/kernel/exports/EmuKrnlNt.cpp
@@ -1371,11 +1371,14 @@ XBSYSAPI EXPORTNUM(215) xbox::ntstatus_xt NTAPI xbox::NtQuerySymbolicLinkObject
 		LOG_FUNC_ARG_OUT(ReturnedLength)
 		LOG_FUNC_END;
 
-	NTSTATUS ret = 0;
+	NTSTATUS ret = STATUS_INVALID_HANDLE;
 	EmuNtSymbolicLinkObject* symbolicLinkObject = NULL;
 
-	// Check that we actually got an EmuHandle :
-	ret = STATUS_INVALID_HANDLE;
+	// We expect LinkHandle to always be an EmuHandle
+	if (!EmuHandle::IsEmuHandle(LinkHandle)) {
+		LOG_UNIMPLEMENTED();
+		return ret;
+	}
 
 	auto iEmuHandle = (EmuHandle*)LinkHandle;
 	// Retrieve the NtSymbolicLinkObject and populate the output arguments :
--- a/src/core/kernel/exports/EmuKrnlNt.cpp
+++ b/src/core/kernel/exports/EmuKrnlNt.cpp
@@ -145,7 +145,7 @@ XBSYSAPI EXPORTNUM(187) xbox::ntstatus_xt NTAPI xbox::NtClose
 	if (EmuHandle::IsEmuHandle(Handle))
 	{
 		// delete 'special' handles
-		EmuHandle *iEmuHandle = HandleToEmuHandle(Handle);
+		auto iEmuHandle = (EmuHandle*)Handle;
 		ret = iEmuHandle->NtClose();
 
 		LOG_UNIMPLEMENTED(); // TODO : Base this on the Ob* functions
@@ -683,7 +683,7 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
 	NTSTATUS ret = xbox::status_success;
 
 	if (EmuHandle::IsEmuHandle(SourceHandle)) {
-		EmuHandle* iEmuHandle = HandleToEmuHandle(SourceHandle);
+		auto iEmuHandle = (EmuHandle*)SourceHandle;
 		ret = iEmuHandle->NtDuplicateObject(TargetHandle, Options);
 /*
 		PVOID Object;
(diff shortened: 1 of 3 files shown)


src/core/kernel/support/EmuFile.cpp
```diff
@@ -255,9 +255,50 @@ EmuHandle::EmuHandle(EmuNtObject* ntObject)
...
+		EmuHandleLookup.erase(EmuHandleToHandle(this));
+		EmuHandleLookupLock.unlock();
+	}
+
+	return status;
+}
+
+bool EmuHandle::IsEmuHandle(HANDLE Handle)
+{
+	auto iter = EmuHandleLookup.find(Handle);
```
PatrickvL, line 288: Is `EmuHandleLookup.find` thread-safe? If not, here too `EmuHandleLookupLock.lock` / `.unlock` calls must wrap this code.

LukeUsher (dismissed):
Looks great, thanks

src/core/kernel/support/EmuFile.cpp
```diff
@@ -255,9 +255,50 @@ EmuHandle::EmuHandle(EmuNtObject* ntObject)
...
 }
 
+std::unordered_map<HANDLE, EmuHandle*> EmuHandle::EmuHandleLookup = {};
+std::mutex EmuHandle::EmuHandleLookupLock = {};
+
+EmuHandle* EmuHandle::CreateEmuHandle(EmuNtObject* ntObject) {
+	auto emuHandle = new EmuHandle(ntObject);
+
+	// Register EmuHandle
+	EmuHandleLookupLock.lock();
```
LukeUsher, line 265: Why do you manually lock/unlock rather than using std::lock_guard?

src/core/kernel/support/EmuFile.cpp
```diff
@@ -255,9 +255,50 @@ EmuHandle::EmuHandle(EmuNtObject* ntObject)
 	NtObject = ntObject;
 }
 
+std::unordered_map<HANDLE, EmuHandle*> EmuHandle::EmuHandleLookup = {};
+std::mutex EmuHandle::EmuHandleLookupLock = {};
```
PatrickvL, line 259: Would it be possible to have this zero-initialization as part of the class declaration?

src/core/kernel/support/EmuFile.cpp
NZJenkins: what's a std::lock_guard? :P I'll check it out

src/core/kernel/support/EmuFile.h
```diff
@@ -160,10 +162,21 @@ NTSTATUS CxbxConvertFilePath(std::string RelativeXboxPath, OUT std::wstring &Rel
...
+	static bool IsEmuHandle(HANDLE Handle);
+private:
+	EmuHandle(EmuNtObject* ntObject);
+	// Keep track of every EmuHandle wrapper
+	// If we remember what EmuHandle objects we hand out, we can tell the difference between an EmuHandle and garbage
+	// We used to rely on the high bit to differentiatean EmuHandles
+	// But titles may attempt to operate on invalid handles with the high bit set
+	// Test case: Amped sets a handle value to 0xFDFDFDFD (coincidentally a VS debugger guard value)
+	static std::unordered_map<HANDLE, EmuHandle*> EmuHandleLookup;
+	static std::mutex EmuHandleLookupLock;
```
PatrickvL, line 179: So if I understand this correctly, each `class EmuHandle` will use the same `EmuHandleLookup` and `EmuHandleLookupLock`, because of the `static` declaration, right? If so, then all access to it must be thread-safe, and especially the `EmuHandleLookupLock` has to be reentrant (allow calls from multiple threads and not break because of concurrency). But it is, right?

src/core/kernel/support/EmuFile.cpp
```diff
@@ -255,9 +255,50 @@ EmuHandle::EmuHandle(EmuNtObject* ntObject)
...
+{
+	return (EmuHandle*)((uint32_t)Handle & 0x7FFFFFFF);
+}
+
+HANDLE EmuHandleToHandle(EmuHandle* emuHandle)
+{
+	// Set the high bit
+	// Avoids collisions with real Xbox handles..?
+	return (HANDLE)((uint32_t)emuHandle | 0x80000000);
 }
```
PatrickvL, line 302: I may be mistaken, but now that you keep track of all EmuHandle's through the `EmuHandleLookup`, there's no real need anymore for these two functions (`HandleToEmuHandle` and `EmuHandleToHandle`), because `IsEmuHandle` can make the distinction without messing with the high bit.

What would happen if these two functions were simply deleted?

PatrickvL (commented):
I guess renaming EmuHandle into XboxHandle would be too much of a change for this?

Also, it'd be better to turn it around, and not track Xbox handles, but instead start tracking host (Windows) handles, but I guess that'd an even bigger change, right?

I'll hold off my approval until the review-questions are cleared up, if you don't mind.

Good job still, ofc! 👍

src/core/kernel/support/EmuFile.cpp
PatrickvL: I guess it cannot be using a guard, because the same `EmuHandleLookupLock` needs to be locked in the other function as well.

src/core/kernel/support/EmuFile.cpp
NZJenkins: Tried to but C++ didn't seem to like it. I think somewhere I read only `int` and `enum` can be static const class members? 🤷

src/core/kernel/support/EmuFile.cpp
NZJenkins: per my vague comment, I _assume_ since they were written this way the Xbox kernel won't create handles with the high bit set - so if anything it lowers the chance of collision?

src/core/kernel/support/EmuFile.cpp
PatrickvL: Not even the `->` syntax works? I don't know.

But if nothing else works, then at least add a comment, please, explaining why it's zero-initialized here, so that if someone figures out a way some day in the future, it'll be clear that this was the original intent of this code anyway.

src/core/kernel/support/EmuFile.cpp
PatrickvL: No, I wrote this shit way back when, and I'm fairly certain abusing the high bit was just done so that we could discern Xbox from Windows handles. So if the lookup changes that assumption, there's no need to mess with the high bit anymore.

src/core/kernel/support/EmuFile.cpp
PatrickvL: On Discord, @LukeUsher suggested a `shared_mutex`, which would be even better (as it will also allow the distinction between read and write locking)

src/core/kernel/support/EmuFile.cpp
NZJenkins: it is not! added a lock

PatrickvL:
Actually, this might solve a lot of thus-far unexplained crashes, and by that, raise our compatibility scores quite a bit!

src/core/kernel/support/EmuFile.cpp
NZJenkins: dropped it 👍

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -1377,7 +1377,7 @@ XBSYSAPI EXPORTNUM(215) xbox::ntstatus_xt NTAPI xbox::NtQuerySymbolicLinkObject
 	// Check that we actually got an EmuHandle :
 	ret = STATUS_INVALID_HANDLE;
 
-	EmuHandle* iEmuHandle = HandleToEmuHandle(LinkHandle);
+	auto iEmuHandle = (EmuHandle*)LinkHandle;
```
PatrickvL, line 1380: Hm, now that you've changed this, perhaps here a check on `if (EmuHandle::IsEmuHandle(LinkHandle))` is needed as well, especially give there's a comment above stating 'Check that we actually got an EmuHandle'... Would you agree that'd be safer here?

PatrickvL (commented):
One more thing to look into, if you please.

src/core/kernel/support/EmuFile.cpp
PatrickvL: If you're not going to add a comment here, then just resolve this conversation - I won't keep bringing it up 😛

src/core/kernel/exports/EmuKrnlNt.cpp
NZJenkins: what do we do if it's not 😕

PatrickvL: Looking at `NtOpenSymbolicLinkObject` this won't happen, so I guess it'll suffice (for now) to call `LOG_UNIMPLEMENTED();`

src/core/kernel/support/EmuFile.cpp
NZJenkins: There's no intent tbh. I don't understand the significance 😁
I'll delete the initialise atuff?

src/core/kernel/exports/EmuKrnlNt.cpp
NZJenkins: done

PatrickvL (approved):

PR #2151, opened 2021-03-04
Use the backbuffer format requested by the title

NZJenkins:
Fixes NBA Live 2K5 alpha blending issue

Seems the previous behaviour ensured the backbuffer format matched the display mode in windowed mode - this might be a holdover from D3D8 🤷‍♂️ 
> https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dpresent-parameters
> For windowed applications, the back buffer format no longer needs to match the display-mode format

The _host_ backbuffer is now unspecified.
The _xbox_ backbuffer is now whatever format the title requested

- Remove code that aligns the host backbuffer with the xbox backbuffer
- Remove code related to setting the backbuffer format
- Always convert D3DFMT_A8 to ARGB
In D3D8, the RGB values appear to be undefined, and the Xbox uses 1.
Whereas in D3D9, the RGB values are 0

Before:
image

After:
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2182,8 +2182,9 @@ static void SetupPresentationParameters
     // TODO: Investigate the best option for this
     params.SwapEffect = D3DSWAPEFFECT_COPY;
 
-    // Attempt to match backbuffer format, this is not *required*, but leads to faster blitting/swapping
-    params.BackBufferFormat = EmuXB2PC_D3DFormat(pXboxPresentationParameters->BackBufferFormat);
+    // Any backbuffer format should do, since we render to a separate xbox backbuffer
+    // Rather than directly to the host backbuffer
+    params.BackBufferFormat = D3DFMT_UNKNOWN;
 
     params.PresentationInterval = g_XBVideo.bVSync ? D3DPRESENT_INTERVAL_ONE : D3DPRESENT_INTERVAL_IMMEDIATE;
     g_Xbox_PresentationInterval_Default = pXboxPresentationParameters->PresentationInterval;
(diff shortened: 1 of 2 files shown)


RadWolfie:
> Does this make the GUI defined format obsolete? Perhaps that needs some cleanup then.

While I agree with the cleanup for both settings and GUI's end. However, I personally believe they should leave intact for at least 6 months. Then do the cleanup afterward for the sake of backward compatibility between today and up to 6 months. Instead of do the cleanup in here, let's add a TODO note for the future cleanup.

RadWolfie:
Wait a minute... There's 5 total from the format string in sscanf. Something we didn't do something right.

ESPPIRAL:
Everything I've tested (AMD GPU) crashes with the following message.

image

ESPPIRAL:
Upon further testing it seems it only happens if you select display resolution other than the automatic (default xbox)

ESPPIRAL:
Selecting exclusive full screen mode doesn't seem to crash even if you select an arbitrary display resolution

medievil1:
fix for CreateDeviceEX failing seems to be to add add
```params.FullScreen_RefreshRateInHz = 0;``` 
after ``` params.BackBufferFormat = EmuXB2PC_D3DFormat(pXboxPresentationParameters->BackBufferFormat);``` 
all the games I tried that were failing run with it in place

medievil1:
ok verified, some games won't run with this, even with refresh rate set... question is, does PC support all xbox backbuffer formats or is an unsupported format trying to get set and causing a crash back to UI??

medievil1:
yea confused emoji...lol.. helps if you continue reading down the code, since right under it checks formats ..so,,it is something else

medievil1:
though backup up a bit, I see now the change is a bit redundant, it was already there, first thing it tries to match
```    // Attempt to match backbuffer format, this is not *required*, but leads to faster blitting/swapping
    params.BackBufferFormat = EmuXB2PC_D3DFormat(pXboxPresentationParameters->BackBufferFormat);``` so it SHOULD have already worked ...hmmmm

medievil1:
working as it should!

medievil1:
ooops found a typo.... line 916 should be X_D3DFMT_A8 ->  not X_D3DFMT_LIN_A8 ->

PatrickvL:
@ESPPIRAL is your original test result (everything crashes on AMD GPU) resolved?

PatrickvL (approved):
The code looks fine, but let's postpone merging this until @ESPPIRAL confirms it works for him too.

Btw, can we somehow confirm that Xbox reads RGB as 1 from A8 textures?

In the future, when all texture sampling is under our control,  we should really perform channel swizzling and defaults in the shader, so that we require much less format conversions.

ESPPIRAL:
I've tested a few games, no more crashes.

RadWolfie:
While I, along with another contributor, was making a new feature. I notice exclusive full screen will crash on start. I also had verified previous merged pull request, before this merged pull request, is working as intended.

PR #2153, opened 2021-03-05
Measure the string length to get the correct size the string should be

NZJenkins:
Fix accidental use of string constructor with length, that causes null bytes to be included in the string


CookiePLMonster:
To be fair `strlen` returns the number of characters **without** the null terminator, so is this really the real issue?

CookiePLMonster (approved):
I can't read.

LGTM

PR #944, opened 2018-02-24
Format partitions by emptying the contents rather than deleting the folder

LukeUsher:
This is intended to fix #942

PR #2155, opened 2021-03-06
Fix partition formatting

CookiePLMonster:
1. Fixes a long standing bug introduced by #944 where partition formatting occassionally failed, resulting either in errors printed to the log or in a game crash. In my case, I often had to re-run the game multiple times to get partitions to format properly.
2. Modernizes `CxbxGetPartitionNumberFromHandle` and `CxbxGetPartitionDataPathFromHandle` to use Unicode and support long paths properly. This saves on internal Unicode -> ANSI -> Unicode conversions which could potentially also be lossy.

### DIFF
--- a/src/core/kernel/support/EmuFile.cpp
+++ b/src/core/kernel/support/EmuFile.cpp
@@ -175,17 +175,12 @@ void CxbxFormatPartitionByHandle(HANDLE hFile)
 
 	// Format the partition, by iterating through the contents and removing all files/folders within
 	// Previously, we deleted and re-created the folder, but that caused permission issues for some users
-	try
-	{
-		for (auto& directoryEntry : std::filesystem::directory_iterator(partitionPath)) {
-			std::filesystem::remove_all(directoryEntry);
+	std::error_code er;
+	for (const auto& directoryEntry : std::filesystem::directory_iterator(partitionPath, er)) {
+		if (!er) {
+			std::filesystem::remove_all(directoryEntry, er);
 		}
 	}
-	catch (std::filesystem::filesystem_error fsException)
-	{
-		printf("std::filesystem failed with message: %s\n", fsException.what());
-	}
-
 
 	printf("Formatted EmuDisk Partition%d\n", CxbxGetPartitionNumberFromHandle(hFile));
 }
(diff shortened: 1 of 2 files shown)


LukeUsher (approved):

src/core/kernel/support/EmuFile.cpp
```diff
@@ -132,60 +132,68 @@ FATX_SUPERBLOCK CxbxGetFatXSuperBlock(int partitionNumber)
...
+int CxbxGetPartitionNumberFromHandle(HANDLE hFile)
 {
 	// Get which partition number is being accessed, by parsing the filename and extracting the last portion 
-	char buffer[MAX_PATH] = {0};
-	if (!GetFinalPathNameByHandle(hFile, buffer, MAX_PATH, VOLUME_NAME_DOS)) {
-		CxbxKrnlCleanup("CxbxGetPartitionDataPathFromHandle Failed:\nUnable to determine path for HANDLE 0x%08X", hFile);
-	}
+	const std::wstring path = CxbxGetFinalPathNameByHandle(hFile);
+
+	const std::wstring_view partitionString = L"\\Partition";
```
PatrickvL, line 161: What if `\Partition` appears multiple times in the path?

src/core/kernel/support/EmuFile.cpp
```diff
@@ -132,60 +132,68 @@ FATX_SUPERBLOCK CxbxGetFatXSuperBlock(int partitionNumber)
...
 
-	std::string bufferString(buffer);
-	std::string partitionString = "\\Partition";
-	std::string partitionPath = bufferString.substr(0, bufferString.find(partitionString) + partitionString.length() + 1);
+std::wstring CxbxGetPartitionDataPathFromHandle(HANDLE hFile)
+{
+	// Get which partition number is being accessed, by parsing the filename and extracting the last portion 
+	const std::wstring path = CxbxGetFinalPathNameByHandle(hFile);
+
+	const std::wstring_view partitionString = L"\\Partition";
```
PatrickvL, line 173: Same here

src/core/kernel/support/EmuFile.cpp
CookiePLMonster: That's how it was originally, but you are right that it's technically a possibility. Since we seem to only ever use this function to obtain paths to `PartitionX.bin` files, I think the best way to resolve this edge case is to use `rfind`, which finds the first occurrence of the substring starting from the **end**. What do you think?

PatrickvL: That'd be safer indeed

PR #2156, opened 2021-03-07
Further improve partition path functions

CookiePLMonster:
* Close a possible edge case of "\\Partition" being present in the path
* Close a possible edge case of an invalid path being shorter than the string "\\Partition"
* Allow for partition formatting to work fine when EmuDisk has been relocated
* Allow for partition formatting to work fine when symlinks are in place (now the "safeguard" is aware of symlinks)

### DIFF
--- a/src/core/kernel/support/EmuFile.cpp
+++ b/src/core/kernel/support/EmuFile.cpp
@@ -153,34 +153,55 @@ static std::wstring CxbxGetFinalPathNameByHandle(HANDLE hFile)
 	return path;
 }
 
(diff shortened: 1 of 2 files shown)


PatrickvL (approved):
Looks solid to me. Only one worry left : Should we forbid users to set CxbxBasePath to some system folder?

CookiePLMonster:
I would not worry about that, as it's:
1. Setting yourself up for failure
2. Might not work due to permissions anyway and Cxbx-Reloaded warns users not to be ran elevated.
3. We only wipe `EmuDisk\PartitionX` directories anyway so unless those system directories have directories named that, it's safe.

PR #2157, opened 2021-03-11
Tidy mip code and work around DXT crash with small textures

NZJenkins:
Some cleanup work while looking at an issue with DXT conversion when textures are smaller than a 4x4 block
Added a hack to avoid crashes, but stopped short of fixing the conversion code!

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -5985,7 +5985,7 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 			for (unsigned int mipmap_level = 0; mipmap_level < dwMipMapLevels; mipmap_level++) {
 
 				// Calculate size of this mipmap level
-				auto numRows = pxMipHeight;
+				DWORD numRows = pxMipHeight;
 
 				if (bCompressed) {
 					// Each row contains a 4x4 pixel blocks, instead of single pixels
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -5982,6 +5982,12 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
             DWORD dwMipRowPitch = dwRowPitch;
 			DWORD dwSrcSlicePitch = dwMipRowPitch * pxMipHeight; // TODO
 
+			// Block size only applies to compressed DXT formats
+			// DXT1 block size is 8 bytes
+			// Other Xbox DXT formats are 16 bytes
+			DWORD blockSize = 0;
+			blockSize = X_Format == xbox::X_D3DFMT_DXT1 ? 8 : 16;
+
 			for (unsigned int mipmap_level = 0; mipmap_level < dwMipMapLevels; mipmap_level++) {
 
 				// Calculate size of this mipmap level
@@ -5991,15 +5997,6 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 					// Each row contains a 4x4 pixel blocks, instead of single pixels
 					// So divide by 4 to get the number of rows
 					numRows = (numRows + 3) / 4;
-
-					// The row pitch can't be smaller than a single 4x4 block
-					// even if we need to store just one pixel
-					// DXT1 block size is 8 bytes
-					// Other Xbox DXT formats are 16 bytes
-					auto blockSize = X_Format == xbox::X_D3DFMT_DXT1 ? 8 : 16;
-					if (dwMipRowPitch < blockSize) {
-						dwMipRowPitch = blockSize;
-					}
 				}
 
 				DWORD dwMipSize = dwMipRowPitch * numRows;
@@ -6155,6 +6152,11 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 
 					// Update the row pitch
 					dwMipRowPitch /= 2;
+
+					// The pitch can't be less than a block
+					if (dwMipRowPitch < blockSize) {
+						dwMipRowPitch = blockSize;
+					}
 				}
 
 				if (pxMipHeight > 1) {
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -5991,6 +5991,15 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 					// Each row contains a 4x4 pixel blocks, instead of single pixels
 					// So divide by 4 to get the number of rows
 					numRows = (numRows + 3) / 4;
+
+					// The row pitch can't be smaller than a single 4x4 block
+					// even if we need to store just one pixel
+					// DXT1 block size is 8 bytes
+					// Other Xbox DXT formats are 16 bytes
+					auto blockSize = X_Format == xbox::X_D3DFMT_DXT1 ? 8 : 16;
+					if (dwMipRowPitch < blockSize) {
+						dwMipRowPitch = blockSize;
+					}
 				}
 
 				DWORD dwMipSize = dwMipRowPitch * numRows;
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6145,17 +6141,30 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
...
+					if (bCompressed) {
+						// The row pitch can't be smaller than a single 4x4 block
+						// even if we need to store just one pixel
+						// DXT1 block size is 8 bytes
+						// Other supported DXT formats are 16 bytes
+						auto blockSize = X_Format == xbox::X_D3DFMT_DXT1 ? 8 : 16;
+						if (dwMipRowPitch < blockSize) {
+							dwMipRowPitch = blockSize;
+						}
+					}
```
PatrickvL, line 6159: This block could be sped up & simplified, by renaming `blockSize` into `minRowPitch`, and assigning it's value upfront (somewhere around where numRows is set). There, set it to 1 for non-compressed formats.

With that, this block here can skip the check for `bCompressed` and the `blockSize` determination, and be reduced to :
```
	if (dwMipRowPitch < minRowPitch) {
		dwMipRowPitch = minRowPitch;
	}
```

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5985,20 +5976,25 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 		for (int face = D3DCUBEMAP_FACE_POSITIVE_X; face <= last_face; face++) {
 			// As we iterate through mipmap levels, we'll adjust the source resource offset
 			DWORD dwMipOffset = 0;
-			DWORD dwMipWidth = dwWidth;
-			DWORD dwMipHeight = dwHeight;
-			DWORD dwMipDepth = dwDepth;
+			DWORD pxMipWidth = dwWidth;
+			DWORD pxMipHeight = dwHeight;
+			DWORD pxMipDepth = dwDepth;
```
PatrickvL, line 5981: This rename clouds the diff a little

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5985,20 +5976,25 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
...
             DWORD dwMipRowPitch = dwRowPitch;
-			DWORD dwSrcSlicePitch = dwMipRowPitch * dwMipHeight; // TODO
+			DWORD dwSrcSlicePitch = dwMipRowPitch * pxMipHeight; // TODO
 
 			for (unsigned int mipmap_level = 0; mipmap_level < dwMipMapLevels; mipmap_level++) {
 
 				// Calculate size of this mipmap level
-				DWORD dwMipSize = dwMipRowPitch * dwMipHeight;
+				auto numRows = pxMipHeight;
```
PatrickvL, line 5988: The use of `auto` is a force of habit I guess? (Often, it's better to adhere to local coding style)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1542,6 +1542,14 @@ bool ConvertD3DTextureToARGBBuffer(
 		AdditionalArgument = DstRowPitch;
 
 	if (EmuXBFormatIsCompressed(X_Format)) {
+		if (SrcWidth < 4 || SrcHeight < 4) {
+			// HACK: The compressed DXT conversion code currently writes more pixels than it should, which can cause a crash.
+			// This code will get hit when converting compressed texture mipmaps on hardware that somehow doesn't support DXT natively
+			// (or lied when Cxbx asked it if it does!)
+			EmuLog(LOG_LEVEL::WARNING, "Converting DXT textures smaller than a block is not currently implemented. Ignoring conversion!");
```
PatrickvL, line 1549: Loving the fact you added an in-depth explanation here!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: I type it when I don't care about the type, which turns out to be most of the time! 😛 
Yeah maybe I can change this back to DWORD then

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: I'm not sure it's simpler?
We replace a special case with a more general `minRowPitch` behaviour, but we don't care about `minRowPitch` outside the special case. The special case is leaking and interleaving itself into other code imo

```c++
// Only relevant unless we're dealing with a compressed texture
DWORD minRowPitch = 0;
if (bCompressed) {
    // The row pitch can't be smaller than a single 4x4 block
    // even if we need to store just one pixel
    // DXT1 block size is 8 bytes
    // Other supported DXT formats are 16 bytes
    minRowPitch = X_Format == xbox::X_D3DFMT_DXT1 ? 8 : 16;
}
```

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
LukeUsher: I'd prefer the use of auto where possible, it's good to modernise the code we work on.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Thing is, in execution there's already an if being performed, so we might as well perform the one that really matters, instead of at least one, at most two (and the blocksize determination in-between).

PatrickvL: Perhaps if you still want it to stand out, call it `minRowPitchForCompressedTextures` or something like that

PatrickvL (changes requested):
One more attempt to get you to reconsider optimizing the inner loop ;)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: True, but such things are better done in a one-shot-kills-all fashion, as now, it's getting spread over many commits & pull requests, which makes it harder to track if things go sour all of a sudden.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: ... are you going to keep te rename, or change it back? (Which would have the benefit of a smaller commit, leading to focus on what actually changed)

CookiePLMonster (dismissed):
Do note this will most likely result in the lowest mipmaps of those uncrompressed textures being black or transparent. Other than that, LGTM as a temporary safeguard before we figure out how to make this code... not suck.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: was tossing up splitting it out and cleaning up some other syntax too, but eh. will keep it in mind for next time

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: that would make it read :
```
	if (dwMipRowPitch < minRowPitchForCompressedTextures) {
		dwMipRowPitch = minRowPitchForCompressedTextures;
	}
```
which avoids the `if (bCompressed)` and `blockSize` determination in a hot spot, at the cost of determining this upfront.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5985,20 +5976,31 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
...
+			DWORD pxMipDepth = dwDepth;
             DWORD dwMipRowPitch = dwRowPitch;
-			DWORD dwSrcSlicePitch = dwMipRowPitch * dwMipHeight; // TODO
+			DWORD dwSrcSlicePitch = dwMipRowPitch * pxMipHeight; // TODO
+
+			// Block size only applies to compressed DXT formats
+			// DXT1 block size is 8 bytes
+			// Other Xbox DXT formats are 16 bytes
+			DWORD blockSize = 0;
+			blockSize = X_Format == xbox::X_D3DFMT_DXT1 ? 8 : 16;
```
PatrickvL, line 5989: Shouldn't (bCompressed) be checked here still? 
Also,  this blocksize variable declaration and assignment could be moved up or level,  towards before the loop over all faces starts

PatrickvL (changes requested):
Nearly there!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: compiler will probably hoist it. Should we care if it doesn't?

PatrickvL (approved):

PR #2158, opened 2021-03-11
Fix Azurik render issue

NZJenkins:
Remove #ifndef blocks that were driven by the opposite define than they should have been
Also provide the opposite flag as a comment, next to where the 'driver' define

Looks like this...
```hlsl
bool alphakill[4] = {false, false, false, false};
#define PS_COMBINERCOUNT 2
// enable PS_COMBINERCOUNT_SAME_C0
#undef PS_COMBINERCOUNT_UNIQUE_C0
// enable PS_COMBINERCOUNT_SAME_C1
#undef PS_COMBINERCOUNT_UNIQUE_C1
#define PS_COMBINERCOUNT_MUX_MSB
#define PS_COMPAREMODE_0(in) CM_LT(in.x) CM_LT(in.y) CM_LT(in.z) CM_LT(in.w)
```

### How this fixes Azurik:
Azurik intends to use the mode: `PS_COMBINERCOUNT_SAME_C0`
We do this by undefining the opposite flag e.g.
```c++
#undef PS_COMBINERCOUNT_UNIQUE_C0
```
But, later we have code that accidentally checked if `PS_COMBINERCOUNT_SAME_C0` was be undefined, rather than `PS_COMBINERCOUNT_UNIQUE_C0`
```c++
#ifndef PS_COMBINERCOUNT_SAME_C0
	#define PS_COMBINERCOUNT_UNIQUE_C0
#endif
```

This would cause `PS_COMBINERCOUNT_UNIQUE_C0` to _always_ be defined, which caused an incorrect constant to be used in Azurik shaders.

Since the code is currently only driven by whether `PS_COMBINERCOUNT_UNIQUE_C0` is defined or not, the `#ifndef` section was removed.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/PixelShader.cpp
+++ b/src/core/hle/D3D8/Direct3D9/PixelShader.cpp
@@ -219,12 +219,24 @@ void FinalCombinerStageHlsl(std::stringstream& hlsl, RPSFinalCombiner& fc)
 	hlsl << "\n	" << opcode_comment[5][0] << "(" << arguments.str() << "); // " << opcode_comment[5][1];
 }
 
-void OutputDefine(std::stringstream& hlsl, std::string define_str, bool enabled)
+void OutputDefineFlag(std::stringstream& hlsl, bool enabled, std::string_view define_enabled, std::string_view define_disabled = "")
 {
-	if (enabled)
-		hlsl << "\n#define " << define_str;
-	else
-		hlsl << "\n#undef " << define_str;
+	if (define_disabled.length() > 0) {
+		if (enabled) {
+			hlsl << "\n#define " << define_enabled << " // not " << define_disabled;
+		}
+		else {
+			hlsl << "\n#define " << define_disabled << " // not " << define_enabled;
+		}
+	}
+	else {
+		if (enabled) {
+			hlsl << "\n#define " << define_enabled;
+		}
+		else {
+			hlsl << "\n#undef " << define_enabled;
+		}
+	}
 }
 
 /* Disabled, until BumpDemo is fixed (which with this code, inadvertedly skips stage 1 and 2 dotproducts) :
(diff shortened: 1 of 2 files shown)


PatrickvL:
Could you make it read
```
#undef PS_COMBINERCOUNT_UNIQUE_C0 // enables PS_COMBINERCOUNT_SAME_C0
#undef PS_COMBINERCOUNT_UNIQUE_C1 // enables PS_COMBINERCOUNT_SAME_C1
```
?

src/core/hle/D3D8/Direct3D9/PixelShader.cpp
```diff
@@ -219,12 +219,20 @@ void FinalCombinerStageHlsl(std::stringstream& hlsl, RPSFinalCombiner& fc)
 	hlsl << "\n	" << opcode_comment[5][0] << "(" << arguments.str() << "); // " << opcode_comment[5][1];
 }
 
-void OutputDefine(std::stringstream& hlsl, std::string define_str, bool enabled)
+void OutputDefineFlag(std::stringstream& hlsl, bool enabled, std::string define_enabled, std::string define_disabled = "")
```
CookiePLMonster, line 222:
```suggestion
void OutputDefineFlag(std::stringstream& hlsl, bool enabled, std::string_view define_enabled, std::string_view define_disabled = "")
```

If stringstream supports string views, of course.

PatrickvL (changes requested):
If possible, try to output one-liners, like the suggestion I made above

src/core/hle/D3D8/Direct3D9/PixelShader.cpp
NZJenkins: interesting... I guess string almost never needs to be passed then?

src/core/hle/D3D8/Direct3D9/PixelShader.cpp
```diff
@@ -219,12 +219,19 @@ void FinalCombinerStageHlsl(std::stringstream& hlsl, RPSFinalCombiner& fc)
...
-void OutputDefine(std::stringstream& hlsl, std::string define_str, bool enabled)
+void OutputDefineFlag(std::stringstream& hlsl, bool enabled, std::string_view define_enabled, std::string_view define_disabled = "")
 {
-	if (enabled)
-		hlsl << "\n#define " << define_str;
-	else
-		hlsl << "\n#undef " << define_str;
+	// Since these are flags, we only need to #define or #undef one define
+	if (enabled) {
+		hlsl << "\n#define " << define_enabled;
```
PatrickvL, line 226: Could you add this :
```
		if (define_disabled.length() > 0) {
			hlsl << " // disables " << define_disabled;
		}
```

src/core/hle/D3D8/Direct3D9/PixelShader.cpp
CookiePLMonster: Yes! It only makes sense when you want to "own" it, e.g. in constructors. For all other cases, `string_view` is superior because it is a near no-op conversion from both `std::string` and char literals.

The only "price" you pay is lack of a null terminator, but if you don't need one then you can use `string_view`.

src/core/hle/D3D8/Direct3D9/PixelShader.cpp
```diff
@@ -219,12 +219,21 @@ void FinalCombinerStageHlsl(std::stringstream& hlsl, RPSFinalCombiner& fc)
...
+	}
+	else {
+		hlsl << "\n#undef " << define_enabled;
+	}
+
+	// Add a comment representing the disabled version of the flag, if any
+	if (define_disabled.length() > 0) {
+		auto action = enabled ? " // disables " : " // implies ";
+		hlsl << action << define_disabled;
+	}
```
PatrickvL, line 236: If we always want to `#define` one of the two, then this implementation could become :

```
	if (define_disabled.length() > 0) {
	    if (enabled) {
			hlsl << "\n#define " << define_enabled << " // not " << define_disabled;
		} else {
			hlsl << "\n#define " << define_disabled << " // not " << define_enabled;
		}
	} else {
		if (enabled) {
			hlsl << "\n#define " << define_enabled;
		} else {
			hlsl << "\n#undef " << define_enabled;
		}
	}
```

PatrickvL (changes requested):

PatrickvL (approved):

PR #2163, opened 2021-03-14
fix up fog mode

medievil1:
fix fog table mode emulation in hlsl vertex shader

before
image

after
image

Azurik is the only game I know for sure it fixes (linear fog is used in some areas so you go no textures at all), but should fix any game using linear distance fog or any of the other special modes in the fog table....
example video as well 

https://user-images.githubusercontent.com/38597905/111089986-6a04e080-8504-11eb-88a9-0c9af318ff43.mp4

as of commit https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2163/commits/cffe1fdd414ea178527f8b47c60c2053dcec878d
saturation on formulas removed as well as abs with abs now being applied to fogFactor value...
there seems to be no actual difference between the two as far as games are concerned, applying to fogFactor allows formulas to remain stock.
Azurik may still randomly lose all textures to black screen, that happens with or without saturation OR abs, so likely caused by another issue unrelated to fog.


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6612,6 +6612,17 @@ xbox::void_xt __fastcall xbox::EMUPATCH(D3DDevice_SetRenderState_Simple)
...
+	// Placed this here Purely because almost every game hits it
+	// frequently enough to keep the data curent
+	// if a better location is viable, i am all for it
+	auto fogStart = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_FOGSTART);
+	auto fogEnd = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_FOGEND);
+	auto fogEnable = XboxRenderStates.GetXboxRenderState(X_D3DRS_FOGENABLE);
+	auto fogTableMode = XboxRenderStates.GetXboxRenderState(X_D3DRS_FOGTABLEMODE);
+	auto density = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_FOGDENSITY);
+	float fogStuff[4] = { fogStart, fogEnd, density, fogTableMode };
+	g_pD3DDevice->SetVertexShaderConstantF(230, fogStuff, 1);
```
PatrickvL, line 6624: Better declare a symbol for this literal numerical value of 230, choose a form and name in line with what's already being used for other VS constants

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6612,6 +6612,17 @@ xbox::void_xt __fastcall xbox::EMUPATCH(D3DDevice_SetRenderState_Simple)
         LOG_FUNC_ARG(Value)
         LOG_FUNC_END;
 
+	// Placed this here Purely because almost every game hits it
+	// frequently enough to keep the data curent
+	// if a better location is viable, i am all for it
+	auto fogStart = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_FOGSTART);
+	auto fogEnd = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_FOGEND);
+	auto fogEnable = XboxRenderStates.GetXboxRenderState(X_D3DRS_FOGENABLE);
```
PatrickvL, line 6620: fogEnable is read but not used here, if it's not needed,  please remove the variable and read.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -326,10 +329,29 @@ R"DELIMITER(
 	// Copy variables to output struct
 	VS_OUTPUT xOut;
 
+    float fogFactor;
+    /* foginfo is broken down into
```
PatrickvL, line 333: For readability, consider declaring const variables for each of the components,  assign the source components to those variables, and use them in the code below. 

This way,  the comment is no longer needed,  the code using the variables becomes more readable,  while the optimizer will still compiler it into the same code (so no additional instructions will be generated for this use of intermediate aliases)

PatrickvL (changes requested):
Nice work!

Could you address the code review remarks and add a few before & after screenshots to the pull request description, to illustrate how this improves fog modes?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
medievil1: done!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
medievil1: removed and done!

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
medievil1: done!

medievil1:
before
image

after
image

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6611,6 +6611,17 @@ xbox::void_xt __fastcall xbox::EMUPATCH(D3DDevice_SetRenderState_Simple)
         LOG_FUNC_ARG(Method)
         LOG_FUNC_ARG(Value)
         LOG_FUNC_END;
+		
+	//Placed this here Purely because almost every game hit is
+	//gequently enough to keep the data curent
+	// if a better location is viable, i am all for it
```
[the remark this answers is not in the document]
medievil1: done

PatrickvL (changes requested):
A few more nitpicks and then we're good

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -326,10 +331,29 @@ R"DELIMITER(
...
+    if(fmode == 0)
+        fogFactor = fdepth;
+    if(fmode == 1)
+        fogFactor = saturate(1 / exp(fdepth * fdensity)); /* / 1 / e^(d * density)*/
+    if(fmode == 2)
+        fogFactor = saturate(1 / exp(pow(fdepth * fdensity, 2))); /* / 1 / e^((d * density)^2)*/
+    if(fmode == 3)
+        fogFactor = saturate((fend - fdepth) / (fend - fstart)) ;
+     /*For now linear mode needs Saturation (other modes may as well, *Needs testing*) 
+       or the results in PS are inconsistant PS fog in may be incorrect*/
```
NZJenkins, line 350: nitpick: this is indented differently to other code

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -326,10 +331,29 @@ R"DELIMITER(
...
+       
+
 	xOut.oPos = reverseScreenspaceTransform(oPos);
 	xOut.oD0 = saturate(oD0);
 	xOut.oD1 = saturate(oD1);
-	xOut.oFog = oFog.x; // Note : Xbox clamps fog in pixel shader
+	xOut.oFog = fogFactor; // Note : Xbox clamps fog in pixel shader /was oFog.x
```
NZJenkins, line 356: thought:
> // Note : Xbox clamps fog in pixel shader

I guess this isn't the case, since you had to saturate the fog factor!
Maybe this whole comment can go...

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1033,11 +1033,12 @@ typedef DWORD X_VERTEXSHADERCONSTANTMODE;
 #define X_D3DSCM_CORRECTION                 96 // Add 96 to arrive at the range 0..191 (instead of -96..95)
 #define X_D3DVS_CONSTREG_COUNT              192
 
+
```
NZJenkins, line 1036: nitpick: extra whitespace

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -28,6 +28,11 @@ struct VS_OUTPUT // Declared identical to pixel shader input (see PS_INPUT)
 // Xbox constant registers
 uniform float4 C[X_D3DVS_CONSTREG_COUNT] : register(c0);
 
+// External Passthru
+float4 foginfo : register(c230);
```
NZJenkins, line 32: suggestion: not sure what this comment means
could be `// Parameters for mapping the shader's fog output value to a fog factor`?

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -28,6 +28,11 @@ struct VS_OUTPUT // Declared identical to pixel shader input (see PS_INPUT)
...
+
+
+
```
NZJenkins, line 35: nitpick: extra whitespace

NZJenkins (commented):
Some nitpicky comments otherwise LGTM 👍 
The fixup commits should probably be squashed before merging

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: Well, in our pixel shader hlsl code for XFC (the final combiner stage, which incorporates fog), we have this : 
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/37195a61da45367c6e8c10220253b24486e8a58c/src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl#L315
and this
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/37195a61da45367c6e8c10220253b24486e8a58c/src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl#L150-L158

which leadds to the conclusion that the fog factor (as passed into the Xbox register combiner XFC stage / our pixel shader HLSL for this), is indeed saturated per-pixel.

So at first we could assume that it won't be necessary to saturate here in the vertex shader.

But on the other hand, it also seems like the fog.a input is never modified inside the PS HLSL code, so perhaps in the future, we could optimize this by removing the saturation operation over fog.a from PS HLSL, and instead saturate here, like you have right now (which is probably more efficient, because vertex shaders run per vertex which is much less than the per-pixel run of pixel shaders).

PatrickvL: (Oh : note, that as long as we're still using host Fixed Function for null pixel shaders, we also better saturate the fog factor in our vertex shaders, because we don't know if we can rely on host FF to do this.)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7458,6 +7458,14 @@ extern float* HLE_get_NV2A_vertex_constant_float4_ptr(unsigned const_index); //
 // remove our patches on D3DDevice_SetVertexShaderConstant (and CxbxImpl_SetVertexShaderConstant)
 void CxbxUpdateHostVertexShaderConstants()
 {
+    // Placed this here until we find a better place
+	auto fogStart = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGSTART);
+	auto fogEnd = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGEND);
+	auto fogTableMode = XboxRenderStates.GetXboxRenderState(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGTABLEMODE);
+	auto density = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGDENSITY);
+	float fogStuff[4] = { fogStart, fogEnd, density, fogTableMode };
```
PatrickvL, line 7466: Maybe a bit obsessive, but since fog table mode is the deciding factor (pun intended) in how the other values are used, it may be clearer to put this mode as a first value in the array.
Also, because fog start and end are only used in the 3rd table mode, it may be best to put these in the final two array slots.

Oh, and for consistency, better rename `density` into `fogDensity`.

This would turn it into this code :
```
	auto fogTableMode = XboxRenderStates.GetXboxRenderState(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGTABLEMODE);
	auto fogDensity = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGDENSITY);
	auto fogStart = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGSTART);
	auto fogEnd = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGEND);
	float fogStuff[4] = { fogTableMode, fogDensity, fogStart, fogEnd };
```

(If you agree, then don't forget to also update the aliasses for them in the shader, as the components they reside in would get shuffled around as well by this!)

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -326,10 +331,29 @@ R"DELIMITER(
 	// Copy variables to output struct
 	VS_OUTPUT xOut;
 
+    float  fstart   =   foginfo.x;
+    float  fend     =   foginfo.y;
+    float  fdensity =   foginfo.z;
+    float  fmode    =   foginfo.w;
+    float  fdepth   =   oFog.x;
```
PatrickvL, line 338: Consider using the same variable names as the C++ code that assigned values to these variables.
When applying that, `foginfo` begins to stand out, so should also be changed into `fogInfo`.

If you're addressing the const prefix and suggested reordering as well, that would lead to this code :
```
    const float fogDepth = oFog.x; 
    const float fogTableMode = fogInfo.x;
    const float fogDensity = fogInfo.y;
    const float fogStart = fogInfo.z;
    const float fogEnd = fogInfo.w;
```

(obviously, also update below code to use the renamed variables)

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: Also, I suggest to change casing into `fogInfo`, or even `CxbxFogInfo`.

Also, please mention the constant for this register index at the end of the line : ` // CXBX_D3DVS_CONSTREG_FOGINFO = 230

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1033,11 +1033,12 @@ typedef DWORD X_VERTEXSHADERCONSTANTMODE;
...
 // Special Registers, used to pass additional information to the shaders
 // TODO co-locate shader workaround constants with shader code
 #define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_BASE      (X_D3DVS_CONSTREG_COUNT)
 #define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_SIZE      16
-
+#define CXBX_D3DVS_FOGINFO_CONST                   230
```
PatrickvL, line 1041: Better make this consistent, and call it `CXBX_D3DVS_CONSTREG_FOGINFO`

PatrickvL (changes requested):

PatrickvL:
> before
> image
> 
> after
> image
> 
> Azurik is the only game I know for sure it fixes (linear fog is used in some areas so you go no textures at all), but should fix any game using linear distance fog or any of the other special modes in the fog table....
> example video as well
> 
>  Desktop.2021.03.12.-.21.17.24.01.mp4

Better move this towards the topmost pull request description (as a comment like this might easily get overlooked)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7458,6 +7458,14 @@ extern float* HLE_get_NV2A_vertex_constant_float4_ptr(unsigned const_index); //
...
+	g_pD3DDevice->SetVertexShaderConstantF(CXBX_D3DVS_FOGINFO_CONST, fogStuff, 1);
+
 	// Copy all constants (as they may have been overwritten with fixed-function mode)
 	// Though we should only have to copy overwritten or dirty constants
 	float* constant_floats = HLE_get_NV2A_vertex_constant_float4_ptr(0);
```
PatrickvL, line 7471: It's nicer if vertex shader constants were set in the same order as in which they are declared. This would mean that the actual constants we set here, should be set first, and only then, below that in the order in which they are declared : `CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_BASE`, `CXBX_D3DVS_SCREENSPACE_SCALE_BASE`, `CXBX_D3DVS_SCREENSPACE_OFFSET_BASE`, `CXBX_D3DVS_TEXTURES_SCALE_BASE` (all set in `CxbxUpdateHostViewPortOffsetAndScaleConstants()`) and only then `CXBX_D3DVS_CONSTREG_FOGINFO` !

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1033,11 +1033,12 @@ typedef DWORD X_VERTEXSHADERCONSTANTMODE;
 #define X_D3DSCM_CORRECTION                 96 // Add 96 to arrive at the range 0..191 (instead of -96..95)
 #define X_D3DVS_CONSTREG_COUNT              192
 
+
 // Special Registers, used to pass additional information to the shaders
 // TODO co-locate shader workaround constants with shader code
 #define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_BASE      (X_D3DVS_CONSTREG_COUNT)
```
PatrickvL, line 1039: Better make FOGINFO consistent with the rest of the CONSTREG's, and call it CXBX_D3DVS_CONSTREG_FOGINFO.

Also, `CXBX_D3DVS_CONSTREG_FOGINFO` should be defined somwhere in this list of CONSTREG's (I guess at the end, or at a logical insertion point), via the same convention as all the others have been declared!

PatrickvL (changes requested):

PatrickvL (changes requested):
Surprising how many improvements could still be found - but don't feel bad about that, most remarks are about (local) consistency in the way the code is written, not about the actual fix, so still kudos for all the work that went into that!

src/core/hle/D3D8/XbD3D8Types.h
medievil1: name change ok, BUT it was already in the constregs section (line 1041)

I did move it to 1043 though just so it didn't break up the group of 4 VREGDEFAULTS

// Special Registers, used to pass additional information to the shaders
`/ TODO co-locate shader workaround constants with shader code
#define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_BASE      (X_D3DVS_CONSTREG_COUNT)
#define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_SIZE      16
#define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_BASE (CXBX_D3DVS_CONSTREG_VREGDEFAULTS_BASE + CXBX_D3DVS_CONSTREG_VREGDEFAULTS_SIZE)
#define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_SIZE 4
#define CXBX_D3DVS_CONSTREG_FOGINFO                230 /`

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
medievil1: so move this to the end of the section :P 
done.....

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
medievil1: done...had debated doing just that....
also reordered them where they are defined.. as:
```auto fogTableMode = XboxRenderStates.GetXboxRenderState(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGTABLEMODE);
	auto fogDensity = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGDENSITY);
	auto fogStart = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGSTART);
	auto fogEnd = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGEND);

File: src/core/hle/D3D8/Direct3D9/Direct3D9.cpp (hunk 10 above)
Comment (medievil1) [reply src/core/hle/D3D8/Direct3D9/Direct3D9.cpp:7466]: done as noted in next review comment...

File: src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl (hunk 8 above)
Comment (medievil1) [reply src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl:32]: sounds good...lol
External passthru to me was just passing the values in 
Didn't really consider how to name it...

File: src/core/hle/D3D8/XbD3D8Types.h (hunk 7 above)
Comment (medievil1) [reply src/core/hle/D3D8/XbD3D8Types.h:1036]: done

File: src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl (hunk 6 above)
Comment (medievil1) [reply src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl:356]: how about a *Needs testing* since I am not 100% sure yet... for Azurik it seemed to need saturation, BUT for other games it might have different results... not enough to know just yet...

File: src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl (hunk 5 above)
Comment (medievil1) [reply src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl:350]: something more like this? 
```    float  fogDepth   =   oFog.x; 
    float  fogMode    =   foginfo.x;
    float  fpgDensity =   foginfo.y;
    float  fogStart   =   foginfo.z;
    float  fogEnd     =   foginfo.w;
    
    float fogFactor;

    if(fogMode == 0)
    fogFactor = fogDepth;
    if(fogMode == 1)
    fogFactor = saturate(1 / exp(fogDepth * fogDensity)); /* / 1 / e^(d * density)*/
    if(fogMode == 2)
    fogFactor = saturate(1 / exp(pow(fogDepth * fogDensity, 2))); /* / 1 / e^((d * density)^2)*/
    if(fogMode == 3)
    fogFactor = saturate((fogEnd - fogDepth) / (fogEnd - fogStart)) ;

**Review by medievil1 — COMMENTED** [review]
somehow initiated a review of myself..lol

**Comment by medievil1** [comment]
> 
> 
> Surprising how many improvements could still be found - but don't feel bad about that, most remarks are about (local) consistency in the way the code is written, not about the actual fix, so still kudos for all the work that went into that!

not surprising to me.. new to all this... surprised there were not even more issues

File: src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl (hunk 6 above)
Comment (medievil1) [reply src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl]: also another thought.... I have the 3 modes saturated currently... should the passthrough oFog.x alone be saturated?? if so I can just removes them int he formula and just saturate fogfactor at the xOut

File: src/core/hle/D3D8/XbD3D8Types.h [hunk 15]
```diff
@@ -1037,10 +1037,9 @@ typedef DWORD X_VERTEXSHADERCONSTANTMODE;
...
 #define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_SIZE      16
-
```
PatrickvL, line 1040: Restore empty separation line

src/core/hle/D3D8/XbD3D8Types.h
```diff
@@ -1037,10 +1037,9 @@ typedef DWORD X_VERTEXSHADERCONSTANTMODE;
...
 #define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_BASE (CXBX_D3DVS_CONSTREG_VREGDEFAULTS_BASE + CXBX_D3DVS_CONSTREG_VREGDEFAULTS_SIZE)
 #define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_SIZE 4
-
+#define CXBX_D3DVS_CONSTREG_FOGINFO                230
```
PatrickvL, line 1042: Enclose with empty separation lines above and below

Replace 230 with a summation of preceding base plus size

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7472,6 +7473,15 @@ void CxbxUpdateHostVertexShaderConstants()
 	// Xbox dashboard (during initial fade from black)
 	// Need for Speed: Hot Pursuit 2 (car select)
 	CxbxUpdateHostViewPortOffsetAndScaleConstants();
+
+	 // Placed this here until we find a better place
```
PatrickvL, line 7477: Correct indentation

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7472,6 +7473,15 @@ void CxbxUpdateHostVertexShaderConstants()
...
+	auto fogTableMode = XboxRenderStates.GetXboxRenderState(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGTABLEMODE);
```
PatrickvL, line 7478: Prefix these variables with `const`

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -28,6 +28,9 @@ struct VS_OUTPUT // Declared identical to pixel shader input (see PS_INPUT)
 // Xbox constant registers
 uniform float4 C[X_D3DVS_CONSTREG_COUNT] : register(c0);
 
+// Parameters for mapping the shader's fog output value to a fog factor
+float4  CxbxFogInfo: register(c230); // = CXBX_D3DVS_CONSTREG_FOGINFO
```
PatrickvL, line 32: Prefix with `uniform`

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
 
 	// Copy variables to output struct
 	VS_OUTPUT xOut;
-
+const float  fogDepth   =   oFog.x;
```
PatrickvL, line 331: Correct indentation

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
 
 	// Copy variables to output struct
 	VS_OUTPUT xOut;
-
+const float  fogDepth   =   oFog.x; 
+const float  fogMode    =   CxbxFogInfo.x;
```
PatrickvL, line 332: Rename into `fogTableMode`, so it's named identical to its C++ counterpart

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
...
+const float  fogDensity =   CxbxFogInfo.y;
+const float  fogStart   =   CxbxFogInfo.z;
+const float  fogEnd     =   CxbxFogInfo.w;  
+ 
+float fogFactor;
```
PatrickvL, line 337: Correct indentation

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
...
+    if(fogMode == 0)
+    fogFactor = abs(fogDepth);
```
PatrickvL, line 339: Correct indentation; each additional nesting level (like following an if) must use an additional indentation

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
...
+    if(fogMode == 1)
+    fogFactor = 1 / exp(abs(fogDepth) * fogDensity); /* / 1 / e^(d * density)*/
```
PatrickvL, line 341: Same here

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
...
+    if(fogMode == 2)
+    fogFactor = 1 / exp(pow(abs(fogDepth) * fogDensity, 2)); /* / 1 / e^((d * density)^2)*/
```
PatrickvL, line 343: Same here

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
...
+    if(fogMode == 3)
+    fogFactor = (fogEnd - abs(fogDepth)) / (fogEnd - fogStart) ;
```
PatrickvL, line 345: Same here

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL, line 331: Notice how all uses of `fogDepth` in below code apply `abs()` function to its value; If that's correct for all cases, the apply the function here once, at assignment, and remove it from all uses.

PatrickvL (commented):
Again a list of remarks. As a fair general suggestion, it may help to review your own changes for consistency before committing.

src/core/hle/D3D8/XbD3D8Types.h
PatrickvL: it's still there.... the nitpick was an indication to remove it, so : not done (yet)

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Note : check consistent use of tab and/or spaces

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Also, current coding style still prefers actual types instead of `auto`. For `fogTableMode` that would mean `uint32_t`, for the others, that would mean `float`.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -7472,6 +7473,15 @@ void CxbxUpdateHostVertexShaderConstants()
...
+	auto fogDensity = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGDENSITY);
+	auto fogStart = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGSTART);
+	auto fogEnd = XboxRenderStates.GetXboxRenderStateAsFloat(xbox::_X_D3DRENDERSTATETYPE::X_D3DRS_FOGEND);
+	float fogStuff[4] = { fogTableMode, fogDensity, fogStart, fogEnd };
```
PatrickvL, line 7482: Note, that the array contains `float` values, but `fogTableMode` is an `uint32_t` value, thus it needs to be cast here; Most code in Cxbx uses the C-style cast syntax `(float)fogTableMode`, although we're gradually migrating to C++-style which would be the somewhat longer syntax `static_cast<float>(fogTableMode)` - I'll let you decide which one is best here.

medievil1: actually, since it was added with ps_hlsl, I can just GetXboxRenderStateAsFloat for it, and it will pull it as float like the others, I think anyway... I will resolve it one way or the other though....

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
medievil1: thats a possibility.... actually in reading through some code I ran across nv2a shaders and noticed there were 6 table modes it is using, linear, linear_ABS, EXP, EXP_ABS and EXP2 and EXP2_ABS... not sure if that was an xbox thing or simply something they did in xqemu... but if xbox used it, they would probably need to be added as possible modes, though currently it seems fine with abs always applied...

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Nope. You can't use `GetXboxRenderStateAsFloat` as that's for reading values stored as a float. Fog table mode is not stored as a float. It is stored as an unsigned 32 bit integer. Thus you should read it like that. But because it's put in a float, it needs casting, so that the integer value gets encoded into a floating point value.

Integers and floats are encoded differently, so you can't read one as if it's the other, but you can convert them, using a cast (this is all very basic coding stuff every contributor should be aware of).

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -325,11 +328,26 @@ R"DELIMITER(
...
 	VS_OUTPUT xOut;
-
+const float  fogDepth   =   oFog.x; 
+const float  fogMode    =   CxbxFogInfo.x;
+const float  fogDensity =   CxbxFogInfo.y;
+const float  fogStart   =   CxbxFogInfo.z;
+const float  fogEnd     =   CxbxFogInfo.w;  
+ 
+float fogFactor;
+    if(fogMode == 0)
```
PatrickvL, line 338: Looking at this again,  it may also be good to replace the 0, 1, 2 and 3 numbers with constants,  or at least, mention the symbol name of each number in an end of line comment

src/core/hle/D3D8/XbD3D8Types.h
medievil1: not exactly sure how to do this... 
I mean I can see how it is setup,  
`#define CXBX_D3DVS_CONSTREG_VREGDEFAULTS_BASE      (X_D3DVS_CONSTREG_COUNT)`
is setting to 192, then size adds 16 for 208 in CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_BASE, etc...  is size just a number to achieve the next result? 
I used 230 because Azurik, used constants all the way up to 228and I wanted to make sure I was outside the "used" range, however from the defines, the last one in the constreg section isn't close to 230

PatrickvL: `#define CXBX_D3DVS_CONSTREG_FOGINFO (CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_BASE + CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_SIZE)`

medievil1: ahhhhhbut that will also entail changing the next in the list (Since it is currently not last) would it not? to simplify I will move it to the end and use the last in list 
`CXBX_D3DVS_SCREENSPACE_SCALE_BASE already uses (CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_BASE + CXBX_D3DVS_CONSTREG_VREGDEFAULTS_FLAG_SIZE)`

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
medievil1: in my defense (auto), they were direct copies from FF in direct3d9.cpp line 6443 :)
I will change them though and add const....

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
medievil1: and get new glasses... my eyesight is getting so poor I could not see they were one space off

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
medievil1: ahhh I misunderstood what `GetXboxRenderStateAsFloat` does then, I assumed it converted to float on the get as the result is a float while `GetXboxRenderState` resulted in uint32_t....

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
medievil1: `const float FOG_TABLE_NONE   = 0;
const float FOG_TABLE_EXP    = 1;
const float FOG_TABLE_EXP2   = 2;
const float FOG_TABLE_LINEAR = 3;`

src/core/hle/D3D8/XbD3D8Types.h
PatrickvL: If that's the last one, then sure. 

Do take care that the shader will be using the same index

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Yeah,  the function name should communicate that as well,  but it's already quite long.  At any rate,  be careful not to assume too much

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: Yeah something along those lines. 

Actually, these and other global symbols should be in an include file, used by both c and hlsl code. But let's address that later.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
medievil1: after further thought, will remove abs from the math and apply it to the value. I know it works as is, and should I find in the future that there needs to be an abs AND non abs version it can be added at that time

medievil1:
> 
> 
> Again a list of remarks. As a fair general suggestion, it may help to review your own changes for consistency before committing.

I thought I had..lol.. as I said new to this so still learning and I appreciate all the feedback!! ... next time will go much smoother!

medievil1:
> 
> 
> Why revert? Did you update the constant index for fog info in hlsl?



> 
> 
> Why revert? Did you update the constant index for fog info in hlsl?

there was an issue I am tracking it down, the game stage that had distance fog was not working so either the constant wasn't being read or an issue with the compare values... wasn't aware before commit because other stages were working fine.... also sorry for delay, internet decided to crap out all of a sudden

medievil1:
> 
> 
> Why revert? Did you update the constant index for fog info in hlsl?

it is resolved..... a couple of issues actually...I updated but was off by one, and static cast couldn't handle uint32_t to float...

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
medievil1: that would be great.... as they are actually already defined in normal c, just not accessible from hlsl shader and it didn't make sense declaring them in constants cause you then have the same issue

Margen67:
Sneakers now renders (mostly) correctly ingame:
image

medievil1:
> 
> 
> Sneakers now renders (mostly) correctly ingame:
> image

wasn't this already fix in h
lsl_ps PR/Branch?

PR #2164, opened 2021-03-17
Vertex buffer cache improvements

CookiePLMonster:
This PR improves the vertex buffer cache again to improve its performance.  Basing on **Group S Challenge** as a test case, I identified our current cache generated a lot of "false positive" matches, continuously evicting buffers and recreating new ones up to 1000 times per frame. Turns out, those cases were caused by vertex buffers with identical vertex data but different vertex declarations.

I modified the vertex buffer cache to key cached resources by a `(dataHash, vertexElementsHash)` pair of elements, which allows such "buffer views" to coexist in the cache without continuously invalidating each other. On top of that, I modified `vertexElementsHash` not to hash some of the vertex stream attributes not relevant for the final form of the vertex buffer - most notably, it wrongly hashed a stream number, effectively evicting buffers from the cache if they are being bound to different stream numbers!

This has potential to improve performance in every game, based on my testing **Group S Challenge** gets a 3x performance improvement in-race, jumping from <20FPS to 55-60FPS - essentially full speed.

Before:
image

After:
image

### DIFF
--- a/src/common/util/hasher.cpp
+++ b/src/common/util/hasher.cpp
@@ -27,7 +27,7 @@ void InitHasher()
     }
 }
 
-uint64_t ComputeHash(void* data, size_t len)
+uint64_t ComputeHash(const void* data, size_t len)
 {
     if (g_HashAlgorithm == HASH_NONE) {
         InitHasher();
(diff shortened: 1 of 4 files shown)


PatrickvL (approved):
LGTM, let's wait on a few more test results before merging

medievil1:
not found any regressions on my end...  haven't really seen any improvements either but could just be my selection of ripped games, most have other issues

PR #2165, opened 2021-03-18
Fix crash in fullscreen mode

NZJenkins:
Set a backbuffer format, since UNKNOWN is only accepted in windowed mode
https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dpresent-parameters

Fixes issue introduced in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2151

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2191,8 +2191,11 @@ static void SetupPresentationParameters
     params.SwapEffect = D3DSWAPEFFECT_COPY;
 
     // Any backbuffer format should do, since we render to a separate xbox backbuffer
-    // Rather than directly to the host backbuffer
-    params.BackBufferFormat = D3DFMT_UNKNOWN;
+    // We need to specify something to support fullscreen exclusive mode
+    // Take the current displaymode format
+    D3DDISPLAYMODE D3DDisplayMode;
+    g_pDirect3D->GetAdapterDisplayMode(g_EmuCDPD.Adapter, &D3DDisplayMode);
+    params.BackBufferFormat = D3DDisplayMode.Format;
 
     params.PresentationInterval = g_XBVideo.bVSync ? D3DPRESENT_INTERVAL_ONE : D3DPRESENT_INTERVAL_IMMEDIATE;
     g_Xbox_PresentationInterval_Default = pXboxPresentationParameters->PresentationInterval;


PatrickvL (dismissed):
Looks simple enough, but here too it's probably wise to await a few rest results

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2191,8 +2191,11 @@ static void SetupPresentationParameters
     params.SwapEffect = D3DSWAPEFFECT_COPY;
 
     // Any backbuffer format should do, since we render to a separate xbox backbuffer
-    // Rather than directly to the host backbuffer
-    params.BackBufferFormat = D3DFMT_UNKNOWN;
+	// We need to specify something to support fullscreen exclusive mode
```
LukeUsher, line 2194: What happened to the indentation ?

NZJenkins: tabs!

CookiePLMonster:
Maybe add this code only for fullscreen and let D3D decide for windowed?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Maybe we should pick a date on which we reformat the entire source tree, instead of updating indentation ad-hock?

RadWolfie: > Maybe we should pick a date on which we reformat the entire source tree, instead of updating indentation ad-hock?

It will require .clang-format. However, it is not an easy task to do. What we need to do is find out how to make the transition easier for all of us without need to fix rebase conflict at all, preferably via git's cli tool commands. Which isn't going to be similar to [newline reformat process](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/wiki/Rebase-Line-Ending-Fix).

RadWolfie:
I can confirm exclusive full screen is working again. 👍

RadWolfie:
> Maybe add this code only for fullscreen and let D3D decide for windowed?

Actually, in the doc it says...
> The back buffer format. For more information about formats, see D3DFORMAT. This value must be one of the render-target formats as validated by CheckDeviceType. You can use GetDisplayMode to obtain the current format.
>
> In fact, D3DFMT_UNKNOWN can be specified for the BackBufferFormat while in windowed mode. This tells the runtime to use the current display-mode format and eliminates the need to call GetDisplayMode.

Which means it is the same method as windowed mode. So, it is unnecessary for make two different methods which will do exactly the same way from the backend in windowed mode.

RadWolfie (approved):
I don't see any problem with this change and it is following identical to the doc. Beside using GetDisplayMode function.

Issue #1885, opened 2020-05-07
Clear Xbox HDD cache partitions from GUI


NZJenkins:
Sometimes, games can break due to the Xbox HDD cache partitions getting into a bad state
This can be fixed by deleting partition folders 3 through 7, and their corresponding `.bin` files
We should able to do this from the GUI with a single click

LukeUsher:
Excellent idea, but make sure only to delete the .bin files for partition 3 through 7, deleting partition0 will delete metadata and some dashboard settings that are not present in eeprom (network stuff, xbl accounts, etc)

LukeUsher:
Interesting to note that this is often a problem on a real Xbox too: Many hacked dashboards have an option to clear the cache, the corruption, in most cases, is NOT an emulation issue.

PR #2170, opened 2021-03-22
Added option to clear the cache partitions from the GUI

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1885


src/gui/WndMain.cpp
```diff
@@ -98,6 +98,36 @@ void ClearSymbolCache(const char sStorageLocation[MAX_PATH])
 	printf("Cleared HLE Cache\n");
 }
 
+unsigned ClearCachePartitions()
+{
+	const std::string &partition_path = g_Settings->GetDataLocation() + "\\EmuDisk\\Partition";
+	std::error_code err;
+	unsigned has_any_err = 0;
+
+	for (int partition_num = 3; partition_num < 8; ++partition_num) {
```
RadWolfie, line 107: Cache partitions are used from 3 - 5. Partitions number 6 and 7 are used something else in homebrew environment if I recall that right. ref: https://xboxdevwiki.net/Xbox_Hard_Disk_Partitioning

Let's reconsider maybe split them up into separate functions to clear.

1) Data (User) Partition (1)
2) Dashboard Partition (2) (unlikely we'll do this.)
3) Cache Partitions (3 - 5)
4) Misc Partitions (6 - 7)

RadWolfie (changes requested):

src/gui/WndMain.cpp
ergo720: It doesn't matter if they are not real cache partitions on the xbox, cxbxr is still using them as such and so we should clear them as well. Source https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/efda45439991a5f377b60637d36534ab780b94d2/src/core/kernel/init/CxbxKrnl.cpp#L1367

RadWolfie (approved):
Tested and confirmed working as intended design even with failure. Although, directory doesn't remove all the content after one file marked as read-only test failure.

Issue #2166, opened 2021-03-19
FS: Thread local storage / EThread objects leak memory


LukeUsher:
## Quick summary
Cxbx-Reloaded does not free TLS/EThread objects when a thread is terminated

## Details
If a game creates and destroys lots of threads during execution, we'll leak memory.

### System Configuration:
N/A

### Additional Information (if any):
N/A

PR #2171, opened 2021-03-23
Free thread resources upon termination

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2166. In addition to free those resources, I also changed the allocation function for those from `ExAllocatePool` to `NtAllocateVirtualMemory`. This, because the former breaks the 16 byte alignment requirement for the TLS and also because it apparently breaks the Intel Graphics Performance Analyzer for some unknown reason.


PatrickvL (commented):
Nice and clean, nothing to remark,  well done.

Let's await a few tests before merging.

PatrickvL (approved):

medievil1:
no issues here.....everything works as it should and GPA seems to be working again

LukeUsher (approved):

LukeUsher:
Nice job!

PR #2172, opened 2021-03-23
Use Win10 CPU Sets API where supported

CookiePLMonster:
**IMPORTANT: Do not merge until this PR is confirmed not to break Windows 7!**

Using [CPU Sets](https://docs.microsoft.com/en-us/windows/win32/procthread/cpu-sets?redirectedfrom=MSDN) (introduced in Windows 10) instead of thread affinity allows to ensure that the entire physical core is reserved ONLY for the game code, so even the worker threads and third party libraries inside the process cannot run code on it.

For Windows 7 and 8.1, the existing thread affinity approach is retained.

### DIFF
--- a/src/common/Timer.cpp
+++ b/src/common/Timer.cpp
@@ -102,9 +102,13 @@ void ClockThread(TimerObject* Timer)
 	if (!Timer->Name.empty()) {
 		CxbxSetThreadName(Timer->Name.c_str());
 	}
-	if (Timer->CpuAffinity != nullptr) {
-		InitXboxThread(*Timer->CpuAffinity);
+	if (Timer->IsXboxTimer) {
+		InitXboxThread();
+		g_AffinityPolicy->SetAffinityXbox();
+	} else {
+		g_AffinityPolicy->SetAffinityOther();
 	}
+
 	NewExpireTime = GetNextExpireTime(Timer);
 
 	while (true) {
@@ -133,7 +137,7 @@ void Timer_Exit(TimerObject* Timer)
 }
 
 // Allocates the memory for the timer object
-TimerObject* Timer_Create(TimerCB Callback, void* Arg, std::string Name, unsigned long* Affinity)
+TimerObject* Timer_Create(TimerCB Callback, void* Arg, std::string Name, bool IsXboxTimer)
 {
 	std::lock_guard<std::mutex>lock(TimerMtx);
 	TimerObject* pTimer = new TimerObject;
@@ -142,8 +146,8 @@ TimerObject* Timer_Create(TimerCB Callback, void* Arg, std::string Name, unsigne
 	pTimer->ExpireTime_MS.store(0);
 	pTimer->Exit.store(false);
 	pTimer->Opaque = Arg;
-	Name.empty() ? pTimer->Name = "Unnamed thread" : pTimer->Name = Name;
-	pTimer->CpuAffinity = Affinity;
+	pTimer->Name = Name.empty() ? "Unnamed thread" : std::move(Name);
+	pTimer->IsXboxTimer = IsXboxTimer;
 	TimerList.emplace_back(pTimer);
 
 	return pTimer;
--- a/src/common/input/SdlJoystick.cpp
+++ b/src/common/input/SdlJoystick.cpp
@@ -84,8 +84,6 @@ namespace Sdl
 		UpdateInputEvent_t = CustomEvent_t + 2;
 		DeviceRemoveAck_t = CustomEvent_t + 3;
 
-		SetThreadAffinityMask(GetCurrentThread(), g_CPUOthers);
-
 		// Drain all joystick add/remove events to avoid creating duplicated
 		// devices when we call PopulateDevices
 		while (SDL_PollEvent(&Event))
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -599,6 +599,7 @@ void CxbxInitWindow(bool bFullInit)
         HANDLE hThread = CreateThread(nullptr, 0, EmuUpdateTickCount, nullptr, 0, nullptr);
         // We set the priority of this thread a bit higher, to assure reliable timing :
         SetThreadPriority(hThread, THREAD_PRIORITY_ABOVE_NORMAL);
+        g_AffinityPolicy->SetAffinityOther(hThread);
 
         CxbxKrnlRegisterThread(hThread);
         CloseHandle(hThread); // CxbxKrnlRegisterThread duplicates the handle so we can close this one
@@ -623,7 +624,7 @@ void CxbxInitWindow(bool bFullInit)
 			EmuShared::Cleanup();
 			ExitProcess(0);
 		}
-		SetThreadAffinityMask(hRenderWindowThread, g_CPUOthers);
+		g_AffinityPolicy->SetAffinityOther(hRenderWindowThread);
 
 		// Wait for the window to create
 		WaitForSingleObject(hStartEvent, INFINITE);
@@ -2092,7 +2093,7 @@ static DWORD WINAPI EmuUpdateTickCount(LPVOID)
 	CxbxSetThreadName("Cxbx Timing Thread");
 
     // since callbacks come from here
-	InitXboxThread(g_CPUOthers); // avoid Xbox1 core for lowest possible latency
+	InitXboxThread();
 
     EmuLog(LOG_LEVEL::DEBUG, "Timing thread is running.");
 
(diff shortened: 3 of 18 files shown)


LukeUsher (approved):
I love this, the AffinityPolicy interface is especially tidy.

Approved tentatively, let's wait for some Windows 7 reports before merging.

src/common/input/SdlJoystick.cpp
```diff
@@ -84,8 +84,6 @@ namespace Sdl
 		UpdateInputEvent_t = CustomEvent_t + 2;
 		DeviceRemoveAck_t = CustomEvent_t + 3;
 
-		SetThreadAffinityMask(GetCurrentThread(), g_CPUOthers);
```
PatrickvL, line 87: Why is affinity setting removed here?

CookiePLMonster: This code runs inside `cxbx.exe`, so the affinity policy is not set up for it in the first place.

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -11,10 +11,6 @@ VertexShaderSource g_VertexShaderSource = VertexShaderSource();
 // (And the ResetD3DDevice method should be removed)
 
 ID3DBlob* AsyncCreateVertexShader(IntermediateVertexShader intermediateShader, ShaderKey key) {
-	// HACK set thread affinity every call to reduce interference with Xbox main thread
```
PatrickvL, line 14: Any idea why this hack was here in the first place?

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -11,10 +11,6 @@ VertexShaderSource g_VertexShaderSource = VertexShaderSource();
...
-	// TODO use a thread pool library for better control over workers
```
PatrickvL, line 15: Shouldn't this TODO be kept?

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -11,10 +11,6 @@ VertexShaderSource g_VertexShaderSource = VertexShaderSource();
...
-	SetThreadAffinityMask(GetCurrentThread(), g_CPUOthers);
```
PatrickvL, line 16: Is this another one of those cases whereby the thread is already running on other threads? Hopefully there IS a place somewhere that this thread is still explicitly set on one of the other threads/ CPU sets?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2092,7 +2093,7 @@ static DWORD WINAPI EmuUpdateTickCount(LPVOID)
 	CxbxSetThreadName("Cxbx Timing Thread");
 
     // since callbacks come from here
-	InitXboxThread(g_CPUOthers); // avoid Xbox1 core for lowest possible latency
+	InitXboxThread();
```
PatrickvL, line 2096: Why no explicit set to Other here?

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -377,7 +375,7 @@ static unsigned int WINAPI CxbxKrnlInterruptThread(PVOID param)
 	CxbxSetThreadName("CxbxKrnl Interrupts");
 
 	// Make sure Xbox1 code runs on one core :
-	InitXboxThread(g_CPUXbox);
```
PatrickvL, line 380: This looses the call to InitXboxThread - is that intentional,  and if so:  why?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -599,6 +599,7 @@ void CxbxInitWindow(bool bFullInit)
         HANDLE hThread = CreateThread(nullptr, 0, EmuUpdateTickCount, nullptr, 0, nullptr);
         // We set the priority of this thread a bit higher, to assure reliable timing :
         SetThreadPriority(hThread, THREAD_PRIORITY_ABOVE_NORMAL);
+        g_AffinityPolicy->SetAffinityOther(hThread);
```
PatrickvL, line 603: Are we sure this thread is not supposed to run Xbox code? Because if it does run Xbox code, it should be set to the corresponding affinity...

src/core/kernel/init/CxbxKrnl.cpp
CookiePLMonster: That's a mistake, will correct.

src/common/Timer.cpp
```diff
@@ -102,9 +102,9 @@ void ClockThread(TimerObject* Timer)
 	if (!Timer->Name.empty()) {
 		CxbxSetThreadName(Timer->Name.c_str());
 	}
-	if (Timer->CpuAffinity != nullptr) {
-		InitXboxThread(*Timer->CpuAffinity);
-	}
+	InitXboxThread();
+	g_AffinityPolicy->SetAffinityOther();
```
PatrickvL, line 106: Even though the only use of this timer stuff indeed seemed to assign non-Xbox affinity, making this change correct,  it still seems odd that right above this thread is initialized as an Xbox thread,  so how come it's set to Other affinity?

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
LukeUsher: The windows 10 threading system we use here acheives this goal, no?

CookiePLMonster: Indeed it does, that's the entire reason I found out about CPU Sets in the first place! The way it was done currently was a hack and it changed the affinity of some random worker thread(s) Cxbx-R shouldn't be touching at all.

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
CookiePLMonster: See comment above.

src/core/kernel/init/CxbxKrnl.cpp
LukeUsher: The reason this probably never showed up in testing/still works regardless, is ALL xbox games spawn a new thread *immediately*, and return as part of the startup routine, meaning that no game code ever runs on the initial thread.

src/common/input/SdlJoystick.cpp
ergo720: This code actually runs in both `cxbx.exe` and `cxbx-emu.dll`. For the latter it's used to provide sdl input to the games from sdl devices The thread calls `SDL_WaitEvent`, which blocks it until there is an sdl event available. This happens when an sdl device receives input or an input message comes from the gui because the user updated the input configuration in the input gui. That being said, since this is not a busy loop, it might not be too critical to set the affinity for it, so this is probably fine as is?

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
CookiePLMonster: See comment above.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: Moved to `CxbxInitWindow`.

src/common/Timer.cpp
ergo720: There's actually another use of it in `OHCI_BusStart`, where the affinity is set to `nullptr` and thus no affinity is required. However, if in the future we create a new timer which runs in the xbox cpu, then indeed this will become wrong, so it's probably better to correct this. In addition, the xbox thread should only be initialized as such if this is going to be an actual xbox thread, since it's possible with the current setup to have a timer running as a host thread instead.

CookiePLMonster: Do you suggest turning this into a `Timer->isXbox` boolean, and alter affinities/xbox thread state basing on that? Sounds good to me.

src/common/input/SdlJoystick.cpp
CookiePLMonster: Yeah I don't think it's worth the "effort" to fix it, unless `if (g_AffinityPolicy)` works in which case I'll do it. Will test.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: I don't know, I kept the affinity the way it was before this PR.

LukeUsher: This thread *does* run some Xbox code: namely, it is the thread that calls Vblank callbacks.
But it is also the thread that detects when vblanks should happen, hence it runs on another thread.

PatrickvL (approved):
I think we're good

PR #2152, opened 2021-03-04
Added option to allow kb/mo input when the rendering window is unfocused

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2142. As the title says, this adds a new option in the input gui that allows keyboard/mouse input to be registered by the game even when the cursor is outside of the rendering window.

Issue #1136, opened 2018-05-05
Re-indent all our code


PatrickvL:
Instead of everyone applying their own indenting style to Cxbx-Reloaded code via relatively small pull requests, update our entire code base to one style, in one go, without any other change, and if possible make it so that future code automatically adheres to this style

x1nixmzeng:
We could hook up https://clang.llvm.org/docs/ClangFormat.html - but it'd be worth defining some style guidelines first.

The downside is any third-party libraries we've imported into the repro (rather than moduled).

PatrickvL:
True. Those would have to be excluded from automatic reformatting (except OpenXDK perhaps, since that's been modified so much, it's hardly the same thing anymore, plus, the project itself is in hibernation)

RadWolfie:
Currently, git doesn't work with intending our code before it is pushed into repo. It seems to require some sort of extension in order to have filter option working. Only new lines are working.

I would say Clang format extension is a requirement plus AppVeyor has support for this? Or would it be TravisCI?

x1nixmzeng:
You can setup commit hooks, which run local scripts before committing.

Personally, I wouldn't bother enforcing standards on a build machine - having an automated build fail because of style is a waste of time in my opinion :stuck_out_tongue:

PatrickvL:
Keeping the style consistent first and foremost requires a choice for an actual style.
Automation is of later concern.

Can we list and discuss a few smart styles here?

RadWolfie:
@x1nixmzeng Not true, it will tells us the contributor isn't following our standard which need to be fix before it can go into master.

Currently, both source code and github repo are showing both spaces and tabs making it look out of order which is not a good design. Using Clang format enforcement will help smooth the workflow better into main branch. I want to mention this, @LukeUsher did state Clang format extension is cross-platform supportive.

I think we need to change the topic to "format standard" with listing in op post.

PatrickvL:
Let's keep it simple. First define a style. Then apply it to all our code. Then document it. And finally evaluate cost versus benefit of automation.

RadWolfie:
We can create `.editorconfig` file with the following code below to force fix github's viewing issue between tab and spaces mixture problem.
```
[*]
indent_size = 4
```
I do not understand why github is using 8 spaces for a single tab.

`.editorconfig` file extension is supported without plugin requirement by the way.

[...]

PR #2174, opened 2021-03-24
Add Basic ImGui Support

### LINKED ISSUE #767: Emulation overlay GUI
At some point in time, it might become useful to add a GUI on top of the emulated screen output.
It could show a debugging command line, performance counters, memory statistics, memory watches, and whatever else we might come up with.

https://github.com/ocornut/imgui seems like The library to use for adding a GUI on top of the Xbox output.

RadWolfie:
**NOTICE: Any new features will be require put into later pull request.**

Currently open for review remarks at the moment.

---

Basic ImGui introduction support into Cxbx-Reloaded.
close #767

The code is partially obtained from Dolphin emulator's source code.

Following features has been implemented:
* Render backends:
  * Direct3D9
  * OpenGL
* Overlay:
  * Main menu bar
  * FPS
  * HLE/LLE stats
  * Vertex Buffer (was from console prints)
  * Audio Buffer Cache (was from console prints)
* Ability to remember windows last position in memory and file (currently disabled) on start and restart of emulation.
*  F1 permanent hotkey to toggle overlay view, any opened windows will remain open.
   * Also will be remembered to keep ImGui focus as well.

With F1 hotkey toggle to show our ImGui overlay, all inputs into the title are ignored.

It also contain few minor bugfixes from inputs' side.

While working on imgui tasks, I found several issues with rendering ImGui which will be create as separate ticket.

### DIFF
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -458,7 +458,7 @@ bool Settings::LoadConfig()
 
 	m_input_general.MoAxisRange = m_si.GetLongValue(section_input_general, sect_input_general.mo_axis_range, MO_AXIS_DEFAULT_RANGE);
 	m_input_general.MoWheelRange = m_si.GetLongValue(section_input_general, sect_input_general.mo_wheel_range, MO_WHEEL_DEFAULT_RANGE);
-	m_input_general.IgnoreKbMoUnfocus = m_si.GetLongValue(section_input_general, sect_input_general.ignore_kbmo_unfocus, 1);
+	m_input_general.IgnoreKbMoUnfocus = m_si.GetBoolValue(section_input_general, sect_input_general.ignore_kbmo_unfocus, true);
 
 	// ==== Input General End ==============
 
@@ -631,7 +631,7 @@ bool Settings::Save(std::string file_path)
 
 	m_si.SetLongValue(section_input_general, sect_input_general.mo_axis_range, m_input_general.MoAxisRange, nullptr, false, true);
 	m_si.SetLongValue(section_input_general, sect_input_general.mo_wheel_range, m_input_general.MoWheelRange, nullptr, false, true);
-	m_si.SetLongValue(section_input_general, sect_input_general.ignore_kbmo_unfocus, m_input_general.IgnoreKbMoUnfocus, nullptr, false, true);
+	m_si.SetBoolValue(section_input_general, sect_input_general.ignore_kbmo_unfocus, m_input_general.IgnoreKbMoUnfocus, nullptr, true);
 
 	// ==== Input General End =========
 
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -112,6 +112,12 @@ static struct {
 	const char* RenderResolution = "RenderResolution";
 } sect_video_keys;
 
+static const char* section_overlay = "overlay";
+static struct {
+	const char* FPS = "FPS";
+	const char* hle_lle_stats = "HLE/LLE Stats";
+} sect_overlay_keys;
+
 static const char* section_audio = "audio";
 static struct {
 	const char* adapter = "adapter";
@@ -523,6 +529,13 @@ bool Settings::LoadConfig()
 
 	// ==== Input Profile End ======
 
+	// ==== Overlay Begin =========
+
+	m_overlay.fps = m_si.GetBoolValue(section_overlay, sect_overlay_keys.FPS, true);
+	m_overlay.hle_lle_stats = m_si.SetBoolValue(section_overlay, sect_overlay_keys.hle_lle_stats, true);
+
+	// ==== Overlay End ===========
+
 	return true;
 }
 
@@ -588,6 +601,7 @@ bool Settings::Save(std::string file_path)
 	m_si.SetBoolValue(section_video, sect_video_keys.FullScreen, m_video.bFullScreen, nullptr, true);
 	m_si.SetBoolValue(section_video, sect_video_keys.MaintainAspect, m_video.bMaintainAspect, nullptr, true);
 	m_si.SetLongValue(section_video, sect_video_keys.RenderResolution, m_video.renderScaleFactor, nullptr, false, true);
+
 	// ==== Video End ===========
 
 	// ==== Audio Begin =========
@@ -700,6 +714,13 @@ bool Settings::Save(std::string file_path)
 
 	// ==== Input Profile End ======
 
+	// ==== Overlay Begin =======
+
+	m_si.SetBoolValue(section_overlay, sect_overlay_keys.FPS, m_overlay.fps, nullptr, true);
+	m_si.SetBoolValue(section_overlay, sect_overlay_keys.hle_lle_stats, m_overlay.hle_lle_stats, nullptr, true);
+
+	// ==== Overlay End =========
+
 	// ==== Hack Begin ==========
 
 	m_si.SetBoolValue(section_hack, sect_hack_keys.DisablePixelShaders, m_hacks.DisablePixelShaders, nullptr, true);
@@ -737,6 +758,7 @@ void Settings::SyncToEmulator()
 
 	// register Video settings
 	g_EmuShared->SetVideoSettings(&m_video);
+	g_EmuShared->SetOverlaySettings(&m_overlay);
 
 	// register Audio settings
 	g_EmuShared->SetAudioSettings(&m_audio);
--- a/src/common/Settings.cpp
+++ b/src/common/Settings.cpp
@@ -744,7 +744,7 @@ void Settings::SyncToEmulator()
 	// register Network settings
 	g_EmuShared->SetNetworkSettings(&m_network);
 
-	// register Input settings
+	// register Input gamepad settings
 	for (int i = 0; i < 4; i++) {
 		g_EmuShared->SetInputDevTypeSettings(&m_input_port[i].Type, i);
 		if (m_input_port[i].Type != to_underlying(XBOX_INPUT_DEVICE::DEVICE_INVALID)) {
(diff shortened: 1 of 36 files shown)


src/core/common/imgui/ui.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+#define LOG_PREFIX CXBXR_MODULE::GUI
+
+#include <thread>
+
+#include "ui.hpp"
+#include "EmuShared.h"
+
+#include "core/kernel/init/CxbxKrnl.h"
+#pragma optimize("", off)
```
RadWolfie: Whoops, forgot to remove this as well.

medievil1:
games give the same error (location)
Halo
indiana jones


image

RadWolfie:
I will investigate with Halo since I haven't given it a try with ImGui build yet.

RadWolfie:
> games give the same error (location)
> Halo
> indiana jones

Hmm... on first `CreateDefaultD3D9Device` call, there is no issue. However, on 2nd call of `CreateDefaultD3D9Device`. It did not like it at all. So, it's calling one of `Direct3D_CreateDevice` patches twice. Further investigation is require in order to fix Halo's issue.

medievil1:
actually makes sense, cause both halo and indiana jones have an issue in GPA where they won't cap when actual 3d (non video) starts(been that way for 7 or 8 months now).... might not be an issue with imgui and instead be cxbxr doing something it shouldn't

RadWolfie:
> Hmm... on first `CreateDefaultD3D9Device` call, there is no issue. However, on 2nd call of `CreateDefaultD3D9Device`. It did not like it at all. So, it's calling one of `Direct3D_CreateDevice` patches twice. Further investigation is require in order to fix Halo's issue.

Okay, the issue is directly from this pull request. Some fixes are require from ImGui implement's end and cleanup as well.

RadWolfie:
Actually, it's simple than I thought it would be. Solved in commit 8194315 for Halo's crash due to some unnecessary shutdown calls on device re-creation.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2403,11 +2403,7 @@ static void CreateDefaultD3D9Device
         // Final release of IDirect3DDevice9 must be called from the window message thread
         // See https://docs.microsoft.com/en-us/windows/win32/direct3d9/multithreading-issues
         RunOnWndMsgThread([] {
-            // TODO: Need implement...
             ImGui_ImplDX9_Shutdown();
-            ImGui_ImplWin32_Shutdown();
-            g_renderbase->Shutdown();
-            g_renderbase.release();
```
RadWolfie: Oops, wasn't my intention to put them in here. I may had been thinking about emulation's terminate process. And this wasn't the place.

RadWolfie:
Two things I need to work on. One is early termination cause a crash. Second is implement shutdown process in CxbxKrnlShutDown function and somewhere in reboot process.

medievil1:
Halo confirmed working with latest commit (imgui not game..lol)
however indiana jones does not, there is no error, just after videos, overlay disappears
image

RadWolfie:
> however indiana jones does not, there is no error, just after videos, overlay disappears. Indy goes black screen at this point normally(cxbxr bug).....

Can you take a look at the ticket I had made for known issue, which has a few different issues for overlay itself, for cause of imgui overlay to disappear?

medievil1:
> 
> 
> > however indiana jones does not, there is no error, just after videos, overlay disappears. Indy goes black screen at this point normally(cxbxr bug).....
> 
> Can you take a look at the ticket I had made for known issue, which has a few different issues for overlay itself, for cause of imgui overlay to disappear? Meaning, make sure it's not the same type of issue you have seen cause overlay to disappeared and use a method that is known to work.

I'll look, honestly though I don't think it is a imgui or this PR issue...I think for that game cxbxr is completely disconnecting from the display device at that point and so nothing gets through, so I wouldn't concentrate or put to much time into this as an issue...

src/common/Settings.cpp
```diff
@@ -737,14 +758,15 @@ void Settings::SyncToEmulator()
...
+	g_EmuShared->SetOverlaySettings(&m_overlay);
 
 	// register Audio settings
 	g_EmuShared->SetAudioSettings(&m_audio);
 
 	// register Network settings
 	g_EmuShared->SetNetworkSettings(&m_network);
 
-	// register Input settings
+	// register Input gamepad settings
```
ergo720, line 769: This addition is wrong. This is registering the input settings for the xbox devices attached to the 4 xbox ports regardless of the xbox device type. So, specifying "gamepad" gives the impression that it only works for those, which it doesn't.

src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -488,13 +488,16 @@ XBSYSAPI EXPORTNUM(49) xbox::void_xt DECLSPEC_NORETURN NTAPI xbox::HalReturnToFi
...
+	bool is_reboot = false;
+
 	switch (Routine) {
 	case ReturnFirmwareHalt:
 		CxbxKrnlCleanup("Emulated Xbox is halted");
 		break;
 
 	case ReturnFirmwareReboot:
 		LOG_UNIMPLEMENTED(); // fall through
+		[[fallthrough]];
```
ergo720, line 500: Missing empty line between the two cases.

src/gui/WndMain.cpp
```diff
@@ -387,6 +387,9 @@ LRESULT CALLBACK WndMain::WndProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lP
 							g_EmuShared->SetIsReady(true);
 						}
 						break;
+						case ID_GUI_STATUS_OVERLAY:
```
ergo720, line 390: Missing empty line between the two cases.

src/common/Settings.cpp
RadWolfie: Well, it set the device type plus selected gamepad settings for each xbox port. If you look down about 20 lines, you can see there's a general Input settings which I combine them into one. Something should *had* been done before.

So, using "Input settings" isn't right and need its purpose. Even put this under general category doesn't fit as well. One other way to say is "register Input active device profile settings".

ergo720: I suggest to use "register xbox device input settings", because it's registering the input settings (type and profile) of the attached xbox device the user has configured.

src/common/input/InputManager.cpp
```diff
@@ -354,8 +355,9 @@ bool InputDeviceManager::UpdateXboxPortInput(int usb_port, void* Buffer, int Dir
 		xid_type < to_underlying(XBOX_INPUT_DEVICE::DEVICE_MAX));
 	bool has_changed = false;
 
+	// First check if ImGui is focus, then ignore any input update occur.
 	// If somebody else is currently holding the lock, we won't wait and instead report no input changes
-	if (m_Mtx.try_lock()) {
+	if (!g_renderbase->IsImGuiFocus() && m_Mtx.try_lock()) {
```
ergo720, line 360:
```suggestion
	if (static_cast<uint8_t>(!g_renderbase->IsImGuiFocus()) & static_cast<uint8_t>(m_Mtx.try_lock())) {
```

It's the same optimization I used in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2152 to reduce the impact of branch misprediction.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -709,6 +744,27 @@ void CxbxReleaseCursor()
 	ClipCursor(nullptr);
 }
 
+static void CxbxUpdateCursor(bool forceShow = false) {
+	// Getting cursor info is a requirement in order to prevent a bug occur with ShowCursor redundant calls.
```
ergo720, line 748: I'm curious: I'm not aware of any bug in `ShowCursor` caused by "redundant calls". Is this something that was introduced with this PR or an existing bug?

src/core/hle/D3D8/XbVertexBuffer.h
```diff
@@ -40,7 +40,7 @@ typedef struct _CxbxDrawContext
     IN     DWORD                 dwStartVertex; // Only D3DDevice_DrawVertices sets this (potentially higher than default 0)
 	IN	   PWORD				 pXboxIndexData; // Set by D3DDevice_DrawIndexedVertices, D3DDevice_DrawIndexedVerticesUP and HLE_draw_inline_elements
 	IN	   DWORD				 dwBaseVertexIndex; // Set to g_Xbox_BaseVertexIndex in D3DDevice_DrawIndexedVertices
-	IN	   INDEX16				 LowIndex, HighIndex; // Set when pXboxIndexData is set
+	IN	   INDEX16               LowIndex, HighIndex; // Set when pXboxIndexData is set
```
ergo720, line 43: Nit: `LowIndex, HighIndex` are now misaligned. Well, actually all the members of this struct are misaligned...

RadWolfie: Nit: They are aligned if you have 4 spaces for a tab. Except there should **NOT** be any tabs between characters. When I made a change to WORD, this is what happen. Except I will not be undo the spaces back to tab.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
RadWolfie: If TRUE or FALSE was input twice. Then the invert value will not cause expected behavior for some reason. And yes it was introduced in this PR bug. I can't remember if I did reproduce it outside of this pull request or not.

ref: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showcursor
> If bShow is TRUE, the display count is incremented by one. If bShow is FALSE, the display count is decremented by one.

src/core/hle/D3D8/XbVertexBuffer.h
ergo720: We should definitely re-indent all our code like this issue [here](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1136) suggests, to avoid such issues in the future. For now this is fine then.

PatrickvL (approved):

Issue #2173, opened 2021-03-24
XInputPoll is not implemented


LukeUsher:
## Quick summary

The XInputPoll function is not implemented, we always treat input as 'automatically polled'

## Details
1. We need to save/cache the setting of pPollingParameters ->fAutoPoll in XInputOpen
2. If pPollingParameters ->fAutoPoll is true,  XInputGetState should automatically update input
3. if pPollingParameters ->fAutoPoll is false, XInputGetState should not update input, just return the last known state
4. if pPollingParameters ->fAutoPoll is true, XInputPoll should should do nothing
5. if pPollingParameters ->fAutoPoll is false, XInputPoll should update the input, so the next call to XInputGetState returns it.

### System Configuration:
N/A

### Additional Information (if any):
This may be the cause of some not-yet solved issues where input is broken in some games, this needs investigation

PR #2176, opened 2021-03-24
Implemented XInputPoll

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2173. Tested with Oddworld: Stranger's Wrath, Jet Set Radio Future, Splinter Cell 2, Panzer Dragoon Orta and 1 (which all call this function) and observed no regressions.

### DIFF
--- a/src/core/hle/XAPI/Xapi.cpp
+++ b/src/core/hle/XAPI/Xapi.cpp
@@ -55,10 +55,10 @@ static CXBX_XINPUT_IN_STATE g_InState[4];
 
 // Global bridge for xbox controller to host, 4 elements for 4 ports.
 CXBX_CONTROLLER_HOST_BRIDGE g_XboxControllerHostBridge[4] = {
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[0], false, false, false, false, { 0, 0, 0, 0, 0 } },
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[1], false, false, false, false, { 0, 0, 0, 0, 0 } },
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[2], false, false, false, false, { 0, 0, 0, 0, 0 } },
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[3], false, false, false, false, { 0, 0, 0, 0, 0 } },
+	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[0], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
+	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[1], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
+	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[2], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
+	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[3], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
 };
 
 
@@ -103,7 +103,7 @@ bool ConstructHleInputDevice(int Type, int Port)
 	g_bIsDevicesEmulating = true;
 	bool ret = true;
 
-	// NOTE: initialize bAutopoll to its default state, which varies depending on the device type
+	// NOTE: initialize bAutoPollDefault to its default state, which varies depending on the device type
 	switch (Type)
 	{
 	case to_underlying(XBOX_INPUT_DEVICE::MS_CONTROLLER_DUKE): {
@@ -112,7 +112,7 @@ bool ConstructHleInputDevice(int Type, int Port)
 		g_XboxControllerHostBridge[Port].bPendingRemoval = false;
 		g_XboxControllerHostBridge[Port].bSignaled = false;
 		g_XboxControllerHostBridge[Port].bIoInProgress = false;
-		g_XboxControllerHostBridge[Port].bAutoPoll = true;
+		g_XboxControllerHostBridge[Port].bAutoPollDefault = true;
 		g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucType = XINPUT_DEVTYPE_GAMEPAD;
 		g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucSubType = XINPUT_DEVSUBTYPE_GC_GAMEPAD;
 		g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucInputStateSize = sizeof(XpadInput);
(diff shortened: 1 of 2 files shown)


src/core/hle/XAPI/Xapi.cpp
```diff
@@ -520,16 +541,18 @@ xbox::dword_xt WINAPI xbox::EMUPATCH(XInputGetState)
...
+			g_XboxControllerHostBridge[Port].bIoInProgress = true;
+			if (g_InputDeviceManager.UpdateXboxPortInput(Port, g_XboxControllerHostBridge[Port].InState, DIRECTION_IN, to_underlying(g_XboxControllerHostBridge[Port].XboxType))) {
+				g_XboxControllerHostBridge[Port].XboxDeviceInfo.dwPacketNumber++;
+			}
+			g_XboxControllerHostBridge[Port].bIoInProgress = false;
+		}
+
+		std::memcpy((void *)&pState->Gamepad, g_XboxControllerHostBridge[Port].InState, g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucInputStateSize);
+		pState->dwPacketNumber = g_XboxControllerHostBridge[Port].XboxDeviceInfo.dwPacketNumber;
+		ret = ERROR_SUCCESS;
```
PatrickvL, line 555: This implementation is nearly identical to `XInputPoll` - would it be much trouble to merge them together into one `CxbxImpl_XInputHandler` function, with the difference between the two controlled by a boolean argument?

PatrickvL (commented):
I can't judge the change the impact of the type-change of dwPacketNumber (maybe @RadWolfie  can?)

Otherwise, only one remark, nice work!

ergo720:
> I can't judge the change the impact of the type-change of dwPacketNumber (maybe @RadWolfie can?)

dwPacketNumber is used by the games to know if the device input state has changed. If it's different than the last known value, then the state changed. dwPacketNumber is DWORD, but it was accidentally set to uint16_t in the SBC PR. Since this value is incremented by one every time the connected host device has its input state changed, setting it to a 16 bit value has the effect that the counter will overflow much sooner than it should. This won't cause problems in practice because, when the value goes back to zero from 0xFFFF, the resulting zero value is still different than 0xFFFF, and so the game will still know that something changed. That being said, it's better to correct it anyway and use the appropriate type for dwPacketNumber.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -336,6 +336,45 @@ void UpdateConnectedDeviceState(xbox::PXPP_DEVICE_TYPE DeviceType) {
...
+template<bool IsXInputPoll>
+xbox::dword_xt CxbxImpl_XInputHandler(xbox::HANDLE hDevice, xbox::PXINPUT_STATE pState)
+{
+	xbox::dword_xt status = ERROR_DEVICE_NOT_CONNECTED;
+	PCXBX_CONTROLLER_HOST_BRIDGE Device = (PCXBX_CONTROLLER_HOST_BRIDGE)hDevice;
+	int Port = Device->XboxPort;
+
+	if constexpr (IsXInputPoll) {
+		if ((g_XboxControllerHostBridge[Port].hXboxDevice == hDevice) && !g_XboxControllerHostBridge[Port].bPendingRemoval) {
+			if (!g_XboxControllerHostBridge[Port].bAutoPoll) {
```
PatrickvL, line 357: Instead of a template, this could've been done by making `IsXInputPoll` an argument, and make this line read:
```
  if (g_XboxControllerHostBridge[Port].bAutoPoll != IsXInputPoll) {
```

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -336,6 +336,45 @@ void UpdateConnectedDeviceState(xbox::PXPP_DEVICE_TYPE DeviceType) {
...
+		}
+	}
+	else {
+		if ((g_XboxControllerHostBridge[Port].hXboxDevice == hDevice) && (!g_XboxControllerHostBridge[Port].bPendingRemoval)) {
+			if (g_XboxControllerHostBridge[Port].bAutoPoll) {
+				GetDeviceInput(Port);
+			}
+
+			std::memcpy((void *)&pState->Gamepad, g_XboxControllerHostBridge[Port].InState, g_XboxControllerHostBridge[Port].XboxDeviceInfo.ucInputStateSize);
+			pState->dwPacketNumber = g_XboxControllerHostBridge[Port].XboxDeviceInfo.dwPacketNumber;
```
PatrickvL, line 370: ..and wrap these two lines in an `if (IsXInputPoll) {` block.

PatrickvL (approved):
Thanks for picking up the previous remark. Allow me to describe what I though was going to happen. You judge what's better. This is fine like it is now too 🤷

ergo720:
It's done this way as an optimization. C++17 has introduced the `if constexpr` statement, which forces the compiler to evaluate the if statement at compile time and throws an error if it can't. Your suggestion to make `IsXInputPoll` an argument will add an additional branching done at runtime, while with this method the branching is optimized away because the compiler does the comparison at compile time. I'm using a template because the result of the conditional statement must be known at compile time, and a template argument satisfies the requirement because it will be instantiated at compile time, while a runtime argument violates it.

PatrickvL:
Interesting, but wouldn't the test be optimized away if coded out like i thought,  but keeping the template argument? I ask, because would allow the two very similar code blocks to be folded into one...

RadWolfie:
> Interesting, but wouldn't the test be optimized away if coded out like i thought, but keeping the template argument? I ask, because would allow the two very similar code blocks to be folded into one...

What I think Patrick is saying... if we understand how compiler construct runtime code will make only one assembly block in the binary. While compiled time code will duplicate assembly blocks for two different things in the binary. Do correct me if I'm wrong please. @PatrickvL 😅

ergo720:
> While compiled time code will duplicate assembly blocks for two different things in the binary. Do correct me if I'm wrong please.

This is not true, the C++ standard says this: "In a constexpr if statement, the value of condition must be a contextually converted constant expression of type bool. If the value is true, then statement-false is discarded (if present), otherwise, statement-true is discarded. ".

Summary:
* Has its condition evaluated at compile time once all necessary arguments have been supplied
* Determines which of the two substatements to compile, discarding the other
* Does not require the discarded substatement to be well-formed

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -330,6 +336,45 @@ void UpdateConnectedDeviceState(xbox::PXPP_DEVICE_TYPE DeviceType) {
...
+
+template<bool IsXInputPoll>
+xbox::dword_xt CxbxImpl_XInputHandler(xbox::HANDLE hDevice, xbox::PXINPUT_STATE pState)
+{
+	xbox::dword_xt status = ERROR_DEVICE_NOT_CONNECTED;
+	PCXBX_CONTROLLER_HOST_BRIDGE Device = (PCXBX_CONTROLLER_HOST_BRIDGE)hDevice;
+	int Port = Device->XboxPort;
+
+	if constexpr (IsXInputPoll) {
+		if ((g_XboxControllerHostBridge[Port].hXboxDevice == hDevice) && !g_XboxControllerHostBridge[Port].bPendingRemoval) {
```
CookiePLMonster, line 356: If I'm reading this correctly, code duplication could be minimally reduced by exchanging the order of this condition and `if constexpr` above - that should make @PatrickvL a bit more happy I believe :)

PatrickvL: This +1

PatrickvL:
Good discussion. I wasn't implying reduction of generated assembly code (sorry Radwolfie). It's just that I noticed both code paths were so similar,  that it could be folded into one.  This template argument 'merely' makes it more optimal than a function argument, at the expense of causing the compiler to emit two variants, but otherwise it's still the same goal : reduce code duplication

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -330,6 +336,33 @@ void UpdateConnectedDeviceState(xbox::PXPP_DEVICE_TYPE DeviceType) {
...
+	if ((g_XboxControllerHostBridge[Port].hXboxDevice == hDevice) && !g_XboxControllerHostBridge[Port].bPendingRemoval) {
+		if (g_XboxControllerHostBridge[Port].bAutoPoll != IsXInputPoll) {
+			g_XboxControllerHostBridge[Port].bIoInProgress = true;
+			if (g_InputDeviceManager.UpdateXboxPortInput(Port, g_XboxControllerHostBridge[Port].InState, DIRECTION_IN, to_underlying(g_XboxControllerHostBridge[Port].XboxType))) {
+				g_XboxControllerHostBridge[Port].XboxDeviceInfo.dwPacketNumber++;
+			}
+			g_XboxControllerHostBridge[Port].bIoInProgress = false;
+		}
+
+		if (!IsXInputPoll) {
```
CookiePLMonster, line 355: Did you miss an `if constexpr` here?

PatrickvL (approved):

PR #2178, opened 2021-03-30
Minor kernel fixes

RadWolfie:
The changes are minor and should not have any impact on general titles. The only function changes are:
- Lower our forward symbolic link interaction into Ob function area.
- Add a warning message for further research necessary to fix xbox::ObInitSystem.
- Finally, found a missing `ObpRootDirectoryObject` assignment for `ObpReferenceObjectByName` function due to unconditional crash occur without it. It doesn't resolve ObInitSystem's issue.
- Fixed an issue with PSTRING contain null pointer for xbox kernel test suite's IoCreateSymbolicLink/IoDeleteSymbolicLink tests.

In summary, there are 4 changes made. As for the rest are cleaning up the format discovered in the process.


src/core/kernel/exports/EmuKrnlOb.cpp
```diff
@@ -772,10 +774,63 @@ XBSYSAPI EXPORTNUM(243) xbox::ntstatus_xt NTAPI xbox::ObOpenObjectByName
 		LOG_FUNC_ARG_OUT(Handle)
 		LOG_FUNC_END;
 
-	LOG_UNIMPLEMENTED();
-	assert(false);
+#if 0 // Enable when ObInitSystem/IoCreateDevice are functional.
```
PatrickvL, line 777: Please, also add a comment that for the current purpose (being : supporting symbolic links), the two TODO's in the `ObSymbolicLinkObjectType` struct should be added (namely : `ObpDeleteSymbolicLink` and `ObpDefaultObject`)

Note : `ObpDeleteSymbolicLink` should probably be implemented as a single call to `ObDereferenceObject` on it's argument object (`((POBJECT_SYMBOLIC_LINK)Object)->LinkTargetObject`) - nothing fancy there.

src/core/kernel/support/EmuFile.cpp
```diff
@@ -461,13 +461,13 @@ NTSTATUS CxbxConvertFilePath(
...
 				}
-				else if (*RootDirectory == (NtDll::HANDLE)-3) {
+				else if (*RootDirectory == ObDosDevicesDirectory()) {
 					// This is a special handle that tells the API that this is a DOS device
 					// We can safely remove it and forward to the Xbe directory.
 					// Test case GTA3
 					NtSymbolicLinkObject = FindNtSymbolicLinkObjectByRootHandle(g_hCurDir);
 				}
-				else if (*RootDirectory == (NtDll::HANDLE)-4) {
+				else if (*RootDirectory == ObWin32NamedObjectsDirectory()) {
```
PatrickvL, line 470: Nice finds!

src/core/kernel/support/EmuFile.cpp
```diff
@@ -705,7 +705,6 @@ NTSTATUS EmuNtSymbolicLinkObject::Init(std::string aSymbolicLinkName, std::strin
 						HostSymbolicLinkPath = HostSymbolicLinkPath + ExtraPath;
 				}
 
-				SHCreateDirectoryEx(NULL, HostSymbolicLinkPath.c_str(), NULL);
```
PatrickvL, line 708: Why was this removed? Is it because `CreateFile` will create all sub-folders if the given path doesn't yet exist on the host file-system? Are we sure about that? And if so, a comment in that regard would be helpful!

PatrickvL (commented):
A few remarks. Otherwise looks good.

src/core/kernel/exports/EmuKrnlOb.cpp
```diff
@@ -772,10 +774,63 @@ XBSYSAPI EXPORTNUM(243) xbox::ntstatus_xt NTAPI xbox::ObOpenObjectByName
...
+			ObfDereferenceObject(Object);
+			result = xbox::status_insufficient_resources;
+		}
+	}
+
+	*Handle = newHandle;
+
+	RETURN(result);
+
+#endif
```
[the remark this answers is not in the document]
RadWolfie: What would you advise to use as defined macro word for this? `XBOX_DEVICE_HW_ENABLE` maybe?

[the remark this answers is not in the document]
PatrickvL: Nah, it's still in it's infancy, `#if 0` will suffice for now. (Local testing would simply mean replacing 0 by 1 for a simple test.)

But if you're going to work on extending the kernel, and want to enable various changes with a single conditional define, then make the name reflect it's intent : CXBX_KERNEL_COMPLETION or something along those lines.

And for everything that's fully implemented, tested, deemed reliable and put into production, the conditional define check can be removed from it (plus any lingering old code).

src/core/kernel/exports/EmuKrnlOb.cpp
```diff
@@ -772,10 +774,63 @@ XBSYSAPI EXPORTNUM(243) xbox::ntstatus_xt NTAPI xbox::ObOpenObjectByName
...
+
+		if (newHandle == nullptr) {
+			ObfDereferenceObject(Object);
+			result = xbox::status_insufficient_resources;
+		}
+	}
+
+	*Handle = newHandle;
+
+	RETURN(result);
```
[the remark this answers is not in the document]
RadWolfie: Huh, I didn't notice `ObOpenObjectByPointer` is unimplemented. The code is almost the same. Though, the initial call is completely different which isn't compatible.

I can make an inline function for the latter method.

src/core/kernel/support/EmuFile.cpp
RadWolfie: Indeed! When I found these `ObWin32NamedObjectsDirectory` and `ObDosDevicesDirectory` references in our codebase. I figured they should be properly used here. 😅

src/core/kernel/support/EmuFile.cpp
RadWolfie: We already have EmuDisk's directory creation and binary file earlier. There's no need to duplicate directory's creation. Beside, the host's file path must exist first or otherwise we could be accepting any non-existing path.

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -873,33 +873,9 @@ XBSYSAPI EXPORTNUM(203) xbox::ntstatus_xt NTAPI xbox::NtOpenSymbolicLinkObject
 	IN POBJECT_ATTRIBUTES ObjectAttributes
 )
 {
-	/* TODO :
 	LOG_FORWARD("ObOpenObjectByName");
 
 	return ObOpenObjectByName(ObjectAttributes, &ObSymbolicLinkObjectType, NULL, LinkHandle);
```
ergo720, line 878: Nit: by using a plain `return`, we no longer log the return value of `NtOpenSymbolicLinkObject`. It might not be too critical since I see that `ObOpenObjectByName` logs it anyway, so you decide what's best.

src/core/kernel/support/EmuFile.cpp
PatrickvL: Notice how above, `ExtraPath` gets appended to the `HostSymbolicLinkPath` - so, potentially, there could be sub-folders mentioned in there, which Xbox code might rely on to get created automatically, or, if we can prove this - perhaps such calls should fail. Only when we know for sure how this scenario is supposed to work, we can update it's implementation to follow suite. That's why I'd caution against removing seemingly unneeded code, based on unproven assumptions. Like I said, at the very least put in a comment to explain what the reasoning is here.

src/core/kernel/exports/EmuKrnlNt.cpp
PatrickvL: Perhaps our `RETURN()` macro can be used for that, although because that's a macro, it requires an intermediate variable.
But then we'd be showing two identical return results in our logs, so I'm not sure that's wise.
Also, because this is a forwarding call (and that fact is already logged), we normally don't log anything in this function, instead we mostly rely on the called function to do the logging for us. (Just my 5 cents)

src/core/kernel/exports/EmuKrnlOb.cpp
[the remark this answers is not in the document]
RadWolfie: > Nah, it's still in it's infancy, `#if 0` will suffice for now. (Local testing would simply mean replacing 0 by 1 for a simple test.)

I kind of disagree with you on this part. Since two or three Io functions involve some lowest level of Ob function to allocate/free handle base on certain device usage. And of course each device interface has its own management which we haven't implement. The commented code I found here and there are actually part of direct interface with internal kernel work. Which should be part of toggle on/off switch with macro define.



> But if you're going to work on extending the kernel, and want to enable various changes with a single conditional define, then make the name reflect it's intent : CXBX_KERNEL_COMPLETION or something along those lines.

I'm thinking `CXBX_KERNEL_DEVICE_ENABLED` since the device system isn't implemented in order for what I said above to work properly.

src/core/kernel/exports/EmuKrnlNt.cpp
RadWolfie: Any forwards should not have any log returns since they belong in forward's function, that's my opinion. Is there a reason we should log forward function's return?

src/core/kernel/support/EmuFile.cpp
RadWolfie: I'm investigating what is using `ExtraPath` argument at the moment. Thanks for bringing it to my attention.

src/core/kernel/exports/EmuKrnlOb.cpp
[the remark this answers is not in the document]
RadWolfie: done

src/core/kernel/exports/EmuKrnlOb.cpp
```diff
@@ -772,10 +800,47 @@ XBSYSAPI EXPORTNUM(243) xbox::ntstatus_xt NTAPI xbox::ObOpenObjectByName
 		LOG_FUNC_ARG_OUT(Handle)
 		LOG_FUNC_END;
 
-	LOG_UNIMPLEMENTED();
-	assert(false);
+#ifdef CXBX_KERNEL_REWORK_ENABLED
+	PVOID Object;
+	HANDLE newHandle = nullptr;
+	ntstatus_xt result = ObpReferenceObjectByName(ObjectAttributes->RootDirectory, ObjectAttributes->ObjectName,
```
PatrickvL, line 806: Stray `newHandle` must be deleted

src/core/kernel/support/EmuFile.cpp
RadWolfie: @PatrickvL I made a test against hardware's IoCreateSymbolicLink. It is confirmed non-existence path will return `STATUS_OBJECT_NAME_NOT_FOUND`.

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -680,25 +679,26 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
...
 
-	NTSTATUS ret = xbox::status_success;
+	NTSTATUS result = xbox::status_success;
 
-	if (EmuHandle::IsEmuHandle(SourceHandle)) {
-		auto iEmuHandle = (EmuHandle*)SourceHandle;
-		ret = iEmuHandle->NtDuplicateObject(TargetHandle, Options);
-/*
-		PVOID Object;
+#ifdef CXBX_KERNEL_DEVICE_ENABLED
```
PatrickvL, line 684: CXBX_KERNEL_REWORK_ENABLED or DEVICE?

RadWolfie: 🤔 It's suppose to be `CXBX_KERNEL_REWORK_ENABLED`. I guess Visual Studio, usually my mistake, missed all the replacement I expect it to do.

RadWolfie: Corrected it in original commit.

ergo720 (approved):

PR #2179, opened 2021-03-31
ci: Use ubuntu-latest for release

Margen67:


RadWolfie:
There's an active incident with Actions and webhooks. That's the reason of no actions activity at the moment.

RadWolfie (approved):

PR #2182, opened 2021-04-05
Detect mismatches between the cxbxr executables

ergo720:
This adds a couple of checks to make sure that cxbxe.exe, cxbxr-emu.dll and cxbxr-ldr.exe all come from the same commit build, and aborts otherwise. This is to identify situations in which the user mixes different executable versions (for example, by dropping in the same folder different versions) and encounters otherwise unexplainable issues.

### DIFF
--- a/projects/cxbxr-ldr/CMakeLists.txt
+++ b/projects/cxbxr-ldr/CMakeLists.txt
@@ -25,6 +25,8 @@ endif()
 file (GLOB HEADERS
  "${CXBXR_ROOT_DIR}/src/common/AddressRanges.h"
  "${CXBXR_ROOT_DIR}/src/common/ReserveAddressRanges.h"
+ "${CXBXR_ROOT_DIR}/src/CxbxVersion.h"
+ "${CXBXR_ROOT_DIR}/src/version.h"
 )
 
 file (GLOB SOURCES
--- a/projects/cxbxr-ldr/CMakeLists.txt
+++ b/projects/cxbxr-ldr/CMakeLists.txt
@@ -70,3 +70,5 @@ target_link_libraries(cxbxr-ldr
 install(TARGETS ${PROJECT_NAME}
   RUNTIME DESTINATION bin
 )
+
+add_dependencies(cxbxr-ldr misc-batch)
--- a/projects/cxbxr-ldr/CMakeLists.txt
+++ b/projects/cxbxr-ldr/CMakeLists.txt
@@ -32,6 +32,7 @@ file (GLOB HEADERS
 file (GLOB SOURCES
  "${CXBXR_ROOT_DIR}/src/common/AddressRanges.cpp"
  "${CXBXR_ROOT_DIR}/src/common/ReserveAddressRanges.cpp"
+ "${CXBXR_ROOT_DIR}/src/CxbxVersion.cpp"
  "${CXBXR_ROOT_DIR}/src/loader/cxbxr-ldr.cpp"
 )
 
--- a/src/common/win32/EmuShared.h
+++ b/src/common/win32/EmuShared.h
@@ -30,8 +30,8 @@
 #include "Mutex.h"
 #include "common\IPCHybrid.hpp"
 #include "common\input\Button.h"
+#include "CxbxVersion.h"
 #include "core/common/imgui/settings.h"
-
 #include <memory.h>
 
 extern HMODULE hActiveModule; // Equals EXE Module handle in (GUI) Cxbx.exe / cxbxr.exe, equals DLL Module handle in cxbxr-emu.dll
@@ -65,6 +65,11 @@ class EmuShared : public Mutex
 		// * Fixed memory allocation size
 		// ******************************************************************
 		unsigned int m_size;
+
+		// ******************************************************************
+		// * Git version string of cxbx.exe
+		// ******************************************************************
+		char m_git_version[GitVersionMaxLength];
 
 		// ******************************************************************
 		// * Each process needs to call this to initialize shared memory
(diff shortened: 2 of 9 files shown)


src/common/win32/EmuShared.h
```diff
@@ -320,28 +330,20 @@ class EmuShared : public Mutex
...
 		float        m_FPS_status; // NOTE: If move into ipc_send_gui_update will spam GUI's message system (one message per frame)
 		bool         m_Krnl_Log_enabled; // Is require in order to preserve previous set for support multi-xbe.
 		bool         m_bDebugging;
 		bool         m_bReady_status;
 		bool         m_bEmulating_status;
 #ifndef CXBX_LOADER // Temporary usage for cxbx.exe's emu
 		unsigned int m_PreviousMmLayout;
-		int          m_Reserved7[3];
-#else
-		int          m_Reserved7[4];
```
PatrickvL, line 334: This reserved variable would still be needed for Emuzshared binary compatibility between loader and gui based emulation,  build from the same git version

src/loader/cxbxr-ldr.cpp
```diff
@@ -28,7 +28,9 @@
 
 #define WIN32_LEAN_AND_MEAN             // Exclude rarely-used stuff from Windows headers
 #include <Windows.h> // For LPTSTR, FormatMessage, GetSystemInfo, etc
+#include "strsafe.h" // For StringCchCopy
```
PatrickvL, line 31: Genuine question: have you checked this doesn't pull in any additional libraries,  not during linking and not during runtime?  If it does, that will sooner or later cause problems with the address range reservation

PatrickvL (changes requested):

src/common/win32/EmuShared.h
ergo720: @RadWolfie asked me to remove these reserved variables, because with the checks it's no longer possible to mix different executable versions and thus we don't need to worry about compatibility of EmuShared

src/loader/cxbxr-ldr.cpp
ergo720: I can't use `std::strncpy` here because the loader does not link against the CRT, so I have no choice but to use a Windows function.

ergo720: Ok, I replaced `StringCchCopy` with a for loop, which does the same.

src/common/win32/EmuShared.h
PatrickvL: Correct me if I'm wrong, but wasn't `CXBX_LOADER` defined for the DLL, but not in the GUI? This would cause a mismatch in the layout of this struct (which is exchanged via EmuShared).

Perhaps a review is required to determine how much use `CXBX_LOADER` still applies. (I guess at the very least is serves as a marker that indicates related pieces of code, which would need alteration when we drop emulation from the GUI executable.)

src/loader/cxbxr-ldr.cpp
PatrickvL: Thanks. Just as long as it's clear as to why this is done...

src/common/win32/EmuShared.h
RadWolfie: @PatrickvL is correct, they need to be reserved because of GUI also load same layout. Otherwise, you will get size mismatch safeguard.

ergo720: I didn't get any in all the tests I did with it removed. I suppose you are referring to this check here https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/52c88d746253c4df7cec24b9b1a9cb085a763603/src/common/win32/EmuShared.cpp#L121

src/common/win32/EmuShared.h
```diff
@@ -355,6 +357,8 @@ class EmuShared : public Mutex
 		overlay_settings m_imgui_overlay_settings;
 		imgui_audio_windows m_imgui_audio_windows;
 		imgui_video_windows m_imgui_video_windows;
+
+		char m_git_version[GitVersionMaxLength];
```
RadWolfie, line 361: Wrong place to put it. It MUST be right below `m_size`, which is at line 67. This is a requirement for compatiblity support and ability to output proper git version output mismatch.

src/common/win32/EmuShared.h
RadWolfie: That's correct. Still, put one reserved back in else block. Eventually, structure will be misalign over time which will trigger the size difference between cxbx and cxbxr-emu because of that extra field.

projects/cxbxr-ldr/CMakeLists.txt
```diff
@@ -68,3 +70,5 @@ target_link_libraries(cxbxr-ldr
 install(TARGETS ${PROJECT_NAME}
   RUNTIME DESTINATION bin
 )
+
+add_dependencies(cxbxr-ldr misc-batch)
```
RadWolfie, line 74: I missed this part, add new line here please.

RadWolfie (changes requested):

RadWolfie (changes requested):
I made a several review remarks. Though, I also made a review remark branch for recommend changes.

review remark branch: https://github.com/RadWolfie/Cxbx-Reloaded/commits/git-version-check-review-remark

I notice there's a merge in there. Disregard that merge commit since it's base from main repo's master branch when I pull the pull request's branch.

Some changes involve such as re-word popup messages, use direct compare between loader/emu than EmuShared first, and relocate git version inside CxbxVersion for reduce time to re-compile whole solution.

src/common/win32/EmuShared.h
```diff
@@ -66,6 +66,11 @@ class EmuShared : public Mutex
 		// ******************************************************************
 		unsigned int m_size;
 
+		// ******************************************************************
+		// * Git version string of cxbx.exe
```
RadWolfie, line 70:
```suggestion
		// * Git version string of first launched initialized
```
Something like ^.

src/common/win32/EmuShared.h
```diff
@@ -286,6 +291,16 @@ class EmuShared : public Mutex
 		void GetOverlaySettings(overlay_settings *value) { Lock(); *value = m_imgui_overlay_settings; Unlock(); }
 		void SetOverlaySettings(const overlay_settings* value) { Lock(); m_imgui_overlay_settings = *value; Unlock(); }
 
+		// ******************************************************************
+		// * Git version Accessor (only the get method is provided because it should not be changed
```
RadWolfie, line 295:
```suggestion
		// * Git version Accessor (only the get method is provided because it should not be changed)
```
Missing closing bracket.

RadWolfie (commented):
Last two changes. Then everything looks good to me.

RadWolfie (approved):

PatrickvL (approved):
Looks fine, thanks. Merging...

PR #2184, opened 2021-04-06
Fixed function pixel shader

NZJenkins:
Placeholder shader until we are able to directly load the pixel shader program used on the Xbox
A shader is generated for each texture op and argument combination

This does not implement any Xbox extensions, so should behave more or less the same as D3D9
F2 toggles the pixel shader - though fogging for D3D9 fixed function is no longer supported

Includes some other graphics-related fixes

~~Needs testing with AMD~~

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -1966,7 +1966,7 @@ static LRESULT WINAPI EmuMsgProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPar
             }
 			else if (wParam == VK_F2)
 			{
-				g_UseFixedFunctionVertexShader = !g_UseFixedFunctionVertexShader;
+				g_UseFixedFunctionPixelShader = !g_UseFixedFunctionPixelShader;
 			}
             else if (wParam == VK_F3)
             {
@@ -6414,11 +6414,6 @@ void UpdateFixedFunctionShaderLight(int d3dLightIndex, Light* pShaderLight, D3DX
 	pShaderLight->SpotIntensityDivisor = cos(d3dLight->Theta / 2) - cos(d3dLight->Phi / 2);
 }
 
-float AsFloat(uint32_t value) {
-	auto v = value;
-	return *(float*)&v;
-}
-
 void UpdateFixedFunctionVertexShaderState()
 {
 	extern xbox::X_VERTEXATTRIBUTEFORMAT* GetXboxVertexAttributeFormat(); // TMP glue
@@ -6506,8 +6501,7 @@ void UpdateFixedFunctionVertexShaderState()
 	// FIXME remove when fixed function PS is implemented
 	// Note if we are using the fixed function pixel shader
 	// We only want to produce the fog depth value in the VS, not the fog factor
-	auto psIsFixedFunction = g_pXbox_PixelShader == nullptr;
-	ffShaderState.Fog.TableMode = psIsFixedFunction ? D3DFOG_NONE : fogTableMode;
+	ffShaderState.Fog.TableMode = !g_UseFixedFunctionPixelShader ? D3DFOG_NONE : fogTableMode;
 
 	// Determine how fog depth is calculated
 	if (fogEnable && fogTableMode != D3DFOG_NONE) {
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -6456,7 +6456,11 @@ void UpdateFixedFunctionVertexShaderState()
 	}
 
 	// Lighting
-	ffShaderState.Modes.Lighting = (float)XboxRenderStates.GetXboxRenderState(X_D3DRS_LIGHTING);
+	// Point sprites aren't lit - 'each point is always rendered with constant colors.'
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/point-sprites
+	bool PointSpriteEnable = XboxRenderStates.GetXboxRenderState(X_D3DRS_POINTSPRITEENABLE);
+	bool LightingEnable = XboxRenderStates.GetXboxRenderState(X_D3DRS_LIGHTING);
+	ffShaderState.Modes.Lighting = LightingEnable && !PointSpriteEnable;
 	ffShaderState.Modes.TwoSidedLighting = (float)XboxRenderStates.GetXboxRenderState(X_D3DRS_TWOSIDEDLIGHTING);
 	ffShaderState.Modes.LocalViewer = (float)XboxRenderStates.GetXboxRenderState(X_D3DRS_LOCALVIEWER);
 
@@ -6472,7 +6476,6 @@ void UpdateFixedFunctionVertexShaderState()
 	ffShaderState.Modes.BackEmissiveMaterialSource = (float)(ColorVertex ? XboxRenderStates.GetXboxRenderState(X_D3DRS_BACKEMISSIVEMATERIALSOURCE) : D3DMCS_MATERIAL);
 
 	// Point sprites; Fetch required variables
-	bool PointSpriteEnable = XboxRenderStates.GetXboxRenderState(X_D3DRS_POINTSPRITEENABLE);
 	float pointSize = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_POINTSIZE);
 	float pointSize_Min = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_POINTSIZE_MIN);
 	float pointSize_Max = XboxRenderStates.GetXboxRenderStateAsFloat(X_D3DRS_POINTSIZE_MAX);
--- a/src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
@@ -174,16 +174,16 @@ TextureArgs ExecuteTextureStage(
 	}
 
 	// Sample the texture
-	float4 t = float4(1, 1, 1, 1);
-	if (stage.IsTextureSet) {
-		int type = TextureSampleType[i];
-		if (type == SAMPLE_2D)
-			t = tex2D(samplers[i], TexCoords[i].xy + offset.xy);
-		else if (type == SAMPLE_3D)
-			t = tex3D(samplers[i], TexCoords[i].xyz + offset.xyz);
-		else if (type == SAMPLE_CUBE)
-			t = texCUBE(samplers[i], TexCoords[i].xyz + offset.xyz);
-	}
+	float4 t;
+	int type = TextureSampleType[i];
+	if (type == SAMPLE_NONE)
+		t = 1; // Test case JSRF
+	else if (type == SAMPLE_2D)
+		t = tex2D(samplers[i], TexCoords[i].xy + offset.xy);
+	else if (type == SAMPLE_3D)
+		t = tex3D(samplers[i], TexCoords[i].xyz + offset.xyz);
+	else if (type == SAMPLE_CUBE)
+		t = texCUBE(samplers[i], TexCoords[i].xyz + offset.xyz);
 
 	// Assign the final value for TEXTURE
 	ctx.TEXTURE = t * factor;
--- a/src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
@@ -1,0 +1,288 @@
+#include "FixedFunctionPixelShader.hlsli"
+
+uniform FixedFunctionPixelShaderState state : register(c0);
+sampler samplers[4] : register(s0);
+
+struct PS_INPUT // Declared identical to vertex shader output (see VS_OUTPUT)
+{
+	float2 iPos : VPOS;   // Screen space x,y pixel location
+	float4 iD0  : COLOR0; // Front-facing primary (diffuse) vertex color (clamped to 0..1)
+	float4 iD1  : COLOR1; // Front-facing secondary (specular) vertex color (clamped to 0..1)
+	float  iFog : FOG;
+	float  iPts : PSIZE;
+	float4 iB0  : TEXCOORD4; // Back-facing primary (diffuse) vertex color (clamped to 0..1)
+	float4 iB1  : TEXCOORD5; // Back-facing secondary (specular) vertex color (clamped to 0..1)
+	float4 iT[4]  : TEXCOORD0; // Texture Coord 0
+	float  iFF : VFACE; // Front facing if > 0
+};
+
+// These 'D3DTA' texture argument values
+// may be used during each texture stage
+struct TextureArgs {
+	float4 CURRENT;
+	float4 TEXTURE;
+	float4 DIFFUSE;
+	float4 SPECULAR;
+	float4 TEMP;
+	float4 TFACTOR;
+};
+
+static float4 TexCoords[4];
+
+// When creating an instance of the fixed function shader
+// we string-replace the assignment below with a value
+// The define keeps the shader compilable without the replacement
+#define TEXTURE_SAMPLE_TYPE {SAMPLE_2D, SAMPLE_2D, SAMPLE_2D, SAMPLE_2D};
+static int TextureSampleType[4] = TEXTURE_SAMPLE_TYPE;
+
+bool HasFlag(float value, float flag) {
+	// http://theinstructionlimit.com/encoding-boolean-flags-into-a-float-in-hlsl
+	return fmod(value, flag) >= flag / 2;
+}
+
+float4 GetArg(float arg, TextureArgs ctx) {
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dta
+	bool alphaReplicate = HasFlag(arg, X_D3DTA_ALPHAREPLICATE);
+	bool complement = HasFlag(arg, X_D3DTA_COMPLEMENT);
+	arg = arg % 16;
+
+	float4 o;
+
+	if (arg == X_D3DTA_DIFFUSE)
+		o = ctx.DIFFUSE;
+	if (arg == X_D3DTA_CURRENT)
+		o = ctx.CURRENT;
+	if (arg == X_D3DTA_TEXTURE)
+		o = ctx.TEXTURE;
+	if (arg == X_D3DTA_TFACTOR)
+		o = ctx.TFACTOR;
+	if (arg == X_D3DTA_SPECULAR)
+		o = ctx.SPECULAR;
+	if (arg == X_D3DTA_TEMP)
+		o = ctx.TEMP;
+
+	if (alphaReplicate)
+		return o.aaaa;
+	else if (complement)
+		return 1 - o;
+	else
+		return o;
+}
+
+float4 ExecuteTextureOp(float op, float4 arg1, float4 arg2, float4 arg0, TextureArgs ctx, PsTextureStageState stage) {
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dtextureop
+
+	// Note if we use ifs here instead of else if
+	// D3DCompile may stackoverflow at runtime
+
+	// X_D3DTOP_DISABLE can only be reached by ALPHAOP
+	// It's documented as undefined behaviour
+	// Test case: DoA:Xtreme menu
+	if (op == X_D3DTOP_DISABLE)
+		return ctx.CURRENT;
+	else if (op == X_D3DTOP_SELECTARG1)
+		return arg1;
+	else if (op == X_D3DTOP_SELECTARG2)
+		return arg2;
+	else if (op == X_D3DTOP_MODULATE)
+		return arg1 * arg2;
+	else if (op == X_D3DTOP_MODULATE2X)
+		return 2 * (arg1 * arg2);
+	else if (op == X_D3DTOP_MODULATE4X)
+		return 4 * (arg1 * arg2);
+	else if (op == X_D3DTOP_ADD)
+		return arg1 + arg2;
+	else if (op == X_D3DTOP_ADDSIGNED)
+		return arg1 + arg2 - 0.5;
+	else if (op == X_D3DTOP_ADDSIGNED2X)
+		return 2 * (arg1 + arg2 - 0.5);
+	else if (op == X_D3DTOP_SUBTRACT)
+		return arg1 - arg2;
+	else if (op == X_D3DTOP_ADDSMOOTH)
+		return arg1 + arg2 * (1 - arg1);
+	else if (op == X_D3DTOP_BLENDDIFFUSEALPHA)
+		return arg1 * ctx.DIFFUSE.a + arg2 * (1 - ctx.DIFFUSE.a);
+	else if (op == X_D3DTOP_BLENDCURRENTALPHA)
+		return arg1 * ctx.CURRENT.a + arg2 * (1 - ctx.CURRENT.a);
+	else if (op == X_D3DTOP_BLENDTEXTUREALPHA)
+		return arg1 * ctx.TEXTURE.a + arg2 * (1 - ctx.TEXTURE.a);
+	else if (op == X_D3DTOP_BLENDFACTORALPHA)
+		return arg1 * ctx.TFACTOR.a + arg2 * (1 - ctx.TFACTOR.a);
+	else if (op == X_D3DTOP_BLENDTEXTUREALPHAPM)
+		return arg1 + arg2 * (1 - ctx.TEXTURE.a);
+	else if (op == X_D3DTOP_PREMODULATE)
+		return arg1; // Note this also multiplies the next stage's CURRENT by its texture
+	else if (op == X_D3DTOP_MODULATEALPHA_ADDCOLOR)
+		return float4(arg1.rgb + arg1.a * arg2.rgb, 1);
+	else if (op == X_D3DTOP_MODULATECOLOR_ADDALPHA)
+		return float4(arg1.rgb * arg2.rgb + arg1.a, 1);
+	else if (op == X_D3DTOP_MODULATEINVALPHA_ADDCOLOR)
+		return float4((1 - arg1.a) * arg2.rgb + arg1.rgb, 1);
+	else if (op == X_D3DTOP_MODULATEINVCOLOR_ADDALPHA)
+		return float4((1 - arg1.rgb) * arg2.rgb + arg1.a, 1);
+	else if (op == X_D3DTOP_DOTPRODUCT3)
+		return dot(arg1.rgb, arg2.rgb).rrrr;
+	// Note arg0 below is arg1 in D3D docs
+	// since it becomes the first argument for operations supporting 3 arguments...
+	else if (op == X_D3DTOP_MULTIPLYADD)
+		return arg0 + arg1 * arg2;
+	else if (op == X_D3DTOP_LERP)
+		return arg0 * arg1 + (1 - arg0) * arg2;
+	else if (op == X_D3DTOP_BUMPENVMAP)
+		return float4(
+			arg1.x * stage.BUMPENVMAT00 + arg1.y * stage.BUMPENVMAT10,
+			arg1.x * stage.BUMPENVMAT01 + arg1.y * stage.BUMPENVMAT11,
+			1, 1);
+	else if (op == X_D3DTOP_BUMPENVMAPLUMINANCE)
+		return float4(
+			arg1.x * stage.BUMPENVMAT00 + arg1.y * stage.BUMPENVMAT10,
+			arg1.x * stage.BUMPENVMAT01 + arg1.y * stage.BUMPENVMAT11,
+			arg1.z * stage.BUMPENVLSCALE + stage.BUMPENVLOFFSET,
+			1);
+
+	// Something is amiss... we should have returned by now!
+	// Return a bright colour
+	return float4(0, 1, 1, 1);
+}
+
+TextureArgs ExecuteTextureStage(
+	int i,
+	TextureArgs ctx,
+	PsTextureHardcodedState s,
+	int previousOp
+)
+{
+	// Early exit if this stage is disabled (and therefore all further stages are too)
+	if (s.COLOROP == X_D3DTOP_DISABLE)
+		return ctx;
+
+	PsTextureStageState stage = state.stages[i];
+
+	// Determine the texture for this stage
+	float3 offset = float3(0, 0, 0);
+	float4 factor = float4(1, 1, 1, 1);
+
+	// Bumpmap special case
+	if (previousOp == X_D3DTOP_BUMPENVMAP ||
+		previousOp == X_D3DTOP_BUMPENVMAPLUMINANCE) {
+		// Assume U, V, L is in CURRENT
+		// Add U', V', to the texture coordinates
+		// And multiply by L'
+		// https://docs.microsoft.com/en-us/windows/win32/direct3d9/bump-mapping-formulas
+		offset.xy = ctx.CURRENT.xy;
+		factor.rgb = ctx.CURRENT.z;
+	}
+
+	// Sample the texture
+	float4 t = float4(1, 1, 1, 1);
+	if (stage.IsTextureSet) {
+		int type = TextureSampleType[i];
+		if (type == SAMPLE_2D)
+			t = tex2D(samplers[i], TexCoords[i].xy + offset.xy);
+		else if (type == SAMPLE_3D)
+			t = tex3D(samplers[i], TexCoords[i].xyz + offset.xyz);
+		else if (type == SAMPLE_CUBE)
+			t = texCUBE(samplers[i], TexCoords[i].xyz + offset.xyz);
+	}
+
+	// Assign the final value for TEXTURE
+	ctx.TEXTURE = t * factor;
+
+	// Premodulate special case
+	if (previousOp == X_D3DTOP_PREMODULATE) {
+		ctx.CURRENT *= ctx.TEXTURE;
+	}
+
+	// Get arguments for the texture operation
+	// Almost all operate on 2 arguments, Arg1 and Arg2
+	// Arg0 is a third argument that seems to have been tacked on
+	// for MULTIPLYADD and LERP
+
+	// Colour operation arguments
+	float4 cArg1 = GetArg(s.COLORARG1, ctx);
+	float4 cArg2 = GetArg(s.COLORARG2, ctx);
+	float4 cArg0 = GetArg(s.COLORARG0, ctx);
+
+	// Alpha operation arguments
+	float4 aArg1 = GetArg(s.ALPHAARG1, ctx);
+	float4 aArg2 = GetArg(s.ALPHAARG2, ctx);
+	float4 aArg0 = GetArg(s.ALPHAARG0, ctx);
+
+	// Execute texture operation
+	float4 value;
+	value.rgb = ExecuteTextureOp(s.COLOROP, cArg1, cArg2, cArg0, ctx, stage).rgb;
+	value.a = ExecuteTextureOp(s.ALPHAOP, aArg1, aArg2, aArg0, ctx, stage).a;
+
+	// Save the result
+	// Note RESULTARG should either be CURRENT or TEMP
+	// But some titles seem to set it to DIFFUSE
+	// Use CURRENT for anything other than TEMP
+	// Test case: DoA 3
+	if (s.RESULTARG == X_D3DTA_TEMP)
+		ctx.TEMP = value;
+	else
+		ctx.CURRENT = value;
+
+	return ctx;
+}
+
+float4 main(const PS_INPUT input) : COLOR {
+
+	TexCoords = input.iT;
+
+	// Each stage is passed and returns
+	// a set of texture arguments
+	// And will usually update the CURRENT value
+	TextureArgs ctx;
+
+	// The CURRENT register
+	// Default to the diffuse value
+	// TODO determine whether to use the front or back colours
+	// and set them here
+	ctx.CURRENT = input.iD0;
+	ctx.DIFFUSE = input.iD0;
+	ctx.SPECULAR = input.iD1;
+	// The TEMP register
+	// Default to 0
+	ctx.TEMP = float4(0, 0, 0, 0);
+	ctx.TFACTOR = state.TextureFactor;
+
+	PsTextureHardcodedState stages[4];
+	stages[0].COLOROP = X_D3DTOP_DISABLE;
+	stages[1].COLOROP = X_D3DTOP_DISABLE;
+	stages[2].COLOROP = X_D3DTOP_DISABLE;
+	stages[3].COLOROP = X_D3DTOP_DISABLE;
+
+	// Define stages
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dtexturestagestatetype
+	// We'll find comment below and insert the definitions after it
+	// STAGE DEFINITIONS
+	// END STAGE DEFINITIONS
+
+	// Run each stage
+	int previousOp = -1;
+	for (int i = 0; i < 4; i++) {
+
+		ctx = ExecuteTextureStage(
+			i,
+			ctx,
+			stages[i],
+			previousOp
+		);
+
+		previousOp = stages[i].COLOROP;
+	}
+
+	// Add fog if enabled
+	if (state.FogEnable) {
+		ctx.CURRENT.rgb = lerp(state.FogColor.rgb, ctx.CURRENT.rgb, saturate(input.iFog));
+	}
+
+	// Add specular if enabled
+	if (state.SpecularEnable) {
+		ctx.CURRENT.rgb += ctx.SPECULAR.rgb;
+	}
+
+	// Output whatever is in current at the end
+	return ctx.CURRENT;
+}
--- a/src/core/hle/D3D8/Direct3D9/TextureStates.cpp
+++ b/src/core/hle/D3D8/Direct3D9/TextureStates.cpp
@@ -124,7 +124,6 @@ void XboxTextureStateConverter::BuildTextureStateMappingTable()
 
 DWORD XboxTextureStateConverter::GetHostTextureOpValue(DWORD Value)
 {
-    bool bOldOrder = g_LibVersion_D3D8 <= 3948; // Verified old order in 3944, new order in 4039
     switch (Value) {
         case xbox::X_D3DTOP_DISABLE: return D3DTOP_DISABLE;
         case xbox::X_D3DTOP_SELECTARG1: return D3DTOP_SELECTARG1;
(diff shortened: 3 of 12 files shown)


src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	float4 DIFFUSE;
+	float4 SPECULAR;
+	float4 TEMP;
+	float4 TFACTOR;
+};
+
+static float4 TexCoords[4];
+
+#define IS_VOLUME_TEXTURE_FLAGS {false, false, false, false}; // define keeps the shader compilable
+static bool IsVolumeTexture[4] = IS_VOLUME_TEXTURE_FLAGS; // This should be replaced with a hardcoded value
```
PatrickvL, line 36: Wow, what happens here!?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+
+float4 GetArg(float arg, TextureArgs ctx) {
+	bool alphaReplicate = fmod(arg, X_D3DTA_ALPHAREPLICATE) >= X_D3DTA_ALPHAREPLICATE / 2;
```
PatrickvL, line 39: This `fmod`-based bit test should be moved towards a tooling macro, if used more often

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	// Each stage is passed and returns
+	// a set of texture arguments
+	// And will usually update the CURRENT value
+	TextureArgs ctx;
+
+	// The CURRENT register
+	// Default to the diffuse value
+	ctx.CURRENT = input.iD0;
+	ctx.DIFFUSE = input.iD0;
+	ctx.SPECULAR = input.iD1;
```
PatrickvL, line 220: If you insert front-face detection here, and use `iB0` / `iB1` as alternative source values, then the above `// TODO handle frontface/backface` would be solved

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	if (arg == X_D3DTA_DIFFUSE)
+		o = ctx.DIFFUSE; // TODO handle frontface/backface
+	if (arg == X_D3DTA_CURRENT)
+		o = ctx.CURRENT;
+	if (arg == X_D3DTA_TEXTURE)
+		o = ctx.TEXTURE;
+	if (arg == X_D3DTA_TFACTOR)
+		o = ctx.TFACTOR;
+	if (arg == X_D3DTA_SPECULAR)
+		o = ctx.SPECULAR;
```
PatrickvL, line 54: This would need a frontface/backface handling TODO too, right?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+
+float4 GetArg(float arg, TextureArgs ctx) {
+	bool alphaReplicate = fmod(arg, X_D3DTA_ALPHAREPLICATE) >= X_D3DTA_ALPHAREPLICATE / 2;
+	bool complement = fmod(arg, X_D3DTA_COMPLEMENT) >= X_D3DTA_COMPLEMENT / 2;
+	arg = arg % 16;
+
+	float4 o;
+
+	if (arg == X_D3DTA_DIFFUSE)
+		o = ctx.DIFFUSE; // TODO handle frontface/backface
```
PatrickvL, line 46: This TODO could be handled when assigning a value to this `DIFFUSE` context structure field, in `main()`

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	else if (op == X_D3DTOP_ADDSIGNED)
+		return arg1 + arg2 - 0.5;
+	else if (op == X_D3DTOP_ADDSIGNED2X)
+		return 2 * (arg1 + arg2 - 0.5);
+	else if (op == X_D3DTOP_SUBTRACT)
+		return arg1 - arg2;
+	else if (op == X_D3DTOP_ADDSMOOTH)
+		return arg1 + arg2 * (1 - arg1);
+	else if (op == X_D3DTOP_BLENDDIFFUSEALPHA)
+		return arg1 * ctx.DIFFUSE.a + arg2 * (1 - ctx.DIFFUSE.a);
```
PatrickvL, line 95: Here, `DIFFUSE` would need to use the backface color when appropriate, right?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	// Add fog if enabled
+	if (state.FogEnable) {
+		ctx.CURRENT.rgb = lerp(state.FogColor.rgb, ctx.CURRENT.rgb, saturate(input.iFog));
+	}
+
+	// Add specular if enabled
+	if (state.SpecularEnable) {
+		ctx.CURRENT.rgb += input.iD1.rgb;
```
PatrickvL, line 257: Better add `ctx.SPECULAR.rgb` here, if you're going to put backface colors in the context diffuse/specular fields.

PatrickvL (commented):
A few questions & suggestions for your consideration.

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	else if (op == X_D3DTOP_MODULATEINVCOLOR_ADDALPHA)
+		return float4((1 - arg1.rgb) * arg2.rgb + arg1.a, 1);
+	else if (op == X_D3DTOP_DOTPRODUCT3)
+		return dot(arg1.rgb, arg2.rgb).rrrr;
+	// Note arg0 below is arg1 in D3D docs
+	// since it becomes the first argument for operations supporting 3 arguments...
+	else if (op == X_D3DTOP_MULTIPLYADD)
+		return arg0 + arg1 * arg2;
+	else if (op == X_D3DTOP_LERP)
+		return arg0 * arg1 + (1 - arg0) * arg2;
```
PatrickvL, line 121: Why not use the actual `lerp()` function here?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	// Premodulate special case
+	if (previousOp == X_D3DTOP_PREMODULATE) {
+		ctx.CURRENT *= ctx.TEXTURE;
+	}
+
+	// Colour operation arguments
+	float4 cArg1 = GetArg(s.COLORARG1, ctx);
+	float4 cArg2 = GetArg(s.COLORARG2, ctx);
+	float4 cArg0 = GetArg(s.COLORARG0, ctx);
+
```
PatrickvL, line 183: Please comment on why ordering 1, 2, 0 is used.

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,264 @@
...
+	// Alpha operation arguments
+	float4 aArg1 = GetArg(s.ALPHAARG1, ctx);
+	float4 aArg2 = GetArg(s.ALPHAARG2, ctx);
+	float4 aArg0 = GetArg(s.ALPHAARG0, ctx);
```
PatrickvL, line 187: Please comment on why ordering 1, 2, 0 is used.

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
```diff
@@ -124,7 +124,6 @@ void XboxTextureStateConverter::BuildTextureStateMappingTable()
 
 DWORD XboxTextureStateConverter::GetHostTextureOpValue(DWORD Value)
 {
-    bool bOldOrder = g_LibVersion_D3D8 <= 3948; // Verified old order in 3944, new order in 4039
```
PatrickvL, line 127: Why was this dropped?

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
```diff
@@ -337,12 +336,37 @@ void XboxTextureStateConverter::Apply()
...
+                return xbox::X_D3DTOP_BLENDTEXTUREALPHAPM;
+            case xbox::X_D3DTOP_BLENDTEXTUREALPHAPM:
+                return xbox::X_D3DTOP_BLENDCURRENTALPHA;
+            }
+        }
+    }
+
+    return value;
+}
+
```
PatrickvL, line 361: Nevermind the previous question on why `bOldOrder` was dropped - it's because you've handled them here. Nice work.

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: `IS_VOLUME_TEXTURE_FLAGS;` gets replaced
Since as far as I could tell we can only use a sampler with Tex2D or Tex3D, not both

PatrickvL: Ah, I see the code searches and replaces `IS_VOLUME_TEXTURE_FLAGS` - better mention that here in a comment!!

PatrickvL (commented):
Some more remarks

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: Yes. Though we don't read the required renderstates from the NV2A to implement it properly yet IIRC so I didn't bother

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: I think we would set DIFFUSE and SPECULAR to front or back in the beginning, and then we can forget about it from then on

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: yep shouldn't need to be using `input` at that point at all!

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: I tried to stick to the D3D9 docs which defines the functions like this

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
NZJenkins: This mapping logic was moved to the `NormalizeValue` function

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: done

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: done

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: added a comment to that end

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2184#discussion_r607745327

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: done

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
PatrickvL: Well, there's this : https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/01b3ce4eec9178e8e704232fd12477a4e557505e/src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl#L328-L335

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,273 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+	EmuCompileShader(finalShader, "ps_3_0", &pShaderBlob, pseudoSourceFile.c_str());
+
+	// Create shader object for the device
+	IDirect3DPixelShader9* pShader = nullptr;
+	auto hRet = g_pD3DDevice->CreatePixelShader((DWORD*)pShaderBlob->GetBufferPointer(), &pShader);
+	if (hRet != S_OK)
+		CxbxKrnlCleanup("Failed to compile fixed function pixel shader");
+	pShaderBlob->Release();
+
+	// Insert the shader into the cahce
```
PatrickvL, line 867: nit : cache (not cahce)

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,273 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+	for (int i = 0; i < 4; i++) {
+		// The stage is initialized to be disabled
+		// We don't have to output anything
+		if (states[i].COLOROP == X_D3DTOP_DISABLE)
+			continue;
+
+		std::string target = "stages[" + std::to_string(i) + "].";
+
+		auto s = states[i];
+		stageSetup << target << "COLOROP = " << D3DTOPToString[(int)s.COLOROP] << ";\n";
```
PatrickvL, line 828: Guard against bounds-overflow?

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,273 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+
+		// TODO handle texture arg flags
+		stageSetup << target << "COLORARG0 = " << d3dTa(s.COLORARG0) << ";\n";
+		stageSetup << target << "COLORARG1 = " << d3dTa(s.COLORARG1) << ";\n";
+		stageSetup << target << "COLORARG2 = " << d3dTa(s.COLORARG2) << ";\n";
+
+		stageSetup << target << "ALPHAOP = " << D3DTOPToString[(int)s.ALPHAOP] << ";\n";
```
PatrickvL, line 835: Guard against bounds-overflow?

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,273 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+			}
+		}
+
+		return str;
+	};
+
+	for (int i = 0; i < 4; i++) {
+		// The stage is initialized to be disabled
+		// We don't have to output anything
+		if (states[i].COLOROP == X_D3DTOP_DISABLE)
```
PatrickvL, line 822: What if ALPHAOP is not disabled?

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,273 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+float AsFloat(uint32_t value) {
+	auto v = value;
+	return *(float*)&v;
+}
+
+// Set constant state for the fixed function pixel shader
+void UpdateFixedFunctionPixelShaderState()
+{
+	using namespace FixedFunctionPixelShader;
+	FixedFunctionPixelShaderState ffPsState;
```
PatrickvL, line 882: nit : empty line between those?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1966,7 +1966,7 @@ static LRESULT WINAPI EmuMsgProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPar
             }
 			else if (wParam == VK_F2)
 			{
-				g_UseFixedFunctionVertexShader = !g_UseFixedFunctionVertexShader;
+				g_UseFixedFunctionPixelShader = !g_UseFixedFunctionPixelShader;
```
PatrickvL, line 1969: So, F2 no longer swaps `g_UseFixedFunctionVertexShader` - aren't users going to notice? Perhaps move that toggle to another hotkey?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: Yep. It doesn't work afaik due to reason above

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: Doesn't matter. if colorop is disabled the stage is

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: Users shouldn't be pressing f2
Imo VS is pretty stable by now - we should purge FF as soon as we can

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: to expand: I am thinking of `TWOSIDEDLIGHTING`, which is not mapped right now in `XbConvert`
Which probably corresponds to `GL_LIGHT_MODEL_TWO_SIDE`?
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glLightModel.xml

But it might be that lighting doesn't have to be enabled to use frontface and backface colours
Don't know any test cases for backface colouring without lighting

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: added some guards and logging too

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
CookiePLMonster: I agree only if those AMD issues are resolved by using a FF PS shader. If not, that hotkey should likely stay.

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,286 @@
...
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dta
+	bool alphaReplicate = HasFlag(arg, X_D3DTA_ALPHAREPLICATE);
+	bool complement = HasFlag(arg, X_D3DTA_COMPLEMENT);
+	arg = arg % 16;
+
+	float4 o;
+
+	if (arg == X_D3DTA_DIFFUSE)
+		o = ctx.DIFFUSE;
+	if (arg == X_D3DTA_CURRENT)
```
CookiePLMonster, line 56: Any reason not to use `else if` in this block?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,286 @@
...
+		// And multiply by L'
+		// https://docs.microsoft.com/en-us/windows/win32/direct3d9/bump-mapping-formulas
+		offset.xy = ctx.CURRENT.xy;
+		factor.rgb = ctx.CURRENT.z;
+	}
+
+	// Sample the texture
+	float4 t = float4(1, 1, 1, 1);
+	if (state.stages[i].IsTextureSet) {
+		t = IsVolumeTexture[i]
```
CookiePLMonster, line 179: What about cubemaps? They need `texCUBE`.

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,286 @@
...
+
+	// Add specular if enabled
+	if (state.SpecularEnable) {
+		ctx.CURRENT.rgb += ctx.SPECULAR.rgb;
+	}
+
+	// Output whatever is in current at the end
+	PS_OUTPUT output;
+	output.oR0 = ctx.CURRENT;
+	return output;
```
CookiePLMonster, line 285: Nitpick, but the function can just return `float4` unless you foresee it'll be returning more in the future at some point. It's unlikely though as iirc xbox doesn't support multiple rendertargets.

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
```diff
@@ -337,12 +336,37 @@ void XboxTextureStateConverter::Apply()
...
 
+// Normalize values which may have different mappings per XDK version
+DWORD NormalizeValue(DWORD xboxState, DWORD value) {
+    if (g_LibVersion_D3D8 <= 3948) {
+        // D3DTOP verified old order in 3948, new order in 4039
+        switch (xboxState) {
+        case xbox::X_D3DTSS_COLOROP:
+        case xbox::X_D3DTSS_ALPHAOP:
+            switch (value) {
+            case xbox::X_D3DTOP_BLENDCURRENTALPHA:
```
CookiePLMonster, line 347: I think this would be easier to understand if switch cases used numbers instead of those constants, as those cases are not supposed to be meaningful stage states (instead being converted to such).

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,297 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+	// TODO does this need to be thread safe?
+	if (!loaded) {
+		loaded = true;
+
+		// Determine the filename and directory for the fixed function shader
+		auto hlslDir = std::filesystem::path(szFilePath_CxbxReloaded_Exe)
+			.parent_path()
+			.append("hlsl");
+
+		auto sourceFile = hlslDir.append("FixedFunctionPixelShader.hlsl").string();
```
CookiePLMonster, line 658:
```suggestion
		auto sourceFile = hlslDir.append("FixedFunctionPixelShader.hlsl").native();
```

This will return a `wstring` and thus not do character conversion, `ifstream` can handle this fine I'm pretty sure.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: I think we should do better than a debug key if for some reason there is still an issue.
It was never intended to be a workaround

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,297 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+		hlsl << hlslStream.rdbuf();
+
+		hlslString = hlsl.str();
+	}
+
+	return hlslString;
+}
+
+std::string_view GetD3DTOPString(int d3dtop) {
+	const std::string opToString[] = {
```
CookiePLMonster, line 672:
```suggestion
	static constexpr std::string_view opToString[] = {
```

Those are constant strings so no need to exercise heap allocations. `string_view` will most likely make them compiled in .rdata as-is.

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,297 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+	}
+
+	return opToString[d3dtop];
+}
+
+// Get a string equivalent of '<Texture Argument> + <Modifier>'
+std::string GetD3DTASumString(int d3dta, bool allowModifier = true) {
+	using namespace FixedFunctionPixelShader;
+
+	const std::string argToString[] = {
```
CookiePLMonster, line 714:
```suggestion
	static constexpr std::string_view argToString[] = {
```

As above.

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,297 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+	// Compile the shader
+	ID3DBlob* pShaderBlob;
+
+	auto hlslDir = std::filesystem::path(szFilePath_CxbxReloaded_Exe)
+		.parent_path()
+		.append("hlsl");
+
+	auto pseudoFileName = "FixedFunctionPixelShader-" + std::to_string(key) + ".hlsl";
+	auto pseudoSourceFile = hlslDir.append(pseudoFileName).string();
+	EmuCompileShader(finalShader, "ps_3_0", &pShaderBlob, pseudoSourceFile.c_str());
```
CookiePLMonster, line 881: Note for the future: Do ensure `EmuCompileShader` understands that the path passed to it is UTF-8. Otherwise make it accept a `wstring` and change the line above to return `.native()`.

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: It's better aligned that's about it. Just pretend it's a switch 😉

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: I don't know. They might just be volume textures? @PatrickvL ?

NZJenkins: yes need to look at handling `X_D3DRTYPE_CUBETEXTURE`

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: you're right we probably won't want to output anything else, but can't seem make it float4 since the compiler wants a semantic i.e. COLOR. will just leave it as is

CookiePLMonster: `float4 main(const PS_INPUT input) : COLOR0`

NZJenkins: oh nice

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: We need `LPCSTR` i.e. 8bit ANSI - that would mean we shouldn't pass wstring?
I am not sure what we are supposed to do if the path is not ANSI - since I think it's required for HLSL includes.
Maybe it should be a relative path since we have control over it i,e. ./hlsl/filename?

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: since we need a string from the return value, I am not sure if there is a benefit, for D3DTA?

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: not sure what character conversion you're referring to

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+	float4 aArg0 = GetArg(s.ALPHAARG0, ctx);
+
+	// Execute texture operation
+	float4 value;
+	value.rgb = Op(s.COLOROP, cArg1, cArg2, cArg0, ctx, stage).rgb;
+	value.a = Op(s.ALPHAOP, aArg1, aArg2, aArg0, ctx, stage).a;
+
+	// Save the result
+	// Note RESULTARG should either be CURRENT or TEMP
+	// But some titles seem to set it to DIFFUSE
```
PatrickvL, line 209: Could you mention a few titles that set `DIFFUSE` instead of `CURRENT` or `TEMP`?

And either add a TODO to handle the `DIFFUSE` case, or actually implement it, or explain in a comment if & why only `TEMP` or `CURRENT` are valid outputs.

This, because `X_D3DTA_` has these suffixes : `DIFFUSE`, `CURRENT`, `TEXTURE`, `TFACTOR`, `SPECULAR`, `TEMP` (besides the flags), so in theory, each of those could be selected as TextureStage output, right?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
PatrickvL: Cubes are not the same as volume textures; A volume texture layers slices on top of eachother, to create a 3D matrix of pixels that can be sampled. A cube is just 6 faces, so only the exterior can be sampled. Calculation of the pixel memory address(es) to sample, is obviously different for these two modes, so yeah, `texCUBE` is required here, if you want to support cube-maps in fixed function ;)

PatrickvL (changes requested):
I know, when people go through your code with a fine comb, it doesn't feel ... great... but it's with the best of intentions that these remarks are made!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6461,7 +6456,11 @@ void UpdateFixedFunctionVertexShaderState()
 	}
 
 	// Lighting
-	ffShaderState.Modes.Lighting = (float)XboxRenderStates.GetXboxRenderState(X_D3DRS_LIGHTING);
+	// Point sprites aren't lit - 'each point is always rendered with constant colors.'
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/point-sprites
+	bool PointSpriteEnable = XboxRenderStates.GetXboxRenderState(X_D3DRS_POINTSPRITEENABLE);
+	bool LightingEnable = (float)XboxRenderStates.GetXboxRenderState(X_D3DRS_LIGHTING);
```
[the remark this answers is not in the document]
NZJenkins: Yep or just implicit cast?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6461,7 +6456,11 @@ void UpdateFixedFunctionVertexShaderState()
 	}
 
 	// Lighting
-	ffShaderState.Modes.Lighting = (float)XboxRenderStates.GetXboxRenderState(X_D3DRS_LIGHTING);
+	// Point sprites aren't lit - 'each point is always rendered with constant colors.'
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/point-sprites
+	bool PointSpriteEnable = XboxRenderStates.GetXboxRenderState(X_D3DRS_POINTSPRITEENABLE);
```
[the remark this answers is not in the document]
NZJenkins: Is it?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+
+	// Note if we use ifs here instead of else if
+	// D3DCompile may stackoverflow at runtime
+
+	// X_D3DTOP_DISABLE can only be reached by ALPHAOP
+	// It's documented as undefined behaviour
+	// Test case: DoA:Xtreme menu
+	if (op == X_D3DTOP_DISABLE)
+		return ctx.CURRENT;
+	else if (op == X_D3DTOP_SELECTARG1)
```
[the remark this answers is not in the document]
NZJenkins: There is a comment
D3DCompile stackoverflows 🤷‍♂️

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+			arg1.x * stage.BUMPENVMAT01 + arg1.y * stage.BUMPENVMAT11,
+			1, 1);
+	else if (op == X_D3DTOP_BUMPENVMAPLUMINANCE)
+		return float4(
+			arg1.x * stage.BUMPENVMAT00 + arg1.y * stage.BUMPENVMAT10,
+			arg1.x * stage.BUMPENVMAT01 + arg1.y * stage.BUMPENVMAT11,
+			arg1.z * stage.BUMPENVLSCALE + stage.BUMPENVLOFFSET,
+			1);
+
+	return float4(0, 1, 1, 1);
```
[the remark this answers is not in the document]
NZJenkins: No reason at all. Will comment

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+
+	if (alphaReplicate)
+		return o.aaaa;
+	else if (complement)
+		return 1 - o;
+	else
+		return o;
+}
+
+float4 Op(float op, float4 arg1, float4 arg2, float4 arg0, TextureArgs ctx, PsTextureStageState stage) {
```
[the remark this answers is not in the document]
NZJenkins: Seems like noise to me

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+struct TextureArgs {
+	float4 CURRENT;
+	float4 TEXTURE;
+	float4 DIFFUSE;
+	float4 SPECULAR;
+	float4 TEMP;
+	float4 TFACTOR;
+};
+
+static float4 TexCoords[4];
```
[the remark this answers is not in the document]
NZJenkins: Nope since it's not needed. Saves passing them around, nothing to see here

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+// The define keeps the shader compilable without the replacement
+#define IS_VOLUME_TEXTURE_FLAGS {false, false, false, false};
+static bool IsVolumeTexture[4] = IS_VOLUME_TEXTURE_FLAGS;
+
+bool HasFlag(float value, float flag) {
+	// http://theinstructionlimit.com/encoding-boolean-flags-into-a-float-in-hlsl
+	return fmod(value, flag) >= flag / 2;
+}
+
+float4 GetArg(float arg, TextureArgs ctx) {
```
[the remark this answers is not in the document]
NZJenkins: Note PS inputs are floats, so having an int arg emits extra instructions to do a conversion.
We can change it to int at no cost, but only since arg will currently be known at compile time

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -638,6 +642,297 @@ constexpr int PSH_XBOX_CONSTANT_FRONTFACE_FACTOR = PSH_XBOX_CONSTANT_LUM + 4; //
...
+	ffPsState.FogColor = (D3DXVECTOR3)((D3DXCOLOR)XboxRenderStates.GetXboxRenderState(xbox::X_D3DRS_FOGCOLOR));
+
+	// Texture state
+	for (int i = 0; i < xbox::X_D3DTS_STAGECOUNT; i++) {
+
+		auto stage = &ffPsState.stages[i];
+
+		stage->COLORKEYOP = XboxTextureStates.Get(i, xbox::X_D3DTSS_COLORKEYOP);
+		stage->COLORSIGN = XboxTextureStates.Get(i, xbox::X_D3DTSS_COLORSIGN);
+		stage->ALPHAKILL = XboxTextureStates.Get(i, xbox::X_D3DTSS_ALPHAKILL);
```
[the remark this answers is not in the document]
NZJenkins: Yes - maybe in the hlsli file so it's closer to the implementation and less likely to become stale

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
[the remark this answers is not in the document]
RadWolfie: > Seems like noise to me

I happen to agree with @PatrickvL, main reason there is `const` existance. If we are not going to modify arguments, then it will perform more optimization than without `const`. On the plus side, it will also prevent us rewrite the arguments unintentionally.

fantesykikachu:
I am testing on an AMD GPU, on master most games crash or have no textures, the last good version was right before the original hlsl vertex shader (non fixed function), I have done a cursory test of all of my games (about 50), and I am happy to report that basically all of them are as good or better than they were back then. The only two regressions that I observed that are also regressions against master with DXVK, are Fuzion Frenzy and Metal Arms: Glitch in the System.

Fuzion Frenzy master:
ff master

Fuzion Frenzy PR:
ff PR

Metal Arms master:
metal arms master

Metal Arms PR:
metal arms PR

RinMaru:
> Metal Arms PR:
> metal arms PR

Metal arms doesnt render anything for me

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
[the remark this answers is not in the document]
NZJenkins: Don't care if it exists. It probably has a place but no reason to blanket add it.

My understanding is it makes no difference to shader asm too. Happy to be shown otherwise

With this I just see future me trying to debug/tweak a shader, and having to remove const to assign an argument...

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: DoA is listed below as a test case
The RESULTARG texture state is documented to be CURRENT or TEMP as noted in the comment

src/core/hle/D3D8/XbPixelShader.cpp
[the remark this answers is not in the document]
NZJenkins: done. There's a comment on the field so it shows up in intellisense wherever used :)

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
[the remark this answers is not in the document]
NZJenkins: done

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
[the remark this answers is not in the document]
NZJenkins: done

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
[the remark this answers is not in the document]
NZJenkins: just removed the cast altogether

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: I've added a TODO since apparently we need an ANSI string, but we don't currently control the path.
If we can coax visual studio to use the .exe location as the working directory, we should be able to use a relative path

src/core/hle/D3D8/Direct3D9/TextureStates.cpp
NZJenkins: done

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: done

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: using `static const std::string` since `string_view` did not seem to be beneficial, here

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+		// Add U', V', to the texture coordinates
+		// And multiply by L'
+		// https://docs.microsoft.com/en-us/windows/win32/direct3d9/bump-mapping-formulas
+		offset.xy = ctx.CURRENT.xy;
+		factor.rgb = ctx.CURRENT.z;
+	}
+
+	// Sample the texture
+	float4 t = float4(1, 1, 1, 1);
+	if (state.stages[i].IsTextureSet) {
```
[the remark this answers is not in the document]
NZJenkins: done

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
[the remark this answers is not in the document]
NZJenkins: done

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+		return float4(
+			arg1.x * stage.BUMPENVMAT00 + arg1.y * stage.BUMPENVMAT10,
+			arg1.x * stage.BUMPENVMAT01 + arg1.y * stage.BUMPENVMAT11,
+			arg1.z * stage.BUMPENVLSCALE + stage.BUMPENVLOFFSET,
+			1);
+
+	return float4(0, 1, 1, 1);
+}
+
+TextureArgs Stage(
```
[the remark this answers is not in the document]
NZJenkins: done

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,288 @@
...
+	float4 t = float4(1, 1, 1, 1);
+	if (stage.IsTextureSet) {
+		int type = TextureSampleType[i];
+		if (type == SAMPLE_2D)
+			t = tex2D(samplers[i], TexCoords[i].xy + offset.xy);
+		else if (type == SAMPLE_3D)
+			t = tex3D(samplers[i], TexCoords[i].xyz + offset.xyz);
+		else if (type == SAMPLE_CUBE)
+			t = texCUBE(samplers[i], TexCoords[i].xyz + offset.xyz);
+	}
```
PatrickvL, line 186: Would it produce more efficient byte code when the final check is omitted?

PatrickvL: Or here's an idea : Remove `IsTextureSet`, and make it a `SAMPLE_NONE` sample type?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -0,0 +1,279 @@
...
+	// Define stages
+	// https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dtexturestagestatetype
+	// We'll find comment below and insert the definitions after it
+	// STAGE DEFINITIONS
+	// END STAGE DEFINITIONS
+
+	// Run each stage
+	int previousOp = -1;
+	for (int i = 0; i < 4; i++) {
+
```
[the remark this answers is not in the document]
NZJenkins: I had added them around the block as it felt a bit cramped

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: The sample type is known at compile time - so this code is compiled away
In fact, the HLSL compiler doesn't allow us to do this branch at runtime (at least in `ps_3_0`)
Each sampler is defined at at the beginning with a particular dimension/type e.g.
```
    dcl_2d s0
    dcl_2d s1
```

NZJenkins: Good idea - will save a few instructions too

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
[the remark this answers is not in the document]
PatrickvL: Sure, fine.

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -788,11 +788,12 @@ IDirect3DPixelShader9* GetFixedFunctionShader()
 		// TODO move XD3D8 resource query functions out of Direct3D9.cpp so we can use them here
 		if (g_pXbox_SetTexture[i]) {
 			auto format = g_pXbox_SetTexture[i]->Format;
-			// SampleType is initialized to SAMPLE_2D
```
PatrickvL: Instead of removing this comment, it could've been updated to `SAMPLE_NONE`

PatrickvL (approved):
Looks good now, great work!

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
PatrickvL: Thanks for applying this - it makes it even cleaner. The less fuss, the better!

PR #2185, opened 2021-04-07
ImGui: couple fixes for different cause of crashes

RadWolfie:
Test cases of crash occur when titles doesn't create Direct3D9 device window through patched function:
- NASCAR Heat 2002 on boot.

The other issue is early termination by intentional testing purpose due to g_renderbase is not created.

### DIFF
--- a/src/core/common/video/RenderBase.hpp
+++ b/src/core/common/video/RenderBase.hpp
@@ -31,7 +31,9 @@ public:
 	}
 
 	void DeviceRelease() {
-		m_device_release();
+		if (m_device_release) {
+			m_device_release();
+		}
 	}
 
 	void SetWindowRelease(const std::function<void()>& func_register) {
@@ -39,7 +41,9 @@ public:
 	}
 
 	void WindowRelease() {
-		m_window_release();
+		if (m_window_release) {
+			m_window_release();
+		}
 	}
 
 protected:
--- a/src/core/kernel/init/CxbxKrnl.cpp
+++ b/src/core/kernel/init/CxbxKrnl.cpp
@@ -1783,9 +1783,10 @@ void CxbxKrnlShutDown(bool is_reboot)
 	g_VMManager.Shutdown();
 
 	// Shutdown the render manager
-	g_renderbase->Shutdown();
-	g_renderbase.release();
-	g_renderbase = nullptr;
+	if (g_renderbase != nullptr) {
+		g_renderbase->Shutdown();
+		g_renderbase = nullptr;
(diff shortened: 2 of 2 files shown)


NZJenkins (approved):
When do the crashes occur?

fantesykikachu:
This causes significant regressions for me. so far in my testing at least Gun Metal and Hulk have significant regressions.

RadWolfie:
> This causes significant regressions for me. so far in my testing at least Gun Metal and Hulk have significant regressions.

I don't have Gun Metal, however I do have Hulk. I will investigate Hulk's issue.

fantesykikachu:
So after testing all my games the ones where I noticed regressions are: Aeon Flux, Darkwatch, Dreamfall, Gauntlet Dark Legacy, Gun Metal, Gunvalkrye, Halo - Combat Evolved, Hulk, Jet Set Radio Future, Kingdom Under Fire, Kingdom Under Fire Heroes, N.U.D.E. Natural Ultimate Digital Experiment, and The Incredible Hulk - Ultimate Destruction.

I think it might be best to revert this for now, however it also looks like it is on the right path as several games had graphical improvements (no major ones, but some that to my knowledge have never rendered before).

PatrickvL:
Yet again the same suggestion as I made on another "bug report" : try to delete all temporary partitions and the symbol cache first - there's a significant chance that will solve the issues you're having...

fantesykikachu:
I always clear symbol cache and cache partitions when I switch builds, and before reporting regressions I confirm by manually deleting both appdata directories. I do use an AMD gpu, but I haven't noticed any real difference since the fixed function pixel shader got merged.

ergo720:
@fantesykikachu The issue has to be on your end. I tried Jet Set Radio Future and NUDE with this build (downloaded from the website) and they both work just fine for me.

RadWolfie:
Title tested with no crash from commit 6f81d21 (which is merged commit from this pull request):
- Halo (XK-009 v1.02, demo)
- Halo (MS-004 v1.08, retail)
- Jet Set Radio Future (SE-010 v1.04)
- The Hulk (VU-017 v1.02)
- Aeon Flux (MJ-020 v1.01)

I'll have to agree with @ergo720 and @PatrickvL's comments. The issue is from your end. Four things could be cause of the issue.
1. Symbol cache's content got corrupted.
2. Cache parition wasn't cleared which is a known issue on xbox hardware.
3. Vendor's graphic driver issue.
4. Third-party wrapper to newer graphic renderer.

PR #2188, opened 2021-04-09
Removed patch menu and SaveXbe(As) options from GUI

ergo720:
This removes the `Allow >64MB` option in the patch menu. This, because it's misleading, as it suggests that, when you turn it on, cxbxr will use 128 MiB of ram, which is not true. All it does is toggling the `bLimit64MB` flag in the xbe, and VMManager only checks for that if the xbe type is debug, and ignores it if it's retail or chihiro. In addition, the patch is only done to the xbe loaded in memory, and won't have any effect if the patched xbe is not saved back to disk before launching the emulation.

The PR is a draft because this is just an idea and needs to be discussed with the other devs first. If it's accepted, the draft status will be removed and then it can be merged.

EDIT: this now also removes the patch menu completely and also the SaveXbe(As) options as well.


LukeUsher:
I'm happy with this: it's mostly a leftover from when Cxbx used to be able to do Exe->Xbe conversions for OpenXDK homebrew.

PatrickvL (approved):

RadWolfie:
If we are getting rid of this, why don't we simply remove the **whole** xbe patches? Along with save modified xbe, which are only purpose for above reason, as well.

src/gui/WndMain.cpp
```diff
@@ -1678,15 +1615,9 @@ void WndMain::RefreshMenus()
...
-            // enable/disable save .xbe file
-            EnableMenuItem(file_menu, ID_FILE_SAVEXBEFILE, MF_BYCOMMAND | MF_WhenXbeLoaded);
-
-            // enable/disable save .xbe file as
-            EnableMenuItem(file_menu, ID_FILE_SAVEXBEFILEAS, MF_BYCOMMAND | MF_WhenXbeLoaded);
-
             // recent xbe files menu
             {
-                HMENU rxbe_menu = GetSubMenu(file_menu, 7);
+                HMENU rxbe_menu = GetSubMenu(file_menu, 4);
```
PatrickvL, line 1620: If two submenu's got removed, then why is this index decreased by 3?

PatrickvL (commented):
Small question, otherwise fine

src/gui/WndMain.cpp
ergo720: Because `MFT_SEPARATOR` also counts as a menu entry, as far as Windows is concerned.

PatrickvL (approved):

PR #2193, opened 2021-04-12
kernel: implement title mount path

RadWolfie:
Finally, it is here! Ability to remember first load title (from CdRom0/Mbfs devices).

I made some changes relative to:
- Move Xbox file path to host path into its own function.
- Added ability to bind cli load's host path to xbox environment path for symbolic link purpose.
- Remember title mount path on first cli load.
  - NOTE: Reboot's cli load is from launchdata page's file path to xbe.
  - Also critical for chihiro titles and retail's xbox live service.
- Add xbox's max_path and apply to relative usage. (I may not get them all, can be fix in later pull request.)
- Due to incomplete implemention of IoCreateDevice, there is a hack in place to prevent dashboard reboot continuously.
  - The reason dashboard reboot continuously is due to unable to find CdRom0 when it should passively return successful.

---

I have personally tested with official dashboard (4920), xbox live, and Exhibition Demo Disc (vol 1). They are able to response very well with this pull request as of commit 152d6be.


src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -27,7 +27,7 @@
 // ******************************************************************
 
 #define LOG_PREFIX CXBXR_MODULE::HAL
-
+#pragma optimize("", off)
```
ergo720, line 30: Why do you need to turn optimizations off for this file? This seems bad.

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -1230,7 +1230,7 @@ void LoadXboxKeys(std::string path)
 	// If we didn't already exit the function, keys.bin could not be loaded
 	EmuLog(LOG_LEVEL::WARNING, "Failed to load Keys.bin. Cxbx-Reloaded will be unable to read Save Data from a real Xbox");
 }
-
+#pragma optimize("", off)
```
ergo720, line 1233: 2/3 Why do you need to turn optimizations off for this file? This seems bad.

src/core/kernel/support/EmuFile.cpp
```diff
@@ -39,9 +39,10 @@
 #pragma warning(default:4005)
 #include "core\kernel\init\CxbxKrnl.h"
 #include "Logging.h"
+#include "common/util/strConverter.hpp"
 
 #include <filesystem>
-
+#pragma optimize("", off)
```
ergo720, line 45: 3/3 Why do you need to turn optimizations off for this file? This seems bad.

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -997,15 +997,15 @@ void CxbxKrnlEmulate(unsigned int reserved_systems, blocks_reserved_t blocks_res
 		// Remove extra slashes.
 		std::string slash_search[] = { "\\\\", "//" };
 		std::string slash_str = "/";
-		for (n = 0, i = 0; i < slash_search->size(); i++, n = 0) {
+		for (n = 0, i = 1; i < slash_search->size(); i++, n = 0) {
```
ergo720, line 1000: Just to be sure of it: is the change  0 -> 1 for `i` intentional or accidental?

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -1384,14 +1406,22 @@ __declspec(noreturn) void CxbxKrnlInit
 	// Create default symbolic links :
 	EmuLogInit(LOG_LEVEL::DEBUG, "Creating default symbolic links.");
 	{
-		// TODO: DriveD should always point to the Xbe Path
+		// TODO: DriveD should auto mount base on launchdata page's ; delimiter xbe path.
```
ergo720, line 1409:
```suggestion
		// TODO: DriveD should auto mount based on the launchdata page's ; delimiter in the xbe path.
```

src/core/kernel/support/EmuFile.h
```diff
@@ -91,8 +91,7 @@ extern const std::string DeviceHarddisk0Partition17;
 extern const std::string DeviceHarddisk0Partition18;
 extern const std::string DeviceHarddisk0Partition19;
 extern const std::string DeviceHarddisk0Partition20;
-extern const char CxbxDefaultXbeDriveLetter;
-extern int CxbxDefaultXbeDriveIndex;
+static constexpr char CxbxDelimiterAutoMountDriveLetter = 'D';
```
ergo720, line 94: No need for `static` here. `constexpr` variables are implicitly `const`, and `const` on global/namespace scope implies `static`.

src/core/kernel/init/CxbxKrnl.cpp
PatrickvL: ...and if intentional, then why not simply remove the first character from `slash_search` instead?

src/core/kernel/support/EmuFile.cpp
```diff
@@ -602,7 +616,44 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
+	int DeviceIndex = CxbxDeviceIndexByHostPath(HostDevicePath.c_str());
+	if (DeviceIndex >= 0)
+		return &Devices[DeviceIndex];
+
+	return nullptr;
+}
+
+std::string CxbxConvertXboxToHostPath(const std::string_view XboxDevicePath) {
+	std::string XbePath;
+	// Convert Xbox XBE Path to Host Path
```
PatrickvL, line 630: This can become a header comment instead of inline.

src/core/kernel/support/EmuFile.cpp
```diff
@@ -602,7 +616,44 @@ XboxDevice *CxbxDeviceByDevicePath(const std::string XboxDevicePath)
...
+	{
```
PatrickvL, line 631: This set of curly brackets can be removed now, and then the indentation of the contained code can be decreased by one level.

src/core/kernel/support/EmuFile.cpp
```diff
@@ -688,31 +741,37 @@ NTSTATUS EmuNtSymbolicLinkObject::Init(std::string aSymbolicLinkName, std::strin
...
 					// Handle the case where a sub folder of the partition is mounted (instead of it's root) :
 					std::string ExtraPath = aFullPath.substr(Devices[DeviceIndex].XboxDevicePath.length(), std::string::npos);
 
-					if (!ExtraPath.empty())
+					if (!ExtraPath.empty()) {
 						HostSymbolicLinkPath = HostSymbolicLinkPath + ExtraPath;
+					}
 				}
 
+
```
PatrickvL, line 774: Duplicate newline. One suffices.

src/core/kernel/support/EmuFile.cpp
```diff
@@ -622,7 +673,8 @@ int CxbxRegisterDeviceHostPath(std::string XboxDevicePath, std::string HostDevic
 
 	// If this path is not a raw file partition, create the directory for it
 	if (!IsFile) {
-		succeeded = std::filesystem::exists(HostDevicePath) || std::filesystem::create_directory(HostDevicePath);
+		std::error_code error;
```
PatrickvL, line 676: Can this `error` safely be ignored? Perhaps comment on it...

src/core/kernel/support/EmuFile.cpp
```diff
@@ -446,6 +445,12 @@ NTSTATUS CxbxConvertFilePath(
 				if (RelativePath.compare(0, 7, "serial:") == 0)
 					return STATUS_UNRECOGNIZED_VOLUME;
 
+				// Raw handle access to the CDROM0:
+				/*if (RelativePath.compare(0, 7, "CDROM0:") == 0) {
```
PatrickvL, line 449: At least, when commented-out code is introduced, add a `TODO` comment along with it, explaining what still needs to be done before this can be enabled.

src/core/kernel/support/EmuFile.cpp
```diff
@@ -39,9 +39,10 @@
 #pragma warning(default:4005)
 #include "core\kernel\init\CxbxKrnl.h"
 #include "Logging.h"
+#include "common/util/strConverter.hpp"
```
PatrickvL, line 42: For what symbols is this #include needed? It helps to add a EOL comment mentioning that (so that if that symbol is ever removed from code, this include can be removed along with it)

src/core/kernel/support/EmuFile.h
PatrickvL, line 94: This variable also doesn't seem to be used as a delimiter (which is normally a sub-string that resides between two separate parts of a string). So, better drop the `Delimiter` part from this symbol name, and simply call it `CxbxAutoMountDriveLetter`. IMHO.

PatrickvL (changes requested):
Nice work! Just a few remarks to consider...

src/core/kernel/support/EmuFile.cpp
RadWolfie: Yes, it can be safely ingored since we expect return value to either true or false. Normally when directory creation fail, it will throw C++ exception error. Since we don't check for C++ exception error nor do we have a "fixed" SEH exception issue duration of the emulation. We need error here to tell fileystem not to throw an exception error.

src/core/kernel/support/EmuFile.h
RadWolfie: We only perform auto mount base on `;` delimiter. Using `CxbxAutoMountDriveLetter` is misleading since kernel don't always auto mount D letter. Well, since XDK does it for titles to automatic assign D letter. I guess I could remove delimiter part.

src/core/kernel/support/EmuFile.cpp
PatrickvL: Okay, thanks. Please summarize that into a comment, so no-one will wonder about this any further.

src/core/kernel/init/CxbxKrnl.cpp
RadWolfie: > Just to be sure of it: is the change 0 -> 1 for `i` intentional or accidental?

Accidental, I was thinking about Windows' network shared directory issue at the time. Change the `i` wouldn't fix it anyway.

RadWolfie: It's revert back to original code.

src/core/kernel/support/EmuFile.cpp
RadWolfie: done

PatrickvL (approved):
Okay,  seems fine now.
Let's await a few test reports before merging this

RadWolfie:
> Okay, seems fine now.
> Let's await a few test reports before merging this

What I would like to hear is titles using xbox live has ability to go into xbox live dashboard and back properly. Though, base on my changes should work as intended. One of title I couldn't test is Call of Duty 3.

RadWolfie:
Eh, I need to make an inline function for this since it need to be done 3 times for different paths.

RadWolfie:
Okay, solved the host's symbolic link to full path issue. I still haven't found working titles able to go into xbox live dashboard.

PatrickvL (approved):

PR #2194, opened 2021-04-13
add user aborted check for sound stream

medievil1:
Work around for  access violation after stream is aborted by user
Crash twinsanity


RadWolfie:
`XMP_STATUS_FLUSHED` is defined as `E_ABORT`. Base on what I had tested on hardware, it is still set the status to flushed state. This is a hack workaround and will cause regression for othe titles. So, this pr changes at this state does not meet the expectuation from reverse engineered notes. Certain titles do expect the statuses to be set as flushed.

Beside, @NZJenkins and I found out there's an issue with the packet's vector corruption when title will submit new packet during the callback process. Using std::vector for packets isn't going to help us. We have to replace it to std::list with a pool system and require some reworks.

RadWolfie:
When you say this fix Crash twinsanity title, it only crash if you skip the cutscene? That's what I'm aware about so far.

medievil1:
> 
> 
> When you say this fix Crash twinsanity title, it only crash if you skip the cutscene? That's what I'm aware about so far.

yes when you skip, the stream stops and status changes to operation aborted so when it tries to get the next status there is no stream, so , access violation error

RadWolfie:
> yes when you skip, the stream stops and status changes to operation aborted so when it tries to get the next status there is no stream, so , access violation error

Okay, similar issue with Obscure title. Only difference there's a hack to flush the packets earlier than expected to avoid access violation occur. So, both titles are on the watchlist for in need to replace std::vector to std::list with a pool system.

PR #2198, opened 2021-04-14
HLSL Pixel Shader tweaks;

PatrickvL:
Various tweaks and cleanup on HLSL Pixel Shader:
* Introduced ENABLE_FF_ALPHAKILL (currently disabled because of graphics corruption), that would add alpha-kill support to fixed function hlsl (for this, all textures stages would need to be fully initialize, to avoid "error X4000: variable 'stages' used without having been completely initialized")
* Use macro-based `alphakill` declaration in regular pixel shader template (same approach as in fixed function hlsl)
* Updated comments to differentiate between texture stage and combiner stage
* Rename `st` into `ts` (texture stage) where appropriate
* Fixed a typo in `Brdf` macro (was `s1`, must be `ts` now)
* Updated `Reflect` to use `C0` (uppercase for Xbox, instead of lowercase `c0` from host)
* Initialized `stage` to zero for use in the `C0` macro, now used by `Reflect`
* Fixed type of `CombinerOutputMapping` to actually be `PS_COMBINEROUTPUT_OUTPUTMAPPING`
* Introduced and used varios PS_*_MASK defines
* Added some comments on pixel shader verification and test-case discovery.
* Fix DolphinClassic sample, by not making fogDepth absolute (thanks NZJenkins!)

This PR might fix a few cases of wrong fog and/or bump mapping, but not much else.


src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
```diff
@@ -218,35 +222,38 @@ float m21(const float input)
 // TODO : Generate sampler status?
 sampler samplers[4] : register(s0);
 
-// Generated alphakill contents are based on X_D3DTSS_ALPHAKILL (we avoid using a constant, to allow false's to be optimized away)
-// bool alphakill[4] = {false, false, false, false}; // Generated by PixelShader.cpp::BuildShader()
+// Declare alphakill as a variable (avoiding a constant, to allow false's to be optimized away) :
+#ifndef ALPHAKILL
```
NZJenkins, line 226: question : Isn't alphakill declared further up?

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
```diff
@@ -189,6 +189,10 @@ TextureArgs ExecuteTextureStage(
 	else if (type == SAMPLE_CUBE)
 		t = texCUBE(samplers[i], TexCoords[i].xyz + offset.xyz);
 
+	if (stage.ALPHAKILL)
+		if (t.a == 0)
```
NZJenkins, line 193: suggestion: use && instead of two ifs

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -671,7 +671,7 @@ std::string GetFixedFunctionShaderTemplate() {
 
 std::string_view GetD3DTOPString(int d3dtop) {
 	static constexpr std::string_view opToString[] = {
-		"UNDEFINED", // 0
+		"X_D3DTOP_DISABLE", // 0 (Was UNDEFINED, but that doesn't compile)
```
NZJenkins, line 674: suggestion: find out why disabled stages are being accessed rather than changing this
Emitting ALPHAKILL in the shader rather than passing it as a constant might do the job
We probably should do that for anything unlikely to change and that causes a branch - that might be all the currently unimplemented extension states

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
NZJenkins: It is but inside `#if 0`

PatrickvL: If you mean `ALPHAKILL`, then yes, it appear in a comment block above, which explains what `#define`'s are emitted by code.
I've chosen to use this `#ifndef ALPHAKILL` construction, so that we're one step closer to ready-compilable HLSL, just like what you did with the Fixed Function Pixel Shader HLSL.

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
PatrickvL: The compiler should solve this. In regular (C/C++) code, having separate checks split over separate lines is a nice convention that allows breakpoints on each step along the way. This wouldn't be possible if all expressions are `&&`-ed together in one `if` expression. I merely applied the same convention here in HSL, because I'm used to writing evaluations like this. Also, writing it in separate steps makes it a little better readable IMHO, but I admit, it's marginal. For now I don't see a big enough benefit to change it.

src/core/hle/D3D8/XbPixelShader.cpp
PatrickvL: Eh, what exactly are you proposing? I've a smooth brain.... ;)

src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl
NZJenkins: Ah I guess I'm just not used to it so find it marginally harder to read. The world's in balance ☯️

src/core/hle/D3D8/XbPixelShader.cpp
NZJenkins: good question - not sure my suggestion makes sense.
Was there any title setting TEXTUREOP to 0? Or it was just that it wasn't set higher up?

Anyway the alphakill error is:
```
error X4000: variable 'stages' used without having been completely initialized
```
It seems the compiler thinks we need to evaluate disabled stages, and errors out?
Maybe it tries to `discard` as early as possible and blows up, even if the `discard` check should never be run?
Some (possibly flakey) workarounds:
* Hardcode the number of active stages since it's known ahead of time. Don't even look at disabled stages
* Store if we should ALPHAKILL in a bool and call `discard` after the stage loop if so
* Put the DISABLED check in the loop instead of ExecuteTextureStage
* Assign everything (including dummy values for UNDEFINED state?)

Probably the smallest change is to move the DISABLED check?

Also, since ALPHAKILL is disabled most of the time,
we can avoid generating any related instructions if we:
* Move `ALPHAKILL` from (the now poorly named) `PsTextureStageState` to `PsTextureHardcodedState` in the .hlsli
* Generate `ALPHAKILL` alongside the other hardcoded states e.g. `COLOROP` and friends, so it is baked in to the shader

PatrickvL: I've simply disabled alphakill for now, as it seems to cause more problems than it solves. I put the related bits behind an `#ifdef ENABLE_FF_ALPHAKILL`, for easy reference.

NZJenkins (approved):

PR #2199, opened 2021-04-14
Log popup test cases in a separate file

ergo720:
This allows logging the popup test cases in a separate file `PopupDebug.txt`, which is created in the same directory that the user has specified for `KrnlDebug.txt`. If the file `PopupDebug.txt` cannot be created/opened for whatever reason, then the popups will still be written to the kernel log file instead. If the user has selected "no logging" or "log to console" for kernel logging in the GUI, then the previous behaviour is retained. The main benefit of this is that it will allow to upload the test cases to the Cxbxr compatibility website (when that becomes a supported feature), instead of polluting the Cxbxr Github tracker with popup issues.


RadWolfie:
I don't agree with the suggestive changes in general. Kernel log file should always have every log output into it, including the test cases. It's almost what Luke asked, except not exactly what he's asking for.

What would be better to do is when test case log is trigger, whenever the console, kernel, or is disabled, make them output into its own file in our own data folder. So, we should make a brand new folder such as `TestCaseLogs`, or `Logs`. Then use SymbolCache folder's method to generate a test case file per title launch.

ergo720:
Converting to draft because this seems to cause a corruption of the EEPROM file for some yet undiscovered reason.

PR #2201, opened 2021-04-19
Implement missing texture behaviour

NZJenkins:
TODO

- [x] Fix Red Dead Revolver

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -768,7 +768,12 @@ IDirect3DPixelShader9* GetFixedFunctionShader()
 
 	bool previousStageDisabled = false;
 	for (int i = 0; i < 4; i++) {
-		// Determine the COLOROP
+		// Determine COLOROP
+		// This controls both the texture operation for the colour of the stage
+		// and when to stop processing
+		// Under certain circumstances we force it to be DISABLE
+		auto colorOp = XboxTextureStates.Get(i, xbox::X_D3DTSS_COLOROP);
+
 		// Usually we execute stages up to the first disabled stage
 		// However, if point sprites are enabled, we just execute stage 3
 		bool forceDisable =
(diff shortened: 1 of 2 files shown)


src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -774,6 +774,21 @@ IDirect3DPixelShader9* GetFixedFunctionShader()
...
+		// When a texture stage has D3DTSS_COLORARG1 equal to D3DTA_TEXTURE
+		// and the texture pointer for the stage is NULL, this stage
+		// and all stages after it are not processed.
+		// Test cases: Red Dead Revolver, JSRF
+		// https://docs.microsoft.com/en-us/windows/win32/direct3d9/texture-blending
+		// Don't follow the D3D9 docs if SELECTARG2 is in use (PC D3D9 behaviour, nvidia quirk?)
+		// Test case: Crash Nitro Kart (engine speed UI)
+		if (!g_pXbox_SetTexture[i]
+			&& XboxTextureStates.Get(i, xbox::X_D3DTSS_COLORARG1) == X_D3DTA_TEXTURE
+			&& XboxTextureStates.Get(i, xbox::X_D3DTSS_COLOROP) != xbox::X_D3DTOP_SELECTARG2)
```
PatrickvL, line 787: You could use `colorOp` here, if you read that beforehand (instead of afterwards).

Perhaps `xbox::X_D3DTSS_COLORARG1` is read twice as well and could be moved into a variable? (I haven't checked)

NZJenkins: will do. COLORARG1 is used but further down so not too fussed on rearranging for it

medievil1:
works great, couldn't find a single issue with any other games

LukeUsher:
Basis that this improves a ton of stuff, and isn't going into master, only hlsl_ps, I think we can get this in now!

LukeUsher (approved):

PR #2203, opened 2021-04-20
Add delayimp lib

scribam:
https://docs.microsoft.com/en-us/cpp/build/reference/delayload-delay-load-import?view=msvc-160
> You must use Delayimp.lib when you link your program, or you can implement your own delay-load helper function.

Fix compilation when using CLion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,10 @@
 [Bb]uild/
 [Bb]uild-*/
 
+# CLion
+.idea/
+cmake-build-*/
+
 # Visual Studio Cache
 .vs/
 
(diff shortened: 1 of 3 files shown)


RadWolfie:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/36b83a8e527fa43a1a414260b618cea29b493ac4/.gitignore#L6-L7

While it's not relative to this pull request fix. Yet, it is understandable necessary for CLion support base on the docs reasons insert into gitignore file.
refs:
- [CLion's usage for CMake build folders](https://www.jetbrains.com/help/clion/cmake-cache.html)
- [.idea directory for CLion usage](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839)

RadWolfie (approved):
Thanks for finding hidden bug which seems was handled by either visual studio or msvc internally.

I can confirm only cxbx and cxbxr-emu are two projects has delay load dlls involve.

PR #2204, opened 2021-04-22
Redefine undefined ALPHAOP behaviour for Crash WoC

NZJenkins:
To aid debugging add a LOG_TEST_CASE , and include ALPHAARGs when ALPHAOP is disabled


LukeUsher (approved):

PatrickvL (approved):
Look good to me. Do note however, that this may be related to the BlueToAlpha feature regular pixel shaders can perform. Kindly check if that's related - fixed function _may_ perform that for the opcodes that perform dot products (`xdd`, and `xdm` not `xmd`) and when `D3DTSS_COLOROP` value equals `D3DTOP_DOTPRODUCT3`.

PR #2206, opened 2021-04-23
check for null device

medievil1:
Check to make sure device is null before creating it, avoids losing device due to unneeded recreation
IE
Indiana jones and buffy menus


PatrickvL (approved):
Approved, nice! (Next time please also document test cases in code comments as well, because commit descriptions are read much less often than the actual code.)

PR #2210, opened 2021-04-25
Few fixes toward title mount path and prepped for chihiro work

RadWolfie:
After feedbacks from @LukeUsher and giving me new knowledge how Chihiro work. This pull request resolve couple issues to work with Chihiro and one regression found from Grand Theft Auto 3 title.

Since Chihiro does have CdRom0 device to mount GD-ROM along with Mbfs. Correction has been made to append mount Mbfs instead of choose one or the other. While testing other titles, I found Grand Theft Auto 3 title regressed due to `g_hCurDir` is zero'd. I made a relocation into symbolic link creation to able allow titles to auto mount themselves.

We should **not** be relying on symbolic link objects 100% of the time. I made a complete replacement `FindNtSymbolicLinkObjectByDevice` to `FindEmuDirPathByDevice`. With this change, ability to support Chihiro is possible.

### DIFF
--- a/src/core/kernel/init/CxbxKrnl.cpp
+++ b/src/core/kernel/init/CxbxKrnl.cpp
@@ -1398,7 +1398,7 @@ __declspec(noreturn) void CxbxKrnlInit
 	CxbxResolveHostToFullPath(relative_path, "xbe's directory");
 
 	CxbxBasePathHandle = CreateFile(CxbxBasePath.c_str(), GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
-	int CxbxTitleDeviceDriveIndex = -1;
+	int CxbxCdrom0DeviceIndex = -1;
 	bool isEmuDisk = _strnicmp(relative_path.c_str(), CxbxBasePath.c_str(), CxbxBasePath.size() - 1) == 0;
 	// Check if title mounth path is already set. This may occur from early boot of Chihiro title.
 	char title_mount_path[sizeof(szFilePath_Xbe)];
@@ -1415,9 +1415,10 @@ __declspec(noreturn) void CxbxKrnlInit
 
 	// TODO: Find a place to make permanent placement for DeviceCdrom0 that does not have disc loaded.
 	if (tmp_buffer[0] != '\0') {
-		CxbxTitleDeviceDriveIndex = CxbxRegisterDeviceHostPath(DeviceCdrom0, tmp_buffer);
+		CxbxCdrom0DeviceIndex = CxbxRegisterDeviceHostPath(DeviceCdrom0, tmp_buffer);
+		// Since Chihiro also map Mbfs to the same path as Cdrom0, we'll map it the same way.
 		if (g_bIsChihiro) {
-			CxbxRegisterDeviceHostPath(DriveMbfs, tmp_buffer);
+			(void)CxbxRegisterDeviceHostPath(DriveMbfs, tmp_buffer);
 		}
 	}
 
@@ -1445,7 +1446,7 @@ __declspec(noreturn) void CxbxKrnlInit
 #else
 		// HACK: It is a hack to override XDK's default mount to CdRom0 which may not exist when launch to dashboard directly.
 		// Otherwise, titles may launch to dashboard, more specifically xbox live title, and back.
-		if (CxbxTitleDeviceDriveIndex == -1 || lastFind != std::string::npos) {
+		if (CxbxCdrom0DeviceIndex == -1 || lastFind != std::string::npos) {
 #endif
 			CxbxCreateSymbolicLink(DriveD, relative_path);
 		}
--- a/src/core/kernel/init/CxbxKrnl.cpp
+++ b/src/core/kernel/init/CxbxKrnl.cpp
@@ -1398,25 +1398,26 @@ __declspec(noreturn) void CxbxKrnlInit
 	CxbxResolveHostToFullPath(relative_path, "xbe's directory");
 
 	CxbxBasePathHandle = CreateFile(CxbxBasePath.c_str(), GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
-	// Titles may assume they are running from CdRom0/Mbfs :
-	std::string_view titleDevice = g_bIsChihiro ? DriveMbfs : DeviceCdrom0;
 	int CxbxTitleDeviceDriveIndex = -1;
 	bool isEmuDisk = _strnicmp(relative_path.c_str(), CxbxBasePath.c_str(), CxbxBasePath.size() - 1) == 0;
-	if (BootFlags == BOOT_NONE) {
-		// Remember our first initialize mount path for CdRom0/Mbfs.
+	// Check if title mounth path is already set. This may occur from early boot of Chihiro title.
+	char title_mount_path[sizeof(szFilePath_Xbe)];
+	const char* tmp_buffer = title_mount_path;
+	g_EmuShared->GetTitleMountPath(title_mount_path);
+
+	if (tmp_buffer[0] == '\0' && BootFlags == BOOT_NONE) {
+		// Remember our first initialize mount path for CdRom0 and Mbfs.
 		if (!isEmuDisk) {
 			g_EmuShared->SetTitleMountPath(relative_path.c_str());
-			CxbxTitleDeviceDriveIndex = CxbxRegisterDeviceHostPath(titleDevice, relative_path);
-		}
-		else {
-			g_EmuShared->SetTitleMountPath("");
+			tmp_buffer = relative_path.c_str();
 		}
 	}
-	else {
-		char szBuffer[sizeof(szFilePath_Xbe)];
-		g_EmuShared->GetTitleMountPath(szBuffer);
-		if (szBuffer[0] != '\0') {
-			CxbxTitleDeviceDriveIndex = CxbxRegisterDeviceHostPath(titleDevice, szBuffer);
+
+	// TODO: Find a place to make permanent placement for DeviceCdrom0 that does not have disc loaded.
+	if (tmp_buffer[0] != '\0') {
+		CxbxTitleDeviceDriveIndex = CxbxRegisterDeviceHostPath(DeviceCdrom0, tmp_buffer);
+		if (g_bIsChihiro) {
+			CxbxRegisterDeviceHostPath(DriveMbfs, tmp_buffer);
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -1398,25 +1398,26 @@ __declspec(noreturn) void CxbxKrnlInit
...
-		char szBuffer[sizeof(szFilePath_Xbe)];
-		g_EmuShared->GetTitleMountPath(szBuffer);
-		if (szBuffer[0] != '\0') {
-			CxbxTitleDeviceDriveIndex = CxbxRegisterDeviceHostPath(titleDevice, szBuffer);
+
+	// TODO: Find a place to make permanent placement for DeviceCdrom0 that does not have disc loaded.
+	if (tmp_buffer[0] != '\0') {
+		CxbxTitleDeviceDriveIndex = CxbxRegisterDeviceHostPath(DeviceCdrom0, tmp_buffer);
+		if (g_bIsChihiro) {
+			CxbxRegisterDeviceHostPath(DriveMbfs, tmp_buffer);
```
PatrickvL, line 1420: Please add a concise comment explaining why the return value of this call isn't stored in `CxbxTitleDeviceDriveIndex`

src/core/kernel/support/EmuFile.cpp
```diff
@@ -1142,6 +1163,52 @@ NTSTATUS NTToXboxFileInformation
 	return result;
 }
 
+void CxbxLaunchNewXbe(const std::string& XbePath) {
```
PatrickvL, line 1166: This function has nothing to do with `EmuFile` and should reside in `EmuKrnlHal`.

src/core/kernel/support/EmuFile.cpp
```diff
@@ -454,58 +459,64 @@ NTSTATUS CxbxConvertFilePath(
...
-					// Handle special root directory constants
-					*RootDirectory = NULL;
-
-					if (OriginalPath.size() == 0){
-						RelativePath = "\\BaseNamedObjects";
-					} else {
-						RelativePath = "\\BaseNamedObjects\\" + OriginalPath;
-					}
+			if (NtSymbolicLinkObject != nullptr || !find_path.HostDirPath.empty()) {
+				/// If found, then we can skip misc checks below.
```
PatrickvL, line 470: ////////////

src/core/kernel/support/EmuFile.cpp
```diff
@@ -833,16 +852,18 @@ EmuNtSymbolicLinkObject* FindNtSymbolicLinkObjectByName(std::string SymbolicLink
 }
 
 
-EmuNtSymbolicLinkObject* FindNtSymbolicLinkObjectByDevice(std::string DeviceName)
+void FindEmuDirPathByDevice(std::string DeviceName, EmuDirPath& hybrid_path)
```
PatrickvL, line 855: Better return bool true on success, bool false on failure

src/core/kernel/support/EmuFile.cpp
```diff
@@ -703,8 +715,12 @@ xbox::ntstatus_xt CxbxCreateSymbolicLink(std::string SymbolicLinkName, std::stri
 	SymbolicLinkObject = new EmuNtSymbolicLinkObject();
 	result = SymbolicLinkObject->Init(SymbolicLinkName, FullPath);
 
-	if (result != xbox::status_success)
+	if (result != xbox::status_success) {
 		SymbolicLinkObject->NtClose();
+	}
+	else if (SymbolicLinkObject->DriveLetter == CxbxAutoMountDriveLetter) {
+		g_hCurDir = SymbolicLinkObject->RootDirectoryHandle;
```
PatrickvL, line 722: It feels wrong to set `g_hCurDir` here,  it's not the task of  `CxbxCreateSymbolicLink()`, instead it's the task of the caller(s)

src/core/kernel/support/EmuFile.cpp
```diff
@@ -795,6 +811,9 @@ NTSTATUS EmuNtSymbolicLinkObject::Init(std::string aSymbolicLinkName, std::strin
 	if (DriveLetter >= 'A' && DriveLetter <= 'Z') {
 		NtSymbolicLinkObjects[DriveLetter - 'A'] = NULL;
 		NtDll::NtClose(RootDirectoryHandle);
+		if (DriveLetter == CxbxAutoMountDriveLetter) {
+			g_hCurDir = NULL;
```
PatrickvL, line 815: See above comment

src/core/kernel/support/EmuFile.cpp
```diff
@@ -679,6 +690,7 @@ int CxbxRegisterDeviceHostPath(const std::string_view XboxDevicePath, std::strin
 	}
 
 	if (succeeded) {
+		newDevice.HostRootHandle = CreateFile(newDevice.HostDevicePath.c_str(), GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
```
PatrickvL, line 693: Please add a comment explaining why this file is created.

Also, verify if the same code doesn't appear elsewhere; if it does, consider moving this into a constructor or function

PatrickvL (dismissed):
Some remarks

src/core/kernel/init/CxbxKrnl.cpp
RadWolfie: There's no reason for need of redundant obtain index. We only need one. Though, I should update variable to `CxbxCdRom0DeviceDriveIndex`.

src/core/kernel/support/EmuFile.cpp
RadWolfie: It has nothing to do with `EmuKrnlHal` to be honest. It is a general kernel function which prepare for launch new process with given xbe.

src/core/kernel/support/EmuFile.cpp
RadWolfie: yagni, check for host path is not empty then be done with it.

src/core/kernel/support/EmuFile.cpp
RadWolfie: You may feel it's wrong, however we do **NOT** assign d letter automatically on first boot. Only when launch data page has delimiter. That's when we do assign automatically. However, due to our crappy code without this hack, Grand Theft Auto 3 will not work correctly.

src/core/kernel/support/EmuFile.cpp
RadWolfie: Due to "newDevice" does not create handle attached just like symoblic link object does when it should have. It causes not able to find host's directory handle when using devices directly than symbolic link.

src/core/kernel/support/EmuFile.cpp
PatrickvL: Well, it doesn't belong in `EmuFile` either, so we'd have to find a best-suited location other than these

src/core/kernel/init/CxbxKrnl.cpp
PatrickvL: I meant a code-comment, not a review-response ;)

src/core/kernel/support/EmuFile.cpp
RadWolfie: If you meant file's creation. This call is not for file creation, it is for open existing directory's handle. It is a copied code from:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/1034185a620886e0b1b10173988c78d2f3921ba3/src/core/kernel/init/CxbxKrnl.cpp#L1400

src/core/kernel/support/EmuFile.cpp
PatrickvL: Then perhaps return `EmuDirPath`, instead of passing in it's address via an argument?

src/core/kernel/support/EmuFile.cpp
PatrickvL: I meant a code-comment, not a review-response ;)

src/core/kernel/support/EmuFile.cpp
RadWolfie: Well, it's relative to the file system, `EmuFile` is the file system though. `CxbxKrnl` file is in need of serious rework and break them down into their own functions. Since `CxbxKrnl` is so large, it's going to confuse me if I need to relocate to earlier line or make declared function earlier. For time being, it's going to reside here until new file is created somewhere that is more organized.

src/core/kernel/support/EmuFile.cpp
RadWolfie: Why would you want unoptimized method for this? It's going to get called a lot and use up stacks for 3 packed member variables. Passing the it's address is more reasonable way and optimized without need to re-assign every fields.

src/core/kernel/support/EmuFile.cpp
RadWolfie: I don't think this need a code comment at all tbh. Otherwise, you will need to comment for other two member variables as well. Such as host path and xbox path. Is that really necessary? I don't think so.

In summary, it's a hidden bug left unchecked.

src/core/kernel/init/CxbxKrnl.cpp
RadWolfie: I preferred rename the variable than unnecessary code comment. 😛

src/core/kernel/support/EmuFile.cpp
RadWolfie: I'll add a comment here to state it's a hack *everywhere* in our emulating file system. 😜

src/core/kernel/support/Emu.cpp
```diff
@@ -42,8 +42,6 @@ CRITICAL_SECTION dbgCritical;
 #endif
 
 // Global Variable(s)
-HANDLE           g_hCurDir    = NULL;
-CHAR            *g_strCurDrive= NULL;
```
RadWolfie, line 46: In the process of relocation, I found g_strCurDrive variable isn't used anywhere. I went ahead remove it from the codebase.

GXTX: Should warnings perhaps be upped to catch such things?

RadWolfie: @GXTX I think you're talking about compiler's warning levels input on compile, right? If so, we don't have it defined in our CMake's projects yet. And yes it should. Currently, there are numerous of warnings which the team haven't covered yet. It's something the team need to fix first before we move to stricter warning level.

src/core/kernel/support/EmuFile.cpp
LukeUsher: If anywhere, I would prefer this to be contained within CxbxKrnl.cpp.

Yes I know that file is a huge monolith of speghetti code, bu tit is where the rest of the XBE loading code currently lives, so it would be nice for it all to be grouped together in a single module.

LukeUsher (approved):

PR #2211, opened 2021-04-27
Remove extra projects we don't need at all and add SDL2 binary to be install

RadWolfie:
With this change, solution explorer become tidier, exclude SDL2 library installation, and no longer need to make further defines for certain subprojects! 🎉

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,14 +13,14 @@ endif()
 
 set(SUBHOOK_TESTS OFF)
 set(SUBHOOK_INSTALL OFF)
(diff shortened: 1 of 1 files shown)


RadWolfie:
CI report an issue about:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Yasm (missing: YASM_EXECUTABLE)
```
Investigating...

PatrickvL (approved):

PR #2212, opened 2021-04-28
Handle D3DTTFF_PROJECTED and when the w texture coordinate component is specified

NZJenkins:
Also move Xbox FF extension states to the fixed function shader so they will be easier to debug

Fixes the ProjectedTexture sample and DoA3 reflections


PatrickvL:
Are the moved extension states part of the identifier used during cache-lookup?

PR #2213, opened 2021-04-29
Reworded fatal error message popup

ergo720:
This is to avoid people assuming the crash is caused by the function mentioned in message, which is almost never the case. ~The original message is still printed in the kernel log.~


LukeUsher:
It might still be useful to print the exception code, and it should probably still print the address

In 99% of cases, it will be 0xC0000005/Access Violation, but not always

LukeUsher (approved):

PR #2215, opened 2021-05-06
Change IVB to an array of 4-float slots

NZJenkins:
Will fix Halo logo and menu text rendering (once it's visible!)
- Remove references to fixed function concepts, since slots might contain anything
e.g. don't exclude the 4th component of the slot corresponding to NORMAL in fixed function mode
- Use the IVB table directly for simplicity. Copy all table data to the vertex buffer

### DIFF
--- a/src/core/hle/D3D8/XbVertexBuffer.cpp
+++ b/src/core/hle/D3D8/XbVertexBuffer.cpp
@@ -706,12 +706,15 @@ void CxbxImpl_End()
 	}
 
 	// Compose an Xbox vertex attribute format to pass through all registers
-	UINT uiStride = 0;
-	g_InlineVertexBuffer_AttributeFormat = {};
-	for (int reg = 0; reg < X_VSH_MAX_ATTRIBUTES; reg++) {
-		g_InlineVertexBuffer_AttributeFormat.Slots[reg].Format = X_D3DVSDT_FLOAT4;
-		g_InlineVertexBuffer_AttributeFormat.Slots[reg].Offset = uiStride;
-		uiStride += sizeof(float) * 4;
+	static UINT uiStride = 0;
+	static bool isIvbFormatInitialized = false;
+	if (!isIvbFormatInitialized) {
+		isIvbFormatInitialized = true;
+		for (int reg = 0; reg < X_VSH_MAX_ATTRIBUTES; reg++) {
+			g_InlineVertexBuffer_AttributeFormat.Slots[reg].Format = X_D3DVSDT_FLOAT4;
+			g_InlineVertexBuffer_AttributeFormat.Slots[reg].Offset = uiStride;
+			uiStride += sizeof(float) * 4;
+		}
 	}
 
 	// Arrange for g_InlineVertexBuffer_AttributeFormat to be returned in CxbxGetVertexDeclaration,
--- a/src/core/hle/D3D8/XbVertexBuffer.cpp
+++ b/src/core/hle/D3D8/XbVertexBuffer.cpp
@@ -50,14 +50,11 @@ CxbxVertexBufferConverter VertexBufferConverter = {};
 
 // Inline vertex buffer emulation
 xbox::X_D3DPRIMITIVETYPE      g_InlineVertexBuffer_PrimitiveType = xbox::X_D3DPT_INVALID;
-uint32_t                      g_InlineVertexBuffer_WrittenRegisters = 0; // A bitmask, indicating which registers have been set in g_InlineVertexBuffer_Table
 xbox::X_VERTEXATTRIBUTEFORMAT g_InlineVertexBuffer_AttributeFormat = {};
 bool                          g_InlineVertexBuffer_DeclarationOverride = false;
 std::vector<D3DIVB>           g_InlineVertexBuffer_Table;
 UINT                          g_InlineVertexBuffer_TableLength = 0;
 UINT                          g_InlineVertexBuffer_TableOffset = 0;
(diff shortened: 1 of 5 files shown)


src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -716,134 +691,26 @@ void CxbxSetVertexAttribute(int Register, FLOAT a, FLOAT b, FLOAT c, FLOAT d)
...
-				// Note, that g_InlineVertexBuffer_AttributeFormat is declared with
-				// data types (in g_InlineVertexBuffer_Table ) that are host-compatible,
-				// so we can do a straight copy here, there's no conversion needed :
-				memcpy(pVertexBufferData, source, size);
-				pVertexBufferData += size;
-			}
-		}
+		g_InlineVertexBuffer_AttributeFormat.Slots[reg].Format = X_D3DVSDT_FLOAT4;
+		g_InlineVertexBuffer_AttributeFormat.Slots[reg].Offset = uiStride;
+		uiStride += sizeof(float) * 4;
```
PatrickvL, line 713: This won't ever differ from one draw to the other, so could be initialized once (which would shave off some per-draw overhead).

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -904,135 +771,31 @@ void CxbxImpl_SetVertexData4f(int Register, FLOAT a, FLOAT b, FLOAT c, FLOAT d)
...
-		g_InlineVertexBuffer_Table[o].Blend[0] = a;
-		g_InlineVertexBuffer_Table[o].Blend[1] = b;
-		g_InlineVertexBuffer_Table[o].Blend[2] = c;
-		g_InlineVertexBuffer_Table[o].Blend[3] = d;
-		break;
-	}
+	// Write vertex data to the IVB table
+	if (Register >= X_D3DVSDE_VERTEX && Register <= X_VSH_MAX_ATTRIBUTES) {
+		// X_D3DVSDE_VERTEX maps to slot 0
+		int index = Register == X_D3DVSDE_VERTEX ? 0 : Register;
```
PatrickvL, line 785: Replace `0` (zero) by `X_D3DVSDE_POSITION`, please.

src/core/hle/D3D8/XbVertexBuffer.cpp
```diff
@@ -904,135 +771,31 @@ void CxbxImpl_SetVertexData4f(int Register, FLOAT a, FLOAT b, FLOAT c, FLOAT d)
...
-	case X_D3DVSDE_BLENDWEIGHT:
-	{
-		g_InlineVertexBuffer_Table[o].Blend[0] = a;
-		g_InlineVertexBuffer_Table[o].Blend[1] = b;
-		g_InlineVertexBuffer_Table[o].Blend[2] = c;
-		g_InlineVertexBuffer_Table[o].Blend[3] = d;
-		break;
-	}
+	// Write vertex data to the IVB table
+	if (Register >= X_D3DVSDE_VERTEX && Register <= X_VSH_MAX_ATTRIBUTES) {
```
PatrickvL, line 783: How can this ever work correctly, given `X_D3DVSDE_VERTEX = 0xFFFFFFFF`, while `X_VSH_MAX_ATTRIBUTES` is 16?

PatrickvL (changes requested):
Wow, soo much code that can be removed - beautiful!

Only a few small remarks. Well done!

src/core/hle/D3D8/XbVertexBuffer.cpp
NZJenkins: As an int that's -1 (as long as we're 32bit?)
Yeah this can be made clearer

medievil1:
KOTOR breaks....
image

medievil1:
actually KOTOR is kinda broken on master so it could be a different commit responsible

PatrickvL:
@medievil1 thanks for looking into this. It'd be great if we could figure out the root cause for the issue you've observed and investigated, because this is one of those changes that would be very beneficial to have merged

PatrickvL:
It might be a memory corruption issue, possibly related to multi threading and resizing buffers

dreimer1986:
Emulating MS-004 v1.09 (Halo) The log is bombed with

[0x1B74] WARN : VTXB    Unknown IVB Register : 9
[0x1B74] WARN : VTXB    Unknown IVB Register : 4
[0x1B74] WARN : VTXB    Unknown IVB Register : 9
etcetcetc

src/core/hle/D3D8/XbVertexBuffer.cpp
NZJenkins: declared `X_D3DVSDE_VERTEX` as -1 instead

PatrickvL (approved):
Besides the potential to pre-initialize some stuff, this looks fine. Great work!

PatrickvL:
If you're really worried about vertex buffer bandwidth overhead, we may need to retain `g_InlineVertexBuffer_WrittenRegisters`, and generate and cache different vertex declarations. When the stride stays the same, we can all still use the same single buffer, or we could declare one buffer per attribute (that would have my preference).

PatrickvL:
I think this is ready for merge now, after some more testing.

@medievil1 could you re-test KOTOR and see if it still crashes, or now works okay?

medievil1:
still some issues
image
image
it is only partially rendering

PatrickvL (approved):
LGTM!

Issue #863, opened 2018-01-15
Call `SetWindowText` from just one place


PatrickvL:
Currently, Cxbx-Reloaded calls `SetWindowText` from multiple locations. These will conflict with eachother.
For this issue:
* build a list of places which currently call `SetWindowText`
* determine which state influences the `lpString` argument
* create a set of  strings that can be part of all text send to `SetWindowText`
* concatenate all these together in one central place
* make sure all states that are needed for this, are communicated from their current locations (possible via `EmuShared`)
* remove all but the central `SetWindowText` call
* test, issue a pull request and close this issue ;)

CakeLancelot:
I've found these 3 instances of SetWindowText for the main window:
- CxbxKrnlSuspend()
- CxbxKrnlResume()
- WndMain::UpdateCaption()

There were other calls for SetWindowText but they were either commented out or for a different window/control (ex. controller setup dialog). I'll try to centralize this to just  WndMain::UpdateCaption() and submit a PR.

bryce13950:
Also, is there are specific reason why `UpdateCaption` is private? If I could make that function public, then I could take care of this in 10 minutes without worry.

PR #2217, opened 2021-05-14
Removed unnecessary calls to SetWindowText

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/863. This removes all calls to `SetWindowText` but the one in `WndMain::UpdateCaption()`, which becomes the only one used to change the title bar. The previous locations that were calling it were these:
1.  In `D3DDevice_Swap`. However, it was disabled by a comment and it was meant to print the `g_dwPrimPerFrame`  variable. In my opinion, we should use ImGui instead of the title bar to print such stats.
2.  In `CxbxKrnlSuspend`. This is only called in `Direct3D9.cpp`, when the emu process receives `WM_SIZE` with a `wParam` of `SIZE_MINIMIZED`, which means the cxbxr window has been minimized. However, I debugged the code and this message is never received, even when I minimize the window (I'm not sure whether or not this is a bug).
3. In `CxbxKrnlResume`. This is called in `Direct3D9.cpp`, when the emu process receives `WM_SIZE` with a `wParam` of `SIZE_MAXIMIZED` or `SIZE_RESTORED`, and also from `CxbxKrnlCleanupEx`. The first two cases are triggered when the cxbxr window is first created, and never afterwards (because it's not possible for the user to maximize the cxbxr window). However, the call only happens when `bAutoPaused` is true, which is never the case because it's a global variable initialized to false, and only becomes true when `SIZE_MINIMIZED` is received, and that never happens like I explained above. This, in turns, means that `CxbxKrnlSuspend` is never called, which is the only function that would set `g_bEmuSuspended` to true, and thus the call in `CxbxKrnlCleanupEx` is also useless as well.

Finally, I think that `CxbxKrnlSuspend` and `CxbxKrnlResume` were originally designed to suspend/resume threads in the old cxbx, where it was possible to do so from the GUI, but since we have removed that ability, they are effectively obsolete now and could probably be removed entirely, unless we decide to use them to implement `KeSuspendThread` and `KeResumeThread`.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -5492,16 +5492,7 @@ xbox::dword_xt WINAPI xbox::EMUPATCH(D3DDevice_Swap)
 
 	if (Flags == CXBX_SWAP_PRESENT_FORWARD) // Only do this when forwarded from Present
 	{
-		// Put primitives per frame in the title
-		/*{
-			char szString[64];
-
-			sprintf( szString, "Cxbx: PPF(%d)", g_dwPrimPerFrame );
-
-			SetWindowText( CxbxKrnl_hEmuParent, szString );
-
-			g_dwPrimPerFrame = 0;
-		}*/
+		// TODO: print the primitives per frame with ImGui
 
 		// TODO : Check if this should be done at Swap-not-Present-time too :
 		// not really accurate because you definately dont always present on every vblank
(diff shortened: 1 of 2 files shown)


CookiePLMonster (approved):

CookiePLMonster:
Even **if** the code path appending "paused" can be reached under some circumstances, I don't think we should care. It's a gimmick and removing it clears up the logic a bit.

PatrickvL (approved):
LGTM

PR #2218, opened 2021-05-15
Fix HLSL PS bump environment mapping

PatrickvL:
(most of the research done by @medieval, thanks!)

Test-case : JSRF boost dash effect (before/after screenshot are added by medieval, in a comment below)

Note : even with this fix, the boost dash effect still looks a bit too strong, compared to real hardware : https://youtu.be/SPZvAsnrNfY?t=201 / https://youtu.be/SPZvAsnrNfY?t=745

TODO : Apply the same change to fixed function pixel shader HLSL!

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
@@ -228,36 +228,46 @@ sampler samplers[4] : register(s0);
 #endif
 static bool alphakill[4] = ALPHAKILL;
 
+float4 PostProcessTexel(const int ts, float4 t)
+{
+	if (alphakill[ts])
+		if (t.a == 0)
+			discard;
+
+	return t;
+}
+
 // Actual texture sampling per texture stage (ts), using the sampling vector (s) as input,
 // abstracting away the specifics of accessing above sampler declarations (usefull for future Direct3D 10+ sampler arrays)
 float4 Sample2D(int ts, float3 s)
 {
 	float4 result = tex2D(samplers[ts], s.xy); // Ignores s.z (and whatever it's set to, will be optimized away by the compiler, see [1] below)
-	if (alphakill[ts])
-		if (result.a == 0)
-			discard;
-
-	return result;
+	return PostProcessTexel(ts, result);
 }
 
 float4 Sample3D(int ts, float3 s)
 {
 	float4 result = tex3D(samplers[ts], s.xyz);
-	if (alphakill[ts])
-		if (result.a == 0)
-			discard;
-
-	return result;
+	return PostProcessTexel(ts, result);
 }
 
 float4 Sample6F(int ts, float3 s)
 {
 	float4 result = texCUBE(samplers[ts], s.xyz);
-	if (alphakill[ts])
-		if (result.a == 0)
-			discard;
+	return PostProcessTexel(ts, result);
+}
 
-	return result;
+// Test-case JSRF (boost-dash effect).
+float3 DoBumpEnv(const float4 TexCoord, const float4 BumpEnvMat, const float4 src)
+{
+	// Convert the input bump map (source texture) value range into two's complement signed values (from (0, +1) to (-1, +1), using s_bx2):
+	const float4 BumpMap = s_bx2(src); // Note : medieval discovered s_bias improved JSRF, PatrickvL changed it into s_bx2 thanks to http://www.rastertek.com/dx11tut20.html
+	// TODO : The above should be removed, and replaced by some form of COLORSIGN handling, which may not be possible inside this pixel shader, because filtering-during-sampling would cause artifacts.
+
+	const float u = TexCoord.x + (BumpEnvMat.x * BumpMap.r) + (BumpEnvMat.z * BumpMap.g); // Or : TexCoord.x + dot(BumpEnvMat.xz, BumpMap.rg)
+	const float v = TexCoord.y + (BumpEnvMat.y * BumpMap.r) + (BumpEnvMat.w * BumpMap.g); // Or : TexCoord.y + dot(BumpEnvMat.yw, BumpMap.rg)
+
+	return float3(u, v, 0);
 }
 
 // Map texture registers to their array elements. Having texture registers in an array allows indexed access to them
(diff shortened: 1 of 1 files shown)


medievil1:
JSRF boost dash
Before
before


after 
image

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
```diff
@@ -228,36 +228,46 @@ sampler samplers[4] : register(s0);
...
-		if (result.a == 0)
-			discard;
+	return PostProcessTexel(ts, result);
+}
+
+// Test-case JSRF (boost-dash effect).
+float3 DoBumpEnv(const float4 dudv, const float4 BumpEnvMat, const float4 src)
+{
+	// Convert the input bump map (source texture) value range into two's complement signed values (from (0, +1) to (-1, +1), using s_bx2):
+	const float4 BumpMap = s_bx2(src); // Note : medieval discovered s_bias improved JSRF, PatrickvL changed it into s_bx2 thanks to http://www.rastertek.com/dx11tut20.html
```
NZJenkins, line 264: Can't this already happen via "dotmapping"?
Does JSRF not set dotmapping to -1 to 1?
It might be possible bx2 is applied twice

PatrickvL: Dotmapping doesn't have to precede bump environment mapping, if I remember correctly. Still, let's investigate this some more...

NZJenkins: Dotmapping is just for dot products - I guess it doesn't apply here? I think I got confused.
Anyway...

In D3D, `texbem` docs state that the texture will be interpreted as signed, and `bx2` is not a valid modifier for the input.
So the input should always be signed.

On Xbox there are signed texture formats, (LVU instead of RGB), but they are aliases of the unsigned formats.
Maybe we should apply a bias, but only when COLORSIGN is disabled.
If it is enabled, we're probably using an LVU texture.

For example, JSRF and BumpEarth both use BUMPENVMAP/BUMPENVMAP_LUM
But JSRF uses an RGB texture, and BumpEarth uses an LVU texture with `COLORSIGN` set

XboxDevWiki looks to have some background on these modes:
https://xboxdevwiki.net/NV2A/Pixel_Combiner

> http://www.graphics.stanford.edu/courses/cs448a-01-fall/nvOpenGLspecs.pdf
> OFFSET_TEXTURE_2D_NV - Transforms the signed (ds,dt) components of a previous texture unit by a 2x2 floating-point matrix and then uses the result to offset the stage's texture coordinates for a 2D non-projective texture.

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
```diff
@@ -228,36 +228,46 @@ sampler samplers[4] : register(s0);
...
+
+	const float u = dudv.x + (BumpEnvMat.x * BumpMap.r) + (BumpEnvMat.z * BumpMap.g);
+	const float v = dudv.y + (BumpEnvMat.y * BumpMap.r) + (BumpEnvMat.w * BumpMap.g);
```
NZJenkins, line 267: suggestion: this should be equivalent?
```hlsl
const float u = dudv.x + dot(BumpEnvMat.xz, BumpMap.rg);
const float v = dudv.y + dot(BumpEnvMat.yw, BumpMap.rg);
```

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
PatrickvL: Good catch on `COLORSIGN`, it's surely possible to handle something like that in our sampling HLSL code. I do wonder if and which host texture formats already perform value range mapping (0..1 to -1..1). If host already does that, we don't need to. But if Xbox COLORSIGN marks some channels as unsigned, and host performs value mapping, we may need to invert the operation on our sampling HLSL.

PatrickvL: https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/sample--sm4---asm- explains about data formats : "For example, if the specified t# was bound with a resource with format DXGI_FORMAT_A8B8G8R8_UNORM_SRGB, then the sampling operation will convert sampled texels from gamma 2.0 to 1.0, apply filtering, and the result will written to the destination register as floating point values in the range [0..1]."

NZJenkins: `COLORSIGN` probably corresponds to this issue linked off the wiki related to controlling filtering https://github.com/xqemu/xqemu/issues/75

Components can arbitrarily be signed or unsigned: https://xboxdevwiki.net/NV2A/Surface_Formats
AFAIK we can't do this easily in a shader, because signed values stored in an unsigned texture format won't filter properly

Maybe we add `COLORSIGN` to the texture hash and convert signed components to unsigned

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
NZJenkins:
```hlsl
float2 uv = dudv.xy + mul(BumpMap.rg, float2x2(BumpEnvMat))
```

nitpick: dudv in d3d docs refers to the perturbation data, but here it is used for the original texture coordinates

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
PatrickvL: Crap. You're right. Filtering is out of our control (or we still have to learn how to get our fingers behind that). And indeed, filtering unsigned values would cause artifacts when the input values that are combined and averaged out during filtering, lay across range-boundaries. Darn. I guess you're right, and we may have to produce various versions of a texture, depending the color-sign of each channel... Not looking forward to the mess of code that would result from this....

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
PatrickvL: Thanks, I'll rename `dudv` into something like `tex_coord`.

NZJenkins (approved):

src/core/hle/D3D8/Direct3D9/CxbxPixelShaderTemplate.hlsl
```diff
@@ -228,36 +228,46 @@ sampler samplers[4] : register(s0);
...
-		if (result.a == 0)
-			discard;
+	return PostProcessTexel(ts, result);
+}
+
+// Test-case JSRF (boost-dash effect).
+float3 DoBumpEnv(const float4 TexCoord, const float4 BumpEnvMat, const float4 src)
+{
+	// Convert the input bump map (source texture) value range into two's complement signed values (from (0, +1) to (-1, +1), using s_bx2):
+	const float4 BumpMap = s_bx2(src); // Note : medieval discovered s_bias improved JSRF, PatrickvL changed it into s_bx2 thanks to http://www.rastertek.com/dx11tut20.html
```
PatrickvL, line 264: Commenting here for posterity : reading the various suggestions JayFoxRox wrote up in the above xqemu link, I wonder if we could try it this way :

The Xbox COLORSIGN texture stages state could be emulated by creating two versions of every texture that requires COLORSIGN treatment:
one copy with a host format that marks all components as being unsigned,
and another copy with a host format that marks all components as being signed.

Then in our pixel shader HLSL code, we could sample both textures, and take the components that Xbox COLORSIGN marks as signed from the signed texture, and the components that Xbox COLORSIGN marks as unsigned from the unsigned texture!

This way we only need two copies of the same texture data, no conversion is required, and sampling can still be allowed to perform filtering, because components will be correctly filtered in both 'signednesses', and for all intents and purposes, won't bleed values across components.

(When no signed texture is needed yet, we can postpone creating it until it's needed.)

Issue #2077, opened 2020-12-12
TASK: Implement Xbox DSP Support


RadWolfie:
## Quick summary
Some titles were report as regression except they're not actually a regression. It is hidden bug we were not aware of.

Part of the reason for following titles crashed:
 - Otogi 1 & 2
   - Using `IDirectSound_DownloadEffectsImage` call.
 - Project Gotham Racing 2
   - Using `XAudioDownloadEffectsImage` call.

Is due to relying on return data from given scrambled DSP binary.

## Details
Eventually, we will need to either implement xbox's DSP support from DirectSound plugin or a new audio plugin.

### System Configuration:
N/A

### Additional Information (if any):
N/A

RadWolfie:
The proper way of doing this is to allow internal DirectSound creation to occur. Then perform trampoline from `IDirectSound_DownloadEffectsImage` side only. Leaving `XAudioDownloadEffectsImage` function may can be unpatched. Internal DSP processing are very complicated and involve encrypted DSP image that need to be decrypt each block in order to receive the values.

As for internal DirectSound creation to occur, we also need some partial LLE audio implement to work properly. Plus communicate back to title for DSP's download process completion.

Finally, forward any patched DSP relative functions to trampoline function for research purpose on return values.

---
Doing the above requirement will properly generate audio correctly for any titles that are using either default or custom DSP image.

jackchentwkh:
Otogi 1 and Otogi 2 use the returned ImageDesc info, which is clear text right in the ImageBuffer instead of encrypted data, what the xdk code does is simply allocate a memory and copy the data from ImageBuffer to it then return the address of allocated memory to game code. and even the original xdk code kept the copy of encrypted code segments directly in DSOUND interface internal data because the decryption is on the fly when written to DSP per my reverse engineer.
here is the possible decryption code and mmio access to DSP.

//Code Reversed from Otogi FS-002 Ver 1.02

// MMIO64_14C18h @ DSOND::001300CE
// if arg1!=0, load arg1, else read MMIO_14C18h, decrypt, write to MMIO_14C18H. 
// if arg2!=0, read MMIO_14c18h and write to arg2 with size arg3.
// write access to const memory has been detected, the output may be wrong!
unsigned int __stdcall MMIO_14C18h_READ_WRITE(__int64 Data_to_write, int *Read_back_buffer, unsigned int read_back_size)
{
  unsigned int v3; // eax
  int v4; // ecx
[...]

PR #2219, opened 2021-05-16
Enable ImageDesc for IDirectSound_DownloadEffectsImage, fix Otogi.

jackchentwkh:
Otogi and some titles uses the returned ImageDesc address after IDirectSound_DownloadEffectsImage was called successfully.
In early stage of Cxbxr, IDirectSound_DownloadEffectsImage returned S_FALSE then Otogi won't check the ImageDesc's address.
sometime last year, develop team decide to change IDirectSound_DownloadEffectsImage to return S_OK in order to enable some titles to proceed in game. But with that change, titles like Otogi which tends to check the returned ImageDesc address will encounter a invalid memory access error since we didn't provide a valid ImageDesc and allocate a memory to return to guest code. 

With Reverse engineer the code of Otogi, I found that the ImageDesc structure is already inside the ImageBuffer. All we have to do is to allocate a memory, calculate the correct offset of ImageDesc, and copy the data from ImageBuffer to the allocated memory. Then return the allocated memory address to guest code. done. 

Tested with Otogi, success. Need further test with other titles which uses IDirectSound_DownloadEffectsImage as well.
I think it's a low risk.

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
@@ -536,26 +536,40 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSound_DownloadEffectsImage)
 
 	// This function is relative to DSP for Interactive 3-D Audio Level 2 (I3DL2)
 
-	LOG_NOT_SUPPORTED();
-	if(ppImageDesc)//If the guest code request a pointer to ImageDesc structure to be returned, then we should allocate a memory and create ImageDesc.
+	LOG_INCOMPLETE();
+
+	if (ppImageDesc) //If the guest code request a pointer to ImageDesc structure to be returned, then we should allocate a memory and create ImageDesc.
 	{
 		// Code block below is reversed from Otogi.
 		// ImageBuffer header starts from offset ox800 and ends in offset 0x817. actual DSP code segment starts from offset 0x818.
 		// DWORD at offset 0x804 and offset 0x80C should be the code segment sizes in dwords.
 		// ImageDesc is appended after code segments. There are additional two DWORDs data for each effect appended after ImageDesc which I have no idea what they are for.
-		DWORD N1 = *(DWORD *)((BYTE *)pvImageBuffer + 0x804); //first code segment size in dwords
-		DWORD N2 = *(DWORD *)((BYTE *)pvImageBuffer + 0x80C); //2nd code segment size in dwords
-		BYTE * pImageDesc = ((BYTE *)pvImageBuffer + 0x818 + 4 * (N1 + N2)); //calculate the starting address of ImageDesc inside the imagebuffer
-		DWORD EffectCount = *(DWORD *)pImageDesc; //the first DWORD in ImageDesc is EffectCount.
-		DWORD ImageDescSize = 8 + 32 * EffectCount; //The size of ImageDesc is two Dwords (8 bytes) + 8 DWORS (32 bytes) for each effects.
-		if (*ppImageDesc = malloc(ImageDescSize)) //allocate a new memory to keep the ImageDesc for guest code since the imagebuffer should be freed after the image were downloaded.
-		{
-			memcpy(*ppImageDesc, pImageDesc, ImageDescSize); //copy the ImageDesc from ImageBuffer.
-		}
+        dword_xt N1 = *(PDWORD)((PBYTE)pvImageBuffer + 0x804); //first code segment size in dwords
+        dword_xt N2 = *(PDWORD)((PBYTE)pvImageBuffer + 0x80C); //2nd code segment size in dwords
+        PBYTE pImageDesc = ((PBYTE)pvImageBuffer + 0x818 + 4 * (N1 + N2)); //calculate the starting address of ImageDesc inside the imagebuffer
+        DWORD EffectCount = *(PDWORD)pImageDesc; //the first DWORD in ImageDesc is EffectCount.
+        dword_xt ImageDescSize = 8 + 32 * EffectCount; //The size of ImageDesc is two Dwords (8 bytes) + 8 DWORS (32 bytes) for each effects.
+
+        // NOTE: this buffer should also be freed in IDirectSound_Release when the ref counter drops to zero, but that function is currently not implemented.
+        static PBYTE ImageDescBuffer = zeroptr;
+        if (ImageDescBuffer != zeroptr) {
+            ExFreePool(ImageDescBuffer);
+        }
+
+        ImageDescBuffer = static_cast<PBYTE>(ExAllocatePool(ImageDescSize));
+        if (ImageDescBuffer == zeroptr) {
+            return E_OUTOFMEMORY;
+        }
+
+        // NOTE: this is very wrong. The dsp image is encrypted, and thus simply copying the original encrypted image won't do any good.
+        *ppImageDesc = ImageDescBuffer;
+        std::memcpy(*ppImageDesc, pImageDesc, ImageDescSize);
+
 		// with the code above, we could easily retrieve the address and size of ImageDesc within the image buffer.
 		// then we can allocate a new memory, copy the imageDesc from the image buffer to the newly allocated memory,
 		// then assign the newly allocated memory to the ppImageDesc. that's all.
 	}
+
 	return S_OK;
 }
 
--- a/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSound.cpp
@@ -518,13 +518,13 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSound_SetEffectData)
 // ******************************************************************
 xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSound_DownloadEffectsImage)
 (
-    LPDIRECTSOUND8  pThis,
-    LPCVOID         pvImageBuffer,
-    dword_xt           dwImageSize,
-    PVOID           pImageLoc,      // TODO: Use this param
-    PVOID*          ppImageDesc)    // TODO: Use this param
+	LPDIRECTSOUND8  pThis,
+	LPCVOID         pvImageBuffer,
+	dword_xt           dwImageSize,
+	PVOID           pImageLoc,      // TODO: Use this param
+	PVOID*          ppImageDesc)    // TODO: Use this param
 {
-    DSoundMutexGuardLock;
+	DSoundMutexGuardLock;
 
 	LOG_FUNC_BEGIN
 		LOG_FUNC_ARG(pThis)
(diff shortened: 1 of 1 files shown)


PatrickvL (changes requested):
Some remarks questions and considerations

jackchentwkh:
> This allocates host memory, while it's an Xbox API, so if this implementation mimicks actual Xbox behaviour, this should be an allocation from Xbox memory pool.
>
> Also, this allocation will have to be freed as well, somewhere. How is that taken into account? Until that's implemented, this allocation will leak memory.

@PatrickvL you're right. this memory should be allocated within xbox memory pool in order to be 100% correct. 
my considerations are:
1. will the memory allocated in emulator be accessed by guest code without access violation? answer is yes. on the other hand, I have no idea how to allocate memory from xbox memory pool. any suggestion?

2. Do we need to free the memory? Per my observation, titles uses this feature call DownloadEffectsImage once, not multiple times. In Otogi's code, when it was called, it will check whether *ppImageDesc is NULL or not, if it's not NULL, which means it has previously allocated memory, so it will free the old memory first. but the titles never reach this situation. and here comes two situation.
   A. If the emulation ends and restarts. Since the memory was allocated in the loader instance, the memory should be freed with the vanish of the loader instance, right?
   B. If the title does call DownloadEffectsImage multiple times. Then all we have to do is to check the *ppImageDesc before we use it and free it if it's not NULL. 

will that be good enough?

jackchentwkh:
> If these image mentions are about the Xbox executable header, then we've got constants and types that can be used to indicate members with

sorry, the image is DSP scrap pad image in the image buffer.

jackchentwkh:
> This log messages should be put in an else block under the below if statement.

you mean the "	LOG_NOT_SUPPORTED();" ? because your comment doesn't linked to the specific lines, so I couldn't understand clearly.

PatrickvL:
Ad 1: Look for other Xbox allocations in other patches/kernel API implementations to see how it's done.

Ad 2 A: You're correct: shutdown of the process frees all process memory. But I guess titles might free the memory this function allocates, after they're done using it, so then we don't have to manage it.
Ad 2 B: It depends what the actual Xbox code does : Does it reuse a given pointer? Does it ignore (and overwrite) the given pointer? Does it free a given pointer?

PatrickvL:
It may be good to declare DSP image struct types, and use those here, so that we can names to the members we know about.

PatrickvL:
And indeed, I was referring to the LOG_NOT_SUPPORTED line

jackchentwkh:
> LOG_NOT_SUPPORTED

I think the  LOG_NOT_SUPPORTED is to indicate that the DSP is not supported by our emulator. 
we're only to give the guest code a fake response, there were no DSP effect at all.
so the  LOG_NOT_SUPPORTED shall be kept.

jackchentwkh:
> Ad 1: Look for other Xbox allocations in other patches/kernel API implementations to see how it's done.
> 
> Ad 2 A: You're correct: shutdown of the process frees all process memory. But I guess titles might free the memory this function allocates, after they're done using it, so then we don't have to manage it.
> Ad 2 B: It depends what the actual Xbox code does : Does it reuse a given pointer? Does it ignore (and overwrite) the given pointer? Does it free a given pointer (and replace it by a new allocation)?

In order to make it more clear. I make a quick definition for the guest code in two parts. 
one is the code that the game programmer wrote, I call it the "game code".
one is the code comes with the XDK library, I call it the "XDK code".

here is the actual code behavior.
DownloadEffectsImage() belongs to XDK code. 
game code read the DSP image file into image buffer, then calls DownloadEffectsImage(),
XDK code take over the image buffer, check whether ppImageDesc points to an existed memory or not. If it pointed to an existed memory, then XDK free the memory first. A new memory will be allocated and assign to *ppImageDesc,
game code will never free the *ppImageDesc, 
XDK code will only free the *ppImageDesc memory only when DownloadEffectsImage() was called again with *ppImageDesc is not NULL.

With the behavior above. I would suggest that we could simply follow the XDK code, only free the memory when DownloadEffectsImage() was called again with *ppImageDesc is not NULL.
and in this situation, does it still matter whether the memory is allocated within xbox memory pool or not?

PatrickvL:
Thanks for the summary. I'd assume there's also a function in the XDK that frees this allocated memory. Such function may have been offered in the form of a define, or inlined function. In both cases it will be hard (or even impossible) to locate this function using our symbol scanning. So, under this assumption, we really should use Xbox memory pool functions for the potential freeing, and allocation you do in here.

jackchentwkh:
> ...I'd assume there's also a function in the XDK that frees this allocated memory. Such function may have been offered in the form of a define, or inlined function. In both cases it will be hard (or even impossible) to locate this function using our symbol scanning. So, under this assumption (that the game code will perform Xbox memory freeing at some later point in time), we really should use Xbox memory pool functions for the potential freeing, and allocation you do in here.

frankly, there is not such function existed in XDK. all documentation available regarding DownloadEffectsImage() barely mentioned the structure of ImageDesc only, that's all. and there is not further detail of how to use the infomation inside ImageDesc, not even when or how to free the memory.

the XDK code itself, however, when allocate memory for ImageDesc, it actually allocate extra 4 bytes in front of Imagedesc and put a 4 bytes of key/signature in that address. So when it encountered a call with *ppImageDesc not NULL, it checks whether the 4 bytes prior the ImageDesc matches the key/signature or not. If the key/signature matches, then it will free the old memory and allocate a new one. If the key/signature does not match, then it won't free the old memory but allocate the new memory directly.

ergo720:
@jackchentwkh I pushed a commit to your PR to kind of fix the memory leak of `malloc`. I also added some comments detailing other issues I found in the implementation.

jackchentwkh:
@PatrickvL thanks a lot for your help. One thing to clarify, the ImageDesc in the image buffer is not encrypted, it's  clear text. only the code segments are encrypted. and the decryption code could be reversed from the XDK code part of the title.

PatrickvL (approved):
Indentation changes weren't really necessary, but otherwise this is a good enough work-around for now. Emulating DSP DMA would be the next step, so that we could unpatch some of this, but that's something for another day ;)

Thanks!

RadWolfie:
Hi @jackchentwkh, had you reviewed with https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2077? It's criticial for both titles, mentioned in the ticket, to work properly. I recently provide some details how to properly get DSP working which do require internal DirectSound library initialize. Base on my notes, there is decryption process involve. Which then is stored pointer of it inside internal DirectSound class and output the same pointer to decrypted descriptor. Which is why @ergo720 add a note of copying straight forward is wrong when it is actually encrypted.

Both @LukeUsher and I had took a depth look in DSP relative functions which end up in agreement of in need for internal DirectSound class creation and let title handle it. Like I said, I recently add details how to get DSP working.

jackchentwkh:
@PatrickvL @ergo720   the XAudioDownloadEffectsImage() was patched and simply returned S_OK, but it also needs to return a proper newly allocated memory with ImageDesc to make game code run correctly, currently PGR2 uses XAudioDownloadEffectsImage(). Auctually in XDK code,  XAudioDownloadEffectsImage() calls IDirectSound_DownloadEffectsImage(). so we have some options.
1. unpatch XAudioDownloadEffectsImage(), let it calls IDirectSound_DownloadEffectsImage(). this will need testing since there are extra codes besides calling  IDirectSound_DownloadEffectsImage().
2. Keep patching  XAudioDownloadEffectsImage(), then we can call our patched IDirectSound_DownloadEffectsImage(), or we can isolate the code which allocate the memory and copy the ImageDesc from ImageBuffer as a subroutine and let both patched  XAudioDownloadEffectsImage() and patched  IDirectSound_DownloadEffectsImage() call this subroutine. 
   what's your comment? @RadWolfie I think you're also involved in this topic, please kindly shade a light here.

jackchentwkh:
Ok, I reversed the PRG2 in further detail and finally figured out what's preventing the returned ImageDesc being working correctly. The ImageBuffer must be kept a copy in guest memory. and the state segment address in the EffectMaps must be rebased to the kept copy. in order to keep things aligned, I also rebased the code segment address in the EffectMaps. 

So the new logic is like:
1. Patch CDirectSound_DownloadEffectsImage(), not patch IDirectSound_DownloadEffectsImage(). because  IDirectSound_DownloadEffectsImage() simply calls CDirectSound_DownloadEffectsImage().
2. Patch XAudioDownloadEffectsImage(); read the image file, create a image buffer, call our version of CDirectSound_DownloadEffectsImage()
3. In our version of CDirectSound_DownloadEffectsImage()
    a. allocate memory in guest memory pool for the backup copy of image buffer. copy the input image buffer to the allocated memory. all image buffer related process will be accessed/applied to the internal backup copy.
    b. calculate the ImageDesc address, get EffectCounts, calculate ImageDesc Size. 
    c. calculate EffectMaps starting address. loop through each EffectMap, rebase Code segment address and State segment address and store the rebased address back to the backed up copy of image buffer.
    c. allocate memory in guest memory pool for ImageDesc. copy ImageDesc from backup copy of image buffer to the newly allocated memory. store &ImageDesc to *ppImageDesc.

with this approach, PGR2 can proceed to menu. but then it will encounter a different access violation which I believe is not related to DSOUND or at least not related to this PR. in debug mode, the crash might show up right after the menu was shown. 
in release build, the crash will show up very late only when you finish the setup of profile/choose the game/and once enter the race, it will crash right away. if the debugger was attached vs VS, it will show a access violation introduced by a "VECTOR" class in which "_mycount" is pointing to an invalid address? 

since the new code is a little bit complicated then the original PR. I leave it in a different branch in my repo. 
@PatrickvL @ergo720 @RadWolfie please take a look at my new code
https://github.com/jackchentwkh/Cxbx-Reloaded/tree/ImageDescWithRebaseAddress
or should I push it here and maybe someone can help me polish it out?

ergo720:
@jackchentwkh FYI, I and other devs have already reversed `IDirectSound_DownloadEffectsImage` and we know how it works internally. The function copies the encrypted image to an internal dsound buffer which was first allocated with `MmAllocateContiguousMemoryEx` when `DirectSoundCreate` is first called. This buffer is always available, and it's only ever released when `IDirectSound_Release` is called and the ref counter drops to zero, which destroys the entire dsound instance. I believe this corresponds to your first usage of `ExAllocatePool` in your linked commit above. Then, a portion of the dsp image in decrypted in that same buffer, and the decrypted portion is copied to another buffer that was allocated during the function call with `ExAllocatePool` (this should correspond to your second usage of it).

ergo720:
> since the new code is a little bit complicated then the original PR. I leave it in a different branch in my repo.
@PatrickvL @ergo720 @RadWolfie please take a look at my new code
https://github.com/jackchentwkh/Cxbx-Reloaded/tree/ImageDescWithRebaseAddress
or should I push it here and maybe someone can help me polish it out?

You can either push it here and change the PR title to reflect this new change or submit it in a separate PR, I don't have a preference.

ergo720:
Superseded by https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2221

PR #2220, opened 2021-05-19
Fix Metal Arms menu

NZJenkins:


PatrickvL (approved):
Nice work, inverting to see how Xbox NV2A deviates from Windows Direct3D specs.

PR #2221, opened 2021-05-19
Keep ImageBuffer backup copy, handle ImageDesc with rebase address, Fix Otogi 1, Otogi 2, PGR2 booting.

jackchentwkh:
Previous PR only allocate memory for ImageDesc and copy the content from ImageBuffer directly. 
that only fixes Otogi 1 and Otogi 2, but not PGR2. 
With further reverse engineering of PGR2, turns out that a copy of ImageBuffer must be kept in guest memory. and Code Segments and State Segments addresses of EffectMaps in ImageDesc must be rebased to the guest memory copy of ImageBuffer.

tested with Otogi 1, Otogi 2, and PGR2. Success. PGR2 can reach the race into. there will be new access violation regarding DSOUND buffer/packet processing, but that's not the concern of this PR.

Thanks a lot for @ergo720 for his help of coding assist.


jackchentwkh:
ha, see your comment, a little bit late. I shall polish the code later. thanks again.

jackchentwkh:
> This is bad: this is an xbox file, and, as such, it should use the kernel file apis, instead of its host counterparts (NtCreateFile in this case). Same for the calls below.

the xbox::NtCreateFile() doesn't take file name as an input. I couldn't figure out a feasible way to implement it properly. any suggestion?

jackchentwkh:
finally found some spare time to polish the code. frankly using host file access is way more cleaner. 
in order to use xbox::CreateFile/ReadFile, I have to reimplement internal logic of CreateFile/ReadFile/GetFileSize/CloseFile. and overall the code becomes very messy. anyway finally it works.

ergo720:
@jackchentwkh I pushed a new commit to untabify your changes (the surrounding code is using spaces instead of tabs) and updated the code style of the `if else` code blocks.
> in order to use xbox::CreateFile/ReadFile, I have to reimplement internal logic of CreateFile/ReadFile/GetFileSize/CloseFile. and overall the code becomes very messy.

I agree, we should provide some wrappers around the Nt kernel file functions to avoid all the extra code. Unfortunately, we don't have any at the moment, so we'll have to live with for now.

jackchentwkh:
@ergo720 thanks bro! Let's say, since I already have some code already completed. If I want to create the xbox::CreateFile/ReadFile/GetFileSize/CloseFile wrapper, where should I put them? EmuFile.cpp?

ergo720:
I'd say yes, `EmuFile.cpp` is the best place for such wrappers. However, it's better that you wait until this [PR](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2210) is merged first, since @RadWolfie is also working on it (shouldn't take too long to finish).

ergo720 (dismissed):

ergo720 (approved):

Issue #1798, opened 2019-12-14
Gamepad not re-detected if connectivity is lost


RedPyramids:
If gamepad loses connectivity, Cxbx will not re-detect it when it establishes connection again.
This wouldn't be a problem if you could Alt+Tab to the input configuration but this causes a black screen.

PR #2222, opened 2021-05-19
Add controller hotplug support

ergo720:
Closes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/1798. This adds controller hotplug support by monitoring device additions/removals by using the RawInput API. I'm using this API because XInput is a poll-driven API, and thus I would need to use a busy loop to monitor for device changes, which I'd rather avoid. Instead, by using RawInput, we can receive a `WM_INPUT_DEVICE_CHANGE` message when it detects a change, thus avoiding the need for a busy loop.


LanHikariDS:
Was discussing something like this a while ago, and I think it might be worth having a checkbox in each controller's config to tie the controller status to the current device. So if a connected controller is disconnected, the emulator will set the controller slot that device was bound to to 'None', causing the game to trigger it's 'controller disconnected' prompt.

ergo720:
I instead don't like this idea for four reasons.
1.  The first one is more conceptual: the xinput apis (`XGetDevices`, `XGetDeviceChanges`) and the xbox devices are not coupled to the host devices. and operate regardless of what happens to them. If the host device is disconnected, it simply means that the bound xbox device will no longer receive input, but it's still attached and that's correct imo because you have set it to the attached state in the input gui.
2.  The current connectivity state at runtime is stored in `EmuShared`. To detach an xbox device, I would then need to use `g_EmuShared->SetInputDevTypeSettings` and call `InputDeviceManager::UpdateDevices`. If you close the emulation at this point and relaunch it, the settings in `EmuShared` will be used for the next launch, and suddenly the xbox device will no longer receive input, because the device state was set to None.
3. Once the host device is re-attached, I would need to reset the xbox device back to what it was before the disconnection, but I can't since the device type was overwritten with None in `EmuShared`. Thus, I would need to re-read it from the settings file, which is bad.
4. Finally, not all games have a 'controller disconnected' prompt, and some will simply not care about the controller being disconnected and they will keep going. So, this won't work with all games anyway.

literalmente-game:
Verified with 5 playable games, all worked properly while disconnecting and reconnecting the controller (tested with an wired XBox One controller)

PR #2223, opened 2021-05-23
Fix DSoundBufferResizeSetSize with DSBuffer creation failure when rem…

jackchentwkh:
…ained hos buffer bytes is less then DSBSIZE_MIN, also correct a wrongly used argument with call to DSound3DBufferCreate. add more sanity checks. this fixs PGR2, now release build can enter the race. but debug build has different issue with vector/iterator.


src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
```diff
@@ -534,7 +549,7 @@ static inline void DSoundBufferResizeUpdate(
     DWORD                       Xb_dwByteLength) {
 
     xbox::EmuDirectSoundBuffer* pThis = pHybridThis->emuDSBuffer;
-    DSoundBufferResizeSetSize(pHybridThis, hRet, Xb_dwByteLength);
+    DSoundBufferResizeSetSize(pHybridThis, hRet, Xb_dwByteLength);
```
ergo720, line 552: Nit: unnecessary space added at the end of the line.

jackchentwkh: I put a note there and removed it later, should be left over. anything else besides this extra space? @ergo720 ?

ergo720:
Your changes are using tabs, while the existing code is using spaces. We haven't agreed yet on which one to use, but as a general rule, we try to match the indentation of the surrounding code.

src/core/hle/DSOUND/DirectSound/DirectSoundInline.hpp
ergo720: I'm not familiar with dsound, I only know about the internals of `IDirectSound_DownloadEffectsImage` and little else really. @RadWolfie is our main sound developer, so they will probably have to review this.

jackchentwkh:
that's strange, I did notice the tab issue and keep not using tab this time. seems VS will use tab automatically by itself. I shall check the editor setting.

ergo720:
Please use `git rebase` to grab the latest changes from master, to avoid introducing merge commits in the PR (which we try not to do).

jackchentwkh:
@ergo720 seems I made a mistake pushing wrong commits to this PR, is there anyway to change it back?

ergo720:
`git reset --hard f54fb15ac8d3da40c29e8946d53fd344b86e43e9` to forcefully reset your branch back to the commit before the merge. Then, `git rebase -i upstream/master` (assuming the your upstream points to the main cxbxr repository of the organization), and solve any conflicts that might arise.

PR #2228, opened 2021-05-27
Implement Texture Stage State xbox extensions in HLSL

PatrickvL:
This PR addresses #1604, introducing most of what's needed to support COLORSIGN, COLORKEYCOLOR, COLORKEYOP alongside with a small ALPHAKILL refactoring.

To get there, X_D3DFMT_L6V5U5 had to be converted into host D3DFMT_X8L8V8U8, and some HLSL code had to be added to implement these extensions.

Some pieces to the puzzle may still be missing, but at least this is a step in the right direction.

The most important test is to see if JSRF boost dash didn't regress because of this. I can't really judge this well, so please give us some feedback.


medievil1:
Purple bumpearth broken(no bumpmapping)
image

older bumpearth is fine
image
boost dash broken (needs bx2 same place as before, src(ts) in the bumpenv (DoBump define)
image

medievil1:
no change with latest commit https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2228/commits/efb4144d560f68f883dc5fcfb8691e9de0097a75, all three exhibit same behavior

medievil1:
just to document, HostTextureFormat never changes, it is always unknown so the reading of it is currently at issue as it is used to determine if host texture is already signed. as it is, that test never succeeds so it always thinks the host is NOT signed 
image

PR #2230, opened 2021-05-31
Reduce cpu usage during waits

NZJenkins:
Takes CPU usage running the dashboard from ~34% down to ~5% on my PC

- [ ] Mitigate potential VBlank timing issues by spinwaiting

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2168,50 +2168,51 @@ static DWORD WINAPI EmuUpdateTickCount(LPVOID)
 
     EmuLog(LOG_LEVEL::DEBUG, "Timing thread is running.");
 
-    // current vertical blank count
-    int curvb = 0;
+	// We check for LLE flag as NV2A handles it's own VBLANK if LLE is enabled!
+	if (bLLE_GPU)
+		return 0;
 
-	// Calculate Next VBlank time
-	auto nextVBlankTime = GetNextVBlankTime();
+    // Create VBlank timer
+	LARGE_INTEGER dueTime;
+	dueTime.QuadPart = -16.6666666667 * 10000; // ~60 fps in 100 nanosecond units
+	HANDLE vblankTimer = CreateWaitableTimer(NULL, false, "cxbx-vblank");
+	if (vblankTimer == NULL) {
+		CxbxKrnlCleanup("Failed to create VBLANK timer");
+	}
+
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5483,17 +5484,23 @@ xbox::dword_xt WINAPI xbox::EMUPATCH(D3DDevice_Swap)
...
-                // This prevents large waits from hogging CPU power, but allows small waits/ to remain precice.
-                while (std::chrono::steady_clock::now() < targetTimestamp) {
-                    ;
-                }
+            // If we have some time remaining until swap time, try to sleep closer to it
+			// This prevents large waits from hogging CPU power, but allows small waits/ to remain precise.
+            if ((targetTimestamp - std::chrono::steady_clock::now()) > std::chrono::duration<double, std::milli>(sleepThresholdMs)) {
+                // We expect sleep to overshoot, so give ourselves some extra time
+                auto sleepTarget = targetTimestamp - std::chrono::duration<double, std::milli>(sleepThresholdMs);
+                std::this_thread::sleep_until(targetTimestamp - std::chrono::duration<double, std::milli>(sleepThresholdMs));
```
PatrickvL, line 5498: I guess you wanted to use sleepTarget  here?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -5483,17 +5484,23 @@ xbox::dword_xt WINAPI xbox::EMUPATCH(D3DDevice_Swap)
...
-                std::this_thread::sleep_until(targetTimestamp);
-            } else {
-                // Otherwise, we fall-through and just keep polling
-                // This prevents large waits from hogging CPU power, but allows small waits/ to remain precice.
-                while (std::chrono::steady_clock::now() < targetTimestamp) {
-                    ;
-                }
+            // If we have some time remaining until swap time, try to sleep closer to it
+			// This prevents large waits from hogging CPU power, but allows small waits/ to remain precise.
+            if ((targetTimestamp - std::chrono::steady_clock::now()) > std::chrono::duration<double, std::milli>(sleepThresholdMs)) {
```
PatrickvL, line 5495: Would `sleep_until` turn out to be wildly inaccurate, this `if` might be better changed into a `while`

NZJenkins: possibly. I assume it more-or-less will sleep for us under the hood, but it might overshoot.
Found an article with an adaptive sleep method which might be good, though simple is easy to trust!

Going to look at using a common function for both of these, then we can change it later if needed

PR #2231, opened 2021-06-02
SleepPrecise for more accurate sleeps

NZJenkins:
We can use it to reduce the amount of spin waiting which burns CPU
Try to sleep as much as we can get away with without overshooting, and then spin wait
Use it for
- VBlank timing
- Frame limiter timing

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -2146,17 +2146,48 @@ static LRESULT WINAPI EmuMsgProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPar
     return S_OK; // = Is not part of D3D8 handling.
 }
 
-std::chrono::time_point<std::chrono::steady_clock, std::chrono::duration<double, std::nano>> GetNextVBlankTime()
+// More precise sleep, but with increased CPU usage
+void SleepPrecise(std::chrono::steady_clock::time_point targetTime)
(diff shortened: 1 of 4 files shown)


ergo720:
Instead of adding `SleepPrecise` to `Direct3D9.cpp`, you can instead add it to `Timer.cpp` and use the same mechanism in `ClockThread` as well (it currently uses `Sleep(1)` instead). This way, all clocks created with `Timer_Create` will benefit from `SleepPrecise`. Also, I think that the vblank clock should use the same clock mechanism offered by `Timer.cpp`, so we can centralize things.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2168,51 +2199,43 @@ static DWORD WINAPI EmuUpdateTickCount(LPVOID)
 
     EmuLog(LOG_LEVEL::DEBUG, "Timing thread is running.");
 
-    // current vertical blank count
-    int curvb = 0;
+	// We check for LLE flag as NV2A handles it's own VBLANK if LLE is enabled!
+	if (bLLE_GPU) return 0;
```
ergo720, line 2203: I think we always add parenthesis to `if` statements even they are only one line long.

LukeUsher: The style varies through the codebase, but I would prefer to always use parenthesis.

LukeUsher (approved):

PR #2233, opened 2021-06-07
Hslsl tss xbox exts

PatrickvL:
[ Replacement for accidentally closed PR#2228 ]

This PR addresses #1604, introducing most of what's needed to support COLORSIGN, COLORKEYCOLOR, COLORKEYOP alongside with a small ALPHAKILL refactoring.

To get there, X_D3DFMT_L6V5U5 had to be converted into host D3DFMT_X8L8V8U8, and some HLSL code had to be added to implement these extensions.

Some pieces to the puzzle may still be missing, but at least this is a step in the right direction.

The most important test is to see if JSRF boost dash didn't regress because of this. I can't really judge this well, so please give us some feedback.


medievil1:
ummm something is afoot...lol 
image

PatrickvL:
> ummm something is afoot...lol
> image

This green color likely is the 'WarningColor' I return when the ColorKey operation reaches the end of the function without hitting any of the known cases. The reason this happens, is probably not because we lack a 'known' case, but because one case is disabled due to it causing a shader compilation error - this, based on the following code :
```
#ifdef ENABLE_FF_ALPHAKILL
	if (ColorKeyOp == 3) // = _KILL
		discard; // When compiled into FixedFunctionPixelShader.hlsl, compile errors arise
```
... whereby `ENABLE_FF_ALPHAKILL` is currently disabled.

PR #2236, opened 2021-06-08
Change Memory/buffer release order. delete X_CDirectSoundStream*   when ref. count decrease to 0

jackchentwkh:
Change Memory/buffer release order to the reverse order of creation. 
delete X_CDirectSoundStream*   pThis in the 2nd call to CDirectSoundStream_Release(). 
this makes CrashTwinsanity happily ingame. 
somehow CrashTwinsanity calls CDirectSoundStream_Release twice with the same X_CDirectSoundStream*, and it only null out it's internal X_CDirectSoundStream* after the 2nd call of release. this modification makes it happy. so far no regressions with other titles. need further test.

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
@@ -176,48 +176,43 @@ xbox::ulong_xt WINAPI xbox::EMUPATCH(CDirectSoundStream_Release)
     LOG_FUNC_ONE_ARG(pThis);
 
     ULONG uRet = 0;
-    if (pThis != 0 && (pThis->EmuDirectSoundBuffer8 != 0)) {
+    if (pThis != nullptr && (pThis->EmuDirectSoundBuffer8 != 0)) {
 
-        //if (uRet == 0) {
-
-            // 1. remove packet/vector/iterator
-            for (auto buffer = pThis->Host_BufferPacketArray.begin(); buffer != pThis->Host_BufferPacketArray.end();) {
-                //CrashTwinsanity crashed here when interrupt cut scene. the vector iterator is still valid, but the buffer->xmp_data is no longer valid. so we have to free the buffer->xmp_data after we delete the packet.
-                DSStream_Packet_Clear(buffer, XMP_STATUS_FLUSHED, pThis->Xb_lpfnCallback, pThis->Xb_lpvContext, pThis);
-            }
-            // 2. remove the EmuBufferDesc
-            if (pThis->EmuBufferDesc.lpwfxFormat != nullptr) {
-                free(pThis->EmuBufferDesc.lpwfxFormat);
-                pThis->EmuBufferDesc.lpwfxFormat = nullptr;
-            }
-            // NOTE: Do not release X_BufferCache! X_BufferCache is using xbox buffer.
-
-            // 3. remove cache entry
-            vector_ds_stream::iterator ppDSStream = std::find(g_pDSoundStreamCache.begin(), g_pDSoundStreamCache.end(), pThis);
-            if (ppDSStream != g_pDSoundStreamCache.end()) {
-                g_pDSoundStreamCache.erase(ppDSStream);
-            }
-
-            // 4. remove direct sound 3D buffer
-            if (pThis->EmuDirectSound3DBuffer8 != nullptr) {
-                pThis->EmuDirectSound3DBuffer8->Release();
-                pThis->EmuDirectSound3DBuffer8 = nullptr;
-            }
-
-            //}
-            // 5. release the direct sound buffer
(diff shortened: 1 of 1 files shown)


src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
```diff
@@ -173,34 +173,51 @@ xbox::ulong_xt WINAPI xbox::EMUPATCH(CDirectSoundStream_Release)
...
+            if (pThis->EmuDirectSoundBuffer8 != nullptr) {
+                uRet = pThis->EmuDirectSoundBuffer8->Release();
+                pThis->EmuDirectSoundBuffer8 = nullptr;
             }
-            // NOTE: Do not release X_BufferCache! X_BufferCache is using xbox buffer.
 
-            delete pThis;
-        }
+            // 6. remove X_CDirectSoundStream. pThis must be delete in the last. once it's been deleted, all it's members will be unaccessible.
+            //delete pThis;
```
LukeUsher, line 215: What happens if a game only calls Release once? Won't the pThis pointer leak? 

it might be worth checking the reference count (uRet) and deleting pThis when the reference count is zero.

src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
```diff
@@ -173,34 +173,51 @@ xbox::ulong_xt WINAPI xbox::EMUPATCH(CDirectSoundStream_Release)
...
+                pThis->EmuDirectSound3DBuffer8 = nullptr;
             }
 
-            if (pThis->EmuBufferDesc.lpwfxFormat != nullptr) {
-                free(pThis->EmuBufferDesc.lpwfxFormat);
+            //}
+            // 5. release the direct sound buffer
+            if (pThis->EmuDirectSoundBuffer8 != nullptr) {
+                uRet = pThis->EmuDirectSoundBuffer8->Release();
+                pThis->EmuDirectSoundBuffer8 = nullptr;
```
LukeUsher, line 211: This might not be safe: You don't check that the reference count is 0 before erasing the pointer, Release may decrement the reference count, it *only* frees the buffer when it returns zero.

src/core/hle/DSOUND/DirectSound/DirectSoundStream.cpp
```diff
@@ -173,34 +173,51 @@ xbox::ulong_xt WINAPI xbox::EMUPATCH(CDirectSoundStream_Release)
...
+            // 3. remove cache entry
             vector_ds_stream::iterator ppDSStream = std::find(g_pDSoundStreamCache.begin(), g_pDSoundStreamCache.end(), pThis);
             if (ppDSStream != g_pDSoundStreamCache.end()) {
                 g_pDSoundStreamCache.erase(ppDSStream);
             }
 
-            for (auto buffer = pThis->Host_BufferPacketArray.begin(); buffer != pThis->Host_BufferPacketArray.end();) {
-                DSStream_Packet_Clear(buffer, XMP_STATUS_FLUSHED, pThis->Xb_lpfnCallback, pThis->Xb_lpvContext, pThis);
+            // 4. remove direct sound 3D buffer
+            if (pThis->EmuDirectSound3DBuffer8 != nullptr) {
```
LukeUsher, line 202: Not safe: see comment on EmuDirectSoundBuffer8.

LukeUsher (changes requested):
Changing the order of operations is useful, but you shouldn't be freeing the related data structures without verifying the reference count is indeed 0.

jackchentwkh:
original code flow:
if pThis!=nullptr && soundbuffer!=0
{
    ref=soundbuffer->release();
    //most guest code will have ref=0 here, no problem. some titles will have ref.=1 such as CrashTwinsanity and Halo, which will introduce trouble later when the guest code call dsstream_release twice with the same pthis
    if(ref==0) {
        packet_clear(); // we'll have access violation here if the guest code call DSStream_release twice with the same pThis. in the 2nd 
        call, the guest side buffer pointed by the packet is invalidate.
        DSStream_cache.erase()
        delete pThis.
    }
}

in the situation of DSStream_release called twice with same pThis, 
packet_clear will have access violation with xbox::XMEDIAPACKET access, like this code.         
 if (buffer->xmp_data.pdwStatus != xbox::zeroptr) {
        (*buffer->xmp_data.pdwStatus) = status;
the xbox::XMEDIAPACKET xmp_data is invalidated before the 2nd time DSStream_Release was called
which means the guest code removes those packets in the time between the fist call and 2nd call to DSStream_Release
if we don't clear host packet in the first call, further access violation might be introduced when the host side packets are being processed and the quest side clears the xbox::XMEDIAPACKET

RadWolfie:
Closing this pull request in favor of https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2257 mimic hack solution.

PR #2237, opened 2021-06-10
fixing 3D mip-map conversion with volume texture, fix Halo

jackchentwkh:
using correct 3D mip map dimention sizes in all 3 dimensions, and calculate correct source size.
this fix Halo, get in game.
note: the cache partition must be cleared prior to emulation, or the title would get stocked in D3DResouce_IsBusy() check. 
debug build is more stable. somehow the release build won't work, seems stocked in the  D3DResouce_IsBusy() check.

 video here https://www.youtube.com/watch?v=ARHHpgDrCts
https://youtu.be/LHXHPia4Pc0
image
image

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -6123,9 +6123,9 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 					if (!skipDueToNoPalette) {
 						if (!ConvertD3DTextureToARGBBuffer(
 							X_Format,
-							pSrc, pxMipWidth, pxMipHeight, dwMipRowPitch, dwMipSize,//use dwMipSize instaed of dwSrcSlicePitch here, because the mipSize changed in 3D mip map.
+							pSrc, pxMipWidth, pxMipHeight, dwMipRowPitch, dwMipSize,//use dwMipSize instead of dwSrcSlicePitch here, because the mipSize changed in 3D mip map.
 							pDst, dwDstRowPitch, dwDstSlicePitch,
-							pxMipDepth,//usd pxMipDepth here because in 3D mip map the 3rd dimension also shrinked to 1/2 at each mip level.
+							pxMipDepth,//used pxMipDepth here because in 3D mip map the 3rd dimension also shrinked to 1/2 at each mip level.
 							iTextureStage)) {
 							CxbxKrnlCleanup("Unhandled conversion!");
 						}
@@ -6218,7 +6218,7 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 				}
 
 				if (pxMipDepth > 1) {
-					pxMipDepth /= 2;//this is for 3D volumeTexture mip-map, it shrinked down to 1/2 in 3 diemesions. this variable should be used.
+					pxMipDepth /= 2;//this is for 3D volumeTexture mip-map, it shrinked down to 1/2 in 3 dimensions. this variable should be used.
 				}
 			} // for mipmap levels
 
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -6123,9 +6123,9 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 					if (!skipDueToNoPalette) {
 						if (!ConvertD3DTextureToARGBBuffer(
 							X_Format,
-							pSrc, pxMipWidth, pxMipHeight, dwMipRowPitch, dwSrcSlicePitch,
+							pSrc, pxMipWidth, pxMipHeight, dwMipRowPitch, dwMipSize,//use dwMipSize instaed of dwSrcSlicePitch here, because the mipSize changed in 3D mip map.
 							pDst, dwDstRowPitch, dwDstSlicePitch,
-							dwDepth,
+							pxMipDepth,//usd pxMipDepth here because in 3D mip map the 3rd dimension also shrinked to 1/2 at each mip level.
 							iTextureStage)) {
 							CxbxKrnlCleanup("Unhandled conversion!");
 						}
@@ -6196,11 +6196,11 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 				}
 
 				if (face == D3DCUBEMAP_FACE_POSITIVE_X) {
-					dwCubeFaceSize += dwDepth * dwMipSize;
+					dwCubeFaceSize += pxMipDepth * dwMipSize;
 				}
 
 				// Calculate the next mipmap level dimensions
-				dwMipOffset += dwMipSize;
+				dwMipOffset += pxMipDepth * dwMipSize;//for 3D volumeTexture, the dwDepth >1 in mip_level 0, pxMipDepth must be multiplied.
 				if (pxMipWidth > 1) {
 					pxMipWidth /= 2;
 
@@ -6218,7 +6218,7 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 				}
 
 				if (pxMipDepth > 1) {
-					pxMipDepth /= 2;
+					pxMipDepth /= 2;//this is for 3D volumeTexture mip-map, it shrinked down to 1/2 in 3 diemesions. this variable should be used.
 				}
 			} // for mipmap levels
 


PatrickvL (approved):
Yes! You did it! Nicely done!

Please add a few screenshots for good measure, to mark this milestone

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6196,11 +6196,11 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 				}
 
 				if (face == D3DCUBEMAP_FACE_POSITIVE_X) {
-					dwCubeFaceSize += dwDepth * dwMipSize;
+					dwCubeFaceSize += pxMipDepth * dwMipSize;
 				}
 
 				// Calculate the next mipmap level dimensions
-				dwMipOffset += dwMipSize;
+				dwMipOffset += pxMipDepth * dwMipSize;//for 3D volumeTexture, the dwDepth >1 in mip_level 0, pxMipDepth must be multiplied.
```
PatrickvL, line 6203: Do we have test cases for this? Asking, because in this commit, this is the only change I'm a little bit in doubt about...

jackchentwkh: frankly I don't know how to verify it neither. the I modify this code and test it directly with Halo then it work out of the box.
I have no idea how to verify whether it's correct or not can you give me some hint? or can you test halo for it?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6123,9 +6123,9 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 					if (!skipDueToNoPalette) {
 						if (!ConvertD3DTextureToARGBBuffer(
 							X_Format,
-							pSrc, pxMipWidth, pxMipHeight, dwMipRowPitch, dwSrcSlicePitch,
+							pSrc, pxMipWidth, pxMipHeight, dwMipRowPitch, dwMipSize,//use dwMipSize instaed of dwSrcSlicePitch here, because the mipSize changed in 3D mip map.
```
LukeUsher, line 6126: typo: instaed -> instead

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6123,9 +6123,9 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
...
 							pDst, dwDstRowPitch, dwDstSlicePitch,
-							dwDepth,
+							pxMipDepth,//usd pxMipDepth here because in 3D mip map the 3rd dimension also shrinked to 1/2 at each mip level.
```
LukeUsher, line 6128: typo: usd -> used or use

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6218,7 +6218,7 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 				}
 
 				if (pxMipDepth > 1) {
-					pxMipDepth /= 2;
+					pxMipDepth /= 2;//this is for 3D volumeTexture mip-map, it shrinked down to 1/2 in 3 diemesions. this variable should be used.
```
LukeUsher, line 6221: typo: diemesions -> dimensions

LukeUsher (changes requested):
Looks great! Nice work, but there's a couple of typos in the comments

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -6196,11 +6196,11 @@ void CreateHostResource(xbox::X_D3DResource *pResource, DWORD D3DUsage, int iTex
 				}
 
 				if (face == D3DCUBEMAP_FACE_POSITIVE_X) {
-					dwCubeFaceSize += dwDepth * dwMipSize;
+					dwCubeFaceSize += pxMipDepth * dwMipSize;
```
NZJenkins, line 6199: question: I know the change here is replacing `dwDepth` with `pxMipDepth`, and nothing else but...

Cubemaps shouldn't have depth - does what this code is doing make sense?
Why is the cube face size a special case? Isn't it just another 2D texture?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins, line 6203: suggestion: again I _suspect_ mip size should include depth already

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: nitpick:
Not sure what is meant by
> this variable should be used

this pr adds a long comment at each change.

I think it'd be best to add comments about what these variables/data is, and make sure they're well defined.
Then what we do at every step should make more sense and need less comments.

In this case we are already dividing all the other dimensions by two, I don't think depth warrants a special comment here

NZJenkins (commented):
nice find!

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: > if there is a way to retrieve the src texture data and examine it with other tools, then it's possible to quickly verify this.

it's possible to examine textures in a frame capture and view each mip level. Can do this with Intel GPA

jackchentwkh: GPA only works for me the 1st time after installation. after that, it won't work. I am still struggle with it. as for this case, we're supposed to check the texture from guest, not host, GPA could only check the texture after conversion. but I would like to check the guest texture first.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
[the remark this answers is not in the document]
jackchentwkh: at mip level 0, mipSize == dwSrcSlicePitch, but when mip level increases, the mipSize decreases accordingly, on the other hand, dwSrcSlicePitch never changes.  thus the source size per layer should follow mipSize instead of dwSrcSlicePitch when the mip_level loops to levels other than 0. you have to be aware that this conversion code is inside a for loop which loops through the mip levels. and the conversion code converts one mip level at a time.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
jackchentwkh: I don't know cube maps. what I am dealing with Halo is volume texture. the case I am dealing with has width 0x20, height 0x20, depth 0x20.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
jackchentwkh: nope, mipSize only calculates row_pitch X height

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
jackchentwkh: yeah, for this specific comment I am actually leave it here for the ease of reviewing. 
because the MipDepth was devided by 2 at each mip level, which means the original code is aware of the 3D mip map.
but somehow the mipDepth wasn't used in the conversion, well, not in the toARGB conversion to be precise.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: Hopefully it's easy to see if there was any issue converting the texture, even without a baseline. I'll have a look

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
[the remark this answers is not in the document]
NZJenkins: Right. I guess it should've always been inside the loop.
Since that was the only use of `dwSrcSlicePitch`, it can be removed entirely! 🧹 

To confirm some terminology:
https://docs.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-resources-textures-intro#3d-textures

The link refers to the mip-level aka slice of a 3D texture as 3D, which was my understanding.
But our conversion code refers to a slice as a single layer of a 3D texture i.e. width * height
IMHO best be consistent and track mip size in pixels as width * height * depth.
Then in the two other places we mip size, we don't have to multiply by depth.
We can divide by depth to make the existing conversion code happy

LukeUsher (approved):

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: Got a capture and found a volume texture that looks good - no mipmap though apparently

PatrickvL:
Merged after confirmation from @NZJenkins 

Note, we still need to check whether volume mipmaps get converted correctly as well

PR #2239, opened 2021-06-10
Add support to load dsp images from xbe sections

ergo720:
Halo 2 tries to load a dsp image with `XAudioDownloadEffectsImage` from an xbe section called "DSPImage", and it was crashing when it attempted to read the nullptr that we were returning from the function. Unfortunately, it still crashes later before showing anything when it tries to execute a dpc set by `KeSetTimer`.


LukeUsher (approved):

jackchentwkh:
Bravo! bro!. I'll take a further look at Halo2 later.

PR #2243, opened 2021-06-20
Upstream Chihiro Non-LLE Code

RadWolfie:
Hopefully this will be the last pull request to move chihiro relative changes. Then any rebases on wip Chihiro will be easier management to continue the work.

With addition to the changes, I added a check if system/xbe are detected to run as Chihiro type will force failure. Chihiro support is still under major work in progress and being work on a separate branch.

I'm requesting @LukeUsher to verify any changes made here are good to go.


src/core/hle/XAPI/Xapi.cpp
```diff
@@ -103,6 +103,12 @@ bool ConstructHleInputDevice(int Type, int Port)
 	g_bIsDevicesEmulating = true;
 	bool ret = true;
 
+	if (g_bIsChihiro) {
+		// Don't emulate XID devices during Chihiro Emulation
+		g_bIsDevicesEmulating = false;
+		return ret;
+	}
```
ergo720, line 110: I believe this will break input in Virtua Cop 3, which does support duke and S controllers. By making the function return success without actually constructing the input device state, this will likely cause issues.

LukeUsher: Chihiro *games* are debug-signed, not Chihiro signed, so this code path won't trigger when running Virtua Cop 3 in 'Xbox' mode, it will only trigger when being run on an emulated chihiro (which isn't yet implemented in this PR)

LukeUsher (approved):
Overall looks good: 

Just a heads up that long-term/in the future:  `g_bIsChihiro` will need to be true for any XBE that runs in Chihiro mode, false otherwise, rather than being set purely based on the Xbe type.

Perhaps we should rename it or add another variable for "IsEmulatingChihiro" or something, to allow Xbox mode to still be used for games like Virtua Cop 3.

PR #2245, opened 2021-06-21
Fix Release Title Default

RadWolfie:
@LukeUsher found a bug caused by #2242 for release creation to return empty name in GitHub API. I recongize recent merged pull request output unintentional default title name on release page. This pull request will fix the issue.

Doc: https://cli.github.com/manual/gh_release_create

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -79,4 +79,4 @@ jobs:
         env:
           GH_REPO: ${{ github.repository }} # https://github.com/cli/cli/issues/3556
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: gh release create ${{ steps.git.outputs.tag_name }} artifacts/*.zip -p --target $GITHUB_SHA
+        run: gh release create ${{ steps.git.outputs.tag_name }} artifacts/*.zip -p --target $GITHUB_SHA --title '${{ steps.git.outputs.tag_name }}'


.github/workflows/CI.yml
```diff
@@ -79,4 +79,4 @@ jobs:
         env:
           GH_REPO: ${{ github.repository }} # https://github.com/cli/cli/issues/3556
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: gh release create ${{ steps.git.outputs.tag_name }} artifacts/*.zip -p --target $GITHUB_SHA
+        run: gh release create ${{ steps.git.outputs.tag_name }} artifacts/*.zip -p --target $GITHUB_SHA --title '${{ steps.git.outputs.tag_name }}'
```
Margen67, line 82:
```suggestion
        run: gh release create ${{ steps.git.outputs.tag_name }} artifacts/*.zip -p --target $GITHUB_SHA -t '${{ steps.git.outputs.tag_name }}'
```

LukeUsher (approved):

.github/workflows/CI.yml
RadWolfie: `-t` is another way but we have `-target` which can lead to confusion if you look at first character letter. That's why I choose `--title` instead.

Margen67: Does the title need single quotation marks around it?

RadWolfie: Either single or double quotation marks are fine. Main reason for this is to be aware if possible space(s) may have been input. Always use quoation marks when you're involve with title and body content strings.

PR #2246, opened 2021-06-21
Added support for memory unit devices

ergo720:
This adds support for a new input device, the memory unit (MU), which becomes available in the input gui when selecting the duke or S controller. For this cxbxr will now create a new folder called EmuMu in the same location where EmuDisk is created, and inside EmuMu there will be 8 folders, one for each MU. Regarding the internal code changes, this implements `XMountMUA`, `XMountMURootA`, `XUnmountMU`, `XReadMUMetaData` and extends the kernel functions `NtQueryVolumeInformationFile`, `NtDeviceIoControlFile` and `NtFsControlFile` to support MUs, in addition to the HDD.  Below is a pic of the dashboard recognizing the MUs:

Capture

Testing is welcome.


LanHikariDS:
Can confirm that TimeSplitters 2 and TimeSplitters: Future Perfect read from/write to them flawlessly-
image
image

src/core/kernel/support/EmuFile.cpp
```diff
@@ -84,7 +84,102 @@ XboxPartitionTable BackupPartTbl =
...
+	if (pos != std::string::npos) {
+		return DeviceType::Harddisk0;
+	}
+
+	pos = path.rfind(L"\\EmuMu");
+	if (pos != std::string::npos) {
+		return DeviceType::MU;
+	}
+
+	EmuNtSymbolicLinkObject *ret = FindNtSymbolicLinkObjectByDevice(DeviceCdrom0);
```
RadWolfie, line 174: I highly recommend do not use `FindNtSymbolicLinkObjectByDevice` solution as it is not intended to be used at all. Instead it is replaced to `FindEmuDirPathByDevice`. Using our own `NtSymbolicLinkObject` objects should be last resort. In fact, FindNtSymbolicLinkObjectByDevice should techinically return list of `NtSymbolicLinkObject` objects. But symbolic link objects are not only target to drive letter. So, using `NtSymbolicLinkObject` only support to drive letters are entirely wrong.

In summary, our `NtSymbolicLinkObject` class and/or functions need a rework to support non-drive letters.

RadWolfie (changes requested):
I had not made a full review, there's some changes need some fixes before I can re-review again.

I notice there are bunch of `dev1` and `port1` and the like which is not suitable to be use and lead confusion. Intead of using dev# and port#, they are actually iterator (in most cases), so using iDev and iPort, or i_dev and i_port, are preferred to be used.

RadWolfie:
Beside code review, I tested the pull request's build with dashboard first, then Monopoly Party title.

I can confirm personalize MU names does work. And able to display the MU name in Monopoly Party title as seen below:

<details>
<summary>MU1A with personalize</summary>

MU1A with personalize
</details>
<details>
<summary>MU1B without personalize (initialized by game title)</summary>

MU1B without personalize
</details>

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -802,19 +830,55 @@ XBSYSAPI EXPORTNUM(200) xbox::ntstatus_xt NTAPI xbox::NtFsControlFile
...
-				CxbxFormatPartitionByHandle(FileHandle);
-				ret = xbox::status_success;
-			}
-			break;
+	switch (FsControlCode)
+	{
+	case fsctl_dismount_volume: {
+		int partitionNumber = CxbxGetPartitionNumberFromHandle(FileHandle);
+		if (partitionNumber > 0) {
+			CxbxFormatPartitionByHandle(FileHandle);
```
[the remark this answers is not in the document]
RadWolfie: I didn't mean by note "NOTE:", I meant by "HACK:" note.

src/common/Settings.hpp
```diff
@@ -147,6 +147,7 @@ class Settings
 
 	struct s_input_port {
 		int Type;
+		int SlotType[2];
```
RadWolfie, line 150:
```suggestion
		int SlotType[XBOX_CTRL_NUM_SLOTS];
```
If not able to do above method, please add `XBOX_CTRL_NUM_SLOTS` in a comment on this line.

src/common/input/InputManager.h
```diff
@@ -47,6 +53,11 @@ inline XBOX_INPUT_DEVICE input_support_list[] = {
 	XBOX_INPUT_DEVICE::ARCADE_STICK,
 };
 
+inline XBOX_INPUT_DEVICE slot_support_list[] = {
+	XBOX_INPUT_DEVICE::DEVICE_INVALID,
+	XBOX_INPUT_DEVICE::MEMORY_UNIT,
+};
```
RadWolfie, line 59: Not a review remark, perhaps include microphone and keyboard (for PSO title) devices in a commented section?

src/common/input/InputManager.h
```diff
@@ -92,44 +103,44 @@ struct SBCOutput {
...
-	CXBX_XINPUT_IN_STATE    *InState;
-	bool                    bPendingRemoval;
-	bool                    bSignaled;
-	bool                    bIoInProgress;
-	bool                    bAutoPoll;         // autopoll on/off, as instructed by the title in XInputOpen
-	bool                    bAutoPollDefault;  // default autopoll value, depending on device type
-	CXBX_XINPUT_DEVICE_INFO XboxDeviceInfo;
-}
-CXBX_CONTROLLER_HOST_BRIDGE, *PCXBX_CONTROLLER_HOST_BRIDGE;
+extern DeviceState g_devs[4 + 8];
```
RadWolfie, line 134:
```suggestion
extern DeviceState g_devs[4 + 4 * XBOX_CTRL_NUM_SLOTS];
```
imo, ^ will better than hardcode numbers.
Although, it appear we don't have macro for `XBOX_HW_NUM_PORT_SLOTS` or the like.

RadWolfie (changes requested):
Few code style changes and base on numerous to input mechanical not relative to memory unit devices. I recommended to perform full test with various titles that had been known to break in the past.

src/common/input/SdlJoystick.cpp
```diff
@@ -160,15 +156,22 @@ namespace Sdl
...
+					PortStr2Int(port, &port_num, &slot);
+
+					g_InputDeviceManager.UpdateDevices(port, false);
+					// Force an update of the entire slot connectivity of this port
+					g_InputDeviceManager.UpdateDevices(port + ".0", false);
+					g_InputDeviceManager.UpdateDevices(port + ".1", false);
 				}
 
-				delete Event.user.data1;
+				delete static_cast<int *>(Event.user.data1);
```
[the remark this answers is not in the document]
ergo720: That's defined by SDL itself, I cannot change it.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -43,24 +43,69 @@
...
 
-// Global bridge for xbox controller to host, 4 elements for 4 ports.
-CXBX_CONTROLLER_HOST_BRIDGE g_XboxControllerHostBridge[4] = {
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[0], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[1], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[2], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
-	{ NULL, PORT_INVALID, XBOX_INPUT_DEVICE::DEVICE_INVALID, &g_InState[3], false, false, false, false, false, { 0, 0, 0, 0, 0 } },
-};
+// Protects access to xpp types
+std::atomic<bool> g_bXppGuard = false;
```
RadWolfie, line 58: I'm not sure if we need this additional guard?

> // Guard against the unfortunate case where XGetDevices or XGetDeviceChanges have already checked for g_bIsDevicesInitializing
	// and g_bIsDevicesEmulating and a thread switch happens to this function

Quote from `while (g_bXppGuard) {}` line.

Can't we simply check both booleans than rely on additional variable, g_bXppGuard?

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -1921,6 +1948,11 @@ void CxbxKrnlShutDown(bool is_reboot)
 	// Shutdown the memory manager
 	g_VMManager.Shutdown();
 
+	if (g_io_mu_metadata) {
+		delete g_io_mu_metadata;
+		g_io_mu_metadata = nullptr;
+	}
+
```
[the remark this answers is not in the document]
ergo720: It cannot be put inside `g_InputDeviceManager.Shutdown();` because this is also called from the GUI, where `g_io_mu_metadata` doesn't exist. Also, this class has nothing to do with the input manager, it just regulates access to the MU partition headers.

src/common/input/SdlJoystick.cpp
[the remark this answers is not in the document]
RadWolfie: Strange, but ok SDL.

src/core/kernel/init/CxbxKrnl.cpp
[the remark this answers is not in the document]
RadWolfie: Okay, except MUs does have XPP type. So, it'll be wise to keep XPP types closely together and before memory manager's shutdown. As in memory manager should be the last one to shutdown as safe guard reason. Of course, before EmuShared's shutdown.

src/common/input/InputManager.h
ergo720: We could but, like the name says, this struct is meant to be used during the initialization code of the xbox controller window, to know which slot devices are to be displayed and we support. So, adding the unsupported devices, even if commented out, doesn't feel right to me, and can be easily added later when we will implement them.

RadWolfie: I don't agree with you. If you can remember, me and other guy had issues trying to find where SBC should be  placed at from existing defines. Beside, having commented out lines here as a "TODO" list will help better. And we do know both microphone and keyboard are insert into the controller's slots. By not having todo/notes here, we will eventually forget all over again. It is intended for a record.

src/core/hle/XAPI/Xapi.cpp
ergo720: `UpdateXppState` is only called from `ConstructHleInputDevice` and `DestructHleInputDevice`, both of which set `g_bIsDevicesEmulating` to true. The unfortunate case happens when `XGetDevices` or `XGetDeviceChanges` are called, find `g_bIsDevicesEmulating = false`, proceed, but then a thread switch happens to `UpdateXppState`. Without the guard, the latter will concurrently access the xpp type, which might cause issues depending on which xpp type `XGetDevices` and `XGetDeviceChanges` are using.

RadWolfie (approved):
Known titles experienced regression in the past has pass test by our tester.

PR #2248, opened 2021-06-22
Raise max resolution scale to 12x

Margen67:
D3D9 has a max texture size of ~8192(?) so this can go a little higher than 8x. I'm not sure if certain GPUs will have a lower max than that. (This can probably be raised again once Cxbx-Reloaded switches to D3D11)


LukeUsher (approved):
It is, however, important to note that the sizes in brackets are assuming that the original XBox game is running at 640x480.

If the game is running at 720p natively, then 12x will be much higher than shown here.

PR #2249, opened 2021-06-23
CI changes

Margen67:


.azure-pipelines.yml
```diff
@@ -4,10 +4,12 @@ trigger:
     - '*'
   paths:
     exclude:
-    - '.github/**'
+    - '.github/*'
+    - '.github/*/*'
```
RadWolfie, line 8: Why change to this way? Base on the doc, the original method is correct.
Doc: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/file-matching-patterns

Margen67: It's unimplemented: https://developercommunity.visualstudio.com/t/support-wildcards-in-trigger-path-filters-1/366363
Even with this it still doesn't ignore the paths:
```
    - doc/**
    - doc/**/
    - doc/**/*
    - doc/**/*/
    - /doc/**
    - /doc/**/
    - /doc/**/*
    - /doc/**/*/
```

RadWolfie: According to fnmatch's method, wildcard will only match any current folder. It would not expand wildcard to multiple subfolders to one, even two, wildcards.

In summary, one wildcard per subfolder (or files) per line is required. In my opinion, we should document the reason why we're not using recursive wildcard in the workflow. Please mention what current method Azure is using, aka `fnmatch` function.

RadWolfie (approved):

PR #2254, opened 2021-06-29
Small Clean Up Portion of EmuFile Usage

RadWolfie:
Moving few and properly rename global variables into EmuFile file. Updated CxbxrGetDeviceTypeFromHandle function to check full path than partial path.

### DIFF
--- a/src/core/kernel/support/EmuFile.cpp
+++ b/src/core/kernel/support/EmuFile.cpp
@@ -47,7 +47,11 @@
 #include <filesystem>
 
 // partition emulation directory handles
-HANDLE           g_hCurDir_hack = NULL; // HACK: We should not be depending on this variable. Instead, we should fix/implement Ob/Io objects such as IoCreateDevice.
+HANDLE g_hCurDir_hack = NULL; // HACK: We should not be depending on this variable. Instead, we should fix/implement Ob/Io objects such as IoCreateDevice.
+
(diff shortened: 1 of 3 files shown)


src/core/kernel/support/EmuFile.cpp
```diff
@@ -157,23 +161,44 @@ void io_mu_metadata::flush(const wchar_t lett)
 	ofs.flush();
 }
 
+// NOTE: root_path input must already had called canonical function for optimization purpose.
+static bool CxbxrIsPathInsideRootPath(const std::filesystem::path& path, const std::filesystem::path& root_path)
+{
+	std::error_code rootError, finalError;
+	const std::filesystem::path rootPath = std::filesystem::canonical(root_path, rootError); // TODO: Replace rootPath to root_path when possible.
+	const std::filesystem::path finalPath = std::filesystem::canonical(path, finalError);
+	if (finalError) {
```
CookiePLMonster, line 170: Any reason behind ignoring `rootError`? The "original" code moved from elsewhere didn't ignore it.

RadWolfie: Was intentionally remove canonical call for root path but std::mismatch didn't exactly like it due to root path has extra slash at the end.

Still, we need to convert our file strings into filesystem::path. Then it will be easier to code the filesystem.

ergo720 (approved):

PR #2259, opened 2021-07-14
Removed unnecessary info from the cxbxr title string

ergo720:
This removes from the title string the date of the build and the strings "Loaded" and "Emulating". This, because the latter are just useless, and the former because the build is already uniquely identified by the hash and it's not reliable because we can have multiple builds on the same day. Very often, this led to excessively long title strings, and you could not read the name of the game because of this.

### DIFF
--- a/src/gui/WndMain.cpp
+++ b/src/gui/WndMain.cpp
@@ -205,7 +205,7 @@ WndMain::WndMain(HINSTANCE x_hInstance) :
 	// initialize members
 	{
 		m_classname = "WndMain";
-		m_wndname   = "Cxbx-Reloaded " + std::string(CxbxVersionStr);
+		m_wndname   = "Cxbx-Reloaded " + std::string(CxbxrHashBuild);
 	}
 
 	// load configuration from settings file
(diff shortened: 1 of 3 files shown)


src/gui/WndMain.cpp
```diff
@@ -1868,16 +1868,9 @@ void WndMain::UpdateCaption()
 {
 	char AsciiTitle[MAX_PATH];
 
-	int i = sprintf(AsciiTitle, "Cxbx-Reloaded %s", CxbxVersionStr);
+	int i = sprintf(AsciiTitle, "Cxbx-Reloaded %s", CxbxrHashBuild);
 	if (m_Xbe != nullptr) {
-		if (m_bIsStarted) {
```
PatrickvL, line 1873: Now that the (m_bIsStarted) based selection of "emulating" beats "loaded" is removed: Is there any other visual feedback that emulation started (especially before any graphics are drawn by the title)?

What we have right now:
* At launch of the GUI : Cxbx backdrop in main window
* Once a titles is loaded : game icon in lower right corner 
* When launching emulation : black screen without any progress while symbol scanning
* Once the title starts running : whatever the title shows (could be black due to bugs or title loading) + active led

So the initial two phases don't require additional info in the caption bar, nor do we need any visual feedback while emulating.

But the symbol scanning phase could use some form of feedback, be it a progress bar or simply a waiting cursor, or whatever would be appropriate.

Perhaps not an addition that should belong in this pull request, but something to contemplate...

PatrickvL (approved):

src/gui/WndMain.cpp
ergo720: Moved to https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2260

gandalfthewhite19890404:
I'm not agree with removing date from title string, it was useful.

LukeUsher:
> I'm not agree with removing date from title string, it was useful.

It is not. It tells you what date the build was compiled but is absolutely meaningless at telling you what that build contained. Only the commit hash provides useful information.

CookiePLMonster:
It can be useful for a quick eyeball of how old the build on the screenshot is though. Retyping the commit hash just to check if it's an out of date build sounds like a lot of effort.

LukeUsher:
> It can be useful for a quick eyeball of how old the build on the screenshot is though. Retyping the commit hash just to check if it's an out of date build sounds like a lot of effort.

It can, but then if a user compiles their own build, the date will not be remotely representative of the contents of the build

ergo720:
The date of the build is still printed in the kernel log file at info level, so you can just retrieve it from there.

Calinou:
> It can, but then if a user compiles their own build, the date will not be remotely representative of the contents of the build

The date could be sourced from the Git commit instead of using the build date. This would also be more friendly to [reproducible builds](https://reproducible-builds.org/) in the long run.

PR #2261, opened 2021-07-14
Removed stdcall from EmuLog functions

ergo720:
`EmuLogEx` and `EmuLogInit` are variadic functions, and thus using stdcall on them is an error and probably caused undefined behavior. This switches them to use cdecl instead.


PatrickvL (approved):
LGTM

PR #2266, opened 2021-07-21
Reworded error message displayed when cxbxr fails to launch an xbe

ergo720:
As the title says, this replaces the error message `"Emulation failed.\n\n If this message repeats, the Xbe is not supported."` with another explaining why the emulation failed to start. This, because the original message is too vague and probably leads to confusion among end users.


LukeUsher (approved):
Excellent use of std::optional

Issue #1450, opened 2018-09-25
Its not possible to play games from Windows share directory.


ruthan:
Pleas,
add support for old game from Windows share directory, other emulators are support it and its nice for playing on multiple machines, i can save lots of space and time.

 Now even i try to load data rom my UNC NAS path:
\\10.0.0.15\Games\Xbox is get:
Received Fatal Message:
Could not open Xbe file for gameName.xbe

RadWolfie:
For windows network sharing, beginning of the path must have double slashes.
`\\__domain_name_or_ip_address__\__path_to_folder_or_file__`

Plus check your settings.ini file if `RecentXbeFiles` key does contain identical to above. If not, add the slash then try again.

LukeUsher:
I've games from a NAS sometimes, and it works just fine: At least I tested with DOAX.
I accessed by share via the UNC path "\\PERSONALCLOUD\\Luke\Xbox\Games\DOAX\default.xbe"

RadWolfie:
Seems github did not output correctly for both @ruthan and @LukeUsher posts.

RadWolfie:
* Are you able to browse on Windows shared directory for xbox game directories?
* Is it a game specific issue?
* Did it happen on start of emulation or later on?

I do not heard of "gameName.xbe". Title's discs are launch from default.xbe at top directory and xboxdash.xbe for official dashboard on harddrive.

haradons:
I have found that when loading games from a NAS that contain a default.xbe that then loads gamename.xbe (Ex. Morrowind.xbe), CXBX will show this error:
![image](https://user-images.githubusercontent.com/13687549/48359320-8ac97c80-e651-11e8-8d5e-87a2bf82f252.png)

RadWolfie:
Hmm... I found the problem (just by looking at the source code).
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/3393ccc684f7814fdcad16472773ab255cac3c45/src/CxbxKrnl/CxbxKrnl.cpp#L1102-L1119

Line 1112 (remove extra slashes) may had replaced double slashes to single slash from very beginning of the string. A solution would be replace `n=0` to `n=1` to ignore first character of the search.

aevans0001:
So I just upgraded form an October build to January Build, and now network locations are acting the same as above.  It no longer sees my Mapped drive and if I use \\mediacenter\games  it finds them but refuses to load , like above

[...]

PR #1413, opened 2018-08-26
Fix both retail and demo discs with given current directory

RadWolfie:
Following titles are functional: (demos didn't work right before)
* Xbox Demos [XK-001 v1.01]
* Whacked
  * Xbox demos
* RalliSport Challenge
  * Xbox demos
* Quantum Redshift
  * Xbox demos

Confirmed does not regress:
* Xbox dashboard
  * Xbox Live

PR #2267, opened 2021-07-26
Fix loading from UNC paths

NZJenkins:
Remove unexplained slash processing
which breaks UNC paths

related #1450


LukeUsher (approved):

RadWolfie:
Not approved because it break Exhibition Demo Disc's Whacked demo launch. https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/1413 contain whole clean up process and this pr only removed half of the clean up process.

I had proposed a possible solution by move starting index to one but I cannot verify it myself. And there was no feedback from it.

PR #2268, opened 2021-07-26
Print the windows error string when CreateProcess fails to create the new emulation process

ergo720:
This builds upon https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2266 to be more explicit. It now prints exactly the error string corresponding to error code reported by Windows when `CreateProcess` fails to create the new emulation process for whatever reason.


RadWolfie (approved):
LGTM!

PR #2271, opened 2021-08-01
Some cleanup kernel process

RadWolfie:
I made some breakdown code blocks into their own functions. Moved several functions into its own files. And rename some functions too.


ergo720:
Why rename `CxbxKrnlCleanup(Ex)` to `CxbxrKrnlAbort(Ex)`? It should just be `CxbxrAbort(Ex)` in my opinion, because we use this function everywhere in the code base to terminate the emulation when a fatal error is encountered, not just to terminate the kernel.

Not related to the above: do we really need the cxbxr prefix in our functions? After all, this is Cxbx-Reloaded already and so, adding the cxbxr prefix feels pretty much superfluous to me.

PatrickvL:
There's a lot more to do, but if we never start, nothing would happen, so thanks for this cleanup.

RadWolfie:
> Why rename `CxbxKrnlCleanup(Ex)` to `CxbxrKrnlAbort(Ex)`? It should just be `CxbxrAbort(Ex)` in my opinion, because we use this function everywhere in the code base to terminate the emulation when a fatal error is encountered, not just to terminate the kernel.

In case you notice other naming scenario, there's `CxbxKrnlPanic`, `CxbxKrnlSuspend`, `CxbxKrnlResume`, `CxbxKrnlShutDown`, etc. Imo, I think it's better to follow this identical way until they're renamed as well. To something like `CxbxrEmu` or `Emu` prefix.

> Not related to the above: do we really need the cxbxr prefix in our functions? After all, this is Cxbx-Reloaded already and so, adding the cxbxr prefix feels pretty much superfluous to me.

There's dxbx, cxbx, and cxbxr prefixes. They're there to tell the difference like where they came from. Beside, I'm following what we already have setup other than use cxbxr than cxbx.

ergo720:
Reminder for the community that this needs some testing before merging.

literalmente-game:
Verified against almost 30 games here, not a single change

medievil1:
not found any issues here either

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -517,6 +517,70 @@ bool HandleFirstLaunch()
...
+			char buffer[16];
+			if (GetConsoleTitle(buffer, 16) != NULL)
+				(void)freopen("nul", "w", stdout);
+		}
+	}
+	return nullptr;
+}
+
+/*! initialize emulation */
+static __declspec(noreturn) void CxbxrKrnlInit(
```
ergo720, line 1009: `__declspec(noreturn)` can be replaced with the C++ attribute `[[noreturn]]`

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -1261,77 +1334,21 @@ static __declspec(noreturn) void CxbxrKrnlInit
...
-		CxbxCdrom0DeviceIndex = CxbxRegisterDeviceHostPath(DeviceCdrom0, p_default_mount_path);
-		// Since Chihiro also map Mbfs to the same path as Cdrom0, we'll map it the same way.
-		if (g_bIsChihiro) {
-			(void)CxbxRegisterDeviceHostPath(DriveMbfs, p_default_mount_path);
-		}
-	}
 
 	CxbxrKrnlRegisterDevicePaths();
 
+	bool isEmuDisk = CxbxrKrnlSetupCdRom0Path(BootFlags, lastFind, relative_path);
```
ergo720, line 1346: `isEmuDisk` is useless because it's not used by anything else inside the function.

src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -853,6 +853,79 @@ static void CxbxrKrnlXbePatchXBEHSig() {
...
+#else
+	// HACK: It is a hack to override XDK's default mount to CdRom0 which may not exist when launch to dashboard directly.
+	// Otherwise, titles may launch to dashboard, more specifically xbox live title, and back.
+	if (CxbxCdrom0DeviceIndex == -1 || lastFind != std::string::npos) {
+#endif
+		CxbxCreateSymbolicLink(DriveD, relative_path.string());
+	}
+}
+
+static bool CxbxrKrnlSetupCdRom0Path(int BootFlags, int lastFind, std::filesystem::path relative_path)
```
ergo720, line 869: The original code that `CxbxrKrnlSetupCdRom0Path` is replacing was only using `isEmuDisk` internally. Thus, I believe that returning the variable is unnecessary.

src/core/kernel/init/CxbxKrnl.cpp
RadWolfie: I don't remember why I had it returned. Might be due to an older revision I had made to compress the changes.
Good finding. 👍

src/core/kernel/init/CxbxKrnl.cpp
RadWolfie: 👍
I'll also update `CxbxrKrnlAbortEx` and any relative code in new commit.

src/core/kernel/init/CxbxKrnl.cpp
RadWolfie: Actually, it is used. About 13 lines down.

ergo720: In that case you can add a call to `CxbxrIsPathInsideEmuDisk` before `CxbxrKrnlSetupCdRom0Path`, and pass `isEmuDisk` as an argument. This way, the call to `CxbxrIsPathInsideEmuDisk` from inside `CxbxrKrnlSetupCdRom0Path` can be removed as well.

RadWolfie: done

src/core/kernel/init/CxbxKrnl.cpp
RadWolfie: done, it wouldn't compile after `static`. So I had to put it before `static`.

RadWolfie:
Two review remarks was made in private discussion which has been submit in this pull request.

- `Relocate src/core/kernel/support/EmuFilePaths.hpp` to `src/common/FilePaths.hpp`
- Rename `CxbxKrnlCleanup(Ex)` to `CxbxKrnlAbort(Ex)` in separate commit.

ergo720 (approved):

PR #2281, opened 2021-09-14
Fix flicker in antialias sample

NZJenkins:
Improve X_D3DSWAP_COPY behaviour

As I understand, the antialias and backbufferscale samples are supposed to render their UI to the frontbuffer, so they don't have to worry about the backbuffer changing resolution (actual or effective).
They call Swap with X_D3DSWAP_COPY before they draw the UI
And when done, they call swap with X_D3DSWAP_FINISH

We would present on every swap call, so presenting before the frame was finished. The UI would flicker since it was only rendered to the frame 50% of the time

This PR
* Skips presenting when swap is called with D3DSWAP_COPY, and let the title continue rendering
* Disables multisampling and backbufferscale scale factors which should behave similarly to rendering to the frontbuffer (not sure this will always be accurate?)

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -183,6 +183,8 @@ static xbox::X_D3DBaseTexture        CxbxActiveTextureCopies[xbox::X_D3DTS_STAGE
 xbox::X_D3DVIEWPORT8 g_Xbox_Viewport = { 0 };
 float g_Xbox_BackbufferScaleX = 1;
 float g_Xbox_BackbufferScaleY = 1;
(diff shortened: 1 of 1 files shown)


LukeUsher:
Any chance of getting a LOG_TEST_CASE specifically on X_D3DSWAP_COPY? that way we can track which titles use this functionality.

literalmente-game:
Tested all the games known to have "screen flickering" issues, none from the list changed:

- MotoGP 2
- MotoGP 3
- Total Immersion Racing
- Without Warning
- The Suffering
- The Suffering: Ties that Bind
- World Series Baseball 2K3
- The Bible Game
- Freaky Flyers
- The Matrix  - Path of Neo
- Enter the Matrix
- Test Drive - Eve of Destruction
- Whiplash

PR #2282, opened 2021-10-03
Fixed invalid handle exception in freopen

ergo720:
This uses `DETACHED_PROCESS` as the `dwCreationFlags` argument of `CreateProcess`, which cxbxr uses when launching a new emulation process. Without this flag, the default behavior would be for the new process to inherit the console of the parent process, which is wrong since we want the emulation process to have its own console allocated with `AllocConsole` instead. This apparently fixes the invalid handle exception that used to always happen in `freopen` during initialization in this [line](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/ac68fd481cf6c4640f6986ce15174b413329fb2c/src/core/kernel/init/CxbxKrnl.cpp#L529).


PatrickvL:
Could you copy-paste this PR description into a comment above the change itself? This, because it's valuable information that people won't know about if they only read the code (pull request descriptions are great to have, but not a source of information that's referenced frequently)

PatrickvL (commented):
Good find!

PatrickvL (approved):

PR #2284, opened 2021-10-07
Tweak Swap behaviour to fix flickering in some titles

NZJenkins:
This is https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2281, but with a hack to fix MotoGP2, which uses the `BYPASSCOPY` flag.

I'm not sure what MotoGP2 was trying to do, so the fix for it might not help with other titles...

For `D3DSWAP_COPY`, don't present, and adjust scale factors to behave as if we are drawing to the frontbuffer.
For `D3DSWAP_BYPASSCOPY`, don't draw anything and hope for the best

@literalmente-game has a list of flickering titles that could be affected: https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2281#issuecomment-920495799


medievil1:
didn't really have any that flickered, one thing I do greatly notice though, game speeds are far, far better....games that showed 30fps but still felt sluggish now feel great...  indiana jones for example, early on you come to a water stage, it always felt stuttery and fps would drop some , now it is butter smooth... wreckless, though most of the screen is black, before it was pretty slow drawing, now it is normal fast, you can actually use the map insert and drive the track at normal speed

PatrickvL (approved):
Looks fine. Any fall-out should be addressed in a later PR

PR #2289, opened 2021-10-30
Added Libusb as input subsystem for passthrough support for xbox cont…

faha223:
…rollers

Added libusb as git submodule

Added LibusbDevice, LibusbXboxController and SteelBattalionController classes. LibusbXboxController and SteelBattalionController both inherit from LibusbDevice and InputDevice.

SteelBattalionController exposes all functionality of the SteelBattalionController for mappings as regular inputs (axes and buttons) but when used the hardware device for an emulated Steel Battalion Controller, no mappings are needed (or used). LEDs are working.

LibusbXboxController behaves the same as the SteelBattalionController but for Xbox Controller and Xbox Controller S, as well as many third party controllers. It is implemented in a way that should enable the use of multiple xbox controllers, but I don't have enough usb breakaway cables to test this.

Rumble is working.

Controllers connected via libusb will reconnect if the breakaway cable become disconnected and then reconnected. Emulated games do not notice the disconnection though.


ergo720:
I'm currently redoing this work locally because, other than the libusb submodule (which is fine), there are major issues with the rest of the code of this PR, but this will take a while. However, I don't have the real SBC to test my changes, so I will need someone to test my code when I'm done.

medievil1:
> 
> 
> I'm currently redoing this work locally because, other than the libusb submodule (which is fine), there are major issues with the rest of the code of this PR, but this will take a while. However, I don't have the real SBC to test my changes, so I will need someone to test my code when I'm done.

it currently works well with my og s controller, pressure sensitive buttons and rumble alike.... no SNC though so can't remark on that

faha223:
It looks like I checked in some changes I hadn't meant to be in the PR relating to toggle buttons on the Steel Battalion Controller. This would probably break keyboard support or other controllers with Line of Contact.

faha223:
> I'm currently redoing this work locally because, other than the libusb submodule (which is fine), there are major issues with the rest of the code of this PR, but this will take a while. However, I don't have the real SBC to test my changes, so I will need someone to test my code when I'm done.

I'm eager to see what you do differently so I can do a better job in the future

LukeUsher:
> > I'm currently redoing this work locally because, other than the libusb submodule (which is fine), there are major issues with the rest of the code of this PR, but this will take a while. However, I don't have the real SBC to test my changes, so I will need someone to test my code when I'm done.
> 
> I'm eager to see what you do differently so I can do a better job in the future

A lot of the work was unnecessary I'm afraid/this implementation is overly complicated; please study our PR when it is available.

There was not the need to do as many changes to the input/mapping systems, and it doesn't need this many changes to the input manager.

Fun fact: you can memcpy the input/output buffer from hardware within XInputGet/SetState directly, because those functions use the exact same memory layout as the real thing.

ergo720:
Superseded by https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2290

PR #2290, opened 2021-11-03
Add support for original xbox gamepads and SBC hardware via USB passthrough with libusb

ergo720:
Mostly the same as https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2289, but with better code. ~~It's a draft because I don't have a SBC to test, so I have no idea if this actually works or it has bugs and/or crashes.~~


medievil1:
seems fine except rumble... either both left and right motors are active when testing the right or the left is just dead(worked in the other incarnation)
gamepad test should activate them separately and did before

ergo720:
@medievil1 I think I found the problem and added a workaround for it. Can you retest to see if it helps?

faha223:
I'll take a look at this with my SBC later to verify that it works

ergo720:
@faha223 A note about the SBC. We are currently using a subtype of 0x2 for the virtual SBC, and that's wrong according to the xboxdevwiki which reports a subtype of 0x1 instead. We can know the correct value by looking at the bSubType member of the xid descriptor of the controller, which is dumped to the kernel log when you select the "libusb" logging option at info level. So, can you also report that value as well please?

faha223:
I'm seeing this in the logfile:
[0x1DFC] INFO : LIBUSB  Xid descriptor dump:
bLength: 0X00000010
bDescriptorType: 0X00000042
bcdXid: 0X00000100
bType: 0X00000080
bSubType: 0X00000001
bMaxInputReportSize: 0X0000001A
bMaxOutputReportSize: 0X00000016
wAlternateProductIds[0]: 0X0000FFFF
wAlternateProductIds[1]: 0X0000FFFF
wAlternateProductIds[2]: 0X0000FFFF
wAlternateProductIds[3]: 0X0000FFFF

So it looks like a real Steel Battalion Controller will have subtype of 0x01.

Inputs seem to all work, but LEDs don't. I had expected either they'd work perfectly, or they'd be off by 2 bytes but they're not working at all.

faha223:
Also, rumble doesn't work on any of my OEM Xbox controllers.
VID 0x045E and PIDs 0x0202 (Duke), 0x0285 (Controller S), and 0x0289 (Controller S). I'm going to try building from source as it's possible I downloaded an artifact from before the rumble fix

ergo720:
@faha223 Yes, I discovered the rumble pointer I was passing was off by two bytes. I also disabled the rumble workaround I added in the previous commit. If it still doesn't work, I'll add it back.

faha223:
Rumble works now. SBC LEDs still don't.

If I switch from libusb_control_transfer to libusb_interrupt_transfer or libusb_bulk_transfer when writing the data to the device then it works fine. It seems that either the parameters for libusb_control_transfer are wrong for the SBC out endpoint, or libusb_control_transfer is the wrong function to use to write to the SBC out endpoint.

medievil1:
Rumble working correctly now.....

faha223:
SBC LEDs are working perfectly now

medievil1:
rumble still good to go as well

ergo720:
I switched to interrupt transfers directly instead of control transfer, in an attempt to simplify `ExecuteIo`. Let's see if this still works, otherwise I'll revert it. Another thing that I don't know is whether or not this works out of the box or it requires the installation of additional drivers/libraries for USB passthrough to actually work. If it's the latter, then the additional requirements should be added to the README of the repository.

faha223:
I installed the WinUSB driver using Zadig to get it working, since libusb on Windows is a wrapper around WinUSB. @medievil1 said he used xb2xinput and that worked too.

faha223:
My Xbox Controller S still works (including Rumble), and my SBC still works (including LEDs)

medievil1:
s controller still works fine (rumble too) via the xb2xinput driver method..which is just libusb pretty much...lol

src/gui/resource/ResCxbx.h
```diff
@@ -94,6 +95,7 @@
 #define IDC_LOG_VSHCACHE                962
 #define IDC_LOG_RINP                    963
 #define IDC_LOG_JVS                     964
+#define IDC_LOG_LIBUSB                  965
```
RadWolfie, line 98: Not sure why this is here. But should be safe to leave it here, maybe.

src/common/input/LibusbDevice.cpp
```diff
@@ -0,0 +1,316 @@
...
+// ******************************************************************
+
+#define LOG_PREFIX CXBXR_MODULE::LIBUSB
+
+#include "LibusbDevice.h"
+#include "InputManager.h"
+#include "core\kernel\support\Emu.h"
+#include "core\hle\XAPI\Xapi.h"
+
+// Sanitiy check: ensure out libusb version is high enough for libusb_get_device_descriptor to succeed and to pass nullptr to libusb_interrupt_transfer
```
RadWolfie, line 35: typo:
```suggestion
// Sanity check: ensure out libusb version is high enough for libusb_get_device_descriptor to succeed and to pass nullptr to libusb_interrupt_transfer
```

RadWolfie (commented):
I made some local modification to see how it react to unfiltered usb devices. It cause a crash which I didn't investigate where as it's not that important at this time.

So, since my Xbox 360 Wireless Controller meet WinUSB compliant driver support. I went with non-standard PC controller to see if at least one title able to detect my inputs. Turns out only one button have response whilst other buttons has no response. However, this was only a test to see how xbox title's may response to "emulated" controller type for non-xbox controller.

The code's functionality looks good to me. Just have some nits here and there.

src/gui/resource/Cxbx.rc
```diff
@@ -201,6 +202,15 @@ BEGIN
     PUSHBUTTON      "Test",IDC_RUMBLE_TEST,95,11,45,14,BS_FLAT
 END
 
+IDD_LIBUSB_CFG DIALOGEX 0, 0, 230, 35
+STYLE DS_SETFONT | DS_MODALFRAME | DS_CENTER | WS_POPUP | WS_CAPTION | WS_SYSMENU
+FONT 8, "Verdana", 0, 0, 0x1
+BEGIN
+    COMBOBOX        IDC_LIBUSB_LIST,15,11,100,15,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    PUSHBUTTON      "Refresh",IDC_REFRESH_DEVICES,120,11,45,14,BS_FLAT
+    PUSHBUTTON      "Test",IDC_LIBUSB_TEST,170,11,45,14,BS_FLAT
```
RadWolfie, line 211: Before I forget, could we have drop down menu to be a bit longer for better readability? Both refresh and test buttons can be move below dropdown menu.

src/common/input/LibusbDevice.cpp
PatrickvL: You probably meant to replace "out" by "our"?

src/gui/resource/Cxbx.rc
ergo720: I can increase the size of the dropdown menu, but putting the other buttons below it is ugly in my opinion as they will be hidden when you open the dropdown. Also, in the xbox controller and the SBC window the refresh button is to the right of the dropdown, so it should still be there for consistency.

RadWolfie: It may look a bit ugly. However, dropdown menu's list isn't easy to read for whole texts. If we were using ImGui method, then it would be fine. It would be nicer to have Win32's GUI to have ability to see whole text from dropdown menu's list.

I think it's better to revisit this issue in the future to resolve it.

src/common/input/LibusbDevice.cpp
RadWolfie: Good catch, I updated suggested change code.

RadWolfie (approved):
Nit review remarks are done.

Base on two reports, I believe it is safe to merge.

PR #2292, opened 2021-11-14
Implement HalEnableSecureTrayEject xbox kernel function.

Fisherman166:
I found this old branch on my hard drive. The function is fully implemented but I never was able to figure out how the flag variable works. I figure something is better than nothing.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlHal.cpp
+++ b/src/core/kernel/exports/EmuKrnlHal.cpp
@@ -828,6 +828,7 @@ XBSYSAPI EXPORTNUM(365) xbox::void_xt NTAPI xbox::HalEnableSecureTrayEject
 		xbox::ulong_xt disable_reset_on_eject = 0;
 		NTSTATUS retcode;
 		do {
+            // TODO: Implement SMC_COMMAND_RESET_ON_EJECT in the SMC.
 			retcode = HalWriteSMBusValue(SMBUS_ADDRESS_SYSTEM_MICRO_CONTROLLER, SMC_COMMAND_RESET_ON_EJECT, write_word, disable_reset_on_eject);
 		} while (retcode != xbox::status_success);
 	}
--- a/src/core/kernel/exports/EmuKrnlHal.cpp
+++ b/src/core/kernel/exports/EmuKrnlHal.cpp
@@ -805,6 +805,9 @@ XBSYSAPI EXPORTNUM(360) xbox::ntstatus_xt NTAPI xbox::HalInitiateShutdown
 	RETURN(S_OK);
 }
 
+// Something else in the kernel sets this flag. Not sure what.
+static xbox::boolean_xt g_SecureTrayEjectAllowed = false;
+
 // ******************************************************************
 // * 0x016D - HalEnableSecureTrayEject()
 // ******************************************************************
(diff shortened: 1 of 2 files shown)


ergo720:
`HalEnableSecureTrayEject` is used to inform the console that it should reboot when it detects a tray opening event. But cxbxr doesn't emulate the dvd drive at this time, so tray transactions are impossible for the time being.

src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -819,7 +822,15 @@ XBSYSAPI EXPORTNUM(365) xbox::void_xt NTAPI xbox::HalEnableSecureTrayEject
 {
 	LOG_FUNC();
 
-	LOG_UNIMPLEMENTED();
+	if (g_SecureTrayEjectAllowed) {
+		g_SecureTrayEjectAllowed = false;
+		xbox::boolean_xt write_word = false;
+		xbox::ulong_xt disable_reset_on_eject = 0;
```
ergo720, line 828: You can just pass the constants to `HalWriteSMBusValue` instead of declaring these variables here.

Fisherman166: I know. But I like to give names to magic numbers. Makes the code easier to read.

Fisherman166:
> 
> 
> `HalEnableSecureTrayEject` is used to inform the console that it should reboot when it detects a tray opening event. But cxbxr doesn't emulate the dvd drive at this time, so tray transactions are impossible for the time being.

That's what I thought at first too. But the comment for the function specifies that it **disables** reset on eject. The more I thought about it, the more writing a 0 to enable a feature didn't make sense unless they were using inverted logic.

src/devices/SMCDevice.h
```diff
@@ -70,7 +70,7 @@
 #define SMC_COMMAND_LED_SEQUENCE 0x08	// LED flashing sequence
 //0x0C	tray eject(0 = eject; 1 = load)
 //0x0E	another scratch register ? seems like an error code.
-//0x19	reset on eject(0 = enable; 1 = disable)
+#define SMC_COMMAND_RESET_ON_EJECT 0x19	// Reset on eject (0 = disable, 1 = enable)
```
ergo720, line 73: Other devs have asked me if you can add a TODO comment here, to explain that the smc device still does not implement this command and thus the command that `HalEnableSecureTrayEject` sends will have no effect in practice.

src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -819,7 +822,15 @@ XBSYSAPI EXPORTNUM(365) xbox::void_xt NTAPI xbox::HalEnableSecureTrayEject
...
+		NTSTATUS retcode;
+		do {
+			retcode = HalWriteSMBusValue(SMBUS_ADDRESS_SYSTEM_MICRO_CONTROLLER, SMC_COMMAND_RESET_ON_EJECT, write_word, disable_reset_on_eject);
```
RadWolfie, line 831: Can you add TODO comment above this line state `SMC_COMMAND_RESET_ON_EJECT` is not implemented?

Doing so will allow us to merge this pull request. Even though it's not been tested. Eventually, we will have tray commands support.

Fisherman166: Done.

RadWolfie (approved):

PR #2294, opened 2021-11-20
Added support for the lightgun input device

ergo720:
This adds support for a new input device, the lightgun, specifically the EMS TopGun II documented in the xboxdevwiki. The laser pointer of the gun is also emulated as a circle drawn as an Imgui overlay over the rendering window, so that players can know where they are shooting at on the rendering window. Note that this will only work with xbox games that use xinput to control the lightgun, and thus it won't work with chihiro games that use JVS input instead.

Known xbox games that support lightguns are:
- Starsky & Hutch
- The House of the Dead III (xbox version, also contain The House of the Dead II)
-  Silent Scope Complete (contain 3 titles).


LanHikariDS:
Tested against GunValkyrie and Steel Battalion, two titles that had controller issues that were fixed. Neither title shows input-based regressions in this PR.

LukeUsher (approved):

src/core/kernel/exports/EmuKrnlHal.cpp
```diff
@@ -535,6 +535,8 @@ XBSYSAPI EXPORTNUM(49) xbox::void_xt DECLSPEC_NORETURN NTAPI xbox::HalReturnToFi
 				#endif
 			}
 
+			// Save lightgun laser status
+			g_EmuShared->SetLightgunLaser(&g_devs->info.ligthgun.laser);
```
RadWolfie, line 539: Not a review remark, usually we save them directly whenever state has change. Although, I do understand the reason to place it here as EmuShared doesn't have kernel function type to sync anything that is not real-time update.

nit: Doing this way will only remember port 1's laser state. Therefore will produce a bug if users are using port 2 - 4 instead. (Such as myself usually do testing pull request's relative input changes on port 3 and 4.) If we are going to merge as-is with this change. I recommended to create ticket for a reminder to fix this.

Also, can we expand InputManager to perform lightgun's saving state? Such as verify if the device is lightgun then save it.

---

EDIT: Actually, I notice there's a union for SBC and lightgun. So, this change is not right.

src/core/common/imgui/ui.cpp
```diff
@@ -15,6 +15,13 @@
 
 #include "core/kernel/init/CxbxKrnl.h"
 
+const ImColor ImGuiUI::m_laser_col[4] = {
+		ImColor(ImVec4(1.0f, 0.0f, 0.0f, 1.0f)), // ply1: red
+		ImColor(ImVec4(0.0f, 1.0f, 0.0f, 1.0f)), // ply2: green
+		ImColor(ImVec4(0.0f, 0.0f, 1.0f, 1.0f)), // ply3: blue
+		ImColor(ImVec4(1.0f, 1.0f, 0.0f, 1.0f))  // ply4: yellow
```
RadWolfie, line 22: nit: use full name instead of ply.

src/common/win32/EmuShared.h
```diff
@@ -364,6 +370,7 @@ class EmuShared : public Mutex
 #else
 		unsigned int m_Reserved;
 #endif
+		uint8_t      m_LightgunLaser;
```
RadWolfie, line 373: nit: This state can stored only one lightgun device instead of up to 4 lightgun devices base on total ports available on hardware.

Plus House of the Dead III's cover art state it can support up to 2 players. Same thing with Starsky & Hutch title, except one being driver and other being shooter (lightgun).

src/common/input/InputDevice.h
```diff
@@ -122,7 +122,8 @@ class InputDevice
 	// retrieves the port this device is attached to
 	bool GetPort(std::string_view Port) const;
 	// sets the port this device is attached to
-	void SetPort(std::string_view Port, bool Connect);
+	// NOTE: using SetPort2 to avoid a collision with the SetPort macro provided by Windows headers
```
PatrickvL, line 125: Good catch, but it might be better to `#undef SetPort` instead of avoiding the name?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -194,6 +194,11 @@ static void CxbxImGui_RenderD3D9(ImGuiUI* m_imgui, IDirect3DSurface9* renderTarg
 
 	m_imgui->DrawMenu();
```
PatrickvL, line 195: Much of this code is repeated in pgraph, perhaps it's possible to move both copies into a single source file (of that allows reuse of the same code)?

src/devices/video/EmuNV2A_PGRAPH.cpp
```diff
@@ -713,6 +715,11 @@ static void CxbxImGui_RenderOpenGL(ImGuiUI* m_imgui, std::nullptr_t unused)
 
 	m_imgui->DrawMenu();
```
PatrickvL, line 716: Much of this code also appears in Direct3D.cpp, consider moving this to a shared location

RadWolfie (changes requested):

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -787,6 +804,88 @@ xbox::dword_xt WINAPI xbox::EMUPATCH(XInputSetState)
...
+	if (g_devs[port].info.hHandle == hDevice && !g_devs[port].bPendingRemoval) {
+		if (g_devs[port].type == XBOX_INPUT_DEVICE::LIGHTGUN) {
+			g_devs[port].info.ligthgun.offset_x = pCalibrationOffsets->wCenterX;
+			g_devs[port].info.ligthgun.offset_y = pCalibrationOffsets->wCenterY;
+			g_devs[port].info.ligthgun.offset_upp_x = pCalibrationOffsets->wUpperLeftX;
+			g_devs[port].info.ligthgun.offset_upp_y = pCalibrationOffsets->wUpperLeftY;
+			ret = ERROR_SUCCESS;
+		}
+		else {
+			ret = ERROR_NOT_SUPPORTED;
```
PatrickvL, line 880: Does Phantasy star online also call this on the keyboard device? In that case a similar note as above might be appropriate

PatrickvL (approved):
Only a few small remarks, and admittedly I didn't review very thoroughly, but this looks fine

src/core/hle/XAPI/Xapi.cpp
RadWolfie: `XInputSetLightgunCalibration` function doesn't exist in PSO title since there's no lightgun support for it.

src/common/input/InputDevice.h
ergo720: That `undef` was already used in `InputManager.h` , but I decided to remove it in favor of changing function name, just like we do for `LOG_LEVEL::ERROR2` + even with the `undef`, when switching to this file, IntelliSense will still report this as a macro anyway.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
ergo720: Maybe, however I fell like that should be done in a separate PR instead of here.

src/core/kernel/exports/EmuKrnlHal.cpp
ergo720: > Not a review remark, usually we save them directly whenever state has change.

There's no need to save this variable whenever it changes, because its state is already saved in `g_devs[port].info.ligthgun.laser`. I only need to store it in EmuShared so that it survives the reboot. And yes, it should indeed check if a lightgun is connected and the state should be saved on a per-port basis.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Yeah, sure. Now that you've nicely moved the `DrawLightgunLaser` calls into `DrawWidgets`, this refactoring suggestion no longer applies to this PR anymore anyway.

src/devices/video/EmuNV2A_PGRAPH.cpp
PatrickvL: Above remark no longer applies.

PatrickvL (approved):

RadWolfie (approved):
LTGM

PR #2297, opened 2021-12-12
Simulate correctly the USB enumeration process in XGetDevices

ergo720:
While `XInitDevices` is initializing the USB drivers, `XGetDevices` should not report any devices attached because the drivers have yet to enumerate them, while our existing implementation was reporting them right away. Implementing this behavior correctly allows Star Wars: Jedi Knight: Jedi Academy to successfully boot instead of crashing at startup.


RadWolfie (approved):
Base on code changes and comments, looks perfect to me and proper procedure.

literalmente-game:
Star Wars Clone Wars also improved from this change, can progress to the menus but still crashing trying to reach in-game in any mode

PR #2298, opened 2021-12-20
Sync XbSymbolDatabase + Add Missing Patch

RadWolfie:
Some optimization for signatures and resolve most errors from self-test OOVPA function. Since changelog is quite big, you can view them from cxbx-reloaded/XbSymbolDatabase#156 pull request.

After checking on existing patches, I notice we are missing one patch. I made two separate commits which will generate two artifact builds to either see the difference or use latter artifact build and modify symbol cache's files to comment out `D3DDevice_GetBackBuffer2_0__LTCG_eax1`.


RadWolfie:
Haven't heard back from tester for list of reports. as far as I'm aware, there are regressions yet some improvement.

LukeUsher (approved):

PR #2302, opened 2021-12-27
Allow ObIniSystem to succeed

ergo720:
This fixes `ObIniSystem` and thus allows it to properly initialize the ob manager. In particular, this extends `NtClose`, `NtCreateDirectoryObject`, `NtOpenDirectoryObject` and `NtDuplicateObject` to support ob handles, in addition to native handles. It also implements some extra internal obp functions of the ob manager. This should not cause any regressions/improvements, but it paves the way to use proper ob handles instead of Windows handles, and will allow to properly implement more kernel functions and avoid having to forward to their Window counterparts, thus reducing the dependency on Windows APIs.


RadWolfie (approved):

RadWolfie:
Can confirm titles are not making any regressions.

PR #2306, opened 2022-01-09
Fixup sampling cube textures with PROJECT2D

NZJenkins:
Use CUBEMAP sampling instead
Fixes lighting in Splinter Cell Chaos Theory

I didn't find anything about what should happen in around these kinds of cases. In OGL it is apparently undefined behaviour.
Developers can set a shader flag to tell XD3D to set the sampling mode automatically, but chaos theory apparently rolled its own shaders and don't set the flag.

Sc:Chaos Theory uses cubemaps extensively for lighting - looks like a normalization cube map

before:
cc-before

after:
cc-after


LukeUsher (approved):

PatrickvL:
This improved the road in Outrun 2 as well : 
before : 
image

after : 
image

PR #2308, opened 2022-01-14
labeler.yml: Consistently use quotation marks

Margen67:


.github/labeler.yml
```diff
@@ -1,74 +1,74 @@
...
 
 timing:
-  - src/common/Timer*
+  - 'src/common/Timer*'
 
 user interface:
-  - src/gui/**
-# - TODO: Dear Imgui location, maybe in src/core/gui path.
+  - 'src/gui/**'
+# - TODO: Dear Imgui location, maybe in src/core/gui.
```
RadWolfie, line 71:
```suggestion
  - 'src/core/common/imgui/*'
  - 'src/gui/**'
```

Since we have imgui in our repository. Could you update to include it? I made a suggestion change above for it.

RadWolfie (approved):

PR #2309, opened 2022-01-15
Update to new issue forms

Margen67:


ergo720:
Have you checked that the auto-close bot still works with this updated template?

.github/ISSUE_TEMPLATE/issue.yaml
```diff
@@ -0,0 +1,46 @@
...
+      required: true
+  - type: textarea
+    id: details
+    attributes:
+      label: Details
+      description: Please describe the problem as accurately as possible.
+    validations:
+      required: true
+  - type: textarea
+    id: system-config
```
ergo720, line 35: In my opinion, the system configuration can just be omitted. This, because the hardware of the user is almost never pertinent to the issue. The only issue we had were it was relevant was the AMD graphics regression, but the user's hardware can always be specified in the additional-info paragraph when it's necessary.

RadWolfie: Plus System Config is optional. We couldn't make this optional in our issue template without breaking the header size.

RadWolfie:
According to GitHub's issue forms, it is currently in beta stage. We do not accept any beta features from GitHub until it is out of beta stage. Same concept with GitHub Actions in the past. For now, I will mark this pull request as draft.

ergo720:
I'm changing this from draft to a regular PR, since now the new issue forms are no longer in beta stage but in public preview. However, Github says that they are still subjected to change, so it's probably better not to merge this yet.

PR #2311, opened 2022-01-18
Fix Outrun 2006 performance issue

NZJenkins:
Outrun 2006 lags entirely due to converting large ADPCM buffers when certain sound effects occur.
This change attempts to convert the minimum range of bytes required
This is probably broken

### DIFF
--- a/src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
+++ b/src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
@@ -1594,52 +1594,59 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_StopEx)
             hRet = pThis->EmuDirectSoundBuffer8->Stop();
             pThis->Xb_rtStopEx = 0LL;
         }
-        else {
-            bool isLooping;
-            if ((pThis->EmuPlayFlags & X_DSBPLAY_LOOPING) > 0) {
-                isLooping = true;
-            }
-            else {
-                isLooping = false;
-            }
+        else if(dwFlags & X_DSBSTOPEX_ENVELOPE) {
+			bool isLooping = pThis->EmuPlayFlags & X_DSBPLAY_LOOPING;
 
-            if ((dwFlags & X_DSBSTOPEX_ENVELOPE) > 0) {
-                if (rtTimeStamp == 0LL) {
-                    xbox::LARGE_INTEGER getTime;
-                    xbox::KeQuerySystemTime(&getTime);
-                    pThis->Xb_rtStopEx = getTime.QuadPart;
-                }
-                else {
-                    pThis->Xb_rtStopEx = rtTimeStamp;
-                }
-                pThis->Xb_rtStopEx += (pThis->Xb_EnvolopeDesc.dwRelease * 512) / 48000;
+			DWORD releaseSamples = pThis->Xb_EnvolopeDesc.dwRelease * 512.0;
+
+            if (rtTimeStamp == 0LL) {
+                xbox::LARGE_INTEGER getTime;
+                xbox::KeQuerySystemTime(&getTime);
+                pThis->Xb_rtStopEx = getTime.QuadPart;
             }
             else {
                 pThis->Xb_rtStopEx = rtTimeStamp;
             }
+			const double samplesToTicks = 10000000 / 48000.0;
+			xbox::REFERENCE_TIME releaseTicks = releaseSamples * samplesToTicks;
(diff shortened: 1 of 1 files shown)


src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
```diff
@@ -1594,52 +1594,59 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_StopEx)
...
+
+            if (rtTimeStamp == 0LL) {
+                xbox::LARGE_INTEGER getTime;
+                xbox::KeQuerySystemTime(&getTime);
+                pThis->Xb_rtStopEx = getTime.QuadPart;
             }
             else {
                 pThis->Xb_rtStopEx = rtTimeStamp;
             }
+			const double samplesToTicks = 10000000 / 48000.0;
```
RadWolfie, line 1610: TODO: This may not be entirely correct. Can you explain where you get `10000000` value from?

src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
```diff
@@ -1594,52 +1594,59 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_StopEx)
...
+
+				if (isLooping) {
+					// Play through the loop
+					bufferRangeStart += pThis->EmuRegionLoopStartOffset;
+					bufferRangeEnd = bufferRangeStart + pThis->EmuRegionLoopLength;
+				}
+				else {
+					// Play out until the end of the release phase
+					DWORD releaseSamples = pThis->Xb_EnvolopeDesc.dwRelease * 512.0;
+					DWORD releaseBytes = releaseSamples / 2; // Assuming 4 bit per sample ADPCM (?)
```
RadWolfie, line 1640: TODO: Hardcode `2` value guess isn't allow. I will check if there's a suitable replacement for it.

src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
```diff
@@ -1650,6 +1657,9 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_StopEx)
                 pThis->Xb_rtStopEx = 0LL;
             }
         }
+		else {
+			LOG_TEST_CASE("Expected DSBSTOPEX_ENVELOPE");
```
RadWolfie, line 1661:
```suggestion
			LOG_TEST_CASE("Expected X_DSBSTOPEX_ENVELOPE");
```

src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
```diff
@@ -1594,52 +1594,59 @@ xbox::hresult_xt WINAPI xbox::EMUPATCH(IDirectSoundBuffer_StopEx)
...
-                }
-
+				// Determine the range of bytes we need to play
+				// Test case: Outrun 2006 - converting large buffers tanks the FPS
+				DWORD bufferRangeStart = pThis->EmuRegionPlayStartOffset;
+				DWORD bufferRangeEnd;
+
+				if (isLooping) {
+					// Play through the loop
+					bufferRangeStart += pThis->EmuRegionLoopStartOffset;
```
RadWolfie, line 1634: TODO: If we are going to add Loop's start offset, then we need to shift our relative position toward our range start position.

RadWolfie (commented):
Beside those TODOs, they need to be submitted into worker thread's query system as the time stamp will be be bind to Xb_rtStopEx trigger the event. So, even the upstream code is wrong as well. It will need another rework to implement into `Xb_rtStopEx`'s query system.

src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
NZJenkins: Convert from seconds to 100 nanosecond ticks

src/core/hle/DSOUND/DirectSound/DirectSoundBuffer.cpp
NZJenkins: Is that the same thing as this step? `DWORD newPos = currentPos + bufferRangeStart`

LanHikariDS:
Just a note, this PR seems to make Halo 2 Multiplayer Map Pack crash with an unhandled exception when wheeling through the menus

PR #2315, opened 2022-01-27
Updated Ps kernel functions to use Ob handles + unpatch thread XAPI functions + added APC support to kernel via Ob

ergo720:
This PR rewrites the kernel thread code to make use of Ob handles, instead of native ones. It also properly implements PsQueryStatistics and enables the execution of thread notification routines. These changes make it possible to unpatch SetThreadPriority, GetThreadPriority, SetThreadPriorityBoost, GetExitCodeThread, XSetProcessQuantumLength, timeSetEvent and timeKillEvent. Finally, this updates NtQueueApcThread to deliver APCs with the kernel, instead of just forwarding to ntdll, which also allows the removal of the thread handle duplication hack that was in place for it before.

~~One thing I notice is that this seems to apparently break the dashboard for me, which now goes to the fatal error screen instead of booting. However Radwolfie is not able to reproduce the issue, so this needs testing from the community.~~

As a side effect, this PR also fixes the stuttering of the dashboard and the slowness of the title screens of Panzer Dragoon Orta currently present in master.


src/core/kernel/support/NativeHandle.cpp
```diff
@@ -51,8 +69,29 @@ void RemoveXboxHandle(xbox::HANDLE xhandle)
 	assert(ret == 1);
 }
 
+template<bool NoConversion>
```
PatrickvL, line 72: For clarity, consider inverting this boolean into `DoConversion` (obviously, also invert the default argument declaration and all calls)

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -51,8 +69,29 @@ void RemoveXboxHandle(xbox::HANDLE xhandle)
...
 std::optional<HANDLE> GetNativeHandle(xbox::HANDLE xhandle)
 {
+	// If SourceHandle is -2 = NtCurrentThread, then we are searching the handle of this thread
+	// Test case: Metal Slug 3
+	if (xhandle == NtCurrentThread()) {
```
PatrickvL, line 77: This special case only applies to thread-handles, which introduces a chance for a false positive match on non-thread handles. How could we avoid this? Perhaps via an additional template boolean `ForThreads`?

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -51,8 +69,29 @@ void RemoveXboxHandle(xbox::HANDLE xhandle)
...
+		}
+		else {
+			xhandle = xbox::PspGetCurrentThread()->UniqueThread;
+		}
+	}
+	// If xhandle is not special handle, check if it's the same as current thread.
+	// Only used for threads as Windows doesn't allow non-special handle for same thread.
+	// This will only triggered within kernel functions i.e. KeSetDisableBoostThread and KeSetBasePriorityThread.
+	else if (NoConversion) {
+		if (xhandle == xbox::PspGetCurrentThread()->UniqueThread) {
```
PatrickvL, line 90: Here a false-positive match could also occur for non-thread handles...

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -222,58 +235,74 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+		result = ObInsertObject(eThread, zeroptr, 0, &eThread->UniqueThread);
+		if (!X_NT_SUCCESS(result)) {
+			ObfDereferenceObject(eThread);
+			RETURN(result);
+		}
+
+		// Call thread notification routine(s)
+		if (g_iThreadNotificationCount) {
+			for (int i = 0; i < PSP_MAX_CREATE_THREAD_NOTIFY; i++) {
+				if (g_pfnThreadNotification[i]) {
```
PatrickvL, line 256: Should access to `g_pfnThreadNotification` be made thread-safe / locked?

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -222,58 +235,74 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+		// Create another handle to pass back to the title in the ThreadHandle argument
+		result = ObOpenObjectByPointer(eThread, &PsThreadObjectType, ThreadHandle);
+		if (!X_NT_SUCCESS(result)) {
+			ObfDereferenceObject(eThread);
+			RETURN(result);
+		}
+
+		if (ThreadId != zeroptr) {
+			*ThreadId = eThread->UniqueThread;
+		}
```
PatrickvL, line 272: Are we sure that the original xbox kernel doesn't set this variable until after above check? Or should it be set before that check?

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -222,58 +235,74 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
 			RETURN(X_STATUS_INSUFFICIENT_RESOURCES);
 		}
-		*ThreadHandle = handle;
-        if (ThreadId != NULL)
-            *ThreadId = dwThreadId;
+
+		KeQuerySystemTime(&eThread->CreateTime);
+		KiUniqueProcess.StackCount++;
+		RegisterXboxHandle(*ThreadHandle, handle);
+		RegisterXboxHandle(eThread->UniqueThread, handle);
```
PatrickvL, line 294: Surely this needs to register `ThreadId` ?

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -222,58 +235,74 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+
+		// Call thread notification routine(s)
+		if (g_iThreadNotificationCount) {
+			for (int i = 0; i < PSP_MAX_CREATE_THREAD_NOTIFY; i++) {
+				if (g_pfnThreadNotification[i]) {
+					EmuLog(LOG_LEVEL::DEBUG, "Calling pfnNotificationRoutine[%d] (0x%.8X)", i, g_pfnThreadNotification[i]);
+					(*g_pfnThreadNotification[i])(eThread, eThread->UniqueThread, TRUE);
+				}
+			}
+		}
```
PatrickvL, line 261: Consider extracting this calling of thread notification routines into a separate method, named `PspCallThreadNotificationRoutines` or something like that, as this code appears in both `PsCreateSystemThreadEx` and `PsTerminateSystemThread` (just pass in the `Create` boolean as an argument, which is the only difference between the two current implementations).

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -312,29 +340,15 @@ XBSYSAPI EXPORTNUM(257) xbox::ntstatus_xt NTAPI xbox::PsSetCreateThreadNotifyRou
...
-	for (int i = 0; i < PSP_MAX_CREATE_THREAD_NOTIFY; i++)
-	{
-		// If we find one, then add it to the array, and break the loop so
-		// that we don't accidently register the same routine twice!
-		if (g_pfnThreadNotification[i] == NULL)
-		{
-			g_pfnThreadNotification[i] = (PVOID)NotifyRoutine;
+	for (int i = 0; i < PSP_MAX_CREATE_THREAD_NOTIFY; i++) {
+		if (g_pfnThreadNotification[i] == zeroptr) {
+			g_pfnThreadNotification[i] = NotifyRoutine;
```
PatrickvL, line 345: Should access to `g_pfnThreadNotification` be made thread-safe / locked?

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -350,34 +364,27 @@ XBSYSAPI EXPORTNUM(258) xbox::void_xt NTAPI xbox::PsTerminateSystemThread
...
+	xbox::PETHREAD eThread = xbox::PspGetCurrentThread();
+	if (eThread->UniqueThread && g_iThreadNotificationCount) {
+		for (int i = 0; i < PSP_MAX_CREATE_THREAD_NOTIFY; i++) {
+			if (g_pfnThreadNotification[i]) {
+				EmuLog(LOG_LEVEL::DEBUG, "Calling pfnNotificationRoutine[%d] (0x%.8X)", i, g_pfnThreadNotification[i]);
+				(*g_pfnThreadNotification[i])(eThread, eThread->UniqueThread, FALSE);
+			}
 		}
-	}*/
+	}
```
PatrickvL, line 376: Consider extracting this calling of thread notification routines into a separate method, named `PspCallThreadNotificationRoutines` or something like that, as this code appears in both `PsCreateSystemThreadEx` and `PsTerminateSystemThread` (just pass in the `Create` boolean as an argument, which is the only difference between the two current implementations).

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -350,34 +364,27 @@ XBSYSAPI EXPORTNUM(258) xbox::void_xt NTAPI xbox::PsTerminateSystemThread
...
+
+	EmuKeFreeThread(ExitStatus);
```
PatrickvL, line 378: Is this called for "cxbxr threads that execute xbox code"? Does that apply here? Please elaborate on this call in a comment.

src/core/kernel/support/EmuFS.cpp
```diff
@@ -188,51 +187,72 @@ void EmuKeSetPcr(xbox::KPCR *Pcr)
...
+	xbox::PKPCR Pcr = EmuKeGetPcr();
+	
+	xbox::PVOID Dummy;
+	xbox::ulong_xt Size;
+	xbox::ntstatus_xt Status;
+	// tls can be nullptr
+	if (Pcr->NtTib.StackBase) {
+		// NOTE: the tls pointer was increased by 12 bytes to enforce the 16 bytes alignment, so adjust it to reach the correct pointer
+		// that was allocated by xbox::NtAllocateVirtualMemory
+		Dummy = static_cast<xbox::PBYTE>(Pcr->NtTib.StackBase) - 12;
```
PatrickvL, line 201: Granted, this code was merely reindented, but please consider replacing the magic number 12 by a descriptively named constant, and use that everywhere it applies to.

RadWolfie:
@NZJenkins, eventually all of handles will be managed by Ob's handle manager. In my opinion, we should accept any form of regressions from handles without add/re-implement hacks to safeguard Windows' handles. It may take about couple weeks to a month worth to resolve regressions relative to handles.

src/core/kernel/support/EmuFS.cpp
```diff
@@ -188,51 +187,72 @@ void EmuKeSetPcr(xbox::KPCR *Pcr)
...
+	xbox::KeQuerySystemTime(&eThread->ExitTime);
+	eThread->Tcb.HasTerminated = 1;
+
+	// Emulate our exit strategy for GetExitCodeThread
+	eThread->ExitStatus = ExitStatus;
+	eThread->Tcb.Header.SignalState = 1;
+
+	xbox::HANDLE UniqueThread = eThread->UniqueThread;
+	if (GetNativeHandle(eThread->UniqueThread)) {
+		xbox::NtClose(eThread->UniqueThread);
```
PatrickvL, line 237: No need to dereference the variable anymore, now that it resides in the local `UniqueThread` variable (likely compiled into a register value) - so use that instead (unless threading requires re-reading the source variable each time?)

src/core/kernel/support/EmuFS.cpp
```diff
@@ -671,16 +691,28 @@ void EmuGenerateFS(Xbe::TLS *pTLS, void *pTLSData)
...
-		xbox::RtlZeroMemory(EThread, sizeof(xbox::ETHREAD));
+		// If it is nullptr, it means we are creating this thread from cxbxr, otherwise we were created from PsCreateSystemThreadEx
+		xbox::PETHREAD EThread = static_cast<xbox::PETHREAD>(Ethread);
+		if (EThread == xbox::zeroptr) {
+			// Since this a host thread that we use to execute xbox code, we do not need to create ob handles for this thread
+			base = xbox::zeroptr;
+			size = sizeof(xbox::ETHREAD);
+			xbox::NtAllocateVirtualMemory(&base, 0, &size, XBOX_MEM_RESERVE | XBOX_MEM_COMMIT, XBOX_PAGE_READWRITE);
+			EThread = (xbox::ETHREAD *)base; // Clear, to prevent side-effects on random contents
+			xbox::RtlZeroMemory(EThread, sizeof(xbox::ETHREAD));
```
PatrickvL, line 702: The end-of-line comment belongs either on a separate line, or behind the actual zeroing-call.

src/core/kernel/support/EmuFS.cpp
```diff
@@ -671,16 +691,28 @@ void EmuGenerateFS(Xbe::TLS *pTLS, void *pTLSData)
...
+			EThread->UniqueThread = reinterpret_cast<xbox::HANDLE>(GetCurrentThreadId());
```
PatrickvL, line 703: Could this assignment cause confusion on further use of `EThread->UniqueThread`?

PatrickvL, line 703: Also, is `GetCurrentThreadId()` a native ThreadId? Or should it be the `-2` from `#define NtCurrentThread() ((HANDLE)-2)` ?

src/core/kernel/support/EmuFS.cpp
```diff
@@ -671,16 +691,28 @@ void EmuGenerateFS(Xbe::TLS *pTLS, void *pTLSData)
...
+		}
 
 		EThread->Tcb.TlsData = pNewTLS;
-		EThread->UniqueThread = GetCurrentThreadId();
```
PatrickvL, line 681: Am I right to understand that when `EThread` is non-null, this `UniqueThread` field is already set by `PsCreateSystemThreadEx` to the "ob handle of the ethread obj"? This might be a valuable comment...

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -39,9 +40,26 @@ std::shared_mutex g_MapMtx;
...
+		auto timeout = now + std::chrono::milliseconds(2000);
+		while (now <= timeout) {
+			lck.unlock();
+			std::this_thread::sleep_for(std::chrono::milliseconds(5));
+			lck.lock();
+			g_RegisteredHandles.try_emplace(xhandle, nhandle);
+			if (ret.second) {
+				return;
+			}
+		}
```
PatrickvL, line 57: When does this case actually happen? When there's a collision between an obsolete handle still being present and a new identical handle being added?? This might need some more comments to become clear.

PatrickvL (changes requested):
Tried to be thorough on this review - although there's only one actual bug (or so I assume)

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -1148,214 +978,6 @@ xbox::LPVOID WINAPI xbox::EMUPATCH(ConvertThreadToFiber)
...
-	PLAUNCH_DATA	pLaunchData
-)
-{
-	// Note : This can be tested using "Innocent tears",
-	// which relaunches different xbes between scenes;
-	// One for menus, one for fmvs, etc.
-	//
-	// Other titles do this too (like "DOA2 Ultimate",
-	// and probably "Panzer Dragoon Orta"), but these
-	// titles don't come this far as-of yet.
```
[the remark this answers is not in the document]
RadWolfie: No need for test, XLaunchNewImageA had been unpatched since 2017.

src/core/hle/XAPI/Xapi.cpp
```diff
@@ -1148,214 +978,6 @@ xbox::LPVOID WINAPI xbox::EMUPATCH(ConvertThreadToFiber)
...
-		{
-			// If no path is specified, then the xbe is rebooting to dashboard
-			char szDashboardPath[xbox::max_path] = { 0 };
-			XboxDevice* rootDevice = CxbxDeviceByDevicePath(DeviceHarddisk0Partition2);
-			if (rootDevice != nullptr)
-				sprintf(szDashboardPath, "%s\\xboxdash.xbe", rootDevice->HostDevicePath.c_str());
-
-			if (PathFileExists(szDashboardPath))
-			{
-				PopupInfo(nullptr, "The title is rebooting to dashboard");
```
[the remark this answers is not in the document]
RadWolfie: Unused since 2017

src/core/hle/XAPI/Xapi.cpp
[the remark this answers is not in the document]
PatrickvL: Ah, it wasn't enclosed in an `#if 0`, so I didn't check if it actually appeared in the patches list.
Still can't hurt to see if these test cases still work...

src/core/kernel/common/ps.h
```diff
@@ -69,7 +68,9 @@ XBSYSAPI EXPORTNUM(257) ntstatus_xt NTAPI PsSetCreateThreadNotifyRoutine
 // ******************************************************************
 XBSYSAPI EXPORTNUM(258) void_xt NTAPI PsTerminateSystemThread(IN ntstatus_xt ExitStatus);
 
-XBSYSAPI EXPORTNUM(259) volatile OBJECT_TYPE PsThreadObjectType;
+XBSYSAPI EXPORTNUM(259) OBJECT_TYPE PsThreadObjectType;
```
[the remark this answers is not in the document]
ergo720: Because it's wrong, `PsThreadObjectType` does not have the volatile qualifier. https://github.com/XboxDev/nxdk/blob/587177545f309695f47f364c173baa1df5413879/lib/xboxkrnl/xboxkrnl.h#L2326

src/core/kernel/exports/EmuKrnlKe.cpp
```diff
@@ -128,24 +130,18 @@ xbox::ulonglong_xt LARGE_INTEGER2ULONGLONG(xbox::LARGE_INTEGER value)
...
-		// Attempt to salvage the situation by calling InitXboxThread to setup KPCR in place
-		InitXboxThread();
-		g_AffinityPolicy->SetAffinityXbox();
-		Pcr = (xbox::PKPCR)__readfsdword(TIB_ArbitraryDataSlot);
-	}
+	xbox::PKPCR Pcr = (xbox::PKPCR)__readfsdword(TIB_ArbitraryDataSlot);
+
+	// If this fails, it's a bug: it means we are executing xbox code from a host thread, and we have forgotten to initialize
+	// the xbox thread first
+	assert(Pcr);
```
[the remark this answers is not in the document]
ergo720: The comment above it explains it: if it ever happens, it's a bug and should be fixed instead of trying to handle it. Also note that this function is called very frequently in the thread code, so this check is probably costly.

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -719,18 +709,18 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
...
 		const ACCESS_MASK DesiredAccess = 0;
 		const ULONG Attributes = 0;
-		dword_xt nativeOptions = (Options | DUPLICATE_SAME_ATTRIBUTES | DUPLICATE_SAME_ACCESS);
+		const ULONG nativeOptions = (Options | DUPLICATE_SAME_ATTRIBUTES | DUPLICATE_SAME_ACCESS);
 
 		if (const auto &nativeHandle = GetNativeHandle(SourceHandle)) {
 			// This was a handle created by Ob
 			PVOID Object;
-			status = ObReferenceObjectByHandle(SourceHandle, /*ObjectType=*/nullptr, &Object);
+			status = ObReferenceObjectByHandle(SourceHandle, zeroptr, &Object);
```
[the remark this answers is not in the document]
RadWolfie: It's unnecessary as our Ob's system are starting to use more of `ObReferenceObjectByHandle`. If unsure what 2nd parameter is, using function's list of references will tell you what it is or at least go in the function's definition to tell easier.

src/core/hle/XAPI/Xapi.cpp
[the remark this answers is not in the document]
PatrickvL: Again, I didn't verify whether `XLaunchNewImageA` still appeared in the list of registered patches. Even so, the remark about this popup still stands - we could still make it reappear in `HalReturnToFirmware` if we want to have some feedback on a reboot (perhaps only in debug runs, but still)...

src/core/kernel/support/NativeHandle.cpp
RadWolfie: Windows doesn't accept non-speical handle for current thread. Therefore require to preserve the special handle. This is the only occur in threading situation. Everything else is handled by xbox's Ob handle.

src/core/kernel/support/NativeHandle.cpp
RadWolfie: It is not false positive... each handle are unqiue there are no same handle values given at any time.

src/core/kernel/common/ps.h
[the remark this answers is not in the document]
PatrickvL: Okay, if that reference is accurate, then fine. Just wanted to make sure 😉

src/core/kernel/exports/EmuKrnlKe.cpp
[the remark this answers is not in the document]
PatrickvL: If you're sure the 'salvage' path was never relied upon, then fine. It's just that there's no mention of test-cases, so we can't verify.

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: No, because that can always be retrieved later at any time with `GetCurrentThreadId`. In addition, registering the thread handle allows this code to work `GetNativeHandle(PspGetCurrentThread()->UniqueThread)`, so we can always retrieve the native handle of the current thread anywhere in the code, instead of the thread id which is not useful since the majority of win api use the thread handle instead of the id.

src/core/kernel/support/EmuFS.cpp
ergo720: Native thread id.

src/core/kernel/exports/EmuKrnlNt.cpp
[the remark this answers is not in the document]
PatrickvL: Mentioning the argument name when the value is non-descriptive, has been a coding convention - I was just wondering why it was removed. And indeed, a lookup of arguments is always possible, but that requires interaction in an editor, hindering the ease of reading, while the comment was right there already.

src/core/kernel/support/NativeHandle.cpp
PatrickvL: Perhaps @ergo720 wants to respond. `GetNativeHandle` is an emulator function, not a WinAPI or xbox function.

src/core/kernel/support/NativeHandle.cpp
PatrickvL: I'm not entirely sure, but it looks like multiple threads might have `UniqueThread` set to -2, which no longer makes it a unique value. (I do admit there's little to no chance -2 or $FFFFFFFD will ever be returned legitimately.)

src/core/kernel/exports/EmuKrnlPs.cpp
RadWolfie: In my opinon, there should be two different handles. Using same `handle` is likely to trigger an issue for Windows making a close once unless `handle` from Windows has a 2nd increment in some way.

PatrickvL: Ah, indeed! That's clever, and should probably be documented in a comment!

src/core/kernel/exports/EmuKrnlNt.cpp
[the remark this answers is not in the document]
ergo720: Because it's useless imo, otherwise, we should have to add it whenever `ObReferenceObjectByHandle` is called, which will happen more and more in the code base as we switch to Ob handles.

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -719,18 +709,18 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
...
 			if (X_NT_SUCCESS(status)) {
 				if (ObpIsFlagSet(Options, DUPLICATE_CLOSE_SOURCE)) {
 					NtClose(SourceHandle);
 				}
 
-				status = ObOpenObjectByPointer(Object, OBJECT_TO_OBJECT_HEADER(Object)->Type, /*OUT*/TargetHandle);
+				status = ObOpenObjectByPointer(Object, OBJECT_TO_OBJECT_HEADER(Object)->Type, TargetHandle);
```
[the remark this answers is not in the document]
ergo720: Because it's useless imo, otherwise, we should have to add it whenever `ObOpenObjectByPointer` is called, which will happen more and more in the code base as we switch to Ob handles.

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -1039,54 +1029,27 @@ XBSYSAPI EXPORTNUM(206) xbox::ntstatus_xt NTAPI xbox::NtQueueApcThread
 		LOG_FUNC_ARG(ApcReserved)
 		LOG_FUNC_END;
 
-	// In order for NtQueueApcThread or QueueUserAPC to work, you must... I repeat...
-	// YOU MUST duplicate the handle with the appropriate permissions first!  So far,
-	// the only game that I know of using this is Metal Slug 3, and it won't launch
-	// without it.  Other SNK games might use it also, beware.
```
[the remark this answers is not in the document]
ergo720: There is only a single test case, Metal Slug 3, which calls `NtQueueApcThread`, that I'm aware off.

src/core/kernel/support/NativeHandle.cpp
ergo720: `UniqueThread` is an xbox handle handed out by Ob, it cannot be a non-thread handle.

RadWolfie: > but it looks like multiple threads might have UniqueThread set to -2

If that's the case, then it is a bug. However, I had not seen any place where UniqueThread used for -2 when I was working with @ergo720.

src/core/kernel/support/NativeHandle.cpp
ergo720: I saw it happening in the dashboard, because of the reason explained in the comment.

RadWolfie: pinging @ergo720 for ^

src/core/hle/XAPI/Xapi.cpp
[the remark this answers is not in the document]
RadWolfie: See pull request https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/721, they had been working ever since. ;P

src/core/kernel/support/EmuFS.cpp
RadWolfie: Honestly, this is how kernel works and there is a check for `UniqueThread` which is happening I believe in XAPI's function.

src/core/kernel/support/NativeHandle.cpp
ergo720: Well, except for the case this is an xbox thread converted from a host thread, that is, not created by the title with `PsCreateSystemThread`. In that case, `UniqueThread` will be the value returned by `GetCurrentThreadId`, so a native handle. @PatrickvL I strongly doubt the ob manager will ever hand out a -2 = 0xFFFFFFFE handle. It starts at zero and increases by 4 at every new handle it creates.

src/core/kernel/exports/EmuKrnl.cpp
```diff
@@ -119,12 +122,6 @@ xbox::PLIST_ENTRY RemoveTailList(xbox::PLIST_ENTRY pListHead)
 	return Result;
 }
 
-// ******************************************************************
-// * Declaring this in a header causes errors with xboxkrnl
-// * namespace, so we must declare it within any file that uses it
-// ******************************************************************
-xbox::KPCR* WINAPI KeGetPcr();
```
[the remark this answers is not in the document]
ergo720: The original problem was because of WINAPI, which is only declared in the windows headers and thus becomes undefined when those are not included. Using NTAPI fixes it because we define that macro ourselves in our xbox headers. I did the renaming because similar functions in EmuFs are prefixed with Emu. Plus this function really only works with the emulator, since it's accessing the fs segment at the arbitrary slot, which is not how it works on hardware.

src/core/kernel/exports/EmuKrnlKi.cpp
```diff
@@ -876,3 +880,53 @@ xbox::void_xt FASTCALL xbox::KiWaitSatisfyAll
...
+			xbox::KiApcListMtx.unlock();
+			return;
+		}
+		xbox::PLIST_ENTRY Entry = kThread->ApcState.ApcListHead[ApcMode].Flink;
+		xbox::PKAPC Apc = CONTAINING_RECORD(Entry, xbox::KAPC, ApcListEntry);
+		RemoveEntryList(Entry);
+		Apc->Inserted = FALSE;
+		xbox::KiApcListMtx.unlock();
+
+		// NOTE: we never use KernelRoutine
```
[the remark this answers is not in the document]
ergo720: Because kernel APCs are only used internally by the kernel, while `NtQueueApcThread` (the only functions available to titles to use APCs), always creates a user APC. Since we never use kernel APCs in our kernel at the moment, there's no reason to use `KernelRoutine` at this time.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2037,27 +2033,10 @@ static LRESULT WINAPI EmuMsgProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPar
...
                 case SIZE_MINIMIZED:
                 {
                     if(g_XBVideo.bFullScreen)
                         CxbxrKrnlAbort(nullptr);
-
-                    if(!g_bEmuSuspended)
-                    {
-                        bAutoPaused = true;
-                        CxbxKrnlSuspend();
-                    }
```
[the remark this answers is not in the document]
ergo720: Well, this is dead code because of the reasons explained in https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2217. So, we are not losing functionality because this doesn't work anyway.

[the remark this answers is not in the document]
RadWolfie: In my opinion, we are still receving eary crash on reboot process due to the fact xbox threads are still running. We should terminate them before we shut down our emulation process. Doing so will prevent any false positive reports.

src/core/kernel/support/EmuFS.cpp
ergo720: Yes, this is documented in a comment I added inside `EmuGenerateFS`

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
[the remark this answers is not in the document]
ergo720: Those crashes are only visible with the debugger, because they only occur in the split of a second after `TerminateProcess` is called and the code still runs. Since the process is going to be killed right now, users will never see the crash anyway.

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: No, the kernel uses the thread notification routines at passive level with no locks acquired, so we shouldn't either.

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: This is just checking if the title wants to have the thread id. If it doesn't, `ThreadId` will be nullptr and setting it unconditionally will cause a crash in that case. Also, `eThread->UniqueThread` is set above with the call to `ObInsertObject` when successful, and if it fails, we won't even reach here.

src/core/kernel/support/EmuFS.cpp
ergo720: No, there's a comment above which explains this is for host threads used as xbox threads.

src/core/kernel/support/EmuFS.cpp
ergo720: Honestly, I'm not sure which name I should give it, this is just used to make the TLS have a 16 byte alignment, like the comment just above explains.

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: This must always be called for both kinds of threads, to free the common resources that have been allocated for the thread.

src/core/kernel/support/NativeHandle.cpp
ergo720: The normal (and most common) case for `GetNativeHandle`, is to do the conversion, while the special case is not to convert `NtCurrentThread` and return it as is. So, in my opinion, it's clearer the way it is now.

src/core/hle/XAPI/Xapi.cpp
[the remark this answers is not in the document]
PatrickvL: There's no mention of Innocent tears in #721, but I'll take your word for it.

src/core/kernel/exports/EmuKrnlKi.cpp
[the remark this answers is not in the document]
PatrickvL: Ah, understood. Thanks. Then best change the comment into "NOTE: we don't use KernelRoutine yet"

src/core/kernel/exports/EmuKrnlNt.cpp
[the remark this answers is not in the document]
PatrickvL: Alright. Let's agree to disagree on the usefulness of this convention then ;)

src/core/kernel/exports/EmuKrnlNt.cpp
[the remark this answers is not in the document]
PatrickvL: Okay, but the remark "Other SNK games might use it also" implies there might still be other test-cases, although I admit it's impossible to mention what's unknown until now. But at least mention "Metal Slug 3" as a "must-test title" in the Pull Request description.

src/core/kernel/support/NativeHandle.cpp
PatrickvL: Fine by me!

src/core/kernel/support/NativeHandle.cpp
PatrickvL: Okay, as long as `NtCurrentThread()` returns `-2`, and all handles that are returned to xbox code are derived from the memory addresses where our kernel allocated the kernel Ob'jects, it's indeed impossible for those values to match (as we'll never allocate an object at memory address -2).

src/core/kernel/support/NativeHandle.cpp
PatrickvL: Yeah, -2 won't ever be a valid Ob memory address.

src/core/kernel/exports/EmuKrnlPs.cpp
PatrickvL: Yeah, that I understand, but I was referring to the preceding check (on `if (!X_NT_SUCCESS(result)) {`) - perhaps the original xbox kernel already assigns `*ThreadID` *before* that check

src/core/kernel/exports/EmuKrnlPs.cpp
PatrickvL: Oh, that wasn't clear to me, because a comment in `EmuKeFreeThread` states : `This functions is to be used for cxbxr threads that execute xbox code`, which sentence I now understand has to be expanded with : ` and xbox threads created via PsCreateSystemThread(Ex)`

src/core/kernel/support/EmuFS.cpp
PatrickvL: Perhaps something like `NTTIB_STACKBASE_ALIGNMENT`? But whatever name we use, it's *having* a symbol name that's important, because once it appears everywhere our code currently has this magic 12, we can (A) easily find all places in code where it applies to this alignment, (B) we have a single declaration that can have a comment why the alignment is set to 12, and (C) other occurrences of the magic value 12 will no longer be misunderstood to have something to do with this alignment.

src/core/kernel/support/EmuFS.cpp
PatrickvL: Eh, how's "this is how kernel works" an answer to the suggestion? What kernel are you referring to? And why does it presumably read the same field three times in a row, while there's already a local variable where we read this field into? Again, I simply suggest to avoid repeated identical memory reads by using the local variable instead of the referenced field.

(Unless, as I stated above, it is a requirement because of some thread-safety concern, but in that case it would help to have a detailed comment explaining the potential problem that could arise if we don't re-read the field each time.)

src/core/kernel/support/EmuFS.cpp
PatrickvL: Oh, excuse me, I meant to ask : How high is the chance that this assignment of a host value, ever gets confused (and could thus potentially cause false positive matches), with the value we assigned to `UniqueThread` in `PsCreateSystemThreadEx` (via a call to `ObInsertObject`)?

PatrickvL: Okay, thanks. (This confirmation made me write a followup question in the preceding conversation.)

src/core/kernel/exports/EmuKrnlNt.cpp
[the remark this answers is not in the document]
PatrickvL: Alright. Let's agree to disagree on the usefulness of this convention then ;)

src/core/kernel/exports/EmuKrnlPs.cpp
PatrickvL: Ehm, isn't it possible that 'passive level' can get replaced by another level, as long as we don't manage thread-switching ourselves? Even worse, when the multi-core hack is in effect, xbox code could be running simultaneously on multiple cores, so we can't rely on the same safety guarantees as an original xbox kernel...

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: No, it's only done if `ObOpenObjectByPointer` succeeds (which is what the check above is for). If that is unsuccessful, it means that the thread creation failed, and thus there's no point in setting the thread id either.

PatrickvL: Great - if that's what the original xbox kernel also does, then this is fine

src/core/kernel/exports/EmuKrnlPs.cpp
PatrickvL: For thread safety, prepend `std::unique_lock lck(PspThreadNotificationMtx);` here (and probably, do a global rename of `PspThreadNotificationMtx` into `g_ThreadNotificationMtx`).

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -350,34 +368,24 @@ XBSYSAPI EXPORTNUM(258) xbox::void_xt NTAPI xbox::PsTerminateSystemThread
...
-			xbox::XTHREAD_NOTIFY_PROC pfnNotificationRoutine = (xbox::XTHREAD_NOTIFY_PROC)g_pfnThreadNotification[i];
-
-			// If the routine doesn't exist, don't execute it!
-			if (pfnNotificationRoutine == NULL)
-				continue;
-
-			EmuLog(LOG_LEVEL::DEBUG, "Calling pfnNotificationRoutine[%d] (0x%.8X)", g_iThreadNotificationCount, pfnNotificationRoutine);
+	// Call the thread notification routine(s)
+	xbox::PETHREAD eThread = xbox::PspGetCurrentThread();
+	if (eThread->UniqueThread && g_iThreadNotificationCount) {
```
PatrickvL, line 373: Please move the check on `g_iThreadNotificationCount` into `PspCallThreadNotificationRoutines` (and in there, right after the mutex is entered), so that will be thread-safe as well.

src/core/hle/XAPI/Xapi.cpp.unused-patches
```diff
@@ -0,0 +1,355 @@
...
+	LOG_FUNC_BEGIN
+		LOG_FUNC_ARG(pThreadNotification)
+		LOG_FUNC_ARG(fRegister)
+		LOG_FUNC_END;
+
+    if(fRegister)
+    {
+		// I honestly don't expect this to happen, but if it does...
+        if(g_iThreadNotificationCount >= 16)
+			CxbxrKrnlAbort("Too many thread notification routines installed\n");
```
PatrickvL, line 138: For thread safety, prepend `std::unique_lock lck(PspThreadNotificationMtx);` here (and probably, do a global rename of `PspThreadNotificationMtx` into `g_ThreadNotificationMtx`).

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -222,58 +236,77 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+		std::memset(eThread, 0, sizeof(ETHREAD) + ThreadExtensionSize);
+
+		// The ob handle of the ethread obj is the thread id we return to the title
+		result = ObInsertObject(eThread, zeroptr, 0, &eThread->UniqueThread);
+		if (!X_NT_SUCCESS(result)) {
+			ObfDereferenceObject(eThread);
+			RETURN(result);
+		}
+
+		// Call the thread notification routine(s)
```
PatrickvL, line 254: This comment can be removed (the function call itself already conveys the same information).

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -222,58 +236,77 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+		if (g_iThreadNotificationCount) {
+			PspCallThreadNotificationRoutines(eThread, TRUE);
+		}
```
PatrickvL, line 257: Please move the check on `g_iThreadNotificationCount` into `PspCallThreadNotificationRoutines` (and in there, right after the mutex is entered), so that will be thread-safe as well.

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -137,13 +134,30 @@ void PspSystemThreadStartup
...
+xbox::PETHREAD xbox::PspGetCurrentThread()
+{
+	// This works because we assign ethread to Prcb->CurrentThread
+	return reinterpret_cast<PETHREAD>(KeGetCurrentThread());
+}
+
+static xbox::void_xt PspCallThreadNotificationRoutines(xbox::PETHREAD eThread, xbox::boolean_xt Create)
+{
+	std::unique_lock lck(PspThreadNotificationMtx);
+	for (int i = 0; i < PSP_MAX_CREATE_THREAD_NOTIFY; i++) {
```
PatrickvL, line 146: Please move all remote checks on `g_iThreadNotificationCount` to here (between entering the mutex and performing the for loop).

src/core/kernel/exports/EmuKrnl.cpp
```diff
@@ -213,6 +210,51 @@ const DWORD IrqlMasks[] = {
...
+				(WaitMode == xbox::UserMode)) {
+				xbox::KiExecuteUserApc();
+				// Queue a native APC to the calling thread to forcefully terminate the wait of the WinApi functions,
+				// in the case it didn't terminate already
+				[[maybe_unused]] BOOL ret = QueueUserAPC(EndWait, *GetNativeHandle(eThread->UniqueThread), 0);
+				assert(ret);
+				EmuKeSetPcr(nullptr);
+				return true;
+			}
+			SwitchToThread();
```
PatrickvL, line 249: Please also call `SwitchToThread` instead of the `Sleep(0)`'s in `xbox::KeWaitForMultipleObjects` and `xbox::KeWaitForSingleObject`.

Note, that our use of `Sleep(0)` is always meant to force a thread-switch, and some of our use of `Sleep(1)` are meant to do that too. However, in selected cases, `Sleep(1)` (and higher constant numbers) it's used to just pause a little while. (There's even a comment in our code explaining this : `// TODO : Improve performance replacing Sleep(1) with YieldProcessor`)

PatrickvL (changes requested):
A partial review, after yesterdays new commits

src/core/hle/XAPI/Xapi.cpp.unused-patches
ergo720: Not sure if you have noticed, but this is the dead code of the unused xapi patches, and thus, it doesn't need any of that. In addition, this code is just wrong: it uses 16 notification routines instead of 8, and it mistakenly aborts the emulation when it exceeds the limit, when is should just return `STATUS_INSUFFICIENT_RESOURCES`

PatrickvL: Oh crap - this comment was meant for `PsSetCreateThreadNotifyRoutine` only

medievil1:
is this suppose to be functional? for me, every game closes back to cxbx ui

ergo720:
@medievil1 With the exception of the dashboard, all games work like before for me, so I'm not sure what your issue might be. One thing I notice however, is that performance dropped considerably compared to master. It now takes a couple of minutes to start any game, because the call to `_beginthreadex` seems to take much longer than it does on master. Also, stopping the emulation now takes a couple of seconds instead of being instantaneous, and that is because of the thread suspending code that was added in some commits. However, since it hurts the user's experience and there's little benefit in it, I'm keen to remove it in the future.

medievil1:
> @medievil1 With the exception of the dashboard, all games work like before for me, so I'm not sure what your issue might be. One thing I notice however, is that performance dropped considerably compared to master. It now takes a couple of minutes to start any game, because the call to `_beginthreadex` seems to take much longer than it does on master. Also, stopping the emulation now takes a couple of seconds instead of being instantaneous, and that is because of the thread suspending code that was added in some commits. However, since it hurts the user's experience and there's little benefit in it, I'm keen to remove it in the future.

very odd then cause I have tried multiple times, even new instance/folder instead of reading caps it craps out, no error, while master and other pr's run fine... I will pull the debug version and see if I can see more detail

RadWolfie:
> is this suppose to be functional? for me, every game closes back to cxbx ui

We recently became aware of this, two of us are able to reproduce the issue (sort of). Except we are not able to use debugger to pinpoint where silent crash occur or self terminated. We will need some extra time to figure out the cause. Thank you for bringing it to our attention. 🙂

Using loader or non-loader method has the same issue without debugger attached.

ergo720:
I can confirm that https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2315/commits/c6672f6fbcc5b21bc95d4c407e6e1de8a78d6fa8 fixes the excessive slowness at emulation shutdown, so now I'm fine with keeping the thread suspending code that was added.

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: I changed `g_iThreadNotificationCount` from `int` to `std::atomic_int`, so it's unnecessary now.

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: I changed `g_iThreadNotificationCount` from `int` to `std::atomic_int`, so it's unnecessary now.

src/core/kernel/exports/EmuKrnlPs.cpp
ergo720: I changed `g_iThreadNotificationCount` from `int` to `std::atomic_int`, so it's unnecessary now. 3/3

PatrickvL (dismissed):

PatrickvL (approved):

PR #2318, opened 2022-02-04
Fix kernel type hints for READ_PORT_BUFFER and friends

GXTX:
Testing this functions thanks to nxdk removing abstractions for port reading I noticed that it was crashing my Xbox, this is due to invalid type hints. Both CXBX-R and nxdk have varying problems (for example nxdk has `PULONG` for some `Count`s and not others).

This patch will fix this for CXBX-R.

Valid ports are from 0x0 - 0xFFFF so a uint16 will suffice.

### DIFF
--- a/src/core/kernel/common/hal.h
+++ b/src/core/kernel/common/hal.h
@@ -133,7 +133,7 @@ XBSYSAPI EXPORTNUM(50) ntstatus_xt NTAPI HalWriteSMBusValue
 // ******************************************************************
 XBSYSAPI EXPORTNUM(329) void_xt NTAPI READ_PORT_BUFFER_UCHAR
 (
-    IN PUCHAR Port,
+    IN word_xt Port,
     IN PUCHAR Buffer,
     IN ulong_xt  Count
 );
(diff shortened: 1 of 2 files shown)


PatrickvL:
Thanks! Could you mention a few testcases?

GXTX:
I'm not aware of any games which actually utilize these functions. I've messaged you some more info off platform.

PatrickvL (approved):
Looks good to me, merging as is

src/core/kernel/common/hal.h
```diff
@@ -133,7 +133,7 @@ XBSYSAPI EXPORTNUM(50) ntstatus_xt NTAPI HalWriteSMBusValue
 // ******************************************************************
 XBSYSAPI EXPORTNUM(329) void_xt NTAPI READ_PORT_BUFFER_UCHAR
 (
-    IN PUCHAR Port,
+    IN word_xt Port,
```
JayFoxRox, line 136: I'm not sure about the ABI here, but this might allow the compiler to only set the lower 16 bits of the input register, so this could lead to errors.

---

These signatures are for imports, so they should match whatever the export (~xboxkrnl.exe) uses, at least for argument size and signedness.

If you want a different signature, you should wrap these functions.

PR #2319, opened 2022-02-04
Move Crash Manager's Init at Beginning of Emulation & Fix Hidden Crash

### LINKED ISSUE #1882: Emulation Project with C++ Exception Doesn't Always Work
After thoroughly testing to find out what reason C++ exception doesn't work with loader's emulation project. It turns out C++ exception does work fine after add test try/catch after `Emulate` function. Plus GUI's internal emulation has the same issue as well.

However, without kernel console output. Emulation is working as intended.

---

Steps to reproduce:
1) Enable kernel console output
2) Run GTA 3 (best title to test with, since always will format the partition.)
3) Monitor which partition is formatted. (In my case it is partition 5.)
4) Stop emulation
5) Mark any files inside partition 5 as "read only".
6) Run GTA 3 again
7) Once get to "Calling XBE entry point..." from console window, silence crash occur or the error message Visual Studio caught below.
`Unhandled exception at 0x75544192 in cxbxr-ldr.exe: Microsoft C++ exception: std::filesystem::filesystem_error at memory location 0x0D23F56C.`

RadWolfie:
Without further ado, we are able to resolve #1882 outstanding issue. 🥳

Special thanks to:
* @medievil1 spot the issue and brought to our attention.
* @NZJenkins for narrow down to `EmuGenerateFS` before and and after `Direct3DCreate9Ex` call from #2315.
* @LukeUsher's suggestion try to comment out StackBase & StackLimit overwrites.
* More importantly @ergo720's #2315 pull request to reproduce the issue more easier.

With this change, I can confirm crash popup window is showing up properly. Plus Dead or Alive 3 demo title is working as intended.

---
**EDIT:** Even with force crash on format partition from GTA 3 title is also working as well. With a bonus, kernel console opened with exception is working as well.

### DIFF
--- a/src/core/kernel/init/CxbxKrnl.cpp
+++ b/src/core/kernel/init/CxbxKrnl.cpp
@@ -1017,6 +1017,10 @@ static bool CxbxrKrnlPrepareXbeMap()
 
 void CxbxKrnlEmulate(unsigned int reserved_systems, blocks_reserved_t blocks_reserved)
 {
+	// This is beginning of emulation process start, therefore require to have exception manager initialized
+	// and capture any crash from this point and beyond. Useful for capture live crash and generate crash report.
+	g_ExceptionManager = new ExceptionManager();
+
 	// First of all, check if the EmuShared version matches the emu version and abort otherwise
 	char GitVersionEmuShared[GitVersionMaxLength];
 	g_EmuShared->GetGitVersion(GitVersionEmuShared);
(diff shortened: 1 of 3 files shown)


ergo720:
This causes Amped to crash when it attempts to use the TLS, which is now no longer properly setup in this PR.

RadWolfie:
Base on few instructions back, we are missing patch for `mov edx,dword ptr fs:4`. With even further check to verify any other FS patches missing is `mov esi,dword ptr fs:4`. Instead of handpicked registers, I went ahead fill in the rest of missing registers for FS's offset 0x00 and 0x04.

Even though some FS patches were missing. It's more important to bundle with this pull request to avoid introduce any regression.

ergo720:
I can confirm that 7d116628c2c271b05f6cec9bbf98705ecf838716 fixes the problem in Amped.

LukeUsher (approved):

PR #2325, opened 2022-03-01
hle: detect when patches fail to apply

LukeUsher:
Although steps could be taken to make this more obvious, I don't want to make a developer feature harm user experience, which is why this is logged rather than quitting emulation.

Previously, failed patches would be totally ignored and we would have no way to detect this,

### DIFF
--- a/src/core/hle/Patches.cpp
+++ b/src/core/hle/Patches.cpp
@@ -438,8 +438,14 @@ inline void EmuInstallPatch(const std::string FunctionName, const xbox::addr_xt 
         return;
     }
 
-	g_FunctionHooks[FunctionName].Install((void*)(FunctionAddr), (void*)patch.patchFunc);
-	printf("HLE: %s Patched\n", FunctionName.c_str());
+	auto result = g_FunctionHooks[FunctionName].Install((void*)(FunctionAddr), (void*)patch.patchFunc);
+	if (!result) {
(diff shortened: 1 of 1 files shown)


src/core/hle/Patches.cpp
```diff
@@ -438,8 +438,14 @@ inline void EmuInstallPatch(const std::string FunctionName, const xbox::addr_xt
...
 
-	g_FunctionHooks[FunctionName].Install((void*)(FunctionAddr), (void*)patch.patchFunc);
-	printf("HLE: %s Patched\n", FunctionName.c_str());
+	auto result = g_FunctionHooks[FunctionName].Install((void*)(FunctionAddr), (void*)patch.patchFunc);
+	if (!result) {
+		printf("HLE: %s Patch Failed\n", FunctionName.c_str());
+	}
+	else {
+		printf("HLE: %s Patched\n", FunctionName.c_str());
+	}
```
RadWolfie, line 447: My advice to change this into...
```suggestion
	auto success = g_FunctionHooks[FunctionName].Install((void*)(FunctionAddr), (void*)patch.patchFunc);
	if (success) {
		printf("HLE: %s Patched\n", FunctionName.c_str());
	}
	else {
		printf("HLE: %s Patch Failed\n", FunctionName.c_str());
	}
```
This way we don't need `not` operator and to make the wording clear.

RadWolfie (approved):

PR #2337, opened 2022-03-28
Fixed wrong type for MmGlobalData kernel export

ergo720:
Triggered by this [PR](https://github.com/XboxDev/nxdk/pull/560) in nxdk.

### DIFF
--- a/src/core/kernel/common/mm.h
+++ b/src/core/kernel/common/mm.h
@@ -17,7 +17,102 @@
 namespace xbox
 {
 
-XBSYSAPI EXPORTNUM(102) PVOID MmGlobalData[8];
+typedef ulong_xt PFN;
+typedef ulong_xt PFN_COUNT;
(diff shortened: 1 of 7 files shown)


PatrickvL:
Is something planning to tie the memory manager to these fields?

ergo720:
No, and it can't be done for most of them since the internal implementation details of our memory manager differ than what the xbox kernel would do. For example, there's no Vad tree and so we can't set the Vad variables of the global.

PatrickvL:
Some homebrew or other software might read these fields and make decisions based on them, so for these cases populating the relevant fields might improve accuracy

ergo720:
The new changes in the memory manager files are simply because I had to add the `xbox::` prefix everywhere because the mm type declarations have been moved inside the xbox namespace, nothing changed code-logic wise.

RadWolfie (approved):

PR #2326, opened 2022-03-02
DirectSoundBuffer streaming

NZJenkins:
Unlike in DirectSound on Windows, Xbox titles can write to audio buffers at any time - with no need to call `lock` APIs to notify XDirectSound
We currently patch `lock`/`unlock`, so on `unlock` we can obtain updates to audio buffers.

But since Xbox titles don't _need_ to ever call `unlock`, there are plenty of titles that don't, and we don't know when the audio buffer has been updated.

This can manifest as a lack of audio, or the first part of audio looping repeatedly.

This PR continuously streams chunks of audio data from audio buffers, keeping just ahead of the audio that is currently playing, so audio is kept up to date.

Improvements:
* NBA Live 2005 (music / various sounds)
* Crash Tag Team Racing (Broken audio loops)
* Madagascar (Broken audio loops)
* Outrun 2006 (Performance related to fix by @RadWolfie, related PR #2311)
* Spyro: a Hero's Tail (missing audio)
* PoP: Sands of Time (missing music)
* PoP: Warrior Within (missing audio)
* Sphinx and the Cursed Mummy  (Broken audio loops)
* Buffy: Chaos Bleeds  (Broken audio loops)
* Need for Speed: Hot Pursuit 2 (missing audio)
* 007 Nightfire (missing audio)
* From Russia with Love (missing audio)
* CoD 3 (crash fix?)
* Max Payne 2 (Broken audio loops)
* Just Cause (unable to lock buffer)
* Robotech: Battlecry (broken audio loops)

Regressions:
* ~~Halo (intro is crackly) (intro latency)~~
* ~~DoA3 (doubled up drum beat entering the menu)~~
* Yourself!Fitness (crash)
* ~~Max Payne 2 (crash due to SetCurrentPosition broken)~~
* Probably lots of things

PR #2339, opened 2022-03-30
Improve Xbox Thread Setup

RadWolfie:
When I started implement `RtlWalkFrameChain` function, I noticed the result return empty. After further investigation, I recently discovered that TlsData reside on the same stack instead of separate memory allocation. With that said, I also found out how we used the xbox's top kernel stack reserved space improperly.

With this pull request fix, I am able to use host's local stack to generate local variable dynamic sized depending on TlsDataSize request and use it for specifically xbox's top kernel stack reserved. Everything else after it are used for cpu instruction requests as normal. And removed the hack method alltogether as well.

I also add `ENABLE_KTHREAD_SWITCHING` macro which works for both method. Except we don't have kthread switching supported.

P.S. `RtlWalkFrameChain` function implemented locally is functional.

---
I had personally verified with DoA 2 and Futurama titles are working as intended.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlPs.cpp
+++ b/src/core/kernel/exports/EmuKrnlPs.cpp
@@ -31,6 +31,7 @@
 
 
 #include <core\kernel\exports\xboxkrnl.h> // For PsCreateSystemThreadEx, etc.
+#include "EmuKrnlPs.hpp"
 #include "core\kernel\exports\EmuKrnlKi.h"
 #include "core\kernel\exports\EmuKrnlKe.h"
 #include <process.h> // For __beginthreadex(), etc.
@@ -55,10 +56,8 @@ namespace NtDll
 // PsCreateSystemThread proxy parameters
 typedef struct _PCSTProxyParam
 {
-	IN xbox::PVOID  StartRoutine;
-	IN xbox::PVOID  StartContext;
-	IN xbox::PVOID  SystemRoutine;
 	IN xbox::PVOID  Ethread;
+	IN xbox::ulong_xt TlsDataSize;
 }
 PCSTProxyParam;
 
@@ -72,7 +71,8 @@ void LOG_PCSTProxy
 	xbox::PVOID StartRoutine,
 	xbox::PVOID StartContext,
 	xbox::PVOID SystemRoutine,
-	xbox::PVOID Ethread
+	xbox::PVOID Ethread,
+	xbox::ulong_xt TlsDataSize
 )
 {
 	LOG_FUNC_BEGIN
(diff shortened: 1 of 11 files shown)


medievil1:
seems to fix gta San Andres (it is locking for other reasons and audio has been absent in game for a while) it no longer slows to a crawl on the first cut scene loading

ergo720:
@medievil1 That sounds odd, as this PR merely initializes stack related members of the `KTHREAD` object and the TLS. Are you it wasn't instead fixed by the [dsound PR](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2326) that was merged yesterday?

medievil1:
> @medievil1 That sounds odd, as this PR merely initializes stack related members of the `KTHREAD` object and the TLS. Are you sure it wasn't instead fixed by the [dsound PR](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2326) that was merged yesterday?

I'll test it and get back, but it hadn't been last I checked the pr before merge, I will check master too

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -244,6 +319,17 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+		// Create kernel stack for xbox title to able write on stack instead of host.
+		PVOID KernelStack = MmCreateKernelStack(KernelStackSize, DebuggerThread);
+
+		if (!KernelStack) {
+			ObfDereferenceObject(eThread);
+			RETURN(X_STATUS_INSUFFICIENT_RESOURCES);
+		}
+
+		// Start thread initialization process here before insert and create thread
+		KeInitializeThread(&eThread->Tcb, KernelStack, KernelStackSize, TlsDataSize, SystemRoutine, StartRoutine, StartContext, &KiUniqueProcess);
```
ergo720, line 331: Initializing the `KTHREAD` with `KeInitializeThread` in the case that either `ObInsertObject` or `ObOpenObjectByPointer` ever fail is wasteful, since in that case the thread initialization has failed. So I suggest to put the call to `KeInitializeThread` after those two functions, since they do not depend on the state of `E/KTHREAD`.

ergo720:
Not caused by this PR specifically, but there's a bug in the call to `PspCallThreadNotificationRoutines` in `PsCreateSystemThreadEx`. In the position it's placed, it will be called even when `ObOpenObjectByPointer` or `_beginthreadex` fail, which is wrong because in that case the thread initialization has failed, and thus the Ps notification routines should not be called. So, it should be placed after `_beginthreadex` instead, the last place where `PsCreateSystemThreadEx` can fail.

src/core/kernel/exports/EmuKrnlPs.cpp
RadWolfie: Technically, that's not how object manager works for multi-threading. You have to initialize thread structure before you can insert into object manager which then will be used in other threads (if any). Therefore won't be corrected.

But the upstream codebase is currently wrong as insert object, do `PspCallThreadNotificationRoutines`, `ObOpenObjectByPointer`, and possibility other missing implement should be at the end of the function. These can be fix in later pull request.

ergo720 (approved):

PR #2342, opened 2022-04-06
Better support for UNC paths

NZJenkins:
Fix launching demos when the title was loaded from a UNC path
e.g. MechAssault with quantum redshift demo

We use GetFinalPathNameByHandle, which returns a "root local device" path beginning with '\\\\?\\'

For UNC paths, this looks like '\\\\?\UNC\bla' - but this path was not handled correctly.

Before, '\\\\?\\' was immediately stripped from the beginning of the path,
resulting in an invalid relative path e.g 'UNC\bla' - causing errors.

Now, we don't strip anything from the path, and accept it as-is.

In some related code, we also need to use the '\\' character instead of '/'
since '/' is not the path separator on Windows and is not valid to add to a path beginning with '\\\\?\\'.

Info about paths windows paths
https://googleprojectzero.blogspot.com/2016/02/the-definitive-guide-on-win32-to-nt.html

### DIFF
--- a/src/core/kernel/init/CxbxKrnl.cpp
+++ b/src/core/kernel/init/CxbxKrnl.cpp
@@ -613,7 +613,7 @@ static bool CxbxrKrnlXbeSystemSelector(int BootFlags, unsigned& reserved_systems
 	// Get title path :
 	std::string xbePath;
 	cli_config::GetValue(cli_config::load, &xbePath);
-	xbePath = std::filesystem::absolute(std::filesystem::path(xbePath)).string();
+	xbePath = std::filesystem::absolute(xbePath).string();
 
 	// NOTE: This is a safety to clean the file path for any malicious file path attempt.
 	// Might want to move this into a utility function.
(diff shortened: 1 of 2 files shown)


src/core/kernel/init/CxbxKrnl.cpp
```diff
@@ -630,7 +630,7 @@ static bool CxbxrKrnlXbeSystemSelector(int BootFlags, unsigned& reserved_systems
 
 	// Once clean up process is done, proceed set to global variable string.
 	strncpy(szFilePath_Xbe, xbePath.c_str(), xbox::max_path - 1);
-	std::replace(xbePath.begin(), xbePath.end(), ';', '/');
+	std::replace(xbePath.begin(), xbePath.end(), ';', '\\');
```
PatrickvL, line 633: The comment above this changed line doesn't really go into details on what's going on - could you elaborate perhaps?

NZJenkins: added some comments

PatrickvL: Thanks, it helps.

PatrickvL:
Perhaps it's good to mention in the description that this change no longer removes the `\\?\` prefix and instead use the UNC path as is? (If I'm not mistaken, because this fact is not clearly described in the PR description.)

PatrickvL (approved):

PatrickvL (approved):

PR #2345, opened 2022-04-09
Fix order of thread initialization

RadWolfie:
The improper order of thread initialization was brought to my attention. Host thread need at least run partially in order to finish thread initialization and terminiation as well. Plus thread initialization should occur before thread object is insert into object manager, along with routine callbacks for creation event.

This pull request resolve above type of issues and it is a hidden bug we didn't recognized for long time.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlPs.cpp
+++ b/src/core/kernel/exports/EmuKrnlPs.cpp
@@ -45,6 +45,8 @@
 #include "core\kernel\support\EmuFS.h" // For EmuGenerateFS
 #include "core\kernel\support\NativeHandle.h"
 
+#include <semaphore>
+
 // prevent name collisions
 namespace NtDll
 {
@@ -58,6 +60,7 @@ typedef struct _PCSTProxyParam
 {
 	IN xbox::PVOID  Ethread;
 	IN xbox::ulong_xt TlsDataSize;
+	IN OUT std::binary_semaphore* signal;
 }
 PCSTProxyParam;
 
@@ -112,6 +115,18 @@ static unsigned int WINAPI PCSTProxy
 #else
 	EmuGenerateFS(eThread);
 #endif
+
+	// NOTE: Native KTHREAD-SWITCHING will not need notification as it should be already done within
+	//       PsCreateSystemThreadEx function. For now, it is an experimental to see if xbox thread setup do work
+	//       without reside in host stack.
+	// Initializing ethread is done, we can allow PsCreateSystemThreadEx process to continue.
+	params.signal->release();
+	params.signal = nullptr;
+	SuspendThread(GetCurrentThread());
+	if (xbox::KeGetCurrentThread()->HasTerminated) {
+		xbox::PsTerminateSystemThread(0);
+	}
+
 	xbox::PKSTART_FRAME StartFrame = reinterpret_cast<xbox::PKSTART_FRAME>(reinterpret_cast<xbox::addr_xt>(eThread->Tcb.KernelStack) + sizeof(xbox::KSWITCHFRAME));
 
 	LOG_PCSTProxy(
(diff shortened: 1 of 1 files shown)


PatrickvL:
To anyone testing this : Be sure to include the mentioned test case: Amped

medievil1:
I do not have amped... not did I find any issues with this with the games I do have ripped, so there is that...lol

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -112,6 +113,18 @@ static unsigned int WINAPI PCSTProxy
...
 	EmuGenerateFS(eThread);
 #endif
+
+	// NOTE: Native KTHREAD-SWITCHING will not need notification as it should be already done within
+	//       PsCreateSystemThreadEx function. For now, it is an experimental to see if xbox thread setup do work
+	//       without reside in host stack.
+	// Initializing ethread is done, we can allow PsCreateSystemThreadEx process to continue.
+	params.cv->notify_one();
+	params.cv = nullptr;
+	SuspendThread(NtCurrentThread());
```
ergo720, line 123: `NtCurrentThread` is intended to be used on the xbox side, and there's no guarantee that the value returned is compatible with the one `SuspendThread` expects. Thus, you should probably use `GetCurrentThread` instead.

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -309,87 +322,99 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+		ObfDereferenceObject(eThread);
+		RETURN(X_STATUS_INSUFFICIENT_RESOURCES);
+	}
 
-		// Start thread initialization process here before insert and create thread
-		KeInitializeThread(&eThread->Tcb, KernelStack, KernelStackSize, TlsDataSize, SystemRoutine, StartRoutine, StartContext, &KiUniqueProcess);
+	// Start thread initialization process here before insert and create thread
+	KeInitializeThread(&eThread->Tcb, KernelStack, KernelStackSize, TlsDataSize, SystemRoutine, StartRoutine, StartContext, &KiUniqueProcess);
+
+	// Create mutex to allow new thread setup non-kthread switching.
```
ergo720, line 344: Nit: the comment is wrong: this is a condition_variable, not a mutex.

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -309,87 +322,99 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
...
+		MmDeleteKernelStack(eThread->Tcb.StackBase, eThread->Tcb.StackLimit);
+		ObfDereferenceObject(eThread);
+		RETURN(X_STATUS_INSUFFICIENT_RESOURCES);
+	}
+	// We are waiting on new thread's non-kthread switching process is done before we can continue on.
+	// Also, in favor for else statement to use instead of naked brackets.
+	else {
+		std::mutex mtx;
+		std::unique_lock lock(mtx);
+		cv.wait(lock);
```
ergo720, line 366: This has the potential to cause a deadlock in my opinion. Since you have now removed the `CREATE_SUSPENDED` flag from `_beginthreadex`, the new thread will run whenever Windows pleases. This means that, if the spawned thread were to reach the call to `notify_one` before this other thread starts the wait here, it will miss the notification, and thus we will wait forever here.

RadWolfie: I had not encounter any problems, however... by having this wait lock right after call new thread function do reduce the risk. Beside, what is your suggestion to resolve this rare chance of possible deadlock? My original idea of using mutex directly for lock/unlock manually posed no deadlock risk at all even though intellisense did show couple warnings.

src/core/kernel/exports/EmuKrnlPs.cpp
RadWolfie: Originally, I was wary with thread usage as some situation require special handle than a given handle will cause spit an error from Windows. After checking again with `GetCurrentThread` shows it is working as intended.

src/core/kernel/exports/EmuKrnlPs.cpp
RadWolfie: resolved

src/core/kernel/exports/EmuKrnlPs.cpp
RadWolfie: I think I know what you're expecting. So, I added a variable check in case of missed notification.
(resolved)

ergo720: This is still not good enough. Your `bThreadInitComplete` variable is checked only outside of `wait`, which blocks the thread and only exits either when it's notified or a spurious wakeup occurs. So, if we miss the notification, we will only exit the wait if the spurious event occurs. The solution here is to just avoid the `condition_variable` entirely, and use a simple busy loop like below:
```
while (!bThreadInitComplete) {
    std::this_thread::sleep_for(milliseconds(1)); // perhaps not necessary
}
```
This works because now it no longer relies on the notification (which you can miss) but only on a boolean (which you can't).

RadWolfie: Changed `std::condition_variable` to `std::binary_semaphore` for simpler signal. Thanks to @NZJenkins' finding for new C++20 semaphore feature.

ergo720 (approved):

PR #2348, opened 2022-04-14
Purge emulation codebase from GUI project

RadWolfie:
Removal of emulation codebase from GUI project benefits are:
- Reduce compile time process
- No longer need to preserve 64MiB size of virtual memory placeholder variable in GUI project's binary file.
- No longer cause Visual Studio 2019+, although we only support Visual Studio 2022+, to softlock (normally take 15+ minutes to response again) when accidentical look in GUI project's xbox space where virtual memory placeholder is located.

However, not everything has been separated from GUI project as some are very dependent on certain interfaces. And therefore remain included which are:
- Xapi (does need to be split, yet unable to do so at this time)
- EmuKrnlXbox
- EmuKrnlXc
- EmuKrnlXe

I can confirm no regression had occured after ran several titles, including dashboard.

---

**NOTE:** Even though there are lots of file changes, only 90% of them are function renames as GUI and emulation projects do need them to coexist, aka CxbxrAbort(Ex) and CxbxrShutdown. Mark them as viewed and review at remaining files as there's only few of them have been transferred to their respective files. Plus some removal of `CXBXR_EMU` marco checks.


ergo720 (approved):

src/core/hle/D3D8/XbConvert.cpp
```diff
@@ -1591,6 +1591,221 @@ Direct3D9 states unused :
 	D3DRS_BLENDOPALPHA = 209   // Blending operation for the alpha channel when D3DRS_SEPARATEDESTALPHAENABLE is TRUE
 */
 
+static xbox::PVOID g_pXbox_Palette_Data[xbox::X_D3DTS_STAGECOUNT] = { xbox::zeroptr, xbox::zeroptr, xbox::zeroptr, xbox::zeroptr }; // cached palette pointer
```
NZJenkins, line 1594: This is not valid?
The palette data is a global in another file
Scaler errors converting a palette texture since this never gets populated

PatrickvL: As far as I remember, icons embedded in xbe sections never use a palette, so perhaps the palette using code path could be compiled into the emulator but not into the GUI?

RadWolfie: Thanks for the input! It is resolved in pull request #2355.

PR #2349, opened 2022-04-14
Save and load compiled vertex shaders to/from disk

NZJenkins:
Which solves some minor stuttering caused by compiling shaders at runtime (or rather reduces it, since this is only for vs shaders)

Also, the FF VS is currently one giant shader whichis quite heavy, and we don't bundle a compiled version (we could whil variations aren't dynamically generated...)
but caching alleviates issues with it which can help render the first few frames right after boot e.g. the dashboard "zoom" effect at boot

Shaders get saved per title in a "Shaders" folder under the data path
TODO:
- [x] Test the cache is rebuilt for new cxbx versions (in dev, need to regen cmake to change the version stamp...?)
- [x] Make optional


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1685,6 +1686,53 @@ uint8_t *ConvertD3DTextureToARGB(
...
+	uint64_t uiHash = ComputeHash((void*)&CxbxKrnl_Xbe->m_Header, sizeof(Xbe::Header));
+	std::stringstream sstream;
+	std::wstring wTitleName = CxbxKrnl_Xbe->m_Certificate.wsTitleName;
+	std::string titleName = std::string(wTitleName.begin(), wTitleName.end());
+	CxbxKrnl_Xbe->PurgeBadChar(titleName);
+
+	std::stringstream fileName;
+	fileName << titleName << "-" << std::hex << uiHash << ".vscache";
+
+	return fs::path(g_DataFilePath) / "Shaders" / fileName.str();
```
RadWolfie, line 1704: just a nit change here:
```suggestion
	return fs::path(g_DataFilePath) / "ShaderCache" / fileName.str();
```

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1685,6 +1686,53 @@ uint8_t *ConvertD3DTextureToARGB(
...
+	else {
+		EmuLog(LOG_LEVEL::INFO, "Could not open vertex shader cache path: %s", cachePath.string());
+	}
+}
+
+void SaveShaderCache() {
+	namespace fs = std::filesystem;
+
+	auto cachePath = GetVshCachePath();
+	fs::create_directory(cachePath.parent_path());
```
RadWolfie, line 1724: imho, `fs` shortcut shouldn't be used. Beside, there's only one function used from filesystem library.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1685,6 +1686,53 @@ uint8_t *ConvertD3DTextureToARGB(
...
+	CxbxKrnl_Xbe->PurgeBadChar(titleName);
+
+	std::stringstream fileName;
+	fileName << titleName << "-" << std::hex << uiHash << ".vscache";
+
+	return fs::path(g_DataFilePath) / "Shaders" / fileName.str();
+}
+
+void LoadShaderCache() {
+	namespace fs = std::filesystem;
```
RadWolfie, line 1708: I don't see any filesystem calls in this function. This line need to be remove.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1685,6 +1686,53 @@ uint8_t *ConvertD3DTextureToARGB(
 
 extern void HLE_init_pgraph_plugins(); // implemented in XbPushBuffer.cpp
 
+// FIXME move cache functions elsewhere
+std::filesystem::path GetVshCachePath() {
+	namespace fs = std::filesystem;
```
RadWolfie, line 1691: imho, fs shortcut shouldn't be used. Beside, there's only one function used from filesystem library.

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -165,8 +176,126 @@ void VertexShaderSource::ReleaseShader(ShaderKey key)
...
+			EmuLog(LOG_LEVEL::INFO, "Serializing shader %x...", c.key);
+
+			out << c.key << "\n";
+			out << c.size << "\n";
+			out.write(shaderBuffer.data(), c.size);
+			out << "\n";
+		}
+	}
+	catch (...) {
+		EmuLog(LOG_LEVEL::INFO, "An error occurred serializing the VSH cache");
```
ergo720, line 237: This is an error condition right? So it should use the error log level.

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -165,8 +176,126 @@ void VertexShaderSource::ReleaseShader(ShaderKey key)
...
+	}
+}
+
+void VertexShaderSource::DeserializeAndLoad(IDirect3DDevice9* pD3DDevice, std::istream& in) {
+	EmuLog(LOG_LEVEL::INFO, "Loading VSH cache...");
+
+	Header hExpected = CreateHeader();
```
PatrickvL, line 244: This could be made a const or possibly even a constexpr

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -85,13 +88,29 @@ ShaderKey VertexShaderSource::CreateShader(const xbox::dword_xt* pXboxFunction,
...
+void VertexShaderSource::RegisterShader(ShaderKey key, IDirect3DVertexShader9* pHostVertexShader) {
+	LazyVertexShader s;
+	s.isReady = true;
+	s.pHostVertexShader = pHostVertexShader;
+
+	LazyVertexShader* existing;
+	if (_FindShader(key, &existing)) {
+		EmuLog(LOG_LEVEL::WARNING, "Overwriting vertex shader %x", key);
+		if (pHostVertexShader)
+		existing->pHostVertexShader->Release();
```
PatrickvL, line 100: Indentation level one extra please

RadWolfie:
> Test the cache is rebuilt for new cxbx versions (in dev, need to regen cmake to change the version stamp...?)

You don't need to touch cmake files to regenerate version stamp (which is git hash base on commit + tag, if any, itself). This is done through batch script as separate project and is run before project's build occur. Every time you build projects, you will see git hash output in the output window. If you are not creating commits, it will only append "-dirty" flag. "-dirty" flag is what you're looking for from GetGitVersionStr() function, if you need to force regen shader cache every time.

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1155,16 +1155,18 @@ void CxbxUpdateHostVertexShader()
 		HRESULT hRet;
 
 		if (g_UseFixedFunctionVertexShader) {
-			static IDirect3DVertexShader* ffHlsl = nullptr;
-			if (ffHlsl == nullptr) {
+			const ShaderKey ffKey = 0xFF;
```
PatrickvL, line 1158: Aren't you afraid of hash collisions on this seemingly randomly chosen value?

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -134,8 +148,6 @@ IDirect3DVertexShader* VertexShaderSource::GetShader(ShaderKey key)
 
 	if (pCompiledShader) {
 		pCompiledShader->Release();
-
-		// TODO compile the shader at a higher optimization level in a background thread?
```
PatrickvL, line 138: Why is this remark removed? Is it not longer relevant? Or is the suggested functionality now delegated to another function?

PatrickvL (approved):
Functionality this looks fine. Perhaps some removed comments could be kept or moved, but otherwise this can be merged after successful testing

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: Nah just one more place to assume hash collisions don't happen...
It's 0xFF for fixed function ;)

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
NZJenkins: Didn't seem relevant - we compile at the highest optimisation level currently (iirc not true in the past)
Compiling with less optimisation might reduce stutter a bit - but this caching should also

RadWolfie:
@PatrickvL, please mark your reivews as resolve if you believe they are resolved. Thanks

RadWolfie:
> the tag doesn't seem to match when I build 🤷‍♂️

Please elaborate on this. Or was it misunderstood from your end?

src/core/hle/D3D8/Direct3D9/VertexShaderSource.cpp
```diff
@@ -165,8 +177,125 @@ void VertexShaderSource::ReleaseShader(ShaderKey key)
...
+	try {
+		std::getline(in, hActual.fileType);
+		std::getline(in, hActual.version);
+
+		if (hActual.fileType != hExpected.fileType) {
+			EmuLog(LOG_LEVEL::ERROR2, "VSH cache did not have the expected header.");
+			return;
+		}
+		if (hActual.version != hExpected.version) {
+			EmuLog(LOG_LEVEL::INFO, "VSH cache version mismatch. The cache will not be loaded.");
```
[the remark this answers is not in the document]
PatrickvL: Last commit fixed this

PatrickvL (changes requested):
GitHub says conflicts need to be resolved before this can be merged, please considered a rebase.

src/common/Settings.hpp
```diff
@@ -107,7 +107,8 @@ class Settings
 		bool allowAdminPrivilege;
 		bool bLogPopupTestCase;
 		bool Reserved4 = 0;
-		int  Reserved99[10] = { 0 };
+		bool UseShaderDiskCache = 1;
+		int  Reserved99[9] = { 0 };
```
RadWolfie, line 111: Misused of how reserved work. Reserved were originally meant to align properly. Since you take one bool out of int array. You left out 3 additional bools reserved.

Please use `bUnused_WasUseLoaderExec` or `Reserved4` variable. The latter is preferred.

src/common/win32/EmuShared.h
```diff
@@ -164,6 +164,11 @@ class EmuShared : public Mutex
 		void GetInputGeneralSettings(Settings::s_input_general *input_general) { Lock(); *input_general = m_input_general; Unlock(); }
 		void SetInputGeneralSettings(const Settings::s_input_general *input_general) { Lock(); m_input_general = *input_general; Unlock(); }
 
+		// ******************************************************************
+		// * Shader disk cache Accessors
+		// ******************************************************************
+		bool GetUseShaderDiskCache() { Lock(); bool s = m_core.UseShaderDiskCache; Unlock(); return s; }
```
RadWolfie, line 170: None of our EmuShared APIs are using any return, please fix by using parameter by design.

src/common/FilePaths.hpp
```diff
@@ -24,6 +24,7 @@
 // ******************************************************************
 #pragma once
 
+// FIXME include MAX_PATH
```
RadWolfie, line 27: FIXME keyword is not common that will show up in IDEs. Please append `TODO:` as before text in order for it to appear on task list window.
These are the default tokens:
image

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -1472,6 +1474,52 @@ void GetSurfaceFaceAndLevelWithinTexture(xbox::X_D3DSurface* pSurface, xbox::X_D
...
+	if (f.is_open()) {
+		g_VertexShaderSource.DeserializeAndLoad(g_pD3DDevice, f);
+	}
+	else {
+		EmuLog(LOG_LEVEL::INFO, "Could not open vertex shader cache path: %s", cachePath.string());
+	}
+}
+
+void SaveShaderCache() {
+	if (!g_EmuShared->GetUseShaderDiskCache()) return;
```
RadWolfie, line 1508: Is it absolutely necessary to constant check EmuShared? EmuShared access design should only be access once per change occur.

According this pull request change, I don't see any IPC sync call during emulation which made this unnecessary and cause additional performance impact just for the lock/unlock. Please get the value from `CxbxInitWindow` function.

RadWolfie (changes requested):
Some changes required before able merge.

RadWolfie:
Beside review remark, if we are to implement backend settings for renderer's cache content. Should they need to be in the video settings instead of part of the core settings?

PR #2350, opened 2022-04-18
Partially fix NaN handling in shaders

NZJenkins:
On hw it appears NaN becomes 1 or 0 depending on the sign bit - and doing operations in the VS seems to change -NaN to +NaN.
Otogi appears to rely on this behaviour by loading constant register +62 with (-INF, -NaN) and using it in a MAD to set oFog

I don't _think_ the FF VSH path needs this since it naturally clamps everything IIRC

nxdk test https://github.com/abaire/nxdk_pgraph_tests/pull/63

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
@@ -369,8 +369,8 @@ R"DELIMITER(
 	xOut.oPos = reverseScreenspaceTransform(oPos);
 	xOut.oD0 = saturate(clean(oD0));
 	xOut.oD1 = saturate(clean(oD1));
-	xOut.oFog = clean(fogFactor); // Note : Xbox clamps fog in pixel shader -> *NEEDS TESTING* /was oFog.x 
-	xOut.oPts = clean(oPts.x);
+	xOut.oFog = clean(fogFactor).x; // Note : Xbox clamps fog in pixel shader -> *NEEDS TESTING* /was oFog.x 
+	xOut.oPts = clean(oPts.x).x;
 	xOut.oB0 = saturate(clean(oB0));
 	xOut.oB1 = saturate(clean(oB1));
 	// Scale textures
--- a/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
@@ -289,6 +289,19 @@ float4 reverseScreenspaceTransform(float4 oPos)
 	return oPos;
 }
 
+// Clean up values before they go to the PS
+// FIXME On hw it appears NaN becomes 1 or 0 depending
+// on the sign bit, which might require bitwise ops in SM4+
+float4 clean(float4 x)
+{
+	// Convert NaN into 1
+	// Note IsNaN() just gets optimized away...
+	// https://shader-playground.timjones.io/0bbe04704caddadb52cd4134daab8ac3
+	// Test case: Otogi does math ops on -NaN
+	// which seems to become +NaN -> +1 in the PS
+    return (x < 0.f || x > 0.f || x == 0.f) ? x : 1;
+}
+
 VS_OUTPUT main(const VS_INPUT xIn)
 {
 	// Output variables
@@ -354,17 +367,17 @@ R"DELIMITER(
        fogFactor = (fogEnd - fogDepth) / (fogEnd - fogStart);
        
 	xOut.oPos = reverseScreenspaceTransform(oPos);
-	xOut.oD0 = saturate(oD0);
-	xOut.oD1 = saturate(oD1);
-	xOut.oFog = fogFactor; // Note : Xbox clamps fog in pixel shader -> *NEEDS TESTING* /was oFog.x 
-	xOut.oPts = oPts.x;
-	xOut.oB0 = saturate(oB0);
-	xOut.oB1 = saturate(oB1);
-	// Scale textures (TODO : or should we apply this to the input register values?)
-	xOut.oT0 = oT0 / xboxTextureScale[0];
-	xOut.oT1 = oT1 / xboxTextureScale[1];
-	xOut.oT2 = oT2 / xboxTextureScale[2];
-	xOut.oT3 = oT3 / xboxTextureScale[3];
+	xOut.oD0 = saturate(clean(oD0));
+	xOut.oD1 = saturate(clean(oD1));
+	xOut.oFog = clean(fogFactor); // Note : Xbox clamps fog in pixel shader -> *NEEDS TESTING* /was oFog.x 
+	xOut.oPts = clean(oPts.x);
+	xOut.oB0 = saturate(clean(oB0));
+	xOut.oB1 = saturate(clean(oB1));
+	// Scale textures
+	xOut.oT0 = clean(oT0 / xboxTextureScale[0]);
+	xOut.oT1 = clean(oT1 / xboxTextureScale[1]);
+	xOut.oT2 = clean(oT2 / xboxTextureScale[2]);
+	xOut.oT3 = clean(oT3 / xboxTextureScale[3]);
 
 	return xOut;
 }


PatrickvL (commented):
Nice detective work! Judging from the disassembly that shader playground shows, this cleanup on all components is going to add a lot of additional instructions - if shaders fail because of that, it might become necessary to only clean certain components or only apply it to certain shaders in certain titles. Something to keep in mind

PatrickvL (dismissed):

PatrickvL:
Oh, could you please add a before and after screenshot?

zaykho:
### Before:
--------------------------



unknown
unknown (1)



### After:
--------------------------




unknown (2)
unknown (3)
unknown (4)
unknown (5)



--------------------------

There is a weird issue happening though, if you restart the stage (push start and select restart), the intro scene (where the character is alone in the dark) will take the same lightning setup than the moment where you restarted (should not happen).

**Here an example:**

First initialization (correct colors and lightning, as expected)
unknown

Then, reaching in-game on the stage:

unknown (2)


Now, if you push start and select restart right away, you get that black overlay:

unknown (1)

Then, if you move the camera and rotate in different orientation, you get different colors/lightning:

unknown (3)

And if you restart again, you get this same result  but in intro (should not):

unknown (4)

And again here:

unknown (5)

unknown (6)

In certain situation, it may even crash.

medievil1:
ummm... notice errors
/* 424 */       xOut.oFog = clean(fogFactor); // Note : Xbox clamps fog in pixel shader -> *NEEDS TESTING* /was oFog.x
/* 425 */       xOut.oPts = clean(oPts.x);
are giving 
[0x5134] DEBUG: VTXSH   N:\cxbx\games\spinx and the cursed mummy\CxbxVertexShaderTemplate.hlsl(424,2-29): warning X3206: implicit truncation of vector type
N:\cxbx\games\spinx and the cursed mummy\CxbxVertexShaderTemplate.hlsl(425,2-26): warning X3206: implicit truncation of vector type

which as I understand, clean is a float 4, BUT fogfactor is not and then oPts.x is not.... may not matter in the end...just thought I would mention it

PatrickvL:
@NZJenkins please tell when draft status is removed, and final review approval can be granted

medievil1:
not ran into any issues and the log file is clean now.. no warnings

zaykho:
> the game still has issues and is still black character a lot, but I think it is more to do with lighting/shading being too dark....

That's why it's a partial fix.

On Otogi there is also a lot of stages where objects "seems" to also be impacted by the fog, and they sill appear black after the fix:
unknown

unknown (1)

Moving the camera make some elements darker or not, so I imagine that there is a bigger issue deep down.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -354,17 +367,17 @@ R"DELIMITER(
        fogFactor = (fogEnd - fogDepth) / (fogEnd - fogStart);
        
 	xOut.oPos = reverseScreenspaceTransform(oPos);
-	xOut.oD0 = saturate(oD0);
-	xOut.oD1 = saturate(oD1);
-	xOut.oFog = fogFactor; // Note : Xbox clamps fog in pixel shader -> *NEEDS TESTING* /was oFog.x 
-	xOut.oPts = oPts.x;
-	xOut.oB0 = saturate(oB0);
-	xOut.oB1 = saturate(oB1);
-	// Scale textures (TODO : or should we apply this to the input register values?)
```
LukeUsher, line 363: Is there a reason this comment was removed?

I'm guessing it's because the decision has been made to scale in the shaders rather than scale the inputs; but I'm just double checking

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -289,6 +289,19 @@ float4 reverseScreenspaceTransform(float4 oPos)
...
+// FIXME On hw it appears NaN becomes 1 or 0 depending
+// on the sign bit, which might require bitwise ops in SM4+
+float4 clean(float4 x)
+{
+	// Convert NaN into 1
+	// Note IsNaN() just gets optimized away...
+	// https://shader-playground.timjones.io/0bbe04704caddadb52cd4134daab8ac3
+	// Test case: Otogi does math ops on -NaN
+	// which seems to become +NaN -> +1 in the PS
+    return (x < 0.f || x > 0.f || x == 0.f) ? x : 1;
```
PatrickvL, line 302: I found a faster NaN check that's also extremely simple :
```
function isNaN(val:Number): Boolean
{
    return val != val;
}
```

Source : https://www.jacksondunstan.com/articles/983

PatrickvL: And if that doesn't work, someone stated that saturate () might work as well? https://twitter.com/_Humus_/status/1074978669658075136?t=v-hjaw8oVmIuV9yF8nx1Ng&s=19

PatrickvL: And someone else suggested the step() intrinsic : https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-step

NZJenkins: seems in hlsl x != x gets compiled to false
saturate kills NaN but we ideally want to map -NaN and +NaN to 0 and 1
I'm not sure about how to use step

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
NZJenkins: I don't remember the specifics of this or if it was referring to scaling VS inputs or PS inputs.
If VS inputs - a shader could write any value to a texture output, so it's not really a general solution.
If PS inputs - I assume it's cheaper to calculate per vertex rather than per-pixel, if we can.
We can revisit it when there is a specific issue.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: The not-NaN-detecting `x != x` test is then probably due to the compiler (or your driver or GPU) is not IEEE 754 compliant, which is a pity.

PatrickvL: The step approach comes from https://www.gamedev.net/forums/topic/610591-hlsl-clamp-infinity-and-nan-to-0/4861379/ which says :
```this seems to work on my machine (it's hard to get a definite test going). I'm not sure definitively if this is defined behavior or not, though.


A = step(A, A) * A;


'A' can be a scalar, vector, or matrix. It works on NaNs on the assumption that they won't be >= to themselves. I'm not sure why it'd work on infinities but it does seem to.```

PatrickvL: Perhaps `min(-1.0, max(x, 1.0)) turns NaN's into 1 and negative NaN's into -1.0? (Not sure if that'd be any more efficient, nor if this is the intended result?)

PatrickvL: About negative NaN's: In some CPU tests I did I've seen that the default NaN value has the sign bit set, while a negative NaN (the -NaN literal value) had the sign bit cleared - that's rather counter-intuitive indeed, but I'm not sure if that's also happening in our environment, nor what the effects might be.

PatrickvL: Actually, that hardware test doesn't prove that -INF behaves like zero - it might even behave like minus one, but how would you test that?

PatrickvL: Btw, you write that -NaN should map to 0, but isn't that a typo? Above test you wrote shows that -NaN seems identical to 1. However, -INF does seem to map to 0.0 hence the suspicion of a typo

Triticum0:
### Game that use NaN Handling 
Otogi 1 and 2
Pro Cast: Sports Fishing Game
Trigger man

PatrickvL:
According to https://github.com/mborgerson/xemu/pull/913/files#r888360218 Lakemaster is also impacted by NaN values

Triticum0:
Matt tested Pro Cast: Sports Fishing Game and Trigger man as he has the games

lorn10:
@PatrickvL Maybe the information provided [here](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/12794) and [here](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/16176#note_1358614) could help somehow to understand the "legacy DX math rules" topic better, - eventually. :wink:

LukeUsher:
What's the status on this? It has been pending review for quite some time, 
@NZJenkins do you think this is ready? or is more work required?

medievil1:
I have not seen any ill effects from it... (in fact I have been adding it to my personal builds for a while)

PR #2353, opened 2022-05-02
Fix tls data initialization from cxbxr's emulation threads end

RadWolfie:
I recently discovered this problem while attempting to run rebased chihiro branch. It turns out couple of our emulated kernel threads caller are using TLS in which we didn't cover it. With this correction, I no longer get any error pop up for this type of problem.

> `PsCreateSystemThread` doesn't use xbe's tls data size. It is hardcoded to 0. Our clock thread do call ExecuteDpcQueue(). Using PsCreateSystemThread isn't viable for our emulated kernel thread implement. So... instead of doing this, we should be using Xapi's CreateThread-like method.

### DIFF
--- a/src/common/Timer.cpp
+++ b/src/common/Timer.cpp
@@ -25,8 +25,6 @@
 // *
 // ******************************************************************
 
-#include <core\kernel\exports\xboxkrnl.h>
-
 #ifdef _WIN32
 #include <windows.h>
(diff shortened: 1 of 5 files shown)


PatrickvL (approved):
LGTM (although I haven't ran it - just a casual code review)

PR #2355, opened 2022-05-02
Fix pull request 2348 regression

RadWolfie:
Thanks to @NZJenkins and @PatrickvL for the inputs.

We're able to resolve conflict between gui and emu difference for `ConvertD3DTextureToARGBBuffer` function. Turns out we only need to pass down pallete pointer depending on `iTextureStage` index from emulator's end. GUI does not rely on this input for generate xbe's icons.

fixed pull request #2348

----

Titles regression are confirmed back to working state:
- Haunted Mansion
- Fallout - Brotherhood of Steel
- Red Faction 2


PatrickvL (approved):
Looks fine to me!

PR #2356, opened 2022-05-07
Fixed wrong irql masks

ergo720:
This was noticed while working with cxbxrkrnl. The irql masks from irql 7 to irql 13 are wrong, which will cause firing an incorrect interrupt under some circumstances.


src/core/kernel/exports/EmuKrnl.cpp
```diff
@@ -183,13 +183,13 @@ const DWORD IrqlMasks[] = {
...
-	0x000EFFF0, // IRQL 11
-	0x0007FFF0, // IRQL 12
-	0x0003FFF0, // IRQL 13
+	0x007FFFF0, // IRQL 7
+	0x003FFFF0, // IRQL 8
+	0x001FFFF0, // IRQL 9
+	0x000FFFF0, // IRQL 10
+	0x0007FFF0, // IRQL 11
+	0x0003FFF0, // IRQL 12
+	0x0001FFF0, // IRQL 13
```
PatrickvL, line 192: This makes IRQ 13 identical to 12. Is that intentional?

PatrickvL (commented):
Seems simple enough, just one question above...

src/core/kernel/exports/EmuKrnl.cpp
ergo720: Yes, I verified this numbers with my kernel dump and they are exactly like this.

PatrickvL:
Thanks. Now, should we await testing results, or just merge it right away?

RadWolfie (approved):

PR #2357, opened 2022-05-08
Make the SBC use the cursor mouse mode by default

ergo720:
When bound to a mouse, the aim and sight controls of the SBC works best with the mouse cursor mode, instead of the axis mode. So this makes the cursor mode the default for the SBC, so that users cannot accidentally choose the axis mode and  experience the [re-centering issue](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2135). Note that existing SBC profiles are unaffected by the change, so users will have either to update it manually or create a new profile.


RadWolfie (approved):
After comparing the change to light gun's cursor mode by default. LGTM!

PR #2358, opened 2022-05-09
Add build, title name, and file name to overlay

RadWolfie:
Following items are add to the overlay UI:
- Build hash (will be display on bottom screen)
- Title name (top right)
- File name (top right)

Preview:
image

### DIFF
--- a/src/core/common/imgui/ui.cpp
+++ b/src/core/common/imgui/ui.cpp
@@ -120,8 +120,11 @@ void ImGuiUI::DrawMenu()
 		if (ImGui::BeginMenu("Settings")) {
 			if (ImGui::BeginMenu("Overlay")) {
 				bool bChanged = false;
+				bChanged |= ImGui::MenuItem("Show Build Hash", NULL, &m_settings.build_hash);
 				bChanged |= ImGui::MenuItem("Show FPS", NULL, &m_settings.fps);
 				bChanged |= ImGui::MenuItem("Show HLE/LLE Stats", NULL, &m_settings.hle_lle_stats);
+				bChanged |= ImGui::MenuItem("Show Title Name", NULL, &m_settings.title_name);
+				bChanged |= ImGui::MenuItem("Show File Name", NULL, &m_settings.file_name);
 				if (bChanged) {
 					g_EmuShared->SetOverlaySettings(&m_settings);
 					ipc_send_gui_update(IPC_UPDATE_GUI::OVERLAY, 1);
@@ -139,14 +142,18 @@ void ImGuiUI::DrawMenu()
 
 void ImGuiUI::DrawWidgets()
 {
-	if (m_settings.fps || m_settings.hle_lle_stats) {
+
+	if (m_settings.fps
+	    || m_settings.hle_lle_stats
+		||m_settings.title_name
+	    || m_settings.file_name) {
 
 		ImGui::SetNextWindowPos(ImVec2(ImGui::GetIO().DisplaySize.x - (IMGUI_MIN_DIST_SIDE/* * m_backbuffer_scale*/),
 			IMGUI_MIN_DIST_TOP/* * m_backbuffer_scale*/), ImGuiCond_Always, ImVec2(1.0f, 0.0f));
 
 		ImGui::SetNextWindowSize(ImVec2(200.0f/* * m_backbuffer_scale*/, 0.0f));
 		ImGui::SetNextWindowBgAlpha(0.5f);
-		if (ImGui::Begin("overlay_stats", nullptr, ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoNav |
+		if (ImGui::Begin("overlay_stats_topright", nullptr, ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoNav |
 			ImGuiWindowFlags_NoSavedSettings | ImGuiWindowFlags_NoScrollbar |
 			ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoFocusOnAppearing)) {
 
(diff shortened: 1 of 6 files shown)


src/core/common/imgui/ui.cpp
```diff
@@ -139,14 +142,18 @@ void ImGuiUI::DrawMenu()
 
 void ImGuiUI::DrawWidgets()
 {
-	if (m_settings.fps || m_settings.hle_lle_stats) {
+
+	if (m_settings.fps
+	    || m_settings.hle_lle_stats
+		||m_settings.title_name
```
PatrickvL, line 148: Indentation seems off on this line

src/core/common/imgui/ui.cpp
```diff
@@ -180,8 +187,36 @@ void ImGuiUI::DrawWidgets()
...
-			ImGui::End();
+
+			if (m_settings.title_name) {
+				ImGui::TextColored(ImVec4(1.0f, 1.0f, 0.0f, 1.0f), "Title: %.41s", CxbxKrnl_Xbe->m_szAsciiTitle);
+			}
+
+			if (m_settings.file_name) {
+				ImGui::TextColored(ImVec4(1.0f, 1.0f, 0.0f, 1.0f), "File: %.260s", CxbxKrnl_Xbe->m_szFileName);
+			}
+
```
PatrickvL, line 198: Nit: no empty line here, as multiple code blocks are closed sequentially - that doesn't require separation

src/core/common/imgui/ui.cpp
```diff
@@ -180,8 +187,36 @@ void ImGuiUI::DrawWidgets()
...
+		}
+		ImGui::End();
```
PatrickvL, line 200: Since the line above this statement closed a code block, there should be an empty separation line above here

src/core/common/imgui/ui.cpp
```diff
@@ -180,8 +187,36 @@ void ImGuiUI::DrawWidgets()
...
+	}
+
+	if (m_settings.build_hash) {
+
```
PatrickvL, line 204: No empty line needed here

src/core/common/imgui/ui.cpp
```diff
@@ -180,8 +187,36 @@ void ImGuiUI::DrawWidgets()
...
+		ImGui::SetNextWindowPos(ImVec2(IMGUI_MIN_DIST_SIDE, ImGui::GetIO().DisplaySize.y - IMGUI_MIN_DIST_SIDE/* * m_backbuffer_scale*/),
+			ImGuiCond_Always, ImVec2(0.0f, 1.0f));
+
```
PatrickvL, line 207: No empty line needed here

src/core/common/imgui/ui.cpp
```diff
@@ -180,8 +187,36 @@ void ImGuiUI::DrawWidgets()
...
+		//ImGui::SetNextWindowSize(ImVec2(200.0f/* * m_backbuffer_scale*/, 0.0f));
+		ImGui::SetNextWindowBgAlpha(0.5f);
+		if (ImGui::Begin("overlay_stats_bottom", nullptr, ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoNav |
+			ImGuiWindowFlags_NoSavedSettings | ImGuiWindowFlags_NoScrollbar |
+			ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoFocusOnAppearing)) {
+
```
PatrickvL, line 213: No empty line needed here

src/core/common/imgui/ui.cpp
```diff
@@ -180,8 +187,36 @@ void ImGuiUI::DrawWidgets()
...
+			if (m_settings.build_hash) {
+				ImGui::TextColored(ImVec4(1.0f, 1.0f, 0.0f, 1.0f), "Build: %s", GetGitVersionStr());
+			}
+
```
PatrickvL, line 217: No empty line needed here

src/core/common/imgui/ui.cpp
```diff
@@ -180,8 +187,36 @@ void ImGuiUI::DrawWidgets()
...
 		}
+		ImGui::End();
```
PatrickvL, line 219: Empty line needed above this one

src/core/common/imgui/video.cpp
```diff
@@ -50,8 +50,8 @@ void ImGuiVideo::DrawWidgets(bool is_focus, ImGuiWindowFlags input_handler)
 			if (ImGui::CollapsingHeader("Vertex Buffer Cache", ImGuiTreeNodeFlags_DefaultOpen)) {
 				VertexBufferConverter.DrawCacheStats();
 			}
-			ImGui::End();
 		}
+		ImGui::End();
```
PatrickvL, line 54: This is an actual bugfix, right?

PatrickvL (dismissed):
The change itself looks fine, just some style remarks that would be nice to have tidied up - but approved regardless

src/core/common/imgui/ui.cpp
Margen67: It's mixing spaces and tabs

RadWolfie: You are correct, Visual Studio somehow re-indent the code without my approval. I'll fix that.

src/core/common/imgui/ui.cpp
RadWolfie: Main reason they're separated as they have no relationship to each other. Same thing for fps and hle/lle stats if statements above.

src/core/common/imgui/ui.cpp
RadWolfie: New line is needed to tell the difference for latter code is not part of first if statment due to they have same indent size.

src/core/common/imgui/ui.cpp
RadWolfie: Not sure if this is intended or a mistake review remark?

src/core/common/imgui/ui.cpp
RadWolfie: Never separate "begin" and "end" blocks. If something is very relative, they should remain together than new separate line.

src/core/common/imgui/video.cpp
RadWolfie: Yes it is a bugfix, see [commented from ImGui's header file](https://github.com/ocornut/imgui/blob/60bea052a92cbb4a93b221002fdf04f0da3698e1/imgui.h#L319-L320).
I notice this problem while working with @NZJenkins for the DirectSound's buffer additional details graph causing crashes which was our mistake.

src/core/common/imgui/ui.cpp
PatrickvL: Understood.

src/core/common/imgui/ui.cpp
PatrickvL: Personally, my eyes are 'tuned' to a certain coding style - empty lines being a part of that. That makes seemingly inconsistent styles stand out more than what the actual code is doing. This, because to understand the code, one needs to read the full statement line(s), interpret them, producing a reason for the absence of an empty line - that's distracting me more than spotting which lines belong to which part of the functionality. But that's just me 😜

src/core/common/imgui/ui.cpp
PatrickvL: It was an intentional remarks, but it's motivated by my assumption about coding style convention, which is still not been fully decided upon. So you do as you please for now 😜

src/core/common/imgui/ui.cpp
PatrickvL: Yeah, multi-line expressions within if statements can indeed be difficult to "get right". One way to circumvent that, would be to replace the ImGui flags by a constexpr variable, declared one line prior. That way, the 'if' line won't wrap around, removing the reason you deem it necessary to have an empty line here. But that's probably overkill. (An alternative work-around is to have the 'or' pipe symbols at the same indentation-level as the 'if' statement, but that's a bit ugly as well.)

PatrickvL: Looking much better now!

PatrickvL (approved):

PR #2360, opened 2022-05-15
Do not leave KeDelayExecutionThread prematurely

CookiePLMonster:
Fixes games waking up from Sleep() calls immediately. `NtDelayExecution` could return `STATUS_NO_YIELD_PERFORMED` if the call was a no-op, which should not be treated as failure of any kind, but it was and it was propagated back to games. This caused Sleep() calls to be nearly no-ops.


LukeUsher (approved):
Excellent find; I'm excited to see the compatibility implications of this change

PatrickvL:
Yeah, this will surely help. How long are we going to wait for feedback from testers, before merging this?

PR #2362, opened 2022-05-15
Keep D3D resources alive during resource creation

CookiePLMonster:
Fixes an issue where a game could tear down a resource from another thread while it's still initializing. With this change, the code is still far from being thread safe against games issuing D3D calls from multiple threads, but it's a step in the right direction.

In my test case, **Lamborghini**, I now crash way less on the car select screen - before this PR, I'd frequently crash in `UpdateTexture` since the texture was evicted from the cache while it was still being created. With this PR it'll still happen, but the resource creation function keeps a reference count to the resource for the duration of the call, so it releases gracefully as the function is finished.

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
+++ b/src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
@@ -77,6 +77,9 @@
 #include <unordered_map>
 #include <thread>
 
+#include <wrl/client.h>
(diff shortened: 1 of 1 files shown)


LukeUsher (approved):
Excellent work; we really need to stop using raw pointers everywhere; it's easy for this kind of issue to go un-noticed

Issue #2361, opened 2022-05-15
Task: Implement I/O File System


RadWolfie:
## Quick summary
To able emulate at lowest level of kernel's APIs for file system, we need drivers for device and file system. There will be multiple edits to this post to add more details and/or functions that are in need to implement before have full completion of I/O file system.

## Details
Excluding higher level of Nt/Io implements, what will need to implement are...
Top Priority: (Managed by Object Manager)
- [x] Io(Create|Delete)Device
- [x] IoCreateFile / NtDeleteFile
- [x] Io(Create|Delete)SymbolicLink
- [x] IopParseFile (forward call to IoParseDevice)
- [x] IoParseDevice **(Part 1)**
  (currently emulated with host's `NtDll::NtCreateFile`)

High Priority: (irp work)
- [ ] IoAllocateIrp and its relative Irp functions
- [ ] Implement functions which require `IRP_MJ_` for `MajorFunction` usage.
  - [ ] Emulated device drivers **(currently stub yet has enough to emulate at IoParseDevice function)**
     - [ ] EmuDisk
     - [ ] EmuCdRom
     - [ ] EmuMu
     - [ ] EmuMediaBoard
        (NOTE: One or more hack(s) are required in order to continue support in higher level emulation)
  - [ ] IofCompleteRequest
  - [ ] ...check for any other missing implement... even if it's exported kernel APIs.
  - [ ] IoParseDevice **(Part 2)**
     (use irp process instead of direct call to host's `NtDll::NtCreateFile`)

After above priorities are done: **(assumed lowest level functions of kernel APIs)**
- [x] Io(Check|Set|Removed)SharedAccess functions
   (already implemented locally based on ReactOS)
- [ ] IoDismountVolume(ByName)
- [ ] ...check for any other missing implement... even if it's exported kernel APIs.
- [ ] file system driver (only occur after device driver)
  

### System Configuration:
N/A

### Additional Information (if any):
- Object Manager Planning for [I/O file system](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/projects/3#card-78083578)

PR #2365, opened 2022-05-21
I/O filesystem partial update

RadWolfie:
This is a major change to the filesystem.

With this fix to the filesystem we enable:
- [x] Native object manager enabled support.
- [x] Full symbolic link functionality instead of hardcode hacks for certain case and only had drive letters support.
- [x] Ability to create native devices and its interface.
  - The interfaces require irp work which are not in this pull request.
- [x] New file system is now base on host's native handle attachment to file object which no longer require xbox to host strings conversion checks anymore.
  - With a bonus benefit, we can use below functions with xbox objects to obtain their file path:
    - `Object2AbsolutePath` : Full path which use straightforward device path and file system's remaining path.
    - `Object2RemainingPath`: Like above except exclude device path. This enable support for rename functionality forward.
- [x] Additional verbose details to I/O objects and their ouputs

Two things I'll like to add:
1. There are some events coded in the structure. Except they are inactive due to object manager's event is not implemented yet.
2. Media Board support is included in this pull request. ~~Media Board will need to be implement and upstream into chihiro branch before can rebase after merge. Which is already been implemented locally.~~

See ticket #2361 for Top Priority's list of completions. Plus `Io(Check|Set|Removed)SharedAccess` functions which are currently unused and base on ReactOS.

---
These are critical titles to verify for ensure file system are still functional:
- Grand Theft Auto (3 and later): Always trigger reformat partition every time.
- Exhibition Demo Disc: Multi-xbe titles
  - Run Panzer Dragoon Orta's xbe directly is also critical to check as well. Using default will generate pdodemo folder whilst direct run PDO's xbe will generate in top partition's folder.
- Official dashboard: For verify Memory Units' and harddrive for titles' data listing. Plus edit Memory Units' names.
- Metal Arms: Glitch in the System: Create profile may fail due to user's end does not have "full access" permission. A workaround is implemented but will need to be restore once device driver interface is fully implement.
- Misc titles for verify gameplay testing.

These following titles has been noticeable made progress:
- NBA 2K2 (Sega Sports) no longer crash before reboot to main menu screen.\
  NOTE: Random soft-lock from intro still occur, even from upstream.
- Panzer Dragoon ORTA no longer crash before loading screen.

---
<details>
<summary>Experiment test code for verification used (require to be after CxbxrSetupDrives call)</summary>

```cpp
	// Begin of I/O test
	std::string file_test = DeviceHarddisk0Partition3 + "\\test.txt";
	std::string file_test_rename = DeviceHarddisk0Partition3 + "\\rename.txt";
	xbox::STRING ObjectName;
	xbox::RtlInitAnsiString(&ObjectName, file_test.c_str());
	xbox::HANDLE FileHandle, dupFileHandle;
	xbox::OBJECT_ATTRIBUTES ObjectAttributes;
	X_InitializeObjectAttributes(&ObjectAttributes, &ObjectName, NULL, NULL);
	xbox::IO_STATUS_BLOCK IoStatusBlock;
	xbox::ntstatus_xt result = xbox::NtCreateFile(&FileHandle, FILE_GENERIC_READ | FILE_GENERIC_WRITE, &ObjectAttributes, &IoStatusBlock, nullptr, FILE_ATTRIBUTE_NORMAL, 0, FILE_OVERWRITE_IF, FILE_NON_DIRECTORY_FILE | FILE_SYNCHRONOUS_IO_NONALERT);

	std::string content_write_test = "This is a test.";
	char content_read_test[std::size("This is a test.")]{};
	result = xbox::NtWriteFile(FileHandle, NULL, nullptr, nullptr, &IoStatusBlock, content_write_test.data(), content_write_test.size(), nullptr);

#if 1 // Result currently return as access denied for some reason. Need to fix this test. (PDO for verification use shown to work)
	xbox::FILE_RENAME_INFORMATION FileInformation{ .ReplaceIfExists = false, .RootDirectory = xbox::zeroptr };
	xbox::RtlInitAnsiString(&FileInformation.FileName, file_test_rename.c_str());
	result = xbox::NtSetInformationFile(FileHandle, &IoStatusBlock, &FileInformation, sizeof(FileInformation), xbox::FileRenameInformation);
#else
	const std::string& host_directory = "P:\\Github Sync\\Cxbx-Reloaded\\build-vs22\\bin\\Debug\\EmuDisk\\Partition3";
	std::string host_file_test = host_directory + "\\test.txt";
	std::string host_file_test_rename = host_directory + "\\rename.txt";
	BOOL test = MoveFileA(host_file_test.c_str(), host_file_test_rename.c_str());
#endif

	result = xbox::NtDuplicateObject(FileHandle, &dupFileHandle, DUPLICATE_CLOSE_SOURCE);

	PVOID FileObject;
	if (X_NT_SUCCESS(result)) {
		result = xbox::ObReferenceObjectByHandle(FileHandle, xbox::zeroptr, &FileObject);
	}
	if (X_NT_SUCCESS(result)) {
		const auto& XboxTargetPathFull = Object2AbsolutePath(FileObject);
		const auto& XboxTargetPathRemaning = Object2RemainingPath(FileObject);
		xbox::ObfDereferenceObject(FileObject);
		xbox::NtClose(FileHandle);
	}

	xbox::LARGE_INTEGER ByteOffset{};
	result = xbox::NtReadFile(dupFileHandle, NULL, nullptr, nullptr, &IoStatusBlock, content_read_test, content_write_test.size(), &ByteOffset);

	result = xbox::NtClose(dupFileHandle);

	result = xbox::NtDeleteFile(&ObjectAttributes);

	// Symbolic Link test
	xbox::HANDLE RootDirHandle = xbox::zeroptr;
	std::string CreateDirectories[]{
		DeviceHarddisk0Partition3 + "\\test",
		DeviceHarddisk0Partition3 + "\\test\\1",
		DeviceHarddisk0Partition3 + "\\test\\1\\2",
		DeviceHarddisk0Partition3 + "\\test\\1\\2\\3",
	};
	for (size_t i = 0; i < std::size(CreateDirectories); i++) {
		xbox::STRING ObjectName;
		xbox::RtlInitAnsiString(&ObjectName, CreateDirectories[i].c_str());
		xbox::HANDLE DirHandle;
		xbox::OBJECT_ATTRIBUTES ObjectAttributes;
		X_InitializeObjectAttributes(&ObjectAttributes, &ObjectName, NULL, NULL);
		result = xbox::NtCreateFile(&DirHandle, FILE_LIST_DIRECTORY | SYNCHRONIZE, &ObjectAttributes, &IoStatusBlock, nullptr, FILE_ATTRIBUTE_NORMAL, 0, FILE_CREATE, FILE_DIRECTORY_FILE | FILE_SYNCHRONOUS_IO_NONALERT | FILE_OPEN_FOR_BACKUP_INTENT);
		if (X_NT_SUCCESS(result)) {
			xbox::NtClose(DirHandle);
		}
	}
	std::string CreateSymbolicLinks[]{
		DeviceHarddisk0Partition3,
		DrivePrefix + "W:\\test",
		DrivePrefix + "X:\\1",
		DrivePrefix + "Y:\\2",
		DrivePrefix + "Z:\\3",
	};
	for (size_t i = 1; i < std::size(CreateSymbolicLinks); i++) {
		xbox::STRING ObjectNamePrev, ObjectNameNext;
		xbox::RtlInitAnsiString(&ObjectNamePrev, CreateSymbolicLinks[i-1].c_str());
		std::string mount_drive = CreateSymbolicLinks[i].substr(0, 6);
		xbox::RtlInitAnsiString(&ObjectNameNext, mount_drive.c_str());
		result = xbox::IoCreateSymbolicLink(&ObjectNameNext, &ObjectNamePrev);
		if (!X_NT_SUCCESS(result)) {
			break;
		}
		xbox::PVOID FileObject;
		xbox::HANDLE DirHandle;
		ObjectAttributes.ObjectName = &ObjectNameNext;
		result = xbox::NtOpenFile(&DirHandle, FILE_LIST_DIRECTORY | SYNCHRONIZE, &ObjectAttributes, &IoStatusBlock, 0, FILE_DIRECTORY_FILE | FILE_SYNCHRONOUS_IO_NONALERT | FILE_OPEN_FOR_BACKUP_INTENT);
		if (X_NT_SUCCESS(result)) {
			result = xbox::ObReferenceObjectByHandle(DirHandle, xbox::zeroptr, &FileObject);
		}
		if (X_NT_SUCCESS(result)) {
			const auto& XboxTargetPathFull = Object2AbsolutePath(FileObject);
			const auto& XboxTargetPathRemaning = Object2RemainingPath(FileObject);
			xbox::ObfDereferenceObject(FileObject);
			xbox::NtClose(DirHandle);
		}
	}
	// Deattach leftover symbolic links
	for (std::size_t i = std::size(CreateSymbolicLinks) - 1; i > 0; i--) {
		xbox::STRING SymbolicLink;
		std::string mount_drive = CreateSymbolicLinks[i].substr(0, 6);
		xbox::RtlInitAnsiString(&SymbolicLink, mount_drive.c_str());
		result = xbox::IoDeleteSymbolicLink(&SymbolicLink);
	}
	// Delete leftover folders
	for (signed i = std::size(CreateDirectories) - 1;; i--) {
		xbox::STRING ObjectName;
		xbox::RtlInitAnsiString(&ObjectName, CreateDirectories[i].c_str());
		xbox::HANDLE DirHandle;
		xbox::OBJECT_ATTRIBUTES ObjectAttributes;
		X_InitializeObjectAttributes(&ObjectAttributes, &ObjectName, NULL, NULL);
		result = xbox::NtDeleteFile(&ObjectAttributes);
		if (!X_NT_SUCCESS(result)) {
			break;
		}
		// Exit once we reach to zero.
		if (!i) {
			break;
		}
	}
	// End of I/O test
```
</details>


src/core/kernel/support/EmuMu.cpp
```diff
@@ -0,0 +1,217 @@
...
+	char MuLetter = 'F' + static_cast<char>(MuIndex);
+	PartitionDevice& MuPath = MuPartitionArray[MuIndex];
+	MuPath.HostDevicePath = g_MuBasePath + MuLetter;
+
+	bool succeeded{ true };
+	std::string partitionHeaderPath = MuPath.HostDevicePath + ".bin";
+	if (!std::filesystem::exists(partitionHeaderPath)) {
+		CxbxCreatePartitionHeaderFile(partitionHeaderPath, MuIndex == 0);
+	}
+	MuPath.HostBinHandle = CreateFileA(partitionHeaderPath.c_str(), GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
```
PatrickvL, line 154: Is their an important reason to use the ANSI API? We're living in the Unicode (widestring) world for a while already ..

src/core/kernel/support/EmuMu.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+	static ntstatus_xt NTAPI MuDeviceControl(PDEVICE_OBJECT DeviceObject, PIRP Irp)
+	{
+		LOG_UNIMPLEMENTED();
+		return X_STATUS_SUCCESS;
+	}
+
+	static DRIVER_OBJECT MuDriverObject = {
+		MuStartIo,
+		nullptr,
```
PatrickvL, line 56: Could you please add each member name per value in an end of line comment? That way code searches for these members will also find this initialization.

src/core/kernel/support/EmuMu.cpp
```diff
@@ -0,0 +1,217 @@
...
+	// Make sure we do not extend over # of MUs.
+	if (MuIndex >= MuPartitionArray.size()) {
+		return;
+	}
+
+	char MuLetter = 'F' + static_cast<char>(MuIndex);
+	PartitionDevice& MuPath = MuPartitionArray[MuIndex];
+	MuPath.HostDevicePath = g_MuBasePath + MuLetter;
+
+	bool succeeded{ true };
```
PatrickvL, line 149: Why did you not use an assignment syntax here? (`bool succeeded = true;`)

src/core/kernel/support/EmuMu.cpp
```diff
@@ -0,0 +1,217 @@
...
+	if (!CxbxrIsPathInsideEmuMu(PartitionPath)) {
+		EmuLog(LOG_LEVEL::WARNING, "Attempting to format a path that is not within a Cxbx-Reloaded data folder... Ignoring!\n");
+		return false;
+	}
+
+	// Format the MU's partition, by iterating through the contents and removing all files/folders within
+	// Previously, we deleted and re-created the folder, but that caused permission issues for some users
+	std::error_code er;
+	for (const auto& directoryEntry : std::filesystem::directory_iterator(PartitionPath, er)) {
+		if (!er) {
```
PatrickvL, line 122: Consider printing an error here from https://en.cppreference.com/w/cpp/error/error_code/message

src/core/kernel/support/EmuMu.cpp
```diff
@@ -0,0 +1,217 @@
...
+	std::string DeviceName = DeviceMUPrefix + MuX;
+	xbox::STRING xDeviceName;
+	xbox::RtlInitAnsiString(&xDeviceName, DeviceName.c_str());
+	std::transform(xDeviceName.Buffer, xDeviceName.Buffer + xDeviceName.Length, xDeviceName.Buffer, xbox::RtlLowerChar);
+	xbox::PDEVICE_OBJECT MuDeviceObject;
+	xbox::ntstatus_xt result = xbox::IoCreateDevice(&xbox::MuDriverObject, sizeof(xbox::MU_EXTENSION), &xDeviceName, xbox::FILE_DEVICE_MEMORY_UNIT, FALSE, &MuDeviceObject);
+	EmuBugCheckInline(result);
+
+	MuDeviceObject->Flags |= 0x45; // TODO: What flags are these? Doesn't seem to match with ReactOS's info: https://github.com/reactos/reactos/blob/999345a4feaae1e74533c96fa2cdfa1bce50ec5f/sdk/include/xdk/iotypes.h
+
```
PatrickvL, line 174: These empty separator lines could be removed (many others as well in this pull request)

src/core/kernel/support/EmuDisk.cpp
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	static ntstatus_xt NTAPI DriverIrpReturn(PDEVICE_OBJECT DeviceObject, PIRP Irp)
+	{
+		LOG_UNIMPLEMENTED();
+		return X_STATUS_SUCCESS;
+	}
+
+	static DRIVER_OBJECT DiskDriverObject = {
+		DiskStartIo,
```
PatrickvL, line 69: Could you please add each member name per value in an end of line comment? That way code searches for these members will also find this initialization.

src/core/kernel/support/EmuCdRom.cpp
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	static ntstatus_xt NTAPI DriverIrpReturn(PDEVICE_OBJECT DeviceObject, PIRP Irp)
+	{
+		LOG_UNIMPLEMENTED();
+		return X_STATUS_SUCCESS;
+	}
+
+	static DRIVER_OBJECT CdRomDriverObject = {
+		CdRomStartIo,
```
PatrickvL, line 62: Could you please add each member name per value in an end of line comment? That way code searches for these members will also find this initialization.

src/core/kernel/common/types.h
```diff
@@ -1327,10 +1326,72 @@ typedef struct _DEVICE_OBJECT
...
+typedef void (NTAPI *PDRIVER_STARTIO)(DEVICE_OBJECT* DeviceObject, struct _IRP* Irp);
+typedef void (NTAPI *PDRIVER_DELETEDEVICE)(DEVICE_OBJECT* DeviceObject);
+typedef void (NTAPI *PDRIVER_UNKNOWN)(DEVICE_OBJECT* DeviceObject);
+typedef ntstatus_xt (NTAPI *PDRIVER_DISPATCH)(DEVICE_OBJECT* DeviceObject, struct _IRP* Irp);
+typedef struct _DRIVER_OBJECT
+{
+	PDRIVER_STARTIO      DriverStartIo;
+	PDRIVER_DELETEDEVICE DriverDeleteDevice;
+	PDRIVER_UNKNOWN      DriverUnknown2;
+	PDRIVER_DISPATCH     MajorFunction[IRP_MJ_MAXIMUM_FUNCTION + 1]; // See below for complete list of each offset function use.
```
PatrickvL, line 1378: It might be useful to create a struct DRIVER_DISPATCH_TABLE to replace the array with, so that each offset can be accessed by name, which would give more insight during debugging (member inspection)

src/core/kernel/common/types.h
```diff
@@ -1369,6 +1430,44 @@ typedef struct _FILE_OBJECT {
...
+	ushort_xt                      FileAttributes;
+	ushort_xt                      ShareAccess;
+	ulong_xt                       Options;
+	ulong_xt                       Disposition;
+	ulong_xt                       DesiredAccess;
+	PFILE_NETWORK_OPEN_INFORMATION NetworkInformation;
+	PDUMMY_FILE_OBJECT             LocalFileObject;
+	boolean_xt                     QueryOnly;
+	boolean_xt                     DeleteOnly;
+} OPEN_PACKET, *POPEN_PACKET;
```
PatrickvL, line 1469: Consider adding end of line comments per member containing it's hexadecimal offset, plus at the end the total size of the struct

src/core/kernel/exports/EmuKrnlLogging.h
```diff
@@ -89,20 +89,21 @@ LOGRENDER_HEADER_BY_REF(PULONG);
...
 // TODO : LOGRENDER_HEADER(DVDX2_AUTHENTICATION_PAGE)
 // TODO : LOGRENDER_HEADER(ERWLOCK)
 // TODO : LOGRENDER_HEADER(ETHREAD)
 // TODO : LOGRENDER_HEADER(EXCEPTION_RECORD)
 LOGRENDER_HEADER(FILETIME)
 // TODO : LOGRENDER_HEADER(FILE_DIRECTORY_INFORMATION)
-// TODO : LOGRENDER_HEADER(FILE_FS_SIZE_INFORMATION)
-// TODO : LOGRENDER_HEADER(IO_STATUS_BLOCK) // LOGRENDER_HEADER(PIO_STATUS_BLOCK ->u1.Pointer, ->Information
-// TODO : LOGRENDER_HEADER(KDEVICE_QUEUE)
+LOGRENDER_HEADER(FILE_FS_SIZE_INFORMATION)
```
PatrickvL, line 101: Above, introduce a commented out registration for the newly introduced type `FILE_ATTRIBUTE_TAG_INFORMATION`

Other newly introduced types should be added likewise in their appropriate location

src/core/kernel/common/types.h
```diff
@@ -1327,10 +1326,72 @@ typedef struct _DEVICE_OBJECT
...
+	PDRIVER_STARTIO      DriverStartIo;
+	PDRIVER_DELETEDEVICE DriverDeleteDevice;
+	PDRIVER_UNKNOWN      DriverUnknown2;
+	PDRIVER_DISPATCH     MajorFunction[IRP_MJ_MAXIMUM_FUNCTION + 1]; // See below for complete list of each offset function use.
+	// Base on what has been discovered, all the question marks seem to only return invalid request.
+	// IRP_MJ_CREATE
+	// IRP_MJ_CLOSE
+	// IRP_MJ_READ
+	// IRP_MJ_WRITE
+	// IRP_MJ_???
```
PatrickvL, line 1384: Is the union `Parameters` in `IO_STACK_LOCATION` related to these? That maybe worth a mention in a comment here.
Referring to the various blocks inside Parameters, their names might be associated with some of the unknown ??? marked lines in here.

src/core/kernel/common/types.h
PatrickvL, line 1469: The same might be good to add to all other Xbox kernel struct types currently still lacking offsets and size indications

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -33,43 +33,106 @@
...
+std::unordered_map<xbox::PVOID, RegisteredObject> g_RegisteredPartitions;
+
+void StrHashDereference(const std::size_t& StrHash)
+{
+	if (!StrHash) {
+		return;
+	}
+	const auto& it = g_HashStrings.find(StrHash);
+	if (it != g_HashStrings.end()) {
+	}
```
PatrickvL, line 66: Please add a TODO comment or an explanation why this is empty

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -33,43 +33,106 @@
...
+	// Check if an object already has string.
+	if (ObjectHeader->Flags & OB_FLAG_NAMED_OBJECT) {
+		// Skip it.
+		return;
+	}
 
-		// If we reach here, it means that we could not insert the handle after more than two seconds of trying. This probably means
-		// that we have forgotten to call RemoveXboxHandle on the old handle, or the other thread is waiting/deadlocked, so this is a bug
-		CxbxrAbortEx(CXBXR_MODULE::CXBXR, "Failed to register new xbox handle after more than two seconds!");
+	if (string.empty()) {
```
PatrickvL, line 81: Consider performing this check first, before entering the lock

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -33,43 +33,106 @@
...
-			lck.unlock();
-			std::this_thread::sleep_for(std::chrono::milliseconds(5));
-			lck.lock();
-			ret = g_RegisteredHandles.try_emplace(xhandle, nhandle);
-			if (ret.second) {
-				return;
-			}
-			now += std::chrono::milliseconds(5);
-		}
+	POBJECT_HEADER ObjectHeader = OBJECT_TO_OBJECT_HEADER(xobject);
```
PatrickvL, line 74: Does this check really need to be in the lock? If not, move entering the lock to afterwards this check, to avoid needless contention

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -33,43 +33,106 @@
...
+	if (string.empty()) {
+		return;
+	}
+
+	// Find existing xobject and attach string to them.
+	std::size_t StrHash = std::hash<std::string>{}(string);
+	std::size_t counter{};
+	const auto& it1 = g_RegisteredObjects.find(xobject);
+	if (it1 != g_RegisteredObjects.end()) {
+		it1->second.StrHash = StrHash;
```
PatrickvL, line 90: Existing non-empty and different StrHash should decrease its reference counter, as it's about to be replaced.

Same below

src/core/kernel/support/NativeHandle.cpp
```diff
@@ -33,43 +33,106 @@
...
+		counter++;
+	}
+	const auto& it2 = g_RegisteredPartitions.find(xobject);
+	if (it2 != g_RegisteredPartitions.end()) {
+		it2->second.StrHash = StrHash;
```
PatrickvL, line 95: Also, decreasing the reference count of an existing StrHash (different from the one that's about to be assigned here) also requires handling cleanup when a reference count reaches zero!

PatrickvL (changes requested):
Only done a partial review, as this is a huge change (great to see it done though!)

Some remarks to consider already. To be continued later...

src/core/kernel/exports/EmuKrnlLogging.h
RadWolfie: Fixed in d54ea6a79782857076d2d4510b28909e5d4e9c0a commit, any further amissed will be in new review remark.

src/core/kernel/support/NativeHandle.cpp
RadWolfie: That is my mistake, a problematic area (not from this) caught my attention and forgot about finish this part of code. So, that's fixed as well. Thank you for bringing this to my attention! 😄👍

src/core/kernel/support/NativeHandle.cpp
RadWolfie: You could had selected multiple-lines block to do review remark instead of three separate comments. So I'll response to this beside other two comments.

Normally when I think of this, I always think about pre-planned to reserse the lock until there's nothing needed to update/release. For now, I will skip the lock guard over two pre-checks. **(fixed)**

src/core/kernel/support/NativeHandle.cpp
RadWolfie: I don't think I understand what you're saying. But this HandleHandle has its own mutex so none of three global variables are used at same time between threads. And this isn't place where decrease reference count occur. Perhaps take another look at the file as raw or explain it clearly for me?

src/core/kernel/common/types.h
RadWolfie: updated in review remark commit

RadWolfie: Um... duplicate of https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2365#discussion_r878840053 remark?

src/core/kernel/common/types.h
RadWolfie: Currently in netural decision, I may be okay with this. We are bringing this to the team discussion for feedback.

src/core/kernel/support/EmuMu.cpp
RadWolfie: Added NOTE comment for reason there are multiple separator lines. This also included with EmuDisk, and EmuCdRom files.

src/core/kernel/support/EmuMu.cpp
RadWolfie: Our current strings in codebase are mostly using `std::string`. Though, xbox kernel doesn't have wchar_t strings supported if I remember that right. So, we may need to change `std::string` to `std::u8string` as we know xbox kernel do have utf-8 strings for filesystem.

src/core/kernel/support/EmuMu.cpp
RadWolfie: I had seen some pull requests here that are using `var{};`, instead of `var = {};` which do perform same zero initialization. We can also do this way for set other than default zero as well. Might as well start using new C++ initialization method.

src/core/kernel/support/EmuDisk.cpp
RadWolfie: Suppressed and use https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2365#discussion_r878835825 review remark as central discussion.

src/core/kernel/support/EmuCdRom.cpp
RadWolfie: Suppressed and use https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/pull/2365#discussion_r878835825 review remark as central discussion.

src/core/kernel/support/EmuMu.cpp
RadWolfie: Right now, irp aren't implemented. Any further work for irp will need to be done in separate pull request. For now, I implemented new C++20 method to improve our readability in codebase easier. I hope new change is enough to satisfy for now. Though, I am awaiting on further discussion for union topic before stay with new fix or re-update.

src/core/kernel/common/types.h
RadWolfie: Reminder to @RadWolfie for next fixup remark.

RadWolfie:
After I notice something was off when I try to attempt copy soundtrack onto hdd from "Exhibition Demo Disc". `FileAttributes` argument has `0x88` input with `Options`' has `0x0` input. I checked out bitmask 0x8 to try figure out what it is used for. Seems like it was used in DOS' volume id flag. But `Options` argument has `0x0` input which means there is no parameter's validation check occur. For this, we need to sanction all parameters. I added `FileAtrributes` to the sanction in the new "fix Exhibition Demo Disc problem for copy soundtracks onto hdd" commit.

According to the rest of parameters, only thing left is `Disposition` have not been sanction and unlikely need any sanction.

src/core/kernel/support/EmuMu.cpp
PatrickvL: Okay. I see you've started to use designated initialization, that's fine too (even better, actually, since that will perform compile-time validation of correctly named fields).

src/core/kernel/support/EmuMu.cpp
PatrickvL: Sure thing. I was just wondering if there was a pressing need to do so, but if this is the way forward, that's fine.

src/core/kernel/support/NativeHandle.cpp
PatrickvL: I see you've addressed this in commit e173a5a177d35fad1c9517f27087010bce8d86e3. Thanks!

src/core/kernel/common/types.h
PatrickvL: I see a comment for this was added in commit bfdcd30e9e3c0ab47846e10ba0d6537afbd67cd4 - thanks!

src/core/kernel/support/NativeHandle.cpp
PatrickvL: Nowadays C++ offers [smart pointers](https://en.wikipedia.org/wiki/Reference_counting#C++ ) that could perhaps be used instead.

src/core/kernel/support/NativeHandle.cpp
PatrickvL: I was talking about avoiding memory leaks when reference-count reaches zero - normally, when that happens, the resource being reference counted no longer serves a purpose (as there's no reference to it anymore), so should be disposed.

The locking you mention is a totally separate subject, not relevant to my remark here.

src/core/kernel/common/types.h
PatrickvL: Additional remarks I had on this, copied from the team chat : 

> Obviously the nested union struct has to be the same size as the array, and have the named fields at the same offset as the corresponding indices in the array, and declare all fields of the same type as the array contains 
> The motivation for the suggestion to add a union overlay is, that it makes inspecting these members easier during debugging, because it'd be no longer necessary to remember the indices
> (Since they're function pointers, I'd expect them to show up as addresses, labelled with the corresponding symbol names)

src/core/kernel/support/NativeHandle.cpp
PatrickvL: Is see this is addressed in commit e173a5a177d35fad1c9517f27087010bce8d86e3 - thanks!

src/core/kernel/support/NativeHandle.cpp
RadWolfie: After internal discussion for clarification, I can understand clearly what Patrick is trying to say to me.

Technically, this shouldn't happen since xbox object are always unique and `AttachStringToXboxObject` should only use on new objects than re-use on existing objects. For now, I add `StrHashDereference` calls which will perform self-check if string hash is non-zero in commit 9631804.

src/core/kernel/common/types.h
RadWolfie: Last updated from team chat vote:
Opposed: 1
Netural: 1
Proposed: 1

ergo720:
Reminder for the community that this needs some testing before merging.

LanHikariDS:
PR enables custom soundtracks in all tested games (Dashboard, Burnout 3, GTA3, and GTAVC). One track bugged out in GTA3, but may have been a random fluke, will test further later.
Also seems to improve loading issues with Halo, but doesn't completely solve it
image

LanHikariDS:
Tested a bit further, sitting back and listening to my Custom Soundtrack now to double-check for playback errors, however, I do have some remarks. First and foremost, this PR seems to break stopping emulation. Every time emulation is stopped, Windows makes an error sound, and CXBX-R goes not responding for a moment. From there, a plethora of things can happen when the emulator reverts to 640x480, such as showing a black screen, restarting the current game at 640x480 (with broken aspect, likely dependant on user settings), or not closing the game at all and trimming the image. 
image
image
If it's any help, closing the emulator with an incorrectly closed game (Agent Under Fire) causes an XBE exception 
image

Moreover, I regret to say the Halo improvement seems to have been a false positive. Upon first testing this PR, I got _four_ valid loads in a row, but have since been unable to get a single valid load since.

medievil1:
I am not sure the cxbx closing issues are related to THIS PR, I have been having issues with games hanging (cxbx-ldr, not closing) multiple tries to get a game started, etc on master for a while (not sure if it was the threading pr' or the one that made cxbx-ldr the only option that I noticed it happening)

RadWolfie:
> First and foremost, this PR seems to break stopping emulation. Every time emulation is stopped, Windows makes an error sound, and CXBX-R goes not responding for a moment.

I remember has happened before. If I recall, it is not relative to this pull request issue. The cause is the shutdown process isn't fully feature complete. See #2366 ticket for this type of problem.

> From there, a plethora of things can happen when the emulator reverts to 640x480, such as showing a black screen, restarting the current game at 640x480 (with broken aspect, likely dependant on user settings), or not closing the game at all and trimming the image.

Thanks for report of this problem. It may not be caused from this pull request. I will make attempt to reproduce this problem soon.

ergo720:
> First and foremost, this PR seems to break stopping emulation.

@LanHikariDS I found the issue: it's caused by this line https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/b43f6bbcdfbd87319c805a7d6d4a67f456276aec/src/common/cxbxr.cpp#L109 If you disable it, the shutdown issue goes away, at least for me. That code has been present for a couple of months now, it's just that this PR makes the issue more obvious now.

medievil1:
> > First and foremost, this PR seems to break stopping emulation.
> 
> @LanHikariDS I found the issue: it's caused by this line
> 
> https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/b43f6bbcdfbd87319c805a7d6d4a67f456276aec/src/common/cxbxr.cpp#L109
> If you disable it, the shutdown issue goes away, at least for me. That code has been present for a couple of months now, it's just that this PR makes the issue more obvious now.

makes sense, cxbx-ldr hangs when it fails to stop. could be related to nz's soundbuffer pr too... possibly it is in the middle of flushing or similar when it gets the suspend, hangs, then causes cxbx-ldr to hang...the hang seems more prevalent when audio is streaming if I recall

LanHikariDS:
... why was this closed?

Ehren1337:
> ... why was this closed?

probably distraction

src/core/kernel/common/types.h
RadWolfie: Added in `kernel: add union to able see name of MajorFunction, address, and name of function in the debugger` commit.

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -700,71 +889,51 @@ XBSYSAPI EXPORTNUM(197) xbox::ntstatus_xt NTAPI xbox::NtDuplicateObject
...
-				}
-				else {
-					RegisterXboxHandle(*TargetHandle, NULL);
-				}
-			}
-		}
-		else {
-			status = NtDll::NtDuplicateObject(
+		// If nativeHandle is NULL, then skip NtDll::NtDuplicateObject. This happens when SourceHandle is an xbox handle with a corresponding
+		// NULL native handle (e.g. directory objects hit this case)
```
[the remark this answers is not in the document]
RadWolfie: Directory objects are handled from the Xbox's filesystem directly. There is no host directory involve from `NtCreateDirectoryObject` API. It is handled by the xbox filesystem objects which will reach to `IopParseDevice` function. It is the heart of the device driver that is handled from xbox<->host filesystem sandbox. In the future, `IopParseDevice` will be interface with the actual emulated device driver.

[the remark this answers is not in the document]
ergo720: That's not what I asked. Also, now that I check better, I see that `SourceHandle` is being checked with `GetHandleInformation`, which will surely fail `if SourceHandle == NULL`, because that's the invalid handle for Windows too. So, if we reach here, the handle is surely valid and thus the check `if (SourceHandle != NULL)` can be deleted. The comment above it is incorrect too, since it's talking about a `nativeHandle`, which has now been removed.

src/core/kernel/common/types.h
PatrickvL: Great!

src/core/kernel/exports/EmuKrnlPs.cpp
```diff
@@ -403,10 +403,13 @@ XBSYSAPI EXPORTNUM(255) xbox::ntstatus_xt NTAPI xbox::PsCreateSystemThreadEx
 		ObfReferenceObject(eThread);
 
 		KeQuerySystemTime(&eThread->CreateTime);
-		RegisterXboxHandle(*ThreadHandle, handle);
+		RegisterXboxObject(eThread, handle);
+		/*
+		// TODO: We don't need this anymore, right????????????
```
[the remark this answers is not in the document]
RadWolfie: https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/110 tests has confirmed it is able to suspend/resume thread just fine. However, Cxbx-Reloaded is not in fully passable state sometimes.

Therefore the commented out code is fully removed.

src/core/kernel/exports/EmuKrnlNt.cpp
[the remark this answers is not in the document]
RadWolfie: Extra if statement check has been removed.

src/core/kernel/exports/EmuKrnlNt.cpp
```diff
@@ -141,42 +196,34 @@ XBSYSAPI EXPORTNUM(187) xbox::ntstatus_xt NTAPI xbox::NtClose
...
+	// * Mutant
+	// * Semaphore
+	// * Fiber(?)
+	// * Event
+	// * Timer
+	// * What else?
+	// TODO: Remove "else if" statement once all items from list above is done.
+	else if (DWORD flags = 0; GetHandleInformation(Handle, &flags)) {
+		result = NtDll::NtClose(Handle);
+	}
```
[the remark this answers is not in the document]
ergo720: I went ahead and crafted an nxdk-powered app to specifically trigger the problem I was talking about. Here is the code:
```
#include <hal/xbox.h>
#include <windows.h>


void main()
{
    for (unsigned i = 0; i < 800; ++i) {
        // Create enough directory objects (handled by Ob in cxbxr) so that we collide with Windows handles (around 0x74D on my machine)
        HANDLE Handle;
        NTSTATUS Status = NtCreateDirectoryObject(&Handle, NULL);
        if (Status != STATUS_SUCCESS) {
            XReboot();
        }
    }
    
    // This will trigger the xbox/Windows handle collision
    HANDLE Handle;
    NtCreateEvent(&Handle, NULL, SynchronizationEvent, TRUE);
    NtClose(Handle);
    
    Sleep(1000);
    
    XReboot();
}

```

To be fair, I now notice that both current master and the PR are affected by the problem, so it’s an existing issue. Based on this, we can probably delay the issue for a future PR.

PR #2374, opened 2022-06-17
Fix vertex shader op independence

NZJenkins:
Ensure the MAC op does not interfere with the input of the ILU op.
- Use a temp register to store the output of the MAC op when necessary
- Reorganize vertex shader decoding to better reflect the data.
Decode one vsh instruction to one intermediate instruction, rather than to multiple independent instructions.
Test case:
KOTOR II (menu)
GTA III (lighting)

Cheers to xemu for solving this same issue a while back and medievil tracking down this was the issue with the KOTOR 1 menu (currently still broken)

GTA III sample
```c
temp = r4;
x_dp3(r4,w, r9.wyzz, r9.wyzz);
x_rsq(r1,w, temp.w);
```

### DIFF
--- a/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
+++ b/src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
@@ -323,6 +323,9 @@ VS_OUTPUT main(const VS_INPUT xIn)
 	init_v( 8); init_v( 9); init_v(10); init_v(11);
 	init_v(12); init_v(13); init_v(14); init_v(15);
 
+	// Temp variable for paired VS instruction
+	float4 temp;
+
 	// Xbox shader program)DELIMITER", /* This terminates the header raw string" // */
 
 R"DELIMITER(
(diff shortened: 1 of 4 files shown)


medievil1:
fixes sphinx and the cursed mummies disappearing/reappearing elements
Stubs the zombie, first time I know showing anything
image

breaks otogi, nightcaster (vertex explosion), memorick(trees are missing), dreamfall(menu background is a vertex explosion to the center), azurik (looks like a ship in warp..lol)
image

still testing.....

PatrickvL (changes requested):
Some remarks to look into

medievil1:
issue that breaks the above mentioned games is the fact that oPos is getting saved to temp in the middle of reverse screenspace transform... so the offset isn't subtracted from the oPos used in the divided scale....
  temp = oPos;
  x_mul(oPos,xyz, r12, c(-38));
  x_rcc(r1,x, r12.w);
  oPos = temp;
  x_dp3(r7,w, r6, r6);
  x_mad(oPos,xyz, r12, r1.x, c(-37));

medievil1:
this is another one that seems to offer no bad side, after last commit it seems to be great

Issue #2368, opened 2022-05-26
Massive slowdowns in TES III: Morrowind


gandalfthewhite19890404:
## Quick summary

Massive slowdowns in TES III: Morrowind

## Details

Massive slowdowns in TES III: Morrowind. Appears when player is moving.But when you're changing camera angle -all things are OK, NPC moving (and player is standing) - OK.
Very noticeable at Seyda Neen,FPS drops from 55 to 7 after you're start moving, after you'd stopped at place - it gonna be normal again (until starting move).
That commit cause slowdowns - https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/95b789eb2776543f3c287f6b494e0a2d577c7f9f. Tested several times, https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/5e928e508ac8c3c01d6b11c64e55a541d78484fa - good, https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/commit/95b789eb2776543f3c287f6b494e0a2d577c7f9f - bad.

### System Configuration:

OS - Windows 10 Pro
CPU - Intel Core i5-10400
GPU - NVIDIA RTX 2060 Super 8 GB
RAM - 32 GB DDR4

### Additional Information (if any):

N/A

RadWolfie:
Please ignore above mentioned to the pull request, it is completely irrelevant to this issue. Later pull request may will resolve this problem.

PR #2377, opened 2022-06-20
Fix shutdown instability

ergo720:
This disables the xbox thread suspending code that happens at shutdown, and it's the cause of all the crashes/freezes at shutdown that I have been observing for a while.

### DIFF
--- a/src/common/cxbxr.cpp
+++ b/src/common/cxbxr.cpp
@@ -105,8 +105,9 @@ bool HandleFirstLaunch()
 	g_InputDeviceManager.Shutdown();
 
 #ifdef CXBXR_EMU
+	// NOTE: this code causes freezes/crashes at shutdown, so avoid for now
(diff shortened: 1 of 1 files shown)


RadWolfie (approved):
LGTM, although I do perferred macro `#if`/`#endif` but it'll do for now.

RadWolfie:
I am requesting list of titles that had been regressed to confirm resolve slowness issue.
- [ ] [TES III: Morrowind](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2368)
- [ ] Halo
- [ ] ??? (I remember there's a title I own has slowness issue and doesn't use 100% on xbox core. Currently investigating...)

ergo720:
The fix for Steel Battalion (and possibly the other titles as well) is not included in this PR, but I will submit it separately in the future. Currently, it's in my branch here https://github.com/ergo720/Cxbx-Reloaded/tree/sb_fix

PR #2378, opened 2022-06-20
Fixed slowness in Steel Battalion caused by WaitApc

ergo720:
The problem was that we were waiting far too much time than needed inside `WaitApc` because of the hard-coded 5 milliseconds wait, even if no timeout was specified. This change fixes the excessive slowness in the title menu of Steel Battalion, and it might also fix https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2368, since `KeDelayExecution` is also implemented on top of `WaitApc`.


RadWolfie:
I am requesting list of titles that had been regressed to confirm resolve slowness issue.
- [ ] [TES III: Morrowind](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2368)
- [ ] Halo
- [x] Steel Battalion
- [ ] ??? (I remember there's a title I own has slowness issue and doesn't use 100% on xbox core. Currently investigating...)

gandalfthewhite19890404:
I do confirm - TES 3 Morrowind is fixed with this PR.

medievil1:
buffy the vampire slayer
indiana jones and the emperor's tomb
Buffy chaos bleeds
crash TWOC
Crash twinsanity
KOTOR I and II
Wreckless
Crimson skies
JSRF
Red Dead Revolver
Sphinx and the cursed mummy

all checked and no difference from master (games have issues, but not controller centered

LukeUsher (approved):

PR #2381, opened 2022-06-25
Fix mipmapped cubemaps

NZJenkins:
which had broken faces (other than the first face)
because the slice pitch calculation did not account for mipmaps.
We are iterating the mipmaps already, so just calculate the slice pitch directly
rather than trying to generalize CxbxGetPixelContainerMeasures


PatrickvL (approved):
Looks fine, thanks!

PR #2384, opened 2022-07-06
EmuException: allow skipping of instructions that trigger unhandled exceptions

LukeUsher:
This PR reworks the fatal exception dialog to have three options, instead of two, so facilitate continuing emulation after an exception occurs.

image

In many cases, this will result in more instability, however, it is helpful as a debugging tool:
some games that would otherwise be working are let down by a *single* invalid read or write, and skipping
over that instruction allows the game to be played.

This can enable further research/debugging within the title.

### DIFF
--- a/src/core/kernel/support/Emu.cpp
+++ b/src/core/kernel/support/Emu.cpp
@@ -169,12 +169,9 @@ bool EmuExceptionNonBreakpointUnhandledShow(LPEXCEPTION_POINTERS e)
 		"  Press \"Ignore\" to attempt to continue emulation.",
 		e->ExceptionRecord->ExceptionCode, e->ContextRecord->Eip);
 
-	if (result == PopupReturn::Abort)
-	{
+	if (result == PopupReturn::Abort) {
 		EmuExceptionExitProcess();
-	}
-	else if (result == PopupReturn::Ignore)
-	{
+	} else if (result == PopupReturn::Ignore) {
 		e->ContextRecord->Eip += EmuX86_OpcodeSize((uint8_t*)e->ContextRecord->Eip); // Skip 1 instruction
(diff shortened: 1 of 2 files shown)


src/core/kernel/support/Emu.cpp
```diff
@@ -157,19 +157,29 @@ bool EmuExceptionBreakpointAsk(LPEXCEPTION_POINTERS e)
...
+		"  Press \"Abort\" to terminate emulation.\n"
+		"  Press \"Retry\" to debug.\n"
+		"  Press \"Ignore\" to attempt to continue emulation.",
+		e->ExceptionRecord->ExceptionCode, e->ContextRecord->Eip);
+
+	if (result == PopupReturn::Abort)
 	{
 		EmuExceptionExitProcess();
 	}
+	else if (result == PopupReturn::Ignore)
```
PatrickvL, line 176: Same here

src/core/kernel/support/Emu.cpp
```diff
@@ -185,7 +195,7 @@ bool IsXboxCodeAddress(xbox::addr_xt addr)
 #include "distorm.h"
 bool EmuX86_DecodeOpcode(const uint8_t* Eip, _DInst& info);
 void EmuX86_DistormLogInstruction(const uint8_t* Eip, _DInst& info, LOG_LEVEL log_level);
-void genericException(EXCEPTION_POINTERS *e) {
+bool genericException(EXCEPTION_POINTERS *e) {
```
PatrickvL, line 198: Now this line is changed anyway, you might as well move the opening brace of this function to the next line as well, to confirm to local coding style.

PatrickvL (approved):

PR #2386, opened 2022-07-08
emufile: fix multi-xbe titles when running through xbox-iso-vfs

LukeUsher:
Prior to this commit, the relaunch path would become corrupted when running a game mounted with xbox-iso-vfs, leading to crashes.

The cause was using GetFinalPathNameByHandle, which resolves to GetFinalPathNameByHandleA instead of GetFinalPathNameByHandleW.


PatrickvL (approved):
Assuming `GetFinalPathNameByHandleW()` also emits a terminating null character, this looks fine as-is, although it might be good to also check for a valid return value - you decide. Approved either way.

PR #2387, opened 2022-07-12
Fix heap corruption in NVNetDevice::GetMacAddress

LukeUsher:
This could cause startup of CxbxR to fail for some users.


LukeUsher:
> 

I realise I was unclear, I meant the original issue could cause that :P

CookiePLMonster (approved):

PR #2389, opened 2022-07-17
fix DOT_STR_3D/CUBE mode

medievil1:
when a cube texture is used and dot_str_3d is the texture mode, it is suppose to be flagged by combiner for change to dot_str_cube.
this is not happening. So I added a check for that mode. 

If PS_TEXTUREMODES_DOT_STR_3D is set and a cube texture is used, the mode is changed to  PS_TEXTUREMODES_DOT_STR_CUBE

 test case:
Crimson skies
the plane is textured with this fix

before:
image

after:
image

WORLD RACING

master:
image

PR:
image

### DIFF
--- a/src/core/hle/D3D8/XbPixelShader.cpp
+++ b/src/core/hle/D3D8/XbPixelShader.cpp
@@ -489,10 +489,6 @@ typedef struct s_CxbxPSDef {
 
 	void AdjustTextureModes(DecodedRegisterCombiner &RC)
 	{
-		if (RC.PSTextureModes[3] == PS_TEXTUREMODES_DOT_STR_3D) {
-			RC.TexModeAdjust = true;
-		}
-
 		// if this flag is set, the texture mode for each texture stage is adjusted as follows:
 		if (RC.TexModeAdjust) {
 			for (int i = 0; i < xbox::X_D3DTS_STAGECOUNT; i++) {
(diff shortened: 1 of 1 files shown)


src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -489,6 +489,10 @@ typedef struct s_CxbxPSDef {
 
 	void AdjustTextureModes(DecodedRegisterCombiner &RC)
 	{
+		if (RC.PSTextureModes[3] == PS_TEXTUREMODES_DOT_STR_3D) {
+			RC.TexModeAdjust = true;
```
PatrickvL, line 493: Please insert these comments here to clarify the reason for this special case :
```
// Test-case:  MS-033 Crimson Skies (Plane texturing in-game and selection menu)
// HACK: use the TexModeAdjust path to downgrade PS_TEXTUREMODES_DOT_STR_3D to PS_TEXTUREMODES_DOT_STR_CUBE for cube textures.
```
Alternatively, handle this scenario in the else block.

PatrickvL (changes requested):
Please also add a "before" screenshot to the pull request description, so that the effect of the change can be clearly seen by comparing with the already present "after" screenshot.

PatrickvL (approved):
Looks good to me now!

Issue #981, opened 2018-03-10
std::this_thread::sleep_for (KeStallExecutionProcessor) resolution is inaccurate


Nukem9:
With VS2017 `sleep_for` calls the Win32 `Sleep` function internally. I only found this in KeStallExecutionProcessor, but I was thinking it might cause other problems in the future. 
It seems to be rounded up to the nearest millisecond (excluding 0). Another reference: https://gha.st/short-sleeps/

https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/3c201d74e7afe8bb1b2f78fc1a67af0c0d04614a/src/CxbxKrnl/EmuKrnlKe.cpp#L1672

PatrickvL:
Note, that reference only test nanoseconds and milliseconds, but not our use-case : microseconds. So until we have test-results from that, we should probably leave it as-is for now.

PatrickvL:
Additional resource (some maybe obsolete/outdated) : 
* http://joeduffyblog.com/2006/08/22/priorityinduced-starvation-why-sleep1-is-better-than-sleep0-and-the-windows-balance-set-manager/
* https://social.msdn.microsoft.com/Forums/en-US/d7071ba4-8962-43c6-975a-28cdbce51548/whats-the-difference-between-threadsleep0-and-threadyield?forum=csharplanguage
* https://www.gamedev.net/forums/topic/643255-making-a-busy-loop-not-consume-that-much-cpu-without-sleep/
* https://code-examples.net/en/q/1591fe

PatrickvL:
This issue applies to more than KeStallExecutionProcessor() : almost all waits we have in our code-base should be reconsidered.

There's so many things to take into account here;

* std::this_thread::sleep_for is implemented (until now) using the Windows Sleep API, and could call Sleep() with any number of milliseconds (including zero)
* Sleep(0) however raises thread priority after roughly 4 seconds, resulting in higher CPU load
* Sleep(1) doesn't do that, but even then
* Sleep() has a relatively low accuracy (around 15 milliseconds)
* waiting for shorter periods of time should be possible, but then the timer used to measure the elapsing of time must have a high enough precision to allow accurate waiting
* waiting for longer periods of time shouldn't cost 100% CPU load
* SwitchToThread() gives back the remaining time-slice of the active thread - but there's no telling how long it'll take until the thread is re-scheduled to run 

And that's only part of the problem.

[...]

PR #2390, opened 2022-07-20
Make the wait of KeStallExecutionProcessor more accurate

ergo720:
Fixes https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/981. This function is called from the dashboard when opening the xbox live menu. Adding some time measuring code to it, shows that we are waiting between 1.2 and 4 ms inside it, while the function only requested waits of 10, 50 and 500us. This PR fixes it by implementing the function with a busy wait, which is how Microsoft says it should be done. With this, I'm measuring average waits of 10.2, 50.3 and 500.1us now.

### DIFF
--- a/src/common/Timer.cpp
+++ b/src/common/Timer.cpp
@@ -69,6 +69,24 @@ void SleepPrecise(std::chrono::steady_clock::time_point targetTime)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


PatrickvL:
Have you looked at this? https://doxygen.reactos.org/d0/daa/pchw_8c.html#a87ed493aa6c2817b417a278fd3f60610

PR #2393, opened 2022-08-26
xdm, xdd, xmd component count fix

medievil1:
all 3 use 3 components but on cxbx they were using 4
discovered by NZJenkins
test case Morrowind water
before 
image

after
image
image

water in crimson skies improved
image


PatrickvL:
Good find you both! Thanks for this dedication and perseverance!

PatrickvL (approved):

PR #2398, opened 2022-09-14
hasher: use xxh3 exclusively

LukeUsher:
This PR removes support for the crc32c hash algorithm and uses xxh3 exclusively.

Back when we first implemented xxhash, we used xxh32, but only as a fallback when hardware crc32c was not available.

For a long time now we have been using xxh3, which is significantly faster than even hardware accellerated crc32c.

This should provide a decent performance improvement in games that were bottlenecked by resource hash performance.


src/common/util/hasher.h
```diff
@@ -27,8 +27,7 @@
 #ifndef _HASHER_H
 #define _HASHER_H
 
-#include <stdint.h>
-
-uint64_t ComputeHash(const void* data, size_t len);
+#include "xxhash.h"
+#define  ComputeHash XXH3_64bits
```
LukeUsher, line 31: Note that this was done to allow the possibility to swap out the hash algorithm in future, without updating every caller of ComputeHash.

PatrickvL (approved):
Looks good to me

PR #2401, opened 2022-09-24
correct hemisphere formula

medievil1:
correct hempisphere formula.
the old formula didn't adjust H and L for sign range
after discussion with PatrickVL and his explanation of the formula, I was able to apply it correctly to the dotmapping.
for all hemisphere HL calculations the 16bit component is treated as signed and divided by 127, however normal non hemisphere HL is divided by 255 and not treated as signed.
before:
image
 
 after, per-pixel and per-vertex hemisphere mapping is visibly identical
 after:
 
image


LukeUsher (approved):

LukeUsher:
Fuck it, it's been 19 days with no feedback, I'm going to merge this.

If there were any issues, we'll see them come in afterwards.

The lack of users willing to comment and test PRs is just frustrating.

PR #2403, opened 2022-10-05
HOTFIX: Fix LibusbDevice's initialization process

### LINKED ISSUE #2293: crash to desktop when configuring inputs
## Quick summary
cxbx crashes when configuring input devices.

## Details
Latest version (4457d11) crashes after opening input configuration and either configuring one of the ports or by exiting the input config.

### System Configuration:
* Windows 10 (build 1904.867)
* Ryzen 3700X
* Nvidia GeForce GTX 970

### Additional Information (if any):
Event viewer displays an Error in the Application logs.

> Faulting application name: cxbx.exe, version: 0.0.0.0, time stamp: 0x618aaf9c
Faulting module name: cxbx.exe, version: 0.0.0.0, time stamp: 0x618aaf9c
Exception code: 0xc000041d
Fault offset: 0x04425bbb
Faulting process id: 0x2acc
Faulting application start time: 0x01d7dcb05d20c65e
Faulting application path: E:\GAMES\Xbox\cxbxr\cxbx.exe
Faulting module path: E:\GAMES\Xbox\cxbxr\cxbx.exe
Report Id: 505b0a1f-e645-4b29-8fc3-8362bf98c8f5
Faulting package full name: 
Faulting package-relative application ID: 

And the Information event before the crash

> Fault bucket , type 0
Event Name: APPCRASH
Response: Not available
Cab Id: 0

> Problem signature:
P1: cxbx.exe
P2: 0.0.0.0
P3: 618aaf9c
P4: cxbx.exe
P5: 0.0.0.0
P6: 618aaf9c
P7: c0000005
P8: 04425bbb
P9: 
P10: 

> Attached files:

> These files may be available here:
\\?\C:\ProgramData\Microsoft\Windows\WER\ReportQueue\AppCrash_cxbx.exe_38e38630c005f771188b3ac5bdd85c41d1e2bbc_1db3656f_9b9290c5-d56c-4765-9dee-c98ca5d9b0dd

> Analysis symbol: 
Rechecking for solution: 0
Report Id: 933ae29f-9b3f-47e7-911a-a3dc48780df0
Report Status: 4
Hashed bucket: 
Cab Guid: 0

The file mentioned above doesn't get created.

RadWolfie:
The following fixes should resolve third-party devices conflict:
- whitelist xbox devices didn't compare against user's device. device's vendor get overridden instead.
- Missing check for any error occur after call libusb_open. In case of original xbox device could not be opened. Likely caused by missing WinUSB driver installed and allocated with original xbox device.

fixes #2293

This pull request will resolve any unsupported devices, aka not a real original xbox device, has matching product id from whitelist to not be accepted. This include but not limited to such as DS4 controller.

### DIFF
--- a/src/common/input/LibusbDevice.cpp
+++ b/src/common/input/LibusbDevice.cpp
@@ -185,9 +185,18 @@ namespace Libusb
 
 		if (m_Type == XBOX_INPUT_DEVICE::DEVICE_INVALID) { return; }
 
-		// Duke, S and SBC have 1 configuration, 1 interface and 2 endpoints (input and output) and use the default alternate setting zero.
-		// The code below assumes that third-party controllers follow suit.
-		if (libusb_open(Dev, &m_hDev) == 0) {
+		// check if we are able to open device through libusb
+		if (int err = libusb_open(Dev, &m_hDev)) {
+			// Couldn't open device, create an error log report then don't use it.
+			EmuLog(LOG_LEVEL::ERROR2, "Unable to open original xbox device \"%s\" (%hX:%hX), libusb's error was: %s",
+				m_Name.c_str(), Desc->idVendor, Desc->idProduct, libusb_strerror(err));
+			m_Type = XBOX_INPUT_DEVICE::DEVICE_INVALID;
+			return;
+		}
+		// If we are able to open device, continue with query process.
+		else {
+			// Duke, S and SBC have 1 configuration, 1 interface and 2 endpoints (input and output) and use the default alternate setting zero.
+			// The code below assumes that third-party controllers follow suit.
 			libusb_config_descriptor *ConfDesc;
 			if (libusb_get_active_config_descriptor(Dev, &ConfDesc) == 0) {
 				if (ConfDesc->bNumInterfaces == 1) {
--- a/src/common/input/LibusbDevice.cpp
+++ b/src/common/input/LibusbDevice.cpp
@@ -220,7 +220,7 @@ namespace Libusb
 								}
 							}
 							EmuLog(LOG_LEVEL::INFO, "Out endpoint %s", m_HasEndpointOut ? "present" : "not present");
-							if (int err = libusb_claim_interface(m_hDev, m_IfaceNum) != 0) {
+							if (int err = libusb_claim_interface(m_hDev, m_IfaceNum)) {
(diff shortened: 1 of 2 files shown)


ergo720:
We'll wait for test results from @TakkuZakku before merging this.

TakkuZakku (changes requested):
Confirmed changes prevent crash, and unsupported device with no driver installed is marked as invalid device.

However, shouldn't the modified if statement have parenthesis around "int err = libusb_open(...)"? err came back as 1 instead of -12 as it is currently, looks like it was set to the boolean value of the comparison instead of the result of libusb_open.

src/common/input/LibusbDevice.cpp
```diff
@@ -185,9 +185,18 @@ namespace Libusb
 
 		if (m_Type == XBOX_INPUT_DEVICE::DEVICE_INVALID) { return; }
 
-		// Duke, S and SBC have 1 configuration, 1 interface and 2 endpoints (input and output) and use the default alternate setting zero.
-		// The code below assumes that third-party controllers follow suit.
-		if (libusb_open(Dev, &m_hDev) == 0) {
+		// check if we are able to open device through libusb
+		if (int err = libusb_open(Dev, &m_hDev) != 0) {
```
PatrickvL, line 189: Another review mentioned this should read:

`if ((int err = libusb_open(Dev, &m_hDev)) != 0) {`

If that hides the scope of `err`, the declaration and assignment should precede the `if`

PatrickvL (changes requested):
Great teamwork, everyone!

src/common/input/LibusbDevice.cpp
ergo720: Or better yet, just remove the != 0 since it's superfluous here and causes this error.

RadWolfie: @TakkuZakku, yes you are correct for comparison check unintentionally set boolean to err variable.

The solution was found pretty late at night and didn't realize this mistake was occur since it was a loan from `if (int err = libusb_claim_interface(m_hDev, m_IfaceNum) != 0) {` some lines later. I agree with @ergo720's solution and have applied the correction on one existing commit and add another commit to fix the latter code problem.

ergo720 (approved):

PR #2406, opened 2022-11-19
xbe: fix version number formatting

LukeUsher:
While working on Insignia, we learned the correct formatting of the Xbe certificate version number field.

It actually consists of two values: a Disc version, and an Update/Patch version.

This PR implements this behaviour and renders version numbers as vX.Y where X is the disc version, and Y is the patch version.

### DIFF
--- a/src/common/xbe/Xbe.cpp
+++ b/src/common/xbe/Xbe.cpp
@@ -825,5 +825,15 @@ XbeType Xbe::GetXbeType()
 	return XbeType::xtRetail;
 }
 
+uint32_t Xbe::GetDiscVersion()
+{
+    return m_Certificate.dwVersion & 0xFF;
+}
+
+uint32_t Xbe::GetPatchVersion()
+{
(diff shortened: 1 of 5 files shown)


PatrickvL (approved):
Looks good to me. Will this impact the compatibility database?

RadWolfie:
I made a comparison between this build against compatibility database. There is an impact at this time.
NASCAR Heat 2002 shows as v1.01 on compatibility whilst this build show as v1.0.
Splinter Cell show as v9.0 (compatibility: v1.09)
All Splinter Cell 2+ show as v2.0 (compatibility: v1.02)

According to internal discussion and feedbacks, the patch version increment only occur by getting patch from xbox live downloader. The only exception is the dashboard that has xbox live included from what I'm told.

LukeUsher:
The website can be fixed after this is merged, there's little point doing one without the other

RadWolfie (approved):
Approved for compatibility of having version return as hex for fallback situation and reflected to correct version format.

Issue #2407, opened 2023-02-07
Cxbx ignores PAL EEPROM settings if i load Nightfire EU version


StenApp:
## Quick summary

Cxbx ignores if i set EEPROM settings to PAL/German/25fps

## Details

If i set Region to rest of the world (Europe!), AV Region to PAL 50Hz and language to German. Cxbx ignores this and hangs before reaching the game menu of Nightfire EU version. Debugger says loading something from NTSC!!!! folder and 30_fps!!! folder.

### System Configuration:

* OS Win 10 22H2 64Bit
* CPU Intel i7-7820HQ 2,9GHz
* GPU Radeon 560 Pro
* 16GB RAM

PR #2408, opened 2023-02-07
cxbxkrnl: avoid region patching loaded titles

LukeUsher:
While it sounds ideal from a UX standpoint, region patching does break a number of titles that would otherwise work.

Instead, show a warning that it may not be compatible with instructions on how to configure region settings in eeprom.

Allow the user to attempt to run the title anyway, if the game does not do it's own region checking, it will most likely just work regardless.

This PR was created to address a user issue on our Discord server:

As a result of region patching, the PAL version of 007 Nightfire assumes it's the NTSC version and fails to boot while attempting to access files only present on NTSC versions of the game.

### DIFF
--- a/src/common/xbe/Xbe.cpp
+++ b/src/common/xbe/Xbe.cpp
@@ -732,25 +732,29 @@ void Xbe::PurgeBadChar(std::string& s, const std::string& illegalChars)
 	}
 }
 
-const char *Xbe::GameRegionToString()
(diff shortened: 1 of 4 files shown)


StenApp:
https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/issues/2407#issue-1574995412

ergo720 (approved):

PR #2412, opened 2023-03-01
Fix a priority bug in KeQueryBasePriorityThread and KeSetBasePriorityThread

ergo720:
The existing implementation is wrongfully setting/querying the priority of the current thread, instead of the specified one. This fixes the deadlock on Metal Slug 3, which can now boot without the speed hack. It also somewhat improves the boot of Halo 2, which sometimes can now reach the title menu without the hack too. @RadWolfie reports that it also fixes the deadlock in GTA 3 and GTA Vice City too.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlKe.cpp
+++ b/src/core/kernel/exports/EmuKrnlKe.cpp
@@ -1379,7 +1379,7 @@ XBSYSAPI EXPORTNUM(124) xbox::long_xt NTAPI xbox::KeQueryBasePriorityThread
 	KiLockDispatcherDatabase(&OldIrql);
 
 	// It cannot fail because all thread handles are created by ob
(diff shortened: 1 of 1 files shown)


LukeUsher (approved):

RadWolfie:
With addition to combined pull requests merged, I can confirm Crazy Taxi 3: High Roller can now get to menu instantly instead of wait long time. However, if choose rule & vehicle to enter in-game. I am approached with "now loading..." blinking with steady 60 fps. At least it is one step further progress. I had tested Crazy Taxi 3 long ago and it was able to reach in-game except can't see anything other than exploding textures.

gandalfthewhite19890404:
Hello!
Can you check  GTA San Andreas on weak PC without all-cores-hack?

RadWolfie:
GTA San Andreas didn't have any change from this combined pull request improvement.

PR #2414, opened 2023-03-01
kernel: fix hacked windows handle check to bypass special handle of current process

RadWolfie:
After combined pull requests were merged recently. Area 51 will crash due to ObReferenceObjectByHandle and KeSetBasePriorityThread functions received Handle/Thread input as -1. This cause KeSetBasePriorityThread to crash due to invalid pointer from Thread input.

I ran a test on xbox hardware and received invalid handle status if I attempt to use -1 for Handle with and without thread object type to ObReferenceObjectByHandle function. This pull request made sure to force invalidate -1 value to skip Windows' handle check if it is valid or not.

I made little bit clean up on ObReferenceObjectByHandle function with comments.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlOb.cpp
+++ b/src/core/kernel/exports/EmuKrnlOb.cpp
@@ -1023,41 +1023,52 @@ XBSYSAPI EXPORTNUM(246) xbox::ntstatus_xt NTAPI xbox::ObReferenceObjectByHandle
 	PVOID Object;
 	POBJECT_HEADER ObjectHeader;
 
+	// Check if Handle contain special handle for current thread.
 	if (Handle == NtCurrentThread()) {
-		if ((ObjectType == &PsThreadObjectType) || (ObjectType == NULL)) {
+		// We only accept either thread or null object type.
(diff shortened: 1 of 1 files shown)


src/core/kernel/exports/EmuKrnlOb.cpp
```diff
@@ -1023,42 +1023,53 @@ XBSYSAPI EXPORTNUM(246) xbox::ntstatus_xt NTAPI xbox::ObReferenceObjectByHandle
...
 			return X_STATUS_SUCCESS;
-		} else {
+		}
+		else {
 			result = STATUS_OBJECT_TYPE_MISMATCH;
 		}
 	} else {
 		Object = ObpGetObjectHandleReference(Handle);
 
-		if (Object != NULL) {
```
PatrickvL, line 1039: This part of the pull request isn't strictly needed, it only leads to a bigger change while being functionally identical. Please consider keeping the otherwise untouched code block in the same place as before. That wat the actual fix becomes more apparent

PatrickvL (dismissed):
One suggestion

ergo720 (approved):

PR #2416, opened 2023-03-02
Dpc recursion fix

ergo720:
The current implementation can incur in a dpc recursion issue, which happens when the currently executing dpc queues another one, and we call it right away. If this happens too many times, the title will eventually crash with a stack overflow exception sooner or later. This fixes a crash on boot in Halo 2 (when not using the speed hack), which was caused by the previous wrong behavior.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlKe.cpp
+++ b/src/core/kernel/exports/EmuKrnlKe.cpp
@@ -96,11 +96,10 @@ namespace NtDll
 // TODO : Move towards thread-simulation based Dpc emulation
 typedef struct _DpcData {
 	CRITICAL_SECTION Lock;
-	HANDLE DpcEvent;
 	xbox::LIST_ENTRY DpcQueue; // TODO : Use KeGetCurrentPrcb()->DpcListHead instead
 } DpcData;
 
-DpcData g_DpcData = { 0 }; // Note : g_DpcData is initialized in InitDpcThread()
+DpcData g_DpcData = { 0 }; // Note : g_DpcData is initialized in InitDpcData()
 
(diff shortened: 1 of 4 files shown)


src/core/kernel/exports/EmuKrnlKe.cpp
```diff
@@ -1268,7 +1268,9 @@ XBSYSAPI EXPORTNUM(119) xbox::boolean_xt NTAPI xbox::KeInsertQueueDpc
 		InsertTailList(&(g_DpcData.DpcQueue), &(Dpc->DpcListEntry));
 		// TODO : Instead of DpcQueue, add the DPC to KeGetCurrentPrcb()->DpcListHead
 		// Signal the Dpc handling code there's work to do
-		HalRequestSoftwareInterrupt(DISPATCH_LEVEL);
+		if (g_DpcData.IsDpcActive.test() == false) {
```
PatrickvL, line 1271: This could use `IsDpcActive()`

src/core/kernel/exports/EmuKrnlKe.cpp
```diff
@@ -1289,7 +1291,11 @@ XBSYSAPI EXPORTNUM(121) xbox::boolean_xt NTAPI xbox::KeIsExecutingDpc
 {
 	LOG_FUNC();
 
+	// This is the correct implementation, but it doesn't work because our Prcb is per-thread instead of being per-processor
+#if 0
```
PatrickvL, line 1295: Beter move the `#if 0` above this comment and put an `#else` between the preferred and current implementations, to keep it nice and clean

PatrickvL (changes requested):
Small suggestion

PatrickvL (approved):

PR #2417, opened 2023-03-06
Update XbSymbolDatabase fix and add LTCG patch missing

RadWolfie:
The changes contain for XbSymbolDatabase is mainly to:
- purge some false positive detection
- fix Grand Theft Auto: San Andreas graphic regression

Beside that, I have sync symbol renames alignment and add missing patch for `D3DDevice_DrawVertices_8__LTCG_eax3` variant recently discovered. Some LTCG titles should receive graphic improvement if they contain `D3DDevice_DrawVertices_8__LTCG_eax3` in cache file.


PatrickvL (approved):
Looks good to me

PR #2418, opened 2023-03-09
kernel: fix non-ansii file path conversion for reboot process

RadWolfie:
Somehow NASCAR Heat 2002 no longer reboot properly. After reviewing the code conversion, I'm determined to drop the current method and use c++17 filesystem's conversion by making request of std::string. Oddly enough, it is willing to do so but needs testing to be sure if this pull request change is enough to handle it.

Please test with titles that does multi-xbe functionality.


CookiePLMonster:
I see, SDL was used here which specifically goes for ASCII - but it still may be a better idea to forgo any conversion to begin with.

ergo720:
I tried Tenchu, Dead or Alive Ultimate and Innocent tears, and they all rebooted properly.

PatrickvL (approved):
LGTM

RadWolfie:
Tested with symbols below and works fine for me by using NASCAR Heat 2002's reboot processing to verify host's directory didn't get corruption.
- ™
- !
- ²

PR #2428, opened 2023-09-03
fix COMMAND_TYPE_CALL pushbuffer command handling

jackchentwkh:
reset subr_active flag to indicate we've returned from any COMMAND_TYPE_CALL command.

NV2A used COMMAND_TYPE_JUMP_LONG to return from COMMAND_TYPE_CALL. Otogi uses lot's of COMMAND_TYPE_CALL, this should inprove the pushbuffer handling.


PatrickvL (approved):
Looks good to me, thanks!

PR #2432, opened 2023-09-20
CI: Upgrade checkout to v4

Margen67:
See commits.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -32,7 +32,7 @@ jobs:
         configuration: [Release, Debug]
         vsver: [2022]
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


RadWolfie:
As usual, like what PatrickvL said before:
> Please add a description

ergo720:
@Margen67 I think you should just have linked to this article [here](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20), which explains why the action upgrade is needed (the action was updated to use node20 instead of node16, which is being phased out). Note that the other actions we use (labeler, download-artifact, upload-artifact, and my auto-close-issues fork), should also be updated in the future too. However, since those haven't updated to node20 yet, we can merge this for the time being.

ergo720 (approved):

PR #2436, opened 2023-10-05
experiment for HLE D3DDevice_DrawRectPatch()

jackchentwkh:
this is a draft to share my experiment with D3DDevice_DrawRectPatch()
to make D3DDevice_DrawRectPatch() run correctly, a vertex shader/vertex format and stream source/vertex buffer must be setup prior to calling it.
d3d9 seems unhappy when pRectPatchInfo == nullptr. so I implement a map cache and retrieve pRectPatchInfo with handle when pRectPatchInfo passed in was nullptr.
the returned hRet is S_OK, but there is no rendering output at all. maybe someone can pick from here and find the solution for it.


jackchentwkh:
try to convert RectPatch to indexed triangle strips. result looks ok, but the mesh is in low order.
image

PR #2437, opened 2023-10-07
Slightly reduce build time

NZJenkins:
Building with Visual Studio build insights highlighted time spent processing `filesystem`.
It turns out this is imported in places it's not used - removing them marginally sped up full rebuilds.

Rebuild timing:
before; 01:24.535 minutes
after: 01:07.085 minutes


PatrickvL:
Nice! That's 1/5th less waiting time (20.64% reduction) for each full build

CookiePLMonster (approved):

PatrickvL (approved):

PR #2439, opened 2023-11-09
Hotload shaders

NZJenkins:
Hotload shaders. Just for fun, not sure if this will be useful.
But also consolidated the shader code a bit.

- All shaders are now read from files (instead of some being inline strings or #included strings)
- `ShaderSources` in `Shader.h` has all the shader text for both pixel and vertex shaders, and keeps track of if they were updated.
- On start, Cxbx will watch the HLSL directory for changes.
If a change is detected, shader caches are flushed and _all_ shaders are reloaded.
This is not wired up to an option so can't be switched off.
- `./hlsl/backup` contains backup copies so it's easy to revert changes if someone forgets to make backups or what changes they made etc.

- [x] Add safety net e.g. include backup copies of shaders in cmake?


src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -2190,7 +2195,7 @@ static void CreateDefaultD3D9Device
     DWORD BehaviorFlags;
     if((g_D3DCaps.DevCaps & D3DDEVCAPS_HWTRANSFORMANDLIGHT) && g_EmuCDPD.DeviceType == D3DDEVTYPE_HAL)
     {
-        EmuLog(LOG_LEVEL::DEBUG, "Using hardware vertex processing");
+        EmuLog(LOG_LEVEL::DEBUG, "Using hard ware vertex processing");
```
PatrickvL, line 2198: Why was a space inserted between hard and ware?

src/core/hle/D3D8/XbPixelShader.cpp
```diff
@@ -1010,6 +1008,7 @@ void DxbxUpdateActivePixelShader() // NOPATCH
   // manually read from D3D__RenderState[X_D3DRS_PSTEXTUREMODES] for that one field.
   // See D3DDevice_SetPixelShaderCommon which implements this
 
+  // TODO hlsl hotloading support
```
PatrickvL, line 1011: What does this TODO need, specifically? Or is it a left-over that has already been done?

src/core/hle/D3D8/Direct3D9/Shader.cpp
```diff
@@ -140,3 +147,176 @@ extern HRESULT EmuCompileShader
...
+				break;
+			}
+		}
+
+		EmuLog(LOG_LEVEL::DEBUG, "Shader file watcher exiting...");
+
+		// until there is a way to disable hotloading
+		// this is always an error
+		FindCloseChangeNotification(changeHandle);
+		return 1;
```
PatrickvL, line 318: For consistency, also do `fsWatcherThread = 0;`

src/core/hle/D3D8/Direct3D9/Shader.cpp
```diff
@@ -140,3 +147,176 @@ extern HRESULT EmuCompileShader
...
+				// Will usually be at least two - one for the file and one for the directory
+				while (true) {
+					FindNextChangeNotification(changeHandle);
+					if (WaitForSingleObject(changeHandle, 100) == WAIT_TIMEOUT) {
+						break;
+					}
+				}
+
+				EmuLog(LOG_LEVEL::DEBUG, "Change detected in shader folder");
+				g_ShaderHlsl.shaderVersionOnDisk += 1;
```
PatrickvL, line 305: Why not just `++`?

src/core/hle/D3D8/Direct3D9/Shader.cpp
```diff
@@ -140,3 +147,176 @@ extern HRESULT EmuCompileShader
...
+		HANDLE changeHandle = FindFirstChangeNotification(hlslDir.string().c_str(), false, FILE_NOTIFY_CHANGE_LAST_WRITE);
+
+		if (changeHandle == INVALID_HANDLE_VALUE) {
+			DWORD errorCode = GetLastError();
+			EmuLog(LOG_LEVEL::ERROR2, "Error initializing shader file watcher: %d", errorCode);
+
+			return 1;
+		}
+
+		while (true) {
```
PatrickvL, line 291: Is there a global "shutdown-requested" boolean that could be checked here?

projects/misc/batch.cmake
```diff
@@ -28,13 +28,8 @@ message("Runtime Build Directory: ${TargetRunTimeDir}")
...
 file(COPY ${CXBXR_GLEW_DLL} DESTINATION  ${TargetRunTimeDir})
 
-# Copy certain HLSL files to the output directory, which we will load at runtime
-set(CXBXR_HLSL_FILES
-"${CMAKE_SOURCE_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShaderState.hlsli"
-"${CMAKE_SOURCE_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionVertexShader.hlsl"
-"${CMAKE_SOURCE_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsli"
-"${CMAKE_SOURCE_DIR}/src/core/hle/D3D8/Direct3D9/FixedFunctionPixelShader.hlsl"
-)
+# Copy HLSL files to the output directory, which we will can load at runtime
```
PatrickvL, line 31: "will can" - What's it gonna be? ;)

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1124,54 +1126,70 @@ IDirect3DVertexDeclaration* CxbxCreateHostVertexDeclaration(D3DVERTEXELEMENT *pD
...
+		if (FAILED(hRet)) CxbxrAbort("Failed to create shader: %s", label);
+	}
+
+	return shader;
+}
+
 void CxbxUpdateHostVertexShader()
 {
 	extern bool g_bUsePassthroughHLSL; // TMP glue
+	// TODO move d3d9 state to VertexShader.cpp
```
PatrickvL, line 1146: Would it be much work to actually do this?

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1124,54 +1126,70 @@ IDirect3DVertexDeclaration* CxbxCreateHostVertexDeclaration(D3DVERTEXELEMENT *pD
...
+
+		g_VertexShaderCache.Clear();
+
+		if (fixedFunctionShader) {
+			fixedFunctionShader->Release();
+			fixedFunctionShader = nullptr;
+		}
+		fixedFunctionShader = InitShader(EmuCompileFixedFunction, "Fixed Function Vertex Shader");
+
+		if (passthroughShader) {
```
PatrickvL, line 1166: Would there be benefit in updating these separately, only when requested? That could avoid a needless compile when only one of the two is ever accessed.

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1124,54 +1126,70 @@ IDirect3DVertexDeclaration* CxbxCreateHostVertexDeclaration(D3DVERTEXELEMENT *pD
 	return pHostVertexDeclaration;
 }
 
-static IDirect3DVertexShader* passthroughshader;
+IDirect3DVertexShader* InitShader(void (*compileFunc)(ID3DBlob**), const char* label) {
```
PatrickvL, line 1129: Can't this stay static?

src/core/hle/D3D8/Direct3D9/Shader.h
```diff
@@ -10,3 +10,38 @@ extern HRESULT EmuCompileShader
 	ID3DBlob** ppHostShader,
 	const char* pSourceName = nullptr
 );
+
+struct ShaderHlsl {
```
PatrickvL, line 14: With the rightful rename of `VertexShaderSource` into `VertexShaderCache`, it's perhaps good to change this name into `ShaderSources`, given that this type doesn't manage compiled shaders but the underlying shader (template) source codes (the fact they're in HLSL syntax is less important).

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
```diff
@@ -290,201 +288,34 @@ extern HRESULT EmuCompileVertexShader
...
-	assert(pIntermediateShader->Instructions.size() > 0);
+	hlsl_stream << g_ShaderHlsl.vertexShaderTemplateHlsl[0]; // Start with the HLSL template header
 	BuildShader(pIntermediateShader, hlsl_stream);
-
-	hlsl_stream << hlsl_template[1]; // Finish with the HLSL template footer
+	hlsl_stream << g_ShaderHlsl.vertexShaderTemplateHlsl[1]; // Finish with the HLSL template footer
 	std::string hlsl_str = hlsl_stream.str();
 
-	HRESULT hRet = EmuCompileShader(hlsl_str, g_vs_model, ppHostShader, "CxbxVertexShaderTemplate.hlsl");
+	auto notionalSourceName = "CxbxVertexShaderTemplate.hlsl";
```
PatrickvL, line 300: Would it make sense to change this type from `auto` into `static std::string`?

PatrickvL (changes requested):
Besides the code remarks, a question : would it be possible to move the shared HLSL code towards an include file? That'd avoid having to change anything in that separately for each shader.

src/core/hle/D3D8/XbVertexShader.cpp
RadWolfie: nit: Always include colon after TODO/NOTE, etc comments.

src/core/hle/D3D8/Direct3D9/Shader.cpp
NZJenkins: No, out if scope unless this is unstable for some reason. I was going to make this toggleable opt in via imgui but stopped short

src/core/hle/D3D8/Direct3D9/Shader.cpp
NZJenkins: I don't care either way but can change if there is a good reason

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: I'm not sure but didn't want to touch too much.
If it's incorrect I'll remove it,
though.

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: It would be more efficient but I don't think it's worth it for a niche feature

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
NZJenkins: Idk honestly

src/core/hle/D3D8/Direct3D9/Shader.cpp
```diff
@@ -140,3 +147,176 @@ extern HRESULT EmuCompileShader
...
+
+	EmuLog(LOG_LEVEL::DEBUG, "Starting shader file watcher...");
+
+	auto fsWatcher = [](void* param) -> DWORD {
+		// Determine the filename and directory for the fixed function shader
+		char cxbxExePath[MAX_PATH];
+		GetModuleFileName(GetModuleHandle(nullptr), cxbxExePath, MAX_PATH);
+		auto hlslDir = std::filesystem::path(cxbxExePath)
+			.parent_path()
+			.append("hlsl/");
```
RadWolfie, line 280: Why is the function calls on a separate lines? Aren't they suppose to be on same line?

src/core/hle/D3D8/Direct3D9/Shader.cpp
```diff
@@ -140,3 +147,176 @@ extern HRESULT EmuCompileShader
...
+
+		EmuLog(LOG_LEVEL::DEBUG, "Shader file watcher exiting...");
+
+		// until there is a way to disable hotloading
+		// this is always an error
+		FindCloseChangeNotification(changeHandle);
+		return 1;
+	};
+
+	fsWatcherThread = CreateThread(nullptr, 0, fsWatcher, nullptr, 0, nullptr);
```
RadWolfie, line 321: nit: use C++'s std::thread instead of Windows' exclusive function.

src/core/hle/D3D8/XbVertexShader.cpp
NZJenkins: not sure what you mean specifically. `shader` should be temporary here though

src/core/hle/D3D8/Direct3D9/Shader.cpp
NZJenkins: using `std::thread` means it's not necessary, we can check the thread is running with `joinable`

src/core/hle/D3D8/Direct3D9/Shader.cpp
NZJenkins: just style

src/core/hle/D3D8/Direct3D9/Shader.cpp
NZJenkins: nice, std::thread is much simpler also

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
```diff
@@ -372,4 +369,4 @@ R"DELIMITER(
 	return xOut;
 }
 
-// End of vertex shader footer)DELIMITER" /* This terminates the footer raw string" // */
+// End of vertex shader footer
```
PatrickvL, line 372: Can this comment be removed entirely?

src/core/hle/D3D8/Direct3D9/Shader.h
```diff
@@ -10,3 +10,38 @@ extern HRESULT EmuCompileShader
...
+	void LoadShadersFromDisk();
+
+	// counts upwards on every change detected to the shader source files at runtime
+	volatile int shaderVersionOnDisk = 0;
+	// current loaded shader version
+	// Initialized to < shaderVersionOnDisk
+	int shaderVersionLoadedFromDisk = -1;
+};
+
+extern ShaderSources g_ShaderHlsl;
```
PatrickvL, line 47: It's probably cleaner to rename the `g_ShaderHlsl` global variable into `g_ShaderSources`, along with its also renamed type name.

src/core/hle/D3D8/Direct3D9/Shader.cpp
PatrickvL: It's what's done mostly in this repository, so if you're editing this file anyway, best take it along

src/core/hle/D3D8/Direct3D9/Shader.cpp
```diff
@@ -140,3 +148,174 @@ extern HRESULT EmuCompileShader
...
+	static std::thread fsWatcherThread;
+
+	if (fsWatcherThread.joinable()) {
+		EmuLog(LOG_LEVEL::ERROR2, "Ignoring request to start shader file watcher - it has already been started.");
+		return;
+	}
+
+	EmuLog(LOG_LEVEL::DEBUG, "Starting shader file watcher...");
+
+	fsWatcherThread = std::thread([]{
```
ergo720, line 275: I see that this `std::thread` is neither joined nor detached. In that case, the C++ standard says that `std::terminate` is invoked when the destructor of the object is called. To avoid that, you should either join or detach the thread yourself, or better, just use `std::jthread` from C++20, which automatically joins on destruction.
With the above said, since I see that the thread is static and we exit the process with `TerminateProcess` of Windows (which doesn't call the destructors), then it's probably fine to ignore this problem in practice, but in my opinion, it still makes sense to solve this anyway

src/core/hle/D3D8/Direct3D9/Shader.h
```diff
@@ -10,3 +10,38 @@ extern HRESULT EmuCompileShader
...
+	int Update();
+
+	// Start a thread to watch for changes in the shader folder
+	void InitShaderHotloading();
+
+private:
+	void LoadShadersFromDisk();
+
+	// counts upwards on every change detected to the shader source files at runtime
+	volatile int shaderVersionOnDisk = 0;
```
ergo720, line 41: Why is this `volatile`? Since I see it's being accessed from the `fsWatcherThread` object, I assume it's to ensure thread safety. In that case, `volatile` is not suitable for that purpose and you should use `std::atomic_int` instead

src/core/hle/D3D8/Direct3D9/Shader.cpp
RadWolfie: I did not see this style occur anywhere in our codebase. Can someone else confirm this?

Meaning if this style is not in our codebase, then basically we do not use this style here.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderPassthrough.hlsl
```diff
@@ -0,0 +1,131 @@
...
+       fogFactor = (fogEnd - fogDepth) / (fogEnd - fogStart);
+
+    xOut.oPos = reverseScreenspaceTransform(oPos);
+    xOut.oD0 = saturate(oD0);
+    xOut.oD1 = saturate(oD1);
+    xOut.oFog = fogFactor; // Note : Xbox clamps fog in pixel shader
+    xOut.oPts = oPts.x;
+    xOut.oB0 = saturate(oB0);
+    xOut.oB1 = saturate(oB1);
+    // Scale textures (TODO : or should we apply this to the input register values?)
```
RadWolfie, line 124: No space between TODO and colon.

src/core/hle/D3D8/XbVertexShader.cpp
```diff
@@ -1124,54 +1126,70 @@ IDirect3DVertexDeclaration* CxbxCreateHostVertexDeclaration(D3DVERTEXELEMENT *pD
...
+	}
+
+	return shader;
+}
+
 void CxbxUpdateHostVertexShader()
 {
 	extern bool g_bUsePassthroughHLSL; // TMP glue
+	// TODO move d3d9 state to VertexShader.cpp
+	static IDirect3DVertexShader* fixedFunctionShader = nullptr; // TODO move to shader cache
```
RadWolfie, line 1147: Always add colon after TODO/NOTE, etc comments.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderPassthrough.hlsl
```diff
@@ -0,0 +1,131 @@
...
+    float4 oT0 = v9;
+    float4 oT1 = v10;
+    float4 oT2 = v11;
+    float4 oT3 = v12;
+
+    // Copy variables to output struct
+    VS_OUTPUT xOut;
+
+    // Fogging
+    // TODO deduplicate
```
RadWolfie, line 95: Always add colon after TODO/NOTE, etc comments.

RadWolfie (changes requested):
Even though two code remarks change requested are relocation but it's better to correct them as well.

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
```diff
@@ -682,6 +683,10 @@ void CxbxInitWindow(bool bFullInit)
 	g_renderbase->SetWindowRelease([] {
 		ImGui_ImplWin32_Shutdown();
 	});
+
+	g_ShaderSources.Update();
```
PatrickvL, line 687: Nit: Consider to explicitly ignore return value here (for instance via a cast to void, of whatever other standard exists for this).

PatrickvL: Then what do you reckon is wisest - centralise reused shader code towards an .hlsli include file now, or avoid any further risk and do that later?

src/core/hle/D3D8/XbVertexShader.cpp
PatrickvL: Just ignore this remark, it was incorrect - the previous line contents declared a static global variable, now it's a function header that doesn't need a static marker.

src/core/hle/D3D8/XbVertexShader.cpp
PatrickvL: I agree it's best to avoid changing too much - it's already great what you've done here. Further work can come later, indeed.

src/core/hle/D3D8/Direct3D9/Shader.cpp
PatrickvL: It's what's done mostly in this repository, so if you're editing this file anyway, best take it along

PatrickvL:
> > Besides the code remarks, a question : would it be possible to move the shared HLSL code towards an include file? That'd avoid having to change anything in that separately for each shader.
> 
> Yes it's possible - for example FixedFunctionVertexShader is using #include to share code with Cxbxr already.



> > Besides the code remarks, a question : would it be possible to move the shared HLSL code towards an include file? That'd avoid having to change anything in that separately for each shader.
> 
> Yes it's possible - for example FixedFunctionVertexShader is using #include to share code with Cxbxr already.

Then what do you reckon is wisest - centralise reused shader code towards an .hlsli include file now, or avoid any further risk and do that later?

src/core/hle/D3D8/Direct3D9/Shader.h
PatrickvL: Resolved

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderTemplate.hlsl
PatrickvL: Resolved

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
PatrickvL: If most of this code base is specifying explicit types, and if you're going to touch up this pull request a little more anyway, then perhaps it's good to take this along as well?

src/core/hle/D3D8/Direct3D9/Shader.h
PatrickvL: Resolved

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Resolved

src/core/hle/D3D8/XbPixelShader.cpp
PatrickvL: Resolved

src/core/hle/D3D8/Direct3D9/Shader.cpp
PatrickvL: Acceptable.

projects/misc/batch.cmake
PatrickvL: Resolved

src/core/hle/D3D8/XbVertexShader.cpp
PatrickvL: Acceptable

src/core/hle/D3D8/Direct3D9/Shader.cpp
NZJenkins: I just want a fire and forget thread for this - sounds like that's `std::jthread`

src/core/hle/D3D8/Direct3D9/Shader.cpp
NZJenkins: I was answering your question. But it sounds like you meant "please change this"

src/core/hle/D3D8/Direct3D9/Shader.h
NZJenkins: I marked it `volatile` since my understanding is that calls out variables shared by multiple threads or can change unexpectedly - which I think has some implications in c++ compiler wise, but also makes it easy to see for a reader.

There isn't any synchronization happening as such, so I thought it was fine to ignore locks/atomics etc. here

I can remove volatile or synchronize the value if it makes more sense, I'm not sure.

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
NZJenkins: ok. I was a bit confused because the explicit type would be `const char*`. can be explicit

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderPassthrough.hlsl
NZJenkins: there are 457 `TODO :` in the codebase, 395 `TODO:`, and 59 `TODO `

Not sure why TODO comments are worth commenting on, let alone doubling back to "fix"?

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
NZJenkins: it seems to be `(void)`, though there is also `std::ignore`. `_ =` would be nice!

src/core/hle/D3D8/Direct3D9/Shader.h
NZJenkins: made it atomic anyway

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderPassthrough.hlsl
RadWolfie: Because `TODO:` is grammar correct. When I had read the documentations from past and present. Not once had I seen without colon nor has space between them. Also, if you had look up "todo" keyword in GitHub's search engine. You will find numerous of "todo:" results in various repositories. Some may have `TODO(something something): details here` which isn't correct way to do it when it should be `TODO: (something something) details here`. Some languages may do things differently except doesn't apply to C/C++ language.

For historically education, past Visual Studio IDEs does not list any TODOs doesn't have colon after it. And require no space in-between either.

Other than that, if clang-format is enforced. It will not check for those mistakes which is our responsibility to correct them manually.

The log file may can be different to align colons as a columns view for easier to read which is a different topic.

src/core/hle/D3D8/Direct3D9/Shader.cpp
RadWolfie: confirmed resolved

src/core/hle/D3D8/Direct3D9/Direct3D9.cpp
PatrickvL: Resolved

src/core/hle/D3D8/Direct3D9/VertexShader.cpp
PatrickvL: Resolved (partially, there's other auto's in these changes but those are all moved around, not new - but since there's no guidance in this repo on what to use when, it's fine for now)

PatrickvL (approved):

src/core/hle/D3D8/Direct3D9/Shader.cpp
RadWolfie: Turns out upstream's shader files are affected. Nowhere else used this code style. For now I'll dismiss this.

src/core/hle/D3D8/XbVertexShader.cpp
RadWolfie: code was relocated, allowing to bypass request change.

src/core/hle/D3D8/Direct3D9/CxbxVertexShaderPassthrough.hlsl
RadWolfie: code was relocated, allowing to bypass request change.

PatrickvL (dismissed):

medievil1:
the template isn't being built into the folder currently 
image

medievil1:
`./hlsl/backup contains backup copies so it's easy to revert changes if someone forgets to make backups or what changes they made etc.`
the above is not happening, no backup folder is created when you make changes to vertex or pixel shader

projects/cxbx/CMakeLists.txt
```diff
@@ -198,3 +198,16 @@ set_source_files_properties(
...
+add_custom_command(
+	TARGET cxbx PRE_BUILD
+	COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:cxbx>/hlsl
+	COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CXBXR_HLSL_FILES} $<TARGET_FILE_DIR:cxbx>/hlsl
+	# These files can be edited.
+	# Create backup copies for convenience of restoring original shader behaviour.
+	COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:cxbx>/hlsl/backup
+	COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CXBXR_HLSL_FILES} $<TARGET_FILE_DIR:cxbx>/hlsl/backup
+)
+install(DIRECTORY $<TARGET_FILE_DIR:cxbx>/hlsl DESTINATION bin)
```
RadWolfie, line 213: These should not be in here, cxbx is a GUI project, not an emulator project. Also, they need to remain in root directory as usual.

projects/misc/batch.cmake
```diff
@@ -27,14 +27,3 @@ message("Runtime Build Directory: ${TargetRunTimeDir}")
...
-# Copy HLSL files to the output directory, which are loaded at runtime
-FILE(GLOB CXBXR_HLSL_FILES "${CMAKE_SOURCE_DIR}/src/core/hle/D3D8/Direct3D9/*.hlsl*")
-set(HlslOutputDir ${TargetRunTimeDir}/hlsl)
-file(MAKE_DIRECTORY ${HlslOutputDir})
-file(COPY ${CXBXR_HLSL_FILES} DESTINATION ${HlslOutputDir})
-# These files can be edited.
-# Create backup copies for convenience of restoring original shader behaviour.
-set(HlslBackupOutputDir ${HlslOutputDir}/backup)
-file(MAKE_DIRECTORY ${HlslBackupOutputDir})
-file(COPY ${CXBXR_HLSL_FILES} DESTINATION ${HlslBackupOutputDir})
```
RadWolfie: These files are triggered whenever the files has change on build process. Tbh, they should be remain here whenever file was edited.

medievil1:
anyone every find any functional issues??...everything I tried works as it should, no outstanding issues...

PatrickvL (approved):

PR #2440, opened 2023-12-20
Fix Xbox Kernel Test Suite issues discovered

RadWolfie:
From https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/87 changes. There are two update API tests, RtlCompareUnicodeString and RtlCompareString, which has been change from kernel test suite. I noticed the return results are off. They have been fixed in this pull request to align with accuracy hardware's kernel results.

Plus there is a wrong type of Buffer variable from UNICODE_STRING structure when it should be wchar_xt instead of ushort_xt.

RtlWalkFrameChain somehow include below title's executable dos header, aka null value, from cxbxr which doesn't align with the test's expectation. It has been resolved by add a check for below title's executable dos header which is static value by the way.

> [!WARNING]
> KeInsertQueueDpc test can cause deadlock due to cxbxr's side of the issue. (Remember to disable this test from cxbxr's side for verification if needed.)

### DIFF
--- a/src/core/kernel/common/types.h
+++ b/src/core/kernel/common/types.h
@@ -266,7 +266,7 @@ typedef struct _UNICODE_STRING
 {
     ushort_xt  Length;
     ushort_xt  MaximumLength;
-    ushort_xt *Buffer;
+    wchar_xt  *Buffer;
 }
 UNICODE_STRING, *PUNICODE_STRING;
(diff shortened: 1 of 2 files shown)


PatrickvL (approved):
Nice work! Approved.

PR #2441, opened 2023-12-29
Fixed a bug in RtlTimeFieldsToTime

ergo720:
Triggered by this PR [here](https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/89).

### DIFF
--- a/src/core/kernel/exports/EmuKrnlRtl.cpp
+++ b/src/core/kernel/exports/EmuKrnlRtl.cpp
@@ -42,6 +42,7 @@ namespace NtDll
 
 #include "core\kernel\init\CxbxKrnl.h" // For CxbxrAbort()
 #include "core\kernel\support\Emu.h" // For EmuLog(LOG_LEVEL::WARNING, )
+#include "EmuKrnlKi.h"
 #include <assert.h>
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


PatrickvL (approved):
Thanks!

RadWolfie:
According to the xkts PR 89 (see link above), the test results are basically the same. I took a look at the difference and found no change at all.

i.e. `CLOCK_TIME_INCREMENT = 0x2710` is value of 10,000 which is basically the same as TICKSPERMSEC (10,000) define. So, my question is how is it a bug? According to our team conversation, @ergo720 found missing one digit number. Yet... both values are the same.

----
Below is the test gave the same result before and after this pull request:
```
Kernel Test Suite
build: a5d80e1
Random seed used is 0
Config File Was Loaded. Only running requested tests.
-----------------------------------------------------
0x0130 - RtlTimeFieldsToTime: Tests Starting
  ERROR(line 112): Expected array time_tests[18].time_expected.QuadPart = 0x0, Got time_tests[18].time_actual.QuadPart = 0x26bffb7ecc000
  ERROR(line 113): Expected array time_tests[18].result_expected = 0x0, Got time_tests[18].result_actual = 0x1
0x0130 - RtlTimeFieldsToTime: One or more tests FAILED
------------------------ End of Tests -----------------------
```
time_tests[18] test is checking against non-leap year day to be expected non-existence, yet Cxbx-Reloaded says it exists. However, it is not relative to the bugfix made here.

PR #2451, opened 2024-02-03
Fix hardware model conversions + use respective hardware model based on console type

RadWolfie:
When I was writing extended log info based on documentation by the xboxdevwiki community for xbox kernel test suite. I noticed a flaw in hardware model conversion from Cxbx-Reloaded's end. Such as `HardwareModel::Revision1_5` points to `P2L` flag which then return `P05`, aka `Revision 1.1` string incorrectly. So, I corrected the return string to `P2L` as expected. Plus made major update on the conversions function and used respective hardware models based on console type.

I changed the retail revision 1.5 to 1.6 as the last released revision model.

Reference used by: https://xboxdevwiki.net/Xboxen_Info

---
> [!NOTE]
> Not everything is fully implemented to use respective hardware model accuracy. For now, this is the first move to use proper hardware model.

### DIFF
--- a/src/core/kernel/exports/EmuKrnlXbox.cpp
+++ b/src/core/kernel/exports/EmuKrnlXbox.cpp
@@ -41,15 +41,19 @@ XBSYSAPI EXPORTNUM(321) xbox::XBOX_KEY_DATA xbox::XboxEEPROMKey = { 0 };
 // ******************************************************************
 // * 0x0142 - XboxHardwareInfo
 // ******************************************************************
+// TODO: The main goal is to completely unset custom init values and have
+//       them set from kernel's initialization end and device classes.
+//       Although, device classes does not really set this value but read
+//       from PCI space for Gpu rev, Mcp rev, and possibility INTERNAL_USB flag.
 XBSYSAPI EXPORTNUM(322) xbox::XBOX_HARDWARE_INFO xbox::XboxHardwareInfo =
 {
 	// TODO: What exactly 0xC0000030 flags are? Might need default to null then set them later properly.
 	// NOTE: Will be set by src/devices/Xbox.cpp and maybe other file(s)...
-	0xC0000030, // Flags: 1=INTERNAL_USB, 2=DEVKIT, 4=MACROVISION, 8=CHIHIRO
-	0xD3, // GpuRevision, byte read from NV2A first register, at 0xFD0000000 - see NV_PMC_BOOT_0
-	0, // NOTE: Will be set by src/devices/Xbox.cpp file.
-	0, // unknown
-	0 // unknown
+	.Flags = 0xC0000030, // Flags: 1=INTERNAL_USB, 2=DEVKIT, 4=MACROVISION, 8=CHIHIRO
+	.GpuRevision = 0xD3, // GpuRevision, byte read from NV2A first register, at 0xFD0000000 - see NV_PMC_BOOT_0
+	.McpRevision = 0, // NOTE: Will be set by src/devices/Xbox.cpp file.
+	.Unknown3 = 0, // unknown
+	.Unknown4 = 0 // unknown
 };
 
 // ******************************************************************
--- a/src/core/kernel/exports/EmuKrnlXbox.cpp
+++ b/src/core/kernel/exports/EmuKrnlXbox.cpp
@@ -43,9 +43,11 @@ XBSYSAPI EXPORTNUM(321) xbox::XBOX_KEY_DATA xbox::XboxEEPROMKey = { 0 };
 // ******************************************************************
 XBSYSAPI EXPORTNUM(322) xbox::XBOX_HARDWARE_INFO xbox::XboxHardwareInfo =
 {
-	0xC0000031, // Flags: 1=INTERNAL_USB, 2=DEVKIT, 4=MACROVISION, 8=CHIHIRO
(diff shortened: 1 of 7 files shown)


src/core/kernel/exports/EmuKrnlXbox.cpp
```diff
@@ -43,9 +43,11 @@ XBSYSAPI EXPORTNUM(321) xbox::XBOX_KEY_DATA xbox::XboxEEPROMKey = { 0 };
...
 XBSYSAPI EXPORTNUM(322) xbox::XBOX_HARDWARE_INFO xbox::XboxHardwareInfo =
 {
-	0xC0000031, // Flags: 1=INTERNAL_USB, 2=DEVKIT, 4=MACROVISION, 8=CHIHIRO
-	0xA2, // GpuRevision, byte read from NV2A first register, at 0xFD0000000 - see NV_PMC_BOOT_0
-	0xD3, // McpRevision, Retail 1.6 - see https://github.com/JayFoxRox/xqemu-jfr/wiki/MCPX-and-bootloader
+	// TODO: What exactly 0xC0000030 flags are? Might need default to null then set them later properly.
+	// NOTE: Will be set by src/devices/Xbox.cpp and maybe other file(s)...
+	0xC0000030, // Flags: 1=INTERNAL_USB, 2=DEVKIT, 4=MACROVISION, 8=CHIHIRO
+	0xD3, // GpuRevision, byte read from NV2A first register, at 0xFD0000000 - see NV_PMC_BOOT_0
+	0, // NOTE: Will be set by src/devices/Xbox.cpp file.
```
PatrickvL, line 50: Consider mentioning the function name as well (`InitXboxHardware()`)

src/devices/Xbox.h
```diff
@@ -53,9 +53,20 @@ typedef enum {
 	Revision1_4,
 	Revision1_5,
 	Revision1_6,
-	DebugKit
+	Retail = 0x00,
```
PatrickvL, line 56: Consistent ordering would either have `Retail` at the top, or have all system flags grouped together either at the top or here

src/devices/Xbox.cpp
```diff
@@ -41,66 +45,82 @@ USBDevice* g_USB0;
...
-		return TVEncoder::Conexant;
-	default: 
-		// UNREACHABLE(hardwareModel);
-		return TVEncoder::Focus;
+	default:
+		CxbxrAbort("TVEncoderFromHardwareModel: Unknown conversion for hardware model (0x%02X)", hardwareModel);
+	}
+}
+
+xbox::uchar_xt MCP_PCIRevisionFromHardwareModel(HardwareModel hardwareModel)
```
PatrickvL, line 108: The other functions above all return a dedicated enum type; Consider introducing one for this function as well.

src/devices/Xbox.cpp
```diff
@@ -111,16 +131,29 @@ void InitXboxHardware(HardwareModel hardwareModel)
 	SCMRevision smc_revision = SCMRevisionFromHardwareModel(hardwareModel);
 	TVEncoder tv_encoder = TVEncoderFromHardwareModel(hardwareModel);
 
+	// Only Xbox 1.0 has usb daughterboard supplied, later revisions are integrated onto motherboard.
+	if (GET_HW_REVISION(hardwareModel) == HardwareModel::Revision1_0) {
```
PatrickvL, line 135: If this applies to `Retail` then perhaps be explicit about that (and add an `else` below this block, so that the following if's can't be reached).

If this however also applies to the other systems as well, then perhaps mention that in a comment?

PatrickvL (dismissed):
Nice work! Only a few minor questions / suggestions.

PatrickvL (approved):
No further remarks, thanks!

src/devices/Xbox.cpp
```diff
@@ -41,66 +45,82 @@ USBDevice* g_USB0;
...
-		// EmuLog(LOG_LEVEL::WARNING, "Guessing SCMRevision");
-		return SCMRevision::P2L; // Assumption; Our SCM returns PIC version string "P05"
+		return SCMRevision::P2L;
 	case DebugKit:
-		return SCMRevision::D01; // Our SCM returns PIC version string "DXB"
+		return SCMRevision::D01;
+	case Chihiro_Type1:
+		return SCMRevision::D05;
+	case DebugKit_r1_2:
+	case Chihiro_Type3:
```
GXTX, line 80: Not entirely sure this is appropriate. A 1.1 revision Xbox can be used with a Type-1 media board. It's *possible* SEGA never shipped a Type-3 with a rev1.1, but it'd be hard to really know.

Either way it's just nitpicky and doesn't matter for anything in the long run.

Appreciate you using the info I gathered. :)

RadWolfie: I'm taking an educated guess on Chihiro's Type-1 motherboard's MCPX revision "0xD05" is a 1.1 revision board. This could be the same as the assumed "Debug Kit - 1.1", yet there are no details for it. 😢

As long as xbox community keeps submitting details to https://xboxdevwiki.net/Xboxen_Info page. The better information hardware accuracy we can use with the emulators. 😉

Anyway, I'm assuming Type-3's last released revision board is 1.2. Which is what has done here. We only use the last released revision board for all console types.

We appreciate the owners of their hardware details submitted to help improve accuracy details for useful report info and emulators. 🙂

PR #2452, opened 2024-02-03
fix incorrect setting in pixel shader template

medievil1:
PS_TEXTUREMODES_DOT_RFLCT_DIFF(ts) should be normal2, not normal3...
PS_TEXTUREMODES_DOT_RFLCT_DIFF(ts) works in conjunction with PS_TEXTUREMODES_DOT_RFLCT_SPEC(ts)(after) and PS_TEXTUREMODES_DOTPRODUCT(ts)(before) dotproduct uses it's own product/normal, rflct_diff uses dotproduct and it's own, and rflct_spec uses the previous 2 and it's own ... this corrects light reflection on the floor in Halo


medievil1:
here is an after screen shot of the result, trying to get a before but halo only runs when it wants too... the white light reflections on the floor are missing in the before 
unknown

PatrickvL (approved):

PR #2458, opened 2024-05-22
fix the build on vs2022 17.9.1

LukeUsher:
std::max_element is defined in `<algorithm>` so we were missing this required include
the longest_str call in DlgAbout was unable to be determined at compile time so is not suited for constexpr (not sure why this changed?)


medievil1:
I could have sworn I submitted this like a year ago...lol
but yea, I had to do the exact same thing... and can confirm 100% it fixes the build issue

PatrickvL (approved):
Looks good to me, thanks

PR #2469, opened 2024-12-13
Replace subhook with working mirror

Margen67:
See https://github.com/tianocore/edk2/issues/6398

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "import/subhook"]
 	path = import/subhook
-	url = https://github.com/Zeex/subhook.git
+	url = https://github.com/Cxbx-Reloaded/subhook.git
 	shallow = true
 [submodule "import/cs_x86"]
 	path = import/cs_x86


RadWolfie:
Hi @Margen67, we have our own [subhook forked repo](https://github.com/Cxbx-Reloaded/subhook) containing the same commit as we have here and the last known latest commit as well. Our forked repo has more up-to-date on another branch but both forked repos are not maintaining any bug fixes nor improvements. So, it would be best to use our forked repo from cxbxr's organization here where it will not be going anywhere.

Margen67:
> Hi @Margen67, we have our own [subhook forked repo](https://github.com/Cxbx-Reloaded/subhook) containing the same commit as we have here and the last known latest commit as well. Our forked repo has more up-to-date on another branch but both forked repos are not maintaining any bug fixes nor improvements. So, it would be best to use our forked repo from cxbxr's organization here where it will not be going anywhere.

Done.

RadWolfie (approved):

PR #2470, opened 2024-12-13
CI: Add back VS2022, update actions

Margen67:
VS2022 currently fails to build.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -30,21 +30,22 @@ jobs:
       fail-fast: false
       matrix:
         configuration: [Release, Debug]
-        vsver: [2019]
-        winver: [7]
-        sdkver: [10.0.22621.0]
+        vsver: [2019, 2022]
+        include:
+          - vsver: 2019
+            cmake_args: ',version=10.0.22621.0 -DCMAKE_SYSTEM_VERSION=7'
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
       - name: Generate CMake files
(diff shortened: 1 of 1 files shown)


RadWolfie:
The reason of fallback to VS2019 is due to current VS2022 has defective compiler. Even if VS2022 has successfully compiled, we will receive false positive reports from our testers and users.

If you separate the actions update changes into another pull request, I will be gladly accept them right away. Yet, I will verify to be sure all the other actions are up to date as well before approval.

Margen67:
> The reason of fallback to VS2019 is due to current VS2022 has defective compiler. Even if VS2022 has successfully compiled, we will receive false positive reports from our testers and users. For this reason, it is unlikely we will accept this pull request even if it's parallel builds.
> 
> If you separate the actions update changes into another pull request, I will be gladly accept them right away. Yet, I will verify to be sure all the other actions are up to date as well before approval.

What's wrong with VS2022? If it gets false positives then that's a problem with the virus detection.

If you want I can make it so VS2022 builds are excluded from releases so users don't download them.

RadWolfie:
The problem isn't virus detection. It is that the whole builds from CI (VS2022) are defective (aka the msvc compiler issue). Some titles doesn't get past the loading screen or progress further. We are not using the latest VS2022 version because of this. Once confirmation from testers say the latest VS2022 build is back to working state for able to progress. Then we will replace it back with VS2022 in CI workflow.

Besides the above primary issue, the other side effects such as the "Config Input" window cause a crash and emulation crashes. The users need to re-update the Visual C++ Redist package.

medievil1:
will VS2022 compile code as VS2019? asking cause VS2019 is no longer available to install

PR #2474, opened 2024-12-17
cmake: Replace /Ob2 with /Ob3

Margen67:
Might improve performance.

### DIFF
--- a/projects/cxbx/CMakeLists.txt
+++ b/projects/cxbx/CMakeLists.txt
@@ -121,7 +121,7 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
 
  # Reference: https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically
  # /Zi   = create a PDB file without affecting optimization
- # /Ob2  = Controls inline expansion of functions.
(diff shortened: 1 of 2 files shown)


projects/cxbxr-ldr/CMakeLists.txt
```diff
@@ -56,12 +56,6 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   /ENTRY:rawMain \
   "
  )
-
- # Set optimization options for release build
- set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} \
```
Margen67, line 61: If this is inheriting flags like I think it does, then there's no need for it since `/Oi` is defined elsewhere.

RadWolfie: I only see `/Oi` in `projects/<name>/CMakeLists.txt` files, so how is it defined in the parent directory?

Margen67: I thought it was getting flags from somewhere in `projects/`. Oops.

LukeUsher:
Don't see why not!

PR #2484, opened 2025-03-30
Updated SDL submodule to version 2.30.11

ergo720:
This updates the SDL2 submodule to the latest stable version possible (it doesn't switch to the SDL3 version). This fixes the linking error `error LNK2001: unresolved external symbol __ftol2` and allows to build cxbxr again with VS2022 (I am using the platform toolset Visual Studio 2022 (v143)). The original cause of the SDL2 error is explained in this [upstream PR]( https://github.com/libsdl-org/SDL/pull/5969). Note that capstone still fails to build, but that’s only required by the cxbxr-debugger, which is optional, so it doesn’t prevent building cxbxr itself.


RadWolfie:
**NOTE:** 2.30.11 release is not the latest stable version. SDL2's 2.32.2 release is the latest stable version (29 days ago).

Issue #2477, opened 2025-01-18
[Feature Request] Custom XBE injector into Xbox games


mrmaller1905:
## Quick summary

A suggestion to add custom XBEs (Xbox dynamic link library) injector to make Xbox games widescreen

## Details

Custom XBE injector allows users to make games widescreen and trainer to make ammo and health infinite in games such as Halo 2.

### System Configuration:
N/A

### Additional Information (if any):
N/A

RadWolfie:
Custom XBEs (Xbox **executable***) injector is outside the scope of this project. When you mentioned "Xbox dynamic link library" is not the same as the executable binary which is what xbe is. In this case, what you're asking for is similar to previous request #2471.

We currently have cheat engine script support in cxbxr-debugger project which the community can use. We don't plan on adding per-game patches any time soon.

Issue #2471, opened 2024-12-15
[Feature Request] DLL Injector


mrmaller1905:
## Quick summary
A suggestion to add DLL injector which allows users to make Xbox games widescreen and log filenames and hashes (e.g. common.dat/ .fat of Far Cry Instincts).

## Details
DLL injector allows users to make games widescreen and trainer to make ammo and health infinite in games such as Halo 2.

### System Configuration:
N/A

### Additional Information (if any):
N/A

ergo720:
This is already possible as far as I know. Also, it's not the job of cxbxr to be injected with DLLs. That's up to third-party programs to perform the injection, such as [this one](https://github.com/nefarius/Injector).

mrmaller1905:
Can you anyone add support for custom XBE injection into Xbox games to Cxbx-Reloaded?

Issue #1498, opened 2018-10-29
Cheats?


Demir2007:
the title explains itself........agin
CHEATS

LukeUsher:
Cxbx-Reloaded does not emulate CPU and memory in the same way as a traditional emulator, making it difficult to implement cheats how they would usually exist in emulation. 

You can, however, use tools designed for cheating on PC games (such as Cheat Engine) with no issues, if you know how to use it.

x1nixmzeng:
Cxbx does include a debugger which can load Cheat Engine (.CT) files to apply memory patches.

The setup just needs a bit of developer love :)

LukeUsher:
Closing as won't fix (for now, at least): Cheating can be achieved by using Cheat Engine

PR #2485, opened 2025-03-31
Deprecate the Cxbx-Reloaded Debugger

ergo720:
This deprecates the usage of the Cxbx-Reloaded Debugger by disabling it in the code. This is done because it has several drawbacks with no obvious benefits:
1. it's possible to load cheat tables with cheat engine itself, so no need to replicate that here.
2. it's not useful to debug issues in cxbxr itself, and developers are expected to use the debugger of Visual Studio instead (which is more capable too).
3. Its file events can be replaced by the file logging options of cxbxr.
4. It's written in C# instead of C++, and brings in a huge dependency on dotnet and capstone, which are not required by cxbxr.
5. After it was introduced, no one has really maintained it or developed any further.

With that being said, it can still be built by passing the variable `-DBUILD_CXBXR_DEBUGGER=ON` to the cmake build command, if someone still wants it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ cmake-build-*/
 
 #Test files
 src/[Vv]ersion.h
+src/common/CxbxDebugger.h
 *.xbe
 
 # Last known git hash
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,14 @@ if(NOT DEFINED CMAKE_RUNTIME_OUTPUT_DIRECTORY)
  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${Cxbx-Reloaded_BINARY_DIR}/bin")
 endif()
 
+set(BUILD_CXBXR_DEBUGGER OFF CACHE BOOL "Build the Cxbx-Reloaded debugger")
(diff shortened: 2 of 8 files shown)


RadWolfie:
1- No evidence had been provided for cheat engine is able to support multiple instance of emulation process. Nor has evidence it is able to support emulation reboot.
2- While it is correct, however the community made this request. And we were outvoted. Therefore require their vote to remove cxbxr-debugger.


4- It is completely optional.
5- There wasn't anything needed to be update in its long term. Plus it had been maintained from time to time.

Deprecate definition does not mean completely hidden and exclude from CI build. There are far too many unnecessary changes in the pull request when it is only needed to update a couple of cmake files. The rest of the codebase should not be touched other than add a warning indicator of such feature removal.

---

Pull request had been converted to draft for following reasons:
1. Community vote was not posted in discord server. Which is up to them to decide if they still want to keep it or not. **NOTE: The post of the vote must last at least a week from its creation before make the final decision.**
2. Cheat table support was not migrate into main GUI codebase. Which remain affected due ticket creations:
   - #2477
   - #2471
   - #1498
   - And various external sources had asked about it.

PR #2491, opened 2025-07-11
cmake: Raise cmake_minimum_required to 3.12

Margen67:
Other parts of the project already have it set to 3.12 so might as well for consistency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.8)
+cmake_minimum_required (VERSION 3.12)
 
 project(Cxbx-Reloaded)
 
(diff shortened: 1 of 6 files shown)


RadWolfie (approved):
Technically the upstream's cmake files were updated due to API doesn't exist until later version of CMake build.

> Changed in version 3.31: Compatibility with versions of CMake older than 3.10 is deprecated. Calls to [cmake_minimum_required(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required) or [cmake_policy(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_policy.html#version) that do not specify at least 3.10 as their policy version (optionally via ...<max>) will produce a deprecation warning in CMake 3.31 and above.

However, as state in [CMake's doc](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html), we do not want any warning produced in the CI log. So, this pull request is LGTM.

